PR #11, opened 2013-12-12
Added serialization whitepaper

esteve:
This pull request adds the serialization whitepaper produced during the ROS 2.0 workshop


tfoote:
Do we want to enforce a line wrap of some form? This diff is almost impossible to read due to the long lines not wrapping.  Also if we have long lines potential future merges maybe much more difficult.

dirk-thomas:
I would propose the "one sentence per line" rule. Any other approach will inevitably screw up diffs.

wjwwood:
I agree with @dirk-thomas, it is a pain to write one sentence per line, but it is the best way to review diff's.

There are chrome plugins for enabling soft wrapping in github diffs, as well as side by side diffs:

https://chrome.google.com/webstore/detail/octosplit/mnkacicafjlllhcedhhphhpapmdgjfbb

esteve:
I reformatted the article to make the diff a bit more readable. It looks better, IMHO.

wjwwood:
I forgot about this pull request, we should look at refreshing (make it consistent with other articles) and reviewing this and then publishing it.

Things to do:
- better abstract
- embed abstract in article
- add author, embed in article
- update toc line
- convert to one sentence per line from consistency with the other articles

esteve:
- [x] better abstract
- [x] embed abstract in article
- [x] add author, embed in article
- [x] update toc line
- [x] convert to one sentence per line for consistency with the other articles

esteve:
@wjwwood I've addressed your feedback, though maybe the abstract still needs more work.

wjwwood:
Cool, this seems to be one paragraph per line rather than one sentence per line. I think the abstract is OK, if you think that it is good enough.

dirk-thomas:
Please add newlines after each sentence. I would say the "one sentence per line" does not prevent you from wrapping long lines.

esteve:
@wjwwood I've updated the article adding a newline after each sentence.

@dirk-thomas I originally had long lines wrapped to 80 characters, but I unwrapped them to be consistent with the rest of the articles.

dirk-thomas:
Both fine for me - the GitHub diff wraps the long lines so they are easy to read.

tfoote:
+1 merging

PR #15, opened 2014-06-18
Update 10_why_ros2.md

scpeters:
Here's some minor changes. The article reads very nicely.

Discussion of the three changes in order:
- first one could the `the` or `a` but it feels like an article should be there
- typo
- judgment call; the phrase was hard to parse on first read. this might make it better? open to suggestions


wjwwood:
+1, I'll let @gerkey make the merge decision.

PR #22, opened 2014-08-07
Updated links, RPC4DDS

jwillemsen:
Updated some links to point to the correct documents and refer to the RPC4DDS WiP page


wjwwood:
Thanks!

jwillemsen:
Hi,

Not a problem, interesting effort you are working on. We integrated DDS
using C++11 within our CBDDS solution. We did hide all lower level DDS
entities complete behind the DDS4CCM Event pattern. This is all based on
LwCCM with major input from the ideas for UCM. A beta version of what we
have is available on our OSportal at http://osportal.remedy.nl, there are
some examples of some user components there. We are currently providing
request/reply through CORBA but want to add RPC4DDS support later this
year. All looks pretty similar to what you are trying to accomplish. Also check out http://www.slideshare.net/RemedyIT/

Johnny

jwillemsen:
I did see auto in some of your proto code, if you are going for C++11, check the ideas that are in IDL2C++11, that language mapping is safe to use, no C++ pointers, all reference types, but still high performance through the usage of C++11 move semantics.

PR #23, opened 2014-08-28
draft for middleware interface

dirk-thomas:
@gerkey @esteve @tfoote @wjwwood fyi the current draft for the middleware interface whitepaper. 

I still have to write the last part describing the actually implemented architecture.

### DIFF
--- a/articles/60_ros_middleware_interface.md
+++ b/articles/60_ros_middleware_interface.md
@@ -1,0 +1,376 @@
+---
+layout: default
+title: ROS 2 middleware interface
+permalink: articles/ros_middleware_interface.html
+abstract:
+  This article describes the rational for using an abstract middleware interface between ROS and a specific middleware implementation.
+  It will outline the targeted use cases as well as their requirements and constraints.
+  Based on that the developed middleware interface is explained.
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+published: false
+---
+
+* This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+
+## The *middleware interface*
+
+
+### Why does ROS 2 have a *middleware interface*?
+
+The ROS client library defines an API which exposed communication concepts like publish / subscribe to users.
+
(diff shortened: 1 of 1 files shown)


articles/60_ros_middleware_interface.md
```diff
@@ -0,0 +1,236 @@
...
+    |      user land       |   1) use the ROS message
+    +----------------------+      ^
+    |  ROS client library  |   2) callback passing a ROS message
+    +----------------------+      ^
+    | middleware interface |      ^
+    +----------------------+      ^
+    |      mw impl N       |   3) convert the DDS sample into a ROS message and invoke subscriber callback
+    +----------------------+
+
+Depending on the middleware implementation the extra conversion can be avoided by implementing serialization functions from as well as deserilization functions to the ROS messages.
```
wjwwood: This should probably either be:

```
Depending on the middleware implementation the extra conversion can be avoided by implementing serialization functions from, as well as deserilization functions to, the ROS messages.
```

Or

```
Depending on the middleware implementation the extra conversion can be avoided by implementing serialization functions from ROS messages as well as deserilization functions to the ROS messages.
```

wjwwood:
+1

dirk-thomas:
I have updated the document and added the missing chapter. Please provide any kind of feedback so that we can announce it on the _ros-sig-ng-ros_ mailing list eaerly next week.

gerkey:
+2.  Really nice explanation.

I made some cosmetic edits in b628668ad4c4d7b5b7425115194be0365746042a.

The announcement will separately include the description/video of the demo showing how easy the swap is, right?

wjwwood:
+1 for new stuff too.

tfoote:
+1 overall

I'd like to see a reminder that this is the middleware interface at the top of the design section L231 mentioning that this is the minimum layer enabling a developer to build a client library on top and is not expected to be a user facing API.

esteve:
+1 looks good to me too (including @gerkey's changes)

dirk-thomas:
@tfoote If you think the fact that it is not a user facing API is not clear enough then I would propose that it needs to be clarified way earlier then on L231.

In the first questions I tried to describe that the ROS client library is the user facing API. Also the following layer diagram show that. If you think it needs more clarification can you please propose a patch to the very beginning of the document?

PR #33, opened 2015-06-16
add article about generated C++ interfaces

dirk-thomas:


articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### Setters
+
+Beside the direct access to the member variables for each member a *setter* method is generated.
+They are named after the members with a leading `set_`.
+The setter methods have a single argument to pass the value for the member variable.
+Each setter methods returns the struct itself to enable (method chaining)[https://isocpp.org/wiki/faq/ctors#named-parameter-idiom}.
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> Need non-colliding naming
```
wjwwood: Using a `set__` suffix seems like that most readable and non-colliding name to me.

articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,132 @@
...
+  <b>TODO:</b> const and non-const getters
+</div>
+
+
+#### Setters
+
+Beside the direct access to the member variables for each member a *setter* method is generated.
+They are named after the members with a leading `set_`.
+The setter methods have a single argument to pass the value for the member variable.
+Each setter methods returns the struct itself to enable (method chaining)[https://isocpp.org/wiki/faq/ctors#named-parameter-idiom}.
```
wjwwood: This should be closed with a `]` not a `}`.

articles/112_generated_interfaces_cpp.md
dirk-thomas: I did that already in 4d4b712fc72df81f0d11ac1d9da4791ed2850ce2 but I am not sure I like it.

articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,132 @@
...
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> The *copy constructor* initializes all members based on a reference instance. `=default`?
+</div>
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> *Move constructor*? `=default`?
+</div>
+
+The struct has no constructor with positional arguments for the members.
+The short reason for this is that if code would reply on positional arguments to construct data objects changing a message definition would break existing code in various ways.
```
wjwwood: `reply` => `rely`

articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,132 @@
...
+Since this would discourage to evolve message definitions the data structures should be populared by setting the members separately, e.g. using the setter methods.
```
wjwwood: `discourage to evolve message definitions` is sort of awkward, maybe `discourage {people, users, developers} to evolve message definitions` or `discourage evolution of message definitions`.

articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,132 @@
...
+#### Operators
+
+The comparison operators `==` and `!=` perform the comparison on a per member basis.
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> *Copy / move assignment operators*? `=default`?
+</div>
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> *Swap*? `=default`?
```
wjwwood: Is there a default swap? I thought this only existed in vectors and similar std containers.

articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,127 @@
...
+  <b>TODO:</b> describe separate header files
+</div>
+
+
+### Messages
+
+A message results in a templated `struct` with the same name followed by an underscore.
+The single template argument is the allocator for the data structure.
+
+For ease of use there is a `typedef` with the same name as the message which uses a default allocator (e.g. `std::allocator`).
```
wjwwood, line 46: I've always wondered about this. Why don't we just have a default template argument like with `std::vector`?

http://en.cppreference.com/w/cpp/container/vector

Is it because it's the only template argument?

articles/112_generated_interfaces_cpp.md
dirk-thomas: There is no `=default` but we could rely on the default swap behavior. Not yet sure if we should / need to customize it.

articles/112_generated_interfaces_cpp.md
dirk-thomas, line 46: Afaik you can't refer to the class as `ns:MyMsg` anymore but must explicitly mentioned the template `<>`.

dirk-thomas:
I have update the article based on the discussion in the meeting and removed all references to copy / move constructors / assignment operators. Please notice the change of the pointer types.

@ros2/owners Please review.

esteve:
+1

tfoote:
+1

articles/112_generated_interfaces_cpp.md
dirk-thomas: Fixed.

articles/112_generated_interfaces_cpp.md
dirk-thomas: Fixed.

articles/112_generated_interfaces_cpp.md
dirk-thomas: Fixed in bbebf8d806642d056e944f8ed99650be0d0666d7

articles/112_generated_interfaces_cpp.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Pointer types
+
+The struct contains `typedefs` for the four common pointer types `plain pointer`, `std::shared_ptr`, `std::unique_ptr`, `std::weak_ptr`.
+For each pointer type there a non-const and a const `typedef`:
+
+* `Ptr` and `ConstPtr`
+* `SharedPtr` and `ConstSharedPtr`
+* `UniquePtr` and `ConstUniquePtr`
+* `WeakPtr` and `ConstWeakPtr`
```
dirk-thomas, line 101: Should we also add `auto_ptr` here? Or should we remove any of the above?

wjwwood, line 101: I'd say no; `auto_ptr` is deprecated in C++11. The others look fine. People might find it weird that `Ptr` is now a "raw" pointer rather than a shared pointer, as it is in ROS 1. Not sure what else we could do though. Maybe `RawPtr` and `ConstRawPtr`, but I don't really like that either.

wjwwood:
+1

articles/112_generated_interfaces_cpp.md
jackieokay, line 101: -1 for `RawPtr`. The definitions in the document for `Ptr` types looks fine to me.

PR #35, opened 2015-06-23
Added article about ROS 2.0 in realtime

jackieokay:
This article summarizes my research into real-time software development techniques, with a focus on C/C++ running on RTLinux or a Linux-based RTOS. It lays out a tentative plan for how to benchmark real-time performance in ROS 2.0 and how existing and future code in the core communication stack can be made real-time-safe.

### DIFF
--- a/articles/realtime.md
+++ b/articles/realtime.md
@@ -18,41 +18,32 @@ author: Jackie Kay
 
 Original Author: {{ page.author }}
 
-Robotic systems need to be responsive. In mission critical applications, a delay of less than a
-millisecond in the system can cause a catastrophic failure. For ROS 2.0 to capture the needs
-of the robotics community, the core software components must not interfere with the requirements of real-time computing.
+Robotic systems need to be responsive.
+In mission critical applications, a delay of less than a millisecond in the system can cause a catastrophic failure.
+For ROS 2.0 to capture the needs of the robotics community, the core software components must not interfere with the requirements of real-time computing.
 
 # Definition of Real-time Computing
 
-Real-time software guarantees correct computation at the correct time. Hard real-time software systems
-have a set of strict deadlines, and missing a deadline is considered a failure. Soft real-time systems
-degrade their quality of service if a deadline is missed.
+Real-time software guarantees correct computation at the correct time.
+Hard real-time software systems have a set of strict deadlines, and missing a deadline is considered a failure.
+Soft real-time systems degrade their quality of service if a deadline is missed.
 
-Real-time computer systems are often associated with low-latency systems. Many applications of real-time
-computing are also low-latency applications (for example, automated piloting systems must be reactive to
-sudden changes in the environment). However, it is
-generally agreed upon that a real-time system is not defined by low latency, but by a deterministic schedule:
-it must be guaranteed that the system finishes a certain task by a certain time. Therefore, it is
-important that the latency in the system be measurable and a maximum allowable latency for tasks be
-set.
+Real-time computer systems are often associated with low-latency systems.
+Many applications of real-time computing are also low-latency applications (for example, automated piloting systems must be reactive to sudden changes in the environment).
+However, it is generally agreed upon that a real-time system is not defined by low latency, but by a deterministic schedule: it must be guaranteed that the system finishes a certain task by a certain time.
+Therefore, it is important that the latency in the system be measurable and a maximum allowable latency for tasks be set.
 
-A more useful metric in evaluating the "hardness" of a real-time system is jitter, which is defined
-as the variable deviation between a task's deadline and its actual time of completion. A more jittery
-system is less deterministic, less predictable, and less real-time. Though in practice it is impossible
-to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound
-for jitter.
+A more useful metric in evaluating the "hardness" of a real-time system is jitter, which is defined as the variable deviation between a task's deadline and its actual time of completion.
+A more jittery system is less deterministic, less predictable, and less real-time.
+Though in practice it is impossible to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound for jitter.
 
 # Implementation of Real-time Computing
 
-In general, an operating system can guarantee that the tasks it handles for the developer, such as
-thread scheduling, are deterministic, but the OS may not guarantee that the developer's code will
-run in real-time. Therefore, it is up to the developer to know what the determinstic guarantees
-of an existing system are, and what she must do to write hard real-time code on top of the OS.
+In general, an operating system can guarantee that the tasks it handles for the developer, such as thread scheduling, are deterministic, but the OS may not guarantee that the developer's code will run in real-time.
+Therefore, it is up to the developer to know what the determinstic guarantees of an existing system are, and what she must do to write hard real-time code on top of the OS.
 
-In this section, various strategies for developing on top of a real-time
-OS are explored, since these strategies might be applicable to ROS 2. The patterns focus on the
-use case of C/C++ development on Linux-based real-time OS's (such as RTLinux), but the general
-concepts are applicable to other platforms.
+In this section, various strategies for developing on top of a real-time OS are explored, since these strategies might be applicable to ROS 2.
+The patterns focus on the use case of C/C++ development on Linux-based real-time OS's (such as RTLinux), but the general concepts are applicable to other platforms.
 
 ## Memory management
 
@@ -68,13 +59,12 @@ concepts are applicable to other platforms.
     memset(dummy, 0, MAX_SAFE_STACK);
 ```
 
-`mlockall` is a Linux system call for locking the process's virtual address space into RAM,
-preventing the memory that will be accessed by the process from getting paged into swap space. Paging
-and accessing paged memory is a nondeterministic operation that should be avoided in real-time computation.
+`mlockall` is a Linux system call for locking the process's virtual address space into RAM, preventing the memory that will be accessed by the process from getting paged into swap space.
+Paging and accessing paged memory is a nondeterministic operation that should be avoided in real-time computation.
 
-This code snippet, when run at the beginning of a thread's lifecycle, ensures that no pagefaults occur
-while the thread is running. `mlockall` locks the stack for the thread. The `memset` call pre-loads each
-block of memory of the stack into the cache, so that no pagefaults will occur when the stack is accessed.
+This code snippet, when run at the beginning of a thread's lifecycle, ensures that no pagefaults occur while the thread is running.
+`mlockall` locks the stack for the thread.
+The `memset` call pre-loads each block of memory of the stack into the cache, so that no pagefaults will occur when the stack is accessed.
 
 ### Allocate dynamic memory pool
 
@@ -98,10 +88,9 @@ block of memory of the stack into the cache, so that no pagefaults will occur wh
   free(buffer);
 ```
 
-It is commonly thought that dynamic memory allocations are not permitted in the real-time code path.
-This code snippet shows how to lock the virtual address space, disallow returning deallocated memory
-to the kernel via `sbrk`, and disable `mmap`. It effectively locks a pool of memory in the heap into RAM,
-which prevents page faults due to `malloc` and `free`.
+It is commonly believed that dynamic memory allocations are not permitted in the real-time code path.
+This code snippet shows how to lock the virtual address space, disallow returning deallocated memory to the kernel via `sbrk`, and disable `mmap`.
+It effectively locks a pool of memory in the heap into RAM, which prevents page faults due to `malloc` and `free`.
 
 Pros:
 
@@ -115,12 +104,9 @@ Cons:
 
 ### Custom fixed allocators for STL containers
 
-An alternative to the above approach is to implement custom allocators for STL containers
-that only allocate memory on the stack.
+An alternative to the above approach is to implement custom allocators for STL containers that only allocate memory on the stack.
 
-There are various comprehensive tutorials
-[already written](http://www.codeguru.com/cpp/article.php/c18503/C-Programming-Stack-Allocators-for-STL-Containers.htm)
-for this task.
+There are various comprehensive tutorials [already written](http://www.codeguru.com/cpp/article.php/c18503/C-Programming-Stack-Allocators-for-STL-Containers.htm) for this task.
 
 Pros:
 * Use existing STL code with deterministic computation and without allocating global dynamic memory pool
@@ -131,62 +117,50 @@ Cons:
 
 ### Global variables and (static) arrays
 
-Global variables are preallocated at the start of a process, thus assigning and accessing them
-is "real-time" safe. However, this strategy comes with the many disadvantages of using global
-variables.
+Global variables are preallocated at the start of a process, thus assigning and accessing them is real-time safe.
+However, this strategy comes with the many disadvantages of using global variables.
 
 ### Use inheritance sparingly
 
 Classes with many levels of inheritance may not be real-time safe because of vtable overhead access.
-When executing an inherited function, the program needs to access the data used in the function,
-the vtable for the class, and the instructions for the function, which are all stored in different
-parts of memory, and may or may not be stored in cache together.
+When executing an inherited function, the program needs to access the data used in the function, the vtable for the class, and the instructions for the function, which are all stored in different parts of memory, and may or may not be stored in cache together.
 
 In general, C++ patterns with poor cache locality are not well-suited to real-time environments.
-Another such pattern is the opaque pointer idiom (PIMPL), which is convenient for ABI compatibility
-and speeding up compile times. However, bouncing between the memory location for the object and its
-private data pointer causes the cache to "spill" as it loads one chunk of memory and then another,
-unrelated chunk for almost every function in the PIMPLized object.
+Another such pattern is the opaque pointer idiom (PIMPL), which is convenient for ABI compatibility and speeding up compile times.
+However, bouncing between the memory location for the object and its private data pointer causes the cache to "spill" as it loads one chunk of memory and then another, unrelated chunk for almost every function in the PIMPLized object.
 
 ### Exceptions
 
-Throwing an exception can put large objects on the stack, which is undesirable in real-time
-programming since we cannot allocate memory on the heap. On modern C++ compilers, catching an
-exception has little memory or time overhead, but can lead to unexpected code growth.
+Throwing an exception can put large objects on the stack, which is undesirable in real-time programming since we cannot allocate memory on the heap.
+On modern C++ compilers, catching an exception has little memory or time overhead, but can lead to unexpected code growth.
 
 ## Device I/O
-Interacting with physical devices (disk I/O, printing to the screen, etc.) may introduce
-unacceptable latency in the real-time code path, since the process is often forced to wait
-on slow physical phenomena. Additionally, many I/O calls such as `fopen` result in pagefaults.
+Interacting with physical devices (disk I/O, printing to the screen, etc.) may introduce unacceptable latency in the real-time code path, since the process is often forced to wait on slow physical phenomena.
+Additionally, many I/O calls such as `fopen` result in pagefaults.
 
 Keep disk reads/writes at the beginning or end of the program, outside of the RT code path.
 
-Spin up userspace threads (executing on a different CPU from the RT code) to print output to
-the screen.
+Spin up userspace threads (executing on a different CPU from the RT code) to print output to the screen.
 
 ## Multithreaded Programming and Synchronization
 
 Real-time computation requirements change the typical paradigm of multithreaded programming.
 Program execution may not block asynchronously, and threads must be scheduled deterministically.
-A real-time operating system will fulfill this scheduling requirement, but there are still pitfalls
-for the developer to fall into. Typical assumptions about locking critical sections will fall apart
-if thread execution can be arbitrarily preempted.
+A real-time operating system will fulfill this scheduling requirement, but there are still pitfalls for the developer to fall into.
+This section provides guidelines for avoiding these pitfalls.
 
 ### Thread creation guidelines
 
-Create threads at the start of the program. This confines the nondeterministic overhead of thread
-allocation to a defined point in the process.
+Create threads at the start of the program.
+This confines the nondeterministic overhead of thread allocation to a defined point in the process.
 
 Create high priority (but not 99) threads with a FIFO or Round Robin scheduler.
 
 ### Avoid priority inversion
 Priority inversion can occur on a system with a preemptive task scheduler and results in deadlock.
-It occurs when: a low-priority task acquires a lock and is then pre-empted by a
-medium-priority task, then a high-priority task acquires the lock held by the low-priority task.
+It occurs when: a low-priority task acquires a lock and is then pre-empted by a medium-priority task, then a high-priority task acquires the lock held by the low-priority task.
 
-The three tasks are stuck in a triangle: the high-priority task is blocked on the low-priority task,
-which is blocked on the medium-priority task because it was preempted by a task with a higher
-priority, and the medium-priority task is also blocked on a task with a higher priority.
+The three tasks are stuck in a triangle: the high-priority task is blocked on the low-priority task, which is blocked on the medium-priority task because it was preempted by a task with a higher priority, and the medium-priority task is also blocked on a task with a higher priority.
 
 Here are the some solutions to priority inversion:
 
@@ -197,48 +171,41 @@ Here are the some solutions to priority inversion:
 
 ### Timing shots
 
-One real-time synchronization technique is when a thread calculates its next "shot" (the start of its
-next execution period). For example, if a thread is required
-to provide an update every 10 milliseconds, and it must complete an operation that takes 3-6 milliseconds,
-the thread should get the time before the operation, do the operation, and then wait for the remaining
-7-4 milliseconds, based on the time measured after the operation.
+One real-time synchronization technique is when a thread calculates its next "shot" (the start of its next execution period).
+For example, if a thread is required to provide an update every 10 milliseconds, and it must complete an operation that takes 3-6 milliseconds, the thread should get the time before the operation, do the operation, and then wait for the remaining 7-4 milliseconds, based on the time measured after the operation.
 
-The most important consideration for the developer is to use a high precision timer, such as `nanosleep`
-on Linux platforms, while waiting. Otherwise the system will experience drift.
+The most important consideration for the developer is to use a high precision timer, such as `nanosleep` on Linux platforms, while waiting.
+Otherwise the system will experience drift.
 
 ### Spinlocks
-Spinlocks tend to cause clock drift. The developer should avoid implementing his own spinlocks. The
-RT Preempt patch replaces much of the kernel's spinlocks with mutexes, but this might not be
-guaranteed on all platforms.
+Spinlocks tend to cause clock drift.
+The developer should avoid implementing his own spinlocks.
+The RT Preempt patch replaces much of the kernel's spinlocks with mutexes, but this might not be guaranteed on all platforms.
 
 # Testing and Performance Benchmarking
 
-## Static code checking
-
 ## cyclictest
 
-cyclictest is a simple Linux command line tool for measuring the jitter of a real-time environment. It
-takes as input a number of threads, a priority for the threads, and a scheduler type. It spins up
-`n` threads that sleep regular intervals (the sleep period can also be spceified from the command line).
+`cyclictest` is a simple Linux command line tool for measuring the jitter of a real-time environment.
+It takes as input a number of threads, a priority for the threads, and a scheduler type.
+It spins up `n` threads that sleep regular intervals (the sleep period can also be specified from the command line).
 
-
+For each thread, `cyclictest` measures the time between when the thread is supposed to wake up and when it actually wakes up.
+This statistic is the scheduling jitter in the system.
+If there are processes with non-deterministic blocking behavior running in the system, the jitter will grow to a large number (on the order of milliseconds), since the scheduler cannot meet the deadlines of the periodically sleeping threads profiled in the program.
+An ideal real-time system will have an average scheduling jitter on the order of nanoseconds or tens of nanoseconds.
 
 ## Test Pipeline
 
-Instrumenting real-time code to validate its correctness may be tricky because the time overhead due to
-instrumentation. A simple benchmarking test that does not involve code instrumentation is as follows:
+Instrumenting real-time code to validate its correctness may be tricky because the time overhead due to instrumentation.
+A simple benchmarking test that does not involve code instrumentation is as follows:
 
 On an RTLinux machine, start script that runs the process to be tested for some fixed time interval.
-Collect data with cyclictest to record the jitter in the system. A variety of command line tools could be
-used to add extra stress to the system in a controlled way, such as
-[cache calibrator](http://homepages.cwi.nl/~manegold/Calibrator/) to pollute the cache in an attempt to
-create pagefaults, or `fping`, which generates a large number of interrupts. Tools for adding load to the
-system could also be used to test QoS settings in DDS and how that affects real-time performance.
+Collect data with cyclictest to record the jitter in the system.
+A variety of command line tools could be used to add extra stress to the system in a controlled way, such as [cache calibrator](http://homepages.cwi.nl/~manegold/Calibrator/) to pollute the cache in an attempt to create pagefaults, or `fping`, which generates a large number of interrupts.
+Tools for adding load to the system could also be used to test QoS settings in DDS and how that affects real-time performance.
 
-This procedure could be run on a suite of example cases: a simple benchmark program known to run in real
-time, inter-process DDS communication, intra-process DDS communication,
-inter-process ROS 2 communication, intra-process ROS 2 communication, another benchmark program that
-obviously does not run in real time (lots of disk I/O, dynamic memory allocation, and blocking).
+This procedure could be run on a suite of example cases: a simple benchmark program known to run in real time, inter-process DDS communication, intra-process DDS communication, inter-process ROS 2 communication, intra-process ROS 2 communication, another benchmark program that obviously does not run in real time (lots of disk I/O, dynamic memory allocation, and blocking).
 
 TODO: Latency test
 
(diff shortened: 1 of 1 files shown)


articles/realtime.md
```diff
@@ -0,0 +1,293 @@
...
+system is less deterministic, less predictable, and less real-time. Though in practice it is impossible
+to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound
+for jitter.
+
+# Implementation of Real-time Computing
+
+In general, an operating system can guarantee that the tasks it handles for the developer, such as
+thread scheduling, are deterministic, but the OS may not guarantee that the developer's code will
+run in real-time. Therefore, it is up to the developer to know what the determinstic guarantees
+of an existing system are, and what she must do to write hard real-time code on top of the OS.
```
scpeters: she :+1:

dirk-thomas:
In general please start each sentence on a new line. It simply prevents changes to one location to propagate through a whole paragraph and result in huge diffs.

articles/realtime.md
```diff
@@ -0,0 +1,293 @@
...
+generally agreed upon that a real-time system is not defined by low latency, but by a deterministic schedule:
+it must be guaranteed that the system finishes a certain task by a certain time. Therefore, it is
+important that the latency in the system be measurable and a maximum allowable latency for tasks be
+set.
+
+A more useful metric in evaluating the "hardness" of a real-time system is jitter, which is defined
+as the variable deviation between a task's deadline and its actual time of completion. A more jittery
+system is less deterministic, less predictable, and less real-time. Though in practice it is impossible
+to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound
+for jitter.
```
dirk-thomas: Can we just reference existing definition(s) of real-time rather than describing it on our own?

articles/realtime.md
```diff
@@ -0,0 +1,293 @@
...
+
+```c
+    if (mlockall(MCL_CURRENT|MCL_FUTURE) == -1) {
+            perror("mlockall failed");
+            exit(-2);
+    }
+    unsigned char dummy[MAX_SAFE_STACK];
+
+    memset(dummy, 0, MAX_SAFE_STACK);
+``` `
```
dirk-thomas: Markdown rendering is broken as of here.

articles/realtime.md
```diff
@@ -0,0 +1,293 @@
...
+  buffer = malloc(SOMESIZE);
+
+  for (i=0; i < SOMESIZE; i+=page_size)
+  {
+      buffer[i] = 0;
+  }
+  free(buffer);
+``` `
+
+It is commonly thought that dynamic memory allocations are not permitted in the real-time code path.
```
dirk-thomas: common~~ly~~?

articles/realtime.md
```diff
@@ -0,0 +1,293 @@
...
+Spin up userspace threads (executing on a different CPU from the RT code) to print output to
+the screen.
+
+## Multithreaded Programming and Synchronization
+
+Real-time computation requirements change the typical paradigm of multithreaded programming.
+Program execution may not block asynchronously, and threads must be scheduled deterministically.
+A real-time operating system will fulfill this scheduling requirement, but there are still pitfalls
+for the developer to fall into. Typical assumptions about locking critical sections will fall apart
+if thread execution can be arbitrarily preempted.
```
dirk-thomas: Is it referring to priority inversion? Because even in non real time systems the thread execution can be arbitrarily preempted.

articles/realtime.md
```diff
@@ -0,0 +1,293 @@
...
+
+* Implement a new real-time stack (rclrt, rmwrt, etc.) designed with real-time computing in mind.
+  * Pros:
+    * Easier to design and maintain.
+    * Real-time code is "quarantined" from existing code. Can fully optimize library for real-time application.
+  * Cons:
+    * More packages to write and maintain.
+    * Potentially less convenient for the user.
+
+* Ensure option real-time safety up to a certain point in the stack, and implement a real-time safe language wrapper
```
dirk-thomas: optional?

articles/realtime.md
wjwwood: Commonly is correct. She could reword it, but this is a valid sentence. It's similar to the phrase 'it is a commonly held belief...'.

dirk-thomas: Ok, sounded weird for the non-native. Just keep it as is.

articles/realtime.md
jackieokay: I think I'm trying to say that unexpected errors like priority inversion can occur if the developer switches to using a fully preemptive scheduler and does not change the concurrency model of existing code at all.

dirk-thomas: The point you are aiming for is well covered in the next paragraphs. May be the sentence can be replaces with just pointing to the following guidelines to avoid those pitfalls?

articles/realtime.md
jackieokay: yep. the syntax highlighting was screwed up in my editor, so I stupidly added another backtick. maybe I should switch to sublime.  7d77b05

articles/realtime.md
jackieokay: ok, I removed the sentence in 7d77b05

articles/realtime.md
jackieokay: I tried to summarize the existing definition in this document. Do you prefer a verbatim quote from an existing document?

articles/realtime.md
jackieokay: 67b1a16

articles/realtime.md
dirk-thomas: I would just reference external resources. Less text = less to maintain.

wjwwood: Dissenting opinion: The reference text shouldn't be changing and it is annoying to have to open and read another (potentially large) document to get the meaning. I think it is totally appropriate to quote a snippet of the referenced text or summarize it in this document to make it more self contained. Part of the job of the document is to distill the subject matter into an easy to digest form.

scpeters: dissent +1: I think it is a nice summary

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
```
adolfo-rt, line 20: It would be good to start a summary of the document, stating its scope, goal and structure.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Robotic systems need to be responsive.
+In mission critical applications, a delay of less than a millisecond in the system can cause a catastrophic failure.
```
adolfo-rt, line 25: Move this phrase below, to the definition of real-time. See comments on hard, firm, soft real-time.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+For ROS 2.0 to capture the needs of the robotics community, the core software components must not interfere with the requirements of real-time computing.
+
+# Definition of Real-time Computing
+
+Real-time software guarantees correct computation at the correct time.
```
adolfo-rt, line 36: Maybe introduce key terms in the definition, like _determinism_ and _deadline_. Something like:

real-time software meets computational _deadlines_ ina _deterministic_ way. The correctness depends not only on the result of a computation, but also on the time it was delivered. Thus, failure to respond is as bad as the wrong response.

Paraphrased from http://www.cse.unsw.edu.au/~cs9242/08/lectures/09-realtimex2.pdf

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Hard real-time software systems have a set of strict deadlines, and missing a deadline is considered a failure.
```
adolfo-rt: I like how [this answer](http://stackoverflow.com/questions/17308956/differences-between-hard-real-time-soft-real-time-and-firm-real-time/21903622#21903622) explains hard and soft (and firm, not so used) real-time. It's simple and clear. Can we tolerate missing a deadline?, are results useful after a missed deadline?.

Here would be a good place to present example of systems which are hard or soft real-time.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Soft real-time systems degrade their quality of service if a deadline is missed.
+
+Real-time computer systems are often associated with low-latency systems.
+Many applications of real-time computing are also low-latency applications (for example, automated piloting systems must be reactive to sudden changes in the environment).
+However, it is generally agreed upon that a real-time system is not defined by low latency, but by a deterministic schedule: it must be guaranteed that the system finishes a certain task by a certain time.
```
adolfo-rt: Use an alternative to 'generally agreed'. It gives the impression that this is a subjective statement, based on the opinion of a majority.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Therefore, it is important that the latency in the system be measurable and a maximum allowable latency for tasks be set.
+
+A more useful metric in evaluating the "hardness" of a real-time system is jitter, which is defined as the variable deviation between a task's deadline and its actual time of completion.
+A more jittery system is less deterministic, less predictable, and less real-time.
+Though in practice it is impossible to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound for jitter.
+
+# Implementation of Real-time Computing
```
adolfo-rt: What do you think about rewording the title to something like 'Writing real-time safe code'. From reading the above title, I'm not sure if it refers to writing an RTOS or user code.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+Real-time computer systems are often associated with low-latency systems.
+Many applications of real-time computing are also low-latency applications (for example, automated piloting systems must be reactive to sudden changes in the environment).
+However, it is generally agreed upon that a real-time system is not defined by low latency, but by a deterministic schedule: it must be guaranteed that the system finishes a certain task by a certain time.
+Therefore, it is important that the latency in the system be measurable and a maximum allowable latency for tasks be set.
+
+A more useful metric in evaluating the "hardness" of a real-time system is jitter, which is defined as the variable deviation between a task's deadline and its actual time of completion.
+A more jittery system is less deterministic, less predictable, and less real-time.
+Though in practice it is impossible to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound for jitter.
+
```
adolfo-rt, line 58: Before the next section, it could be mentioned that a system that meets real-time requirements needs both an operating system (kernel, scheduler) **and** user code capable of delivering deterministic execution.

It might be out of scope, but a comment on RTOS alternatives could be interesting. It would be the complement to the next section on writing real-time safe user code. In the Linux world, it's worth mentioning:
- Linux kernel patches like [PREEMPT_RT ](https://rt.wiki.kernel.org/index.php/Main_Page) which '...allows nearly all of the kernel to be preempted'.
- Hard real-time co-kernels (also called hypervisors) that cooperate with the Linux kernel like [Xenomai](http://xenomai.org/) (POSIX compliant) and [RTAI](https://www.rtai.org/). They treat Linux as the idle task of the real-time scheduler.

There are also RTOS for mission-critical and safety-critical systems, like [QNX Neutrino](http://www.qnx.com/products/neutrino-rtos/neutrino-rtos.html) (fully POSIX compliant) and [VxWorks](http://www.windriver.com/vxworks/reinvented/?gclid=COrv7YXrtcYCFWv3wgodfEcKqA) (POSIX compliant through compatibility layer) that are certified to IEC 61508 SIL 3.

For all the above RTOS except Linux+PREEMPT_RT, one must generate binaries that are not compatible with standard Linux, which might be an inconvenience.

adolfo-rt:
General comment. How do you feel about citing references in the running text?. It can quickly become hard to identify where the reference is used in the text.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+    if (mlockall(MCL_CURRENT|MCL_FUTURE) == -1) {
+            perror("mlockall failed");
+            exit(-2);
+    }
+    unsigned char dummy[MAX_SAFE_STACK];
+
+    memset(dummy, 0, MAX_SAFE_STACK);
+```
+
+`mlockall` is a Linux system call for locking the process's virtual address space into RAM, preventing the memory that will be accessed by the process from getting paged into swap space.
```
adolfo-rt: [`mlockall`](http://linux.die.net/man/2/mlockall) hyperlink or [ref], for convenient jumping to the source?.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Paging and accessing paged memory is a nondeterministic operation that should be avoided in real-time computation.
+
+This code snippet, when run at the beginning of a thread's lifecycle, ensures that no pagefaults occur while the thread is running.
+`mlockall` locks the stack for the thread.
+The `memset` call pre-loads each block of memory of the stack into the cache, so that no pagefaults will occur when the stack is accessed.
```
adolfo-rt: [`memset`](http://linux.die.net/man/3/memset)

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Allocate dynamic memory pool
```
adolfo-rt, line 106: Before going into details of the solutions, mention what is the problem with dynamic memory allocation. This [Wikipedia](https://en.wikipedia.org/wiki/Real-time_operating_system#Memory_allocation) entry provides a good summary. It mentions issues with memory fragmentation, and the non-deterministic nature of standard memory allocators.

Generally speaking, I would recommend strategies that depend on the problem at hand:
- Data size is known at complile-time: Use fixed objects allocated on the stack (if reasonably sized).
- Data size is known at runtime, during configuration, before starting RT-scheduled execution. Pre-allocate resources on the heap. Node lifecycle becomes relevant here, as we can clearly state in which transitions a real-time safe node can perform non real-time safe calls.
- Data size is known at (or can change during) RT-scheduled execution. Multiple alternatives exist, which is mostly what you focus on below. Typical strategies are to use object pools (good if you have some context of the problem), general-purpose memory pools like the [TLSF](http://www.gii.upv.es/tlsf/) O(1) allocator, stack allocation, etc. In such cases there's typically a buffer that can wither get full or overflow, so AFAIK, there's no silver bullet.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+  buffer = malloc(SOMESIZE);
+
+  for (i=0; i < SOMESIZE; i+=page_size)
+  {
+      buffer[i] = 0;
+  }
+  free(buffer);
+```
+
+It is commonly believed that dynamic memory allocations are not permitted in the real-time code path.
```
adolfo-rt: Reword 'commonly believed', as it again gives the impression that this is an opinionated statement. If the default allocator is non-deterministic or subject to causing page faults, then it cannot be used in the real-time path. Guidelines for critical system code like [MISRA C](http://www.misra.org.uk/MISRAHome/MISRAC2012/tabid/196/Default.aspx) (paywalled) go as far as recommending not using `malloc`, ever (!!!), because it's too implementation dependent, and very little guarantees can be made on its behavior.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+Cons:
+
+* Must accurately predict bounded memory size for the process!
+* Using STL containers is therefore dangerous (unbounded sizes)
+* In practice, only works for small processes
+
+### Custom fixed allocators for STL containers
+
+An alternative to the above approach is to implement custom allocators for STL containers that only allocate memory on the stack.
```
adolfo-rt: Custom allocators don't necessarily need to allocate on the stack, although that is one solution to the problem.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+Cons:
+* Complex to implement
+
+### Global variables and (static) arrays
+
+Global variables are preallocated at the start of a process, thus assigning and accessing them is real-time safe.
+However, this strategy comes with the many disadvantages of using global variables.
+
+### Use inheritance sparingly
```
adolfo-rt: This subsection talks first about inheritance, then about PIMPL. The common subject is cache-friendliness. Maybe the title should be changed accordingly for consistency.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Classes with many levels of inheritance may not be real-time safe because of vtable overhead access.
+When executing an inherited function, the program needs to access the data used in the function, the vtable for the class, and the instructions for the function, which are all stored in different parts of memory, and may or may not be stored in cache together.
+
+In general, C++ patterns with poor cache locality are not well-suited to real-time environments.
+Another such pattern is the opaque pointer idiom (PIMPL), which is convenient for ABI compatibility and speeding up compile times.
+However, bouncing between the memory location for the object and its private data pointer causes the cache to "spill" as it loads one chunk of memory and then another, unrelated chunk for almost every function in the PIMPLized object.
+
+### Exceptions
+
+Throwing an exception can put large objects on the stack, which is undesirable in real-time programming since we cannot allocate memory on the heap.
```
adolfo-rt: I don't really get this statement (stack vs. heap). My understanding is that when no exceptions are thrown, the overhead is small, but exception handling (stack unwinding, catching) can add a significant penalty. FWIW, this [stack overflow](http://stackoverflow.com/a/2226262) answer states something similar.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+On modern C++ compilers, catching an exception has little memory or time overhead, but can lead to unexpected code growth.
+
+## Device I/O
+Interacting with physical devices (disk I/O, printing to the screen, etc.) may introduce unacceptable latency in the real-time code path, since the process is often forced to wait on slow physical phenomena.
+Additionally, many I/O calls such as `fopen` result in pagefaults.
+
+Keep disk reads/writes at the beginning or end of the program, outside of the RT code path.
+
+Spin up userspace threads (executing on a different CPU from the RT code) to print output to the screen.
```
adolfo-rt: It should be enough to perform the printing from a non-rt scheduled thread. No actual need for CPU isolation.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+Program execution may not block asynchronously, and threads must be scheduled deterministically.
+A real-time operating system will fulfill this scheduling requirement, but there are still pitfalls for the developer to fall into.
+This section provides guidelines for avoiding these pitfalls.
+
+### Thread creation guidelines
+
+Create threads at the start of the program.
+This confines the nondeterministic overhead of thread allocation to a defined point in the process.
+
+Create high priority (but not 99) threads with a FIFO or Round Robin scheduler.
```
adolfo-rt: or `SCHED_DEADLINE`, since version 3.14. [Source](http://man7.org/linux/man-pages/man7/sched.7.html).

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Avoid priority inversion
+Priority inversion can occur on a system with a preemptive task scheduler and results in deadlock.
+It occurs when: a low-priority task acquires a lock and is then pre-empted by a medium-priority task, then a high-priority task acquires the lock held by the low-priority task.
```
adolfo-rt, line 225: The wording is a bit confusing, in particular, the high-prio task does not acquire the lock (but attempts and fails). [This example](https://en.wikipedia.org/wiki/Priority_inversion#Example_of_a_priority_inversion) is longer, but the wording is clearer.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+The three tasks are stuck in a triangle: the high-priority task is blocked on the low-priority task, which is blocked on the medium-priority task because it was preempted by a task with a higher priority, and the medium-priority task is also blocked on a task with a higher priority.
+
+Here are the some solutions to priority inversion:
+
+* Don't use locks (but sometimes, they are necessary)
```
adolfo-rt: Is this meant as don't use any kind of synchronization primitives?.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+* Disable preemption for tasks holding locks (can lead to jitter)
+* Increase priority of task holding a lock
+* Use priority inheritance: a task that owns a lock inherits the priority of a task that tries to acquire the lock
+
```
adolfo-rt, line 236: Another alternative: Use lock-free data structures. They don't have blocking semantics and guarantee forward progress of the active thread. On the flipside, their performance is typically poorer than locks, and can have a larger memory footprint (because of multiple versioned objects being tracked). A [very nice read on the subject](http://www.1024cores.net/home/lock-free-algorithms).

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+### Spinlocks
+Spinlocks tend to cause clock drift.
+The developer should avoid implementing his own spinlocks.
+The RT Preempt patch replaces much of the kernel's spinlocks with mutexes, but this might not be guaranteed on all platforms.
+
+# Testing and Performance Benchmarking
+
+## cyclictest
+
+`cyclictest` is a simple Linux command line tool for measuring the jitter of a real-time environment.
```
adolfo-rt: [`cyclictest`](http://manpages.ubuntu.com/manpages/trusty/man8/cyclictest.8.html)

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+An ideal real-time system will have an average scheduling jitter on the order of nanoseconds or tens of nanoseconds.
+
+## Test Pipeline
+
+Instrumenting real-time code to validate its correctness may be tricky because the time overhead due to instrumentation.
+A simple benchmarking test that does not involve code instrumentation is as follows:
+
+On an RTLinux machine, start script that runs the process to be tested for some fixed time interval.
+Collect data with cyclictest to record the jitter in the system.
+A variety of command line tools could be used to add extra stress to the system in a controlled way, such as [cache calibrator](http://homepages.cwi.nl/~manegold/Calibrator/) to pollute the cache in an attempt to create pagefaults, or `fping`, which generates a large number of interrupts.
```
adolfo-rt: [`fping`](http://linux.die.net/man/8/fping)

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+TODO: Latency test
+
+TODO: Test schemes for different embedded platforms
+
+# Design Guidelines for ROS 2
+
+## Achieving real-time computation across platforms
+
+TODO: How can ROS 2 be real-time friendly and cross-platform?
+Much of the research in this document focuses on achieving real-time on Linux systems with pthreads.
```
adolfo-rt: pthreads → POSIX

articles/realtime.md
adolfo-rt: The RTOS we have more experience with is Xenomai. It comes with its own [testing scripts](http://xenomai.org/2014/07/benchmarking-with-xeno-test/).

There's also the [stress](http://manpages.ubuntu.com/manpages/lucid/man1/stress.1.html) package for subjecting a system to different kinds of load.

adolfo-rt: Also, using Xenomai, it's possible to detect overruns (SIGXCPU signal) and [relaxes](http://xenomai.org/2014/06/finding-spurious-relaxes/) (SIGDEBUG signal), which can prove useful for automated performance testing.

adolfo-rt: Although setting up a Xenomai system is more complex than an RT_PREEMPT one, and you need to generate different binaries (that link against Xenomai libs) there's a clear separation between primary and secondary modes, and it's reasonably easy to detect non real-time safe code. With PREEMPT_RT, one can leverage all the typical Linux tooling, which is great, but I've found it more challenging to detect real-time violations. My experience with PREEMPT_RT is limited, so the problem might be on my end.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+A more jittery system is less deterministic, less predictable, and less real-time.
+Though in practice it is impossible to completely eliminate jitter from a real-time system, it is a worthy goal to determine a hard upper bound for jitter.
+
+# Implementation of Real-time Computing
+
+In general, an operating system can guarantee that the tasks it handles for the developer, such as thread scheduling, are deterministic, but the OS may not guarantee that the developer's code will run in real-time.
+Therefore, it is up to the developer to know what the determinstic guarantees of an existing system are, and what she must do to write hard real-time code on top of the OS.
+
+In this section, various strategies for developing on top of a real-time OS are explored, since these strategies might be applicable to ROS 2.
+The patterns focus on the use case of C/C++ development on Linux-based real-time OS's (such as RTLinux), but the general concepts are applicable to other platforms.
```
adolfo-rt: Just to be sure, by RTLinux are you referring to the microkernel that was maintained by [WindRiver](https://en.wikipedia.org/wiki/RTLinux) until 2011, or the PREEMPT_RT patch?.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+A variety of command line tools could be used to add extra stress to the system in a controlled way, such as [cache calibrator](http://homepages.cwi.nl/~manegold/Calibrator/) to pollute the cache in an attempt to create pagefaults, or `fping`, which generates a large number of interrupts.
+Tools for adding load to the system could also be used to test QoS settings in DDS and how that affects real-time performance.
+
+This procedure could be run on a suite of example cases: a simple benchmark program known to run in real time, inter-process DDS communication, intra-process DDS communication, inter-process ROS 2 communication, intra-process ROS 2 communication, another benchmark program that obviously does not run in real time (lots of disk I/O, dynamic memory allocation, and blocking).
+
+TODO: Latency test
+
+TODO: Test schemes for different embedded platforms
+
+# Design Guidelines for ROS 2
```
adolfo-rt, line 285: No brain left for this part. Generally speaking, the typical things that potentially need special codepaths to be compatible with an RTOS are:
- Synchonization primitives
- Threads
- Clock access, especially getting (Linux) system timestamps.

Also, an RT-friendly string class is invaluable for logging purposes.

articles/realtime.md
```diff
@@ -0,0 +1,259 @@
...
+
+Real-time computation requirements change the typical paradigm of multithreaded programming.
+Program execution may not block asynchronously, and threads must be scheduled deterministically.
+A real-time operating system will fulfill this scheduling requirement, but there are still pitfalls for the developer to fall into.
+This section provides guidelines for avoiding these pitfalls.
+
+### Thread creation guidelines
+
+Create threads at the start of the program.
+This confines the nondeterministic overhead of thread allocation to a defined point in the process.
```
adolfo-rt, line 219: A note based on past experience. One must avoid any calls that can lead to `fork()` from a _process_ running real-time scheduled threads. Some time ago we had this witch hunt in which a _non real-time_ thread was loading `pluginlib` plugins and wrecking havoc on the real-time control thread. It turns out that `fork()` is lazy and does [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) of memory pages. A workaround was applied [here](https://github.com/ros/rospack/pull/34).

gavanderhoorn:
+1 on all comments by @adolfo-rt, especially those about what appear to be opinions. The RT-field has some very clear definitions this article could re-use, so why not do that.

Also +1 on mentioning Xenomai and all its tools and infrastructure. This may be personal, but I like the co-kernel approach - and the strict separation between primary and secondary modes that it allows - as well.

articles/realtime.md
jackieokay, line 58: yeah, it would be great to target ROS 2 to perform in real-time on a variety of RTOS's/real-time environments and set up several machines for performance testing. I have been testing with a machine with the `PREEMPT_RT`kernel installed, hence the focus on that platform in this document.

articles/realtime.md
jackieokay: The `PREEMPT_RT` (`RT_PREEMPT`?) patch. We are using a version maintained by IHMC that is available in [this PPA](https://launchpad.net/~ihmcrobotics/+archive/ubuntu/ppa). I can clarify this in the text.

articles/realtime.md
jackieokay, line 219: Interesting story and nice solution! I will add an item here about avoiding `fork` and anything with `fork` in its call stack.

jackieokay:
Thanks for all the comments, @adolfo-rt! I am addressing them in the text now.

The Markdown/Jekyll system we have set up for these documents doesn't support "proper" inline citations, but I'll probably figure out a numbered footnote system for citations, or just add links to sources in the text. We could consider adding more citations to the other ROS 2 design documents, too, but this article in particular synthesizes a lot of established information that is not common knowledge for the average software engineer.

Thanks too for your feedback, @gavanderhoorn. I'll look into setting up a Xenomai machine for the purposes of ROS 2 performance testing.

gavanderhoorn:
> I'll look into setting up a Xenomai machine for the purposes of ROS 2 performance testing.

@jacquelinekay: perhaps the OSADL can help with that? At the 3rd ROS-Industrial conference one of the presentations included an overview of the kind of software testing they do. Testing real-time systems / applications across a wide variety of HW and SW was part of that. I can't link to the slides, but the [OSADL Realtime QA Farm](http://www.osadl.org/QA-Farm-Realtime.qa-farm-about.0.html) page seems to give a good overview of what he presented.

If you really (or also) want to do it yourself: the MachineKit community has pre-built Xenomai debian kernel pkgs that could save you some work.

articles/realtime.md
adolfo-rt: RT_PREEMPT is correct, my bad.

jackieokay:
@adolfo-rt I've incorporated your suggestions in 0990fa35e1979e5c0b4631a2864c8cfbb71f5d5c and 2a13eeab35d936d32fb617ba0d373e407fdb7ed1.

articles/realtime.md
```diff
@@ -0,0 +1,330 @@
...
+# Definition of Real-time Computing
+
+The definition of real-time computing requires the definition of a few other key terms:
+
+* Determinism: A system is deterministic if it always produces the same output for a known input. The output of a nondeterministic system will have random variations.
+* Deadline: A deadline is the finite window of time in which a certain task must be completed.
+* Quality of Service: The overall performance of a network. Includes factors such as bandwith, throughput, availability, jitter, latency, and error rates.
+
+Real-time software guarantees correct computation at the correct time.
+Hard real-time software systems have a set of strict deadlines, and missing a deadline is considered a system failure.
```
adolfo-rt: Add a line break to separate hard real-time definition to a new paragraph. Also, consider emphasizing the hard, soft, and firm real-time initalics.

articles/realtime.md
```diff
@@ -0,0 +1,330 @@
...
+Firm real-time systems treat information delivered/computations made after a deadline as invalid.
+Like soft real-time systems, they do not fail after a missed deadline, and they may degrade QoS if a deadline is missed. (1)
+Examples of firm real-time systems: financial forecast systems, robotic assembly lines. (2)
+
+Real-time computer systems are often associated with low-latency systems.
+Many applications of real-time computing are also low-latency applications (for example, automated piloting systems must be reactive to sudden changes in the environment).
+However, a real-time system is not defined by low latency, but by a deterministic schedule: it must be guaranteed that the system finishes a certain task by a certain time.
+Therefore, it is important that the latency in the system be measurable and a maximum allowable latency for tasks be set.
+
+A real-time computer system needs both an operating system that can handle scheduling and memory allocation in real-time and user code that delivers deterministic execution.
```
adolfo-rt: I don't think it's necessary to explicitly mention memory allocation here. I would say that handling memory allocations in a real-time safe manner can be considered part of writing real-time safe code.

articles/realtime.md
```diff
@@ -0,0 +1,330 @@
...
+Neither deterministic user code on a non-real-time operating system or nondeterministic code on a real-time operating system will result in real-time performance.
+
+Some examples of real-time environments:
+
+* The `RT_PREEMPT` Linux kernel patch, which modifies the Linux scheduler to be fully preemptible. (3)
+* Xenomai, a POSIX-compliant co-kernel (or hypervisor) that provides a real-time kernel cooperating with the Linux kernel. The Linux kernel is treated as an idle task of the real-time kernel's scheduler.
```
adolfo-rt: as _the_ idle task
(i.e. the one with lowest priority)

articles/realtime.md
```diff
@@ -0,0 +1,330 @@
...
+Cons:
+
+* Platform/implementation dependent
+* Must accurately predict bounded memory size for the process!
+* Using STL containers is therefore dangerous (unbounded sizes)
+* In practice, only works for processes with small memory footprint
+
+### Custom static allocators for STL containers
+
+To use the standard library safely, custom allocators for STL containers must be implemented that only allocate memory on the stack.
```
adolfo-rt: I still don't think that custom allocators must forcibly allocate on the stack to be real-time safe. The [TLSF](http://www.gii.upv.es/tlsf/) allocator cited in the original post is an example. [Brief overview presentation](http://www.gii.upv.es/tlsf/files/papers/tlsf_slides.pdf).

articles/realtime.md
```diff
@@ -0,0 +1,330 @@
...
+
+There are various comprehensive tutorials [already written](http://www.codeguru.com/cpp/article.php/c18503/C-Programming-Stack-Allocators-for-STL-Containers.htm) for this task.
+
+Pros:
+* Use existing STL code with deterministic computation
+* More modular solution
+
+Cons:
+* Complex to implement
```
adolfo-rt: If one leverages an existing allocator, the cons end up being an additional dependency and potentially some extra complexity in setting up the allocator. The latter is allocator-specific, but typical setup parameters are pool size,  or specifying the policy of what to do if pool gets full (allocate more vs. `std::bad_alloc`). More verbose syntax (e.g. specifying allocator in STL container template params) can be alleviated with typedefs.

adolfo-rt:
I added an additional round of minor comments. The document is looking good 👍

articles/realtime.md
jackieokay: thanks for the many in-depth resources. I'll link to this one since many of the finer points of memory allocation are probably out of the scope of this document, but definitely useful for a subtler understanding of safe dynamic memory allocation. And I'll reword this section.

articles/realtime.md
```diff
@@ -0,0 +1,330 @@
...
+Collecting these statistics gives an indication of what events could cause the latency/scheduling jitter measured by the previously described methods.
+
+# Design Guidelines for ROS 2
+
+## Achieving real-time computation across platforms
+
+TODO: How can ROS 2 be real-time friendly and cross-platform?
+Much of the research in this document focuses on achieving real-time on POSIX-compliant Linux systems.
+
+## Implementation strategy
```
adolfo-rt, line 292: Maybe worth mentioning explicitly. Using ROS interfaces should be real-time safe (at least within a limited context, like intra-process), setting them up needs not be. In particular, for pub-sub with dynamic-sized data, it should be possible to pre-allocate resources (e.g., resize payload arrays) at setup time.

jackieokay, line 292: eb93b9a442a9217c90d18e550848e0dc5adeeca5 (and other comments)

gerkey:
+1

wjwwood:
+1

PR #36, opened 2015-06-27
Added draft of an article about QoS

esteve:
Draft of an article about QoS

Comments and questions are more than welcome.

### DIFF
--- a/articles/qos.md
+++ b/articles/qos.md
@@ -1,0 +1,118 @@
+---
+layout: default
+title: ROS 2.0 Quality of Service policies
+abstract:
+  This article describes the approach to provide QoS (Quality of Service) policies for ROS 2.0.
+published: false
+author: '[Esteve Fernandez](https://github.com/esteve)'
+---
+
+* This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+With the advent of inexpensive robots using unreliable wireless networks, developers and users need mechanisms to control how traffic is prioritized across network links.
+
+## Background
+
+ROS 1.x uses TCP as the underlying transport, which is unsuitable for lossy networks such as wireless links.
+With ROS 2.0 relying on DDS which uses UDP as its transport, we can give control over the level of reliability a node can expect and act accordingly.
+
+## DDS Quality of Service policies
+
+DDS provides fine-grained control over the Quality of Service (QoS) setting for each of the entities involved in the system.
+Common entities whose behavior can be modified via QoS settings include: Topic, DataReader, DataWriter, Publisher and Subscriber.
+QoS is enforced based on a Request vs Offerer Model, however Publications and Subscriptions will only match if the QoS settings are compatible.
+
+## ROS 2.0 proposal
+
+Given the complexity of choosing the correct QoS settings for a given scenario, it may make sense for ROS 2.0 to provide a set of predefined QoS profiles for common usecases (e.g. sensor data, real time, etc.), while at the same time give the flexibility to control specific features of the QoS policies for the most common entities.
+
+## QoS profiles
+
+A QoS profile defines a set of policies, including durability, reliability, queue depth and sample history storage.
+The base QoS profile includes settings for the following policies:
+
+* History.
+  * Keep last: only store up to N samples, configurable via the queue depth option.
+  * Keep all: store all samples, subject to the configured resource limits of the DDS vendor.
+* Depth.
+  * Size of the queue: only honored if used together with "keep last".
+* Reliability.
+  * Best effort: attempt to deliver samples, but may lose them if the network is not robust.
+  * Reliable: guarantee that samples are delivered, may retry multiple times.
+* Durability.
+  * Transient local: *only applies to `DataWriter`s*. `DataWriter` becomes responsible of persisting samples until a corresponding `DataReader` becomes available.
+  * Volatile: no attempt is made to persist samples.
+
+Note: for each of the main bullets there is also the option of "system default", which uses whatever setting was defined via the DDS vendor tools (e.g. XML configuration files).
+
+ROS 2.0 will provide QoS profiles based on the following use cases:
+
+* Default QoS settings for publishers and subscriptions.
+
+https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L39
+
+In order to make the transition from ROS1 to ROS2, exercising a similar network behavior is desirable.
+By default, publishers and subscriptions are reliable in ROS2.
+However, many DDS vendors don't support publishing large messages (e.g. images) with reliable connections and the common API.
+These DDS vendors support an alternate API (asynchronous publishers) for which support will be added to ROS2.
+Until the asynchronous API is supported by ROS2, it is advised to use the "best effort" QoS setting for reliability.
+
+* Services.
+
+https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L46
+
+In the same vein as publishers and subscriptions, services are reliable.
+The difference here is that we support some level of durability, so clients can submit requests even before a service is available.
+The durability offered is "transient local", that is, the client's writer will be responsible of persiting the samples until the service can respond to requests. 
+
+* Sensor data.
+
+https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L25
+
+For sensor data, in most cases it's more important to receive readings in a timely fashion, rather than ensuring that all of them arrive.
+That is, developers want the latest samples as soon as they are captured, at the expense of maybe losing some.
+
+Profiles allow developers to focus on their applications without worrying about every QoS setting in the DDS specification.
+
+Note: the values in the profiles are subject to further tweaks, based on the feedback from the community.
+
+## Integration with existing DDS deployments
+
+Both PrismTech OpenSplice and RTI Connext support loading of QoS policies via an external XML file.
+In environments where DDS is already deployed and also to enable more extensibility other than the offered by the ROS 2.0 and the predefined profiles, ROS 2.0 may provide loading of the QoS settings via the same mechanisms the underlying DDS implementations use.
+However, this mechanism will not be added into the common ROS2 API so as to keep the `rmw` layer transport agnostic and let future developers implement it for other transports (e.g. ZeroMQ, TCPROS, etc.)
+To honor the QoS settings of the system, developers can use the `rmw_qos_profile_system_default` QoS profile which delegates the responsibility of the QoS machinery to the underlying DDS vendor.
+This allows developers to deploy ROS2 applications and use DDS vendor tools to configure the QoS settings.
+
+## Open questions
+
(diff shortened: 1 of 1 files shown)


tfoote:
I think this might be a good place to define our standard set of QoS profiles.

esteve:
@tfoote agreed. I'm adding a few to the proposed implementation and experimenting with them to have a sense of what they may look like. So far I have:
- sensor data (last history, depth size 5, best effort)
- parameters (last history, depth size 10, reliable)
- client/service (last history, depth 10, reliable)
- default (all history, depth 10, reliable)

articles/qos.md
```diff
@@ -0,0 +1,57 @@
...
+## Background
+
+ROS 1.x uses TCP as the underlying transport, which is unsuitable for lossy networks such as wireless links.
+With ROS 2.0 relying on DDS which uses UDP as its transport, we can give control over the level of reliability a node can expect and act accordingly.
+
+## DDS Quality of Service policies
+
+DDS provides fine-grained control over the Quality of Service (QoS) setting for each of the entities involved in the system.
+Common entities whose behavior can be modified via QoS settings include: Topic, DataReader, DataWriter, Publisher and Subscriber.
+QoS is enforced based on a Request vs Offerent Model, however Publications and Subscriptions will only match if the QoS settings are compatible.
```
wjwwood: Unless I'm mistaken, in English I believe this is `Offerer` rather than `Offerent`.

wjwwood:
What's there looks good to me, but without more it's hard for me to see where this article is going. Maybe it would help me if you could add some more skeleton structure to the document with TODO's to fill it with details like the different proposed built-in QoS defaults.

articles/qos.md
esteve: Good catch, thanks!

esteve:
@wjwwood yeah, I just used this PR to structure my thoughts a bit. I'm making the API changes in the following PRs and will reflect them in this document:

https://github.com/ros2/rmw_opensplice/pull/42
https://github.com/ros2/rmw/pull/15
https://github.com/ros2/rmw_connext/pull/42

articles/qos.md
```diff
@@ -0,0 +1,57 @@
...
+## ROS 2.0 proposal
+
+Given the complexity of choosing the correct QoS settings for a given scenario, it may make sense for ROS 2.0 to provide a set of predefined QoS profiles for common usecases (e.g. sensor data, real time, etc.), while at the same time give the flexibility to control specific features of the QoS policies for the most common entities.
+
+## Integration with existing DDS deployments
+
+Both PrismTech OpenSplice and RTI Connext support loading of QoS policies via an external XML file.
+In environments where DDS is already deployed and also to enable more extensibility other than the offered by the ROS 2.0 and the predefined profiles, ROS 2.0 may provide loading of the QoS settings via the same mechanisms the underlying DDS implementations use.
+
+## References
```
jackieokay, line 100: +1 for references!

While reviewing the realtime article, Adolfo encouraged me to cite sources inline. I like the idea of mapping information directly to sources, and I think we should come up with a unified way of doing it (if you want to do it in this article).

I see two options for this:
- Provide a link directly in the text
- Footnotes, for example: "DDS provides fine-grained control over QoS (1)", with a list of numbers at the bottom of the document, e.g. "1. Gordon A. Hunt, [DDS - Advanced Tutorial, Using QoS to Solve Real-World Problems](http://www.omg.org/news/meetings/workshops/RT-2007/00-T5_Hunt-revised.pdf)" and maybe with a page number for PDFs, etc.

articles/qos.md
```diff
@@ -0,0 +1,57 @@
...
+
+## DDS Quality of Service policies
+
+DDS provides fine-grained control over the Quality of Service (QoS) setting for each of the entities involved in the system.
+Common entities whose behavior can be modified via QoS settings include: Topic, DataReader, DataWriter, Publisher and Subscriber.
+QoS is enforced based on a Request vs Offerer Model, however Publications and Subscriptions will only match if the QoS settings are compatible.
+
+## ROS 2.0 proposal
+
+Given the complexity of choosing the correct QoS settings for a given scenario, it may make sense for ROS 2.0 to provide a set of predefined QoS profiles for common usecases (e.g. sensor data, real time, etc.), while at the same time give the flexibility to control specific features of the QoS policies for the most common entities.
```
gbiggs, line 36: I like the idea of providing good defaults for different use cases, rather than a global set of defaults expected to be used everywhere. An integrator will be able to say "This topic is carrying range data; I shall use the sensor data QoS set."

BobDean:
What us the reasoning for the chosen depth sizes? And given the history settings, does this mean the dds layer us handling all message queuing, that there will not be a node level "callback queue"?

wjwwood:
@BobDean I think the current defaults are pretty arbitrary, and @esteve is looking into the best defaults for a few cases. DDS is handling all of the queueing, though we might have a ring buffer, whose size is driven by the depth setting, which is used for intra process communications. That's all still up in the air at the moment though.

jackieokay:
Now that the QoS API has been designed and iterated over, it seems like a good time to make this design document more specific so that future changes we make are consistent with each other.

Here are a few questions that have been raised about the current prototype during the review process:
1. What are the default QoS profiles for pub/sub and services and what is the justification for choosing these values?
2. What is the behavior for overriding system defaults for certain QoS profiles? Should it be uniform for all entities? For example, in rmw_connext, the behavior for the `depth` setting for Publishers and Subscriptions is confusing.
3. Should the QoS settings passed to DDS also affect the client library layer (e.g. depth determines the size of ring buffer for intra-process communication)?

esteve:
@jacquelinekay thanks for summarizing the discussion, I'll update the document to address these questions.

esteve:
I've added a section about QoS profiles. I don't know how intraprocess should honor (or if it should) the QoS settings, so I added that as an open question.

Does anyone think it'd worth adding a terminology section explaining every QoS setting implemented in the current API? I think it'd be useful, but I worried this would make the article too long.

tfoote:
> Does anyone think it'd worth adding a terminology section explaining every QoS setting implemented in the current API? I think it'd be useful, but I worried this would make the article too long.

Terminology might be good, but I'd suggest linking to a DDS documentation of QoS definitions, just clearly state which ones we're supporting maybe?

wjwwood:
I actually think this article should stand on it's own and that readers should not have to go to a set of DDS documentation to understand what QoS settings ROS is providing and what they do.

esteve:
I agree with @wjwwood , I prefer documentation that doesn't force me to jump to external resource for core topics. I can add links to the QoS section of the DDS spec in case the readers want to know more.

esteve:
I updated the document removing the QoS profile and instead explained it can store and how the values affect its behavior.

articles/qos.md
```diff
@@ -0,0 +1,102 @@
...
+  * Keep last: only store up to N samples, configurable via the queue depth option.
+  * Keep all: store all samples, subject to the configured resource limits of the DDS vendor.
+* Depth.
+  * Size of the queue: only honored if used together with "keep last".
+* Reliability.
+  * Best effort: attempt to deliver samples, but may lose them if the network is not robust.
+  * Reliable: guarantee that samples are delivered, may retry multiple times.
+* Durability.
+  * Transient local: *only applies to `DataWriter`s*. `DataWriter` becomes responsible of persisting samples until a corresponding `DataReader` becomes available.
+  * Volatile: no attempt is made to persist samples.
```
dirk-thomas, line 53: For each of the main bullets there is also the option of "system default".

esteve, line 53: Added note about this. Thanks.

esteve:
I've rebased this branch and squashed the commits. Is there anything that needs improving?

articles/qos.md
```diff
@@ -0,0 +1,104 @@
...
+
+Note: for each of the main bullets there is also the option of "system default", which uses whatever setting was defined via the DDS vendor tools (e.g. XML configuration files).
+
+ROS 2.0 will provide QoS profiles based on the following use cases:
+
+* Default QoS settings for publishers and subscriptions.
+In order to make the transition from ROS1 to ROS2, exercising a similar network behavior is desirable.
+By default, publishers and subscriptions are reliable in ROS2.
```
dirk-thomas, line 64: Should we add note here that this default value won't work for larger messages for all rmw implementations atm?

esteve, line 64: Good point, will update the document to reflect this.

articles/qos.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Angelo Corsaro, [DDS QoS Unleashed](http://www.slideshare.net/Angelo.Corsaro/dds-qos-unleashed)
+
+* Angelo Corsaro, [The DDS Tutorial Part II](http://www.slideshare.net/Angelo.Corsaro/the-dds-tutorial-part-ii)
+
+* [DDS Spec, section 2.2.3](http://www.omg.org/spec/DDS/1.4/PDF/)
+
+* Heidi Schubert, [QoS policies for configuring reliability](https://community.rti.com/content/forum-topic/qos-policies-configuring-reliability)
+
+* [RTI Connext - Comprehensive Summary of QoS Policies](https://community.rti.com/rti-doc/510/ndds.5.1.0/doc/pdf/RTI_CoreLibrariesAndUtilities_QoS_Reference_Guide.pdf)
```
dirk-thomas: Should we link the current version of the document here (https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/RTI_ConnextDDS_CoreLibraries_QoS_Reference_Guide.pdf)?

esteve: Updated, thanks.

esteve:
Addressed @dirk-thomas 's latest feedback, squashed again and pushed.

articles/qos.md
```diff
@@ -0,0 +1,106 @@
...
+  * Volatile: no attempt is made to persist samples.
+
+Note: for each of the main bullets there is also the option of "system default", which uses whatever setting was defined via the DDS vendor tools (e.g. XML configuration files).
+
+ROS 2.0 will provide QoS profiles based on the following use cases:
+
+* Default QoS settings for publishers and subscriptions.
+In order to make the transition from ROS1 to ROS2, exercising a similar network behavior is desirable.
+By default, publishers and subscriptions are reliable in ROS2.
+However, many DDS vendors don't support publishing large messages (e.g. images) with reliable connections.
```
dirk-thomas: The sentence is a bit misleading. Connext does support publishing large messages with reliable connections. But we have to use a different API for that.

wjwwood:
Todo:
- @esteve to address one more comment
- Link to each current profile with a justification and a note that these are subject to suggestion and feedback.

esteve:
Updated the article with the most recent feedback. I'm going to squash and merge this PR soon unless there are more comments.

dirk-thomas:
Currently the article does not mention the parameter related profiles. Is that intentional? If not, I think especially the depth of 1000 could use some rational.

dirk-thomas:
@esteve Can you please comment on the previous question.

esteve:
The parameters profile is not mentioned in the article because we still haven't officially released the support for parameters. In any case, the value in there is purely arbitrary, the parameters profile has a priori the same rationale as the services one, so it may make sense to use the same values.

dirk-thomas:
Since the parameters are scheduled for the upcoming alpha next week I think the article should be updated to cover the QoS setting for services (and parameters).

Issue #41, opened 2015-07-06
Add article about migration strategy


jackieokay:
Per suggestions from discussion on the sig-ng-ros mailing list, create an article describing the ROS 1->ROS 2 migration strategy. Topics include renaming, build system changes, using the bridge, etc.

PR #43, opened 2015-07-14
add draft of migration guide article

dirk-thomas:
Connects to ros2/design#41

Please provide feedback.

### DIFF
--- a/articles/200_migration_guide_from_ros1.md
+++ b/articles/200_migration_guide_from_ros1.md
@@ -1,0 +1,158 @@
+---
+layout: default
+title: Migration guide from ROS 1
+permalink: articles/migration_guide_from_ros1.html
+abstract:
+  This article describes the high-level steps to migrate a ROS 1 package to ROS 2.
+  It does not aim to be a step-by-step migration instruction.
+published: true
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+---
+
+* This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+
+## Prerequisite
+
+Before being able to migrate a ROS 1 package to ROS 2 all of its dependencies must be available in ROS 2.
+
+
+## Migration steps
+
+### Package naming
+
+Packages containing messages and / or services which should be supported by the [ROS bridge](https://github.com/ros2/ros1_bridge/) need to have a different name in ROS 1 and ROS 2.
+Therefore the suffixes `_msgs` and `_srvs` are replaced (and merged in a single package) with the suffix `_interfaces`.
+
+
+### Package manifests
+
+ROS 2 only support the format 2 of the package specification which is defined in [REP 140](http://www.ros.org/reps/rep-0140.html).
+Therefore the `package.xml` file must be update to format 2 if it uses format 1.
+Since ROS 1 support both formats (1 as well as 2) it is safe to perform that conversion in the ROS 1 package.
+
+Some packages might have different names in ROS 2 so the dependencies might need to be updated accordingly.
+
+
+### Messages and services
+
+Message files must end in `.msg` and must be located in the subfolder `msg`.
+Service files must end in `.srv` and must be located in the subfolder `srv`.
+
+These files might need to be updated to comply with the [ROS Interface definition](http://design.ros2.org/articles/interface_definition.html).
+Some primitive types have been removed and the types `duration` and `time` which were builtin types in ROS 1 have been replaced with normal message definitions and must be used from the `builtin_interfaces` package.
+Also some naming conventions are stricter then in ROS 1.
+
+
+### Build system
+
+The build system in ROS 2 is called [ament](http://design.ros2.org/articles/ament.html).
+
+
+#### Build tool
+
+Instead of using `catkin_make`, `catkin_make_isolated` or `catkin build` ROS 2 uses the command line tool [ament build](https://github.com/ament/ament_tools) to build and install a set of packages.
+
+
+#### Pure Python package
+
+If the ROS 1 package uses CMake only to invoke the `setup.py` file and does not contain anything beside Python code (e.g. also no messages, services, etc.) it should be converted into a pure Python package in ROS 2:
+
+* update the build type in the `package.xml` file:
+
+    <export>
+      <build_type>ament_python</build_type>
+    </export>
+
+* remove the `CMakeLists.txt` file
+* update the `setup.py` file to be a standard Python setup script
+
+ROS 2 supports Python 3 only.
+While each package can choose to also support Python 2 it must invoke executables with Python 3 if it uses any API provided by other ROS 2 packages.
+
+
+#### Update the *CMakeLists.txt* to use *ament_cmake*
+
+Apply the following changes to use `ament_cmake` instead of `catkin`:
+
+* Replace the `find_package` invocation with `catkin` and the `COMPONENTS` with:
+
+    find_package(ament_cmake REQUIRED)
+    find_package(component1 REQUIRED)
+    ...
+    find_package(componentN REQUIRED)
+
+* Move and update the `catkin_package` invocation with:
+
+  * Invoke `ament_package` instead but **after** all targets have been registered.
+  * The only valid argument for [ament_package](https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/cmake/core/ament_package.cmake) is `CONFIG_EXTRAS`.
+    All other arguments are covered by separate functions which all need to be invoked *before* `ament_package`.
+  * The `INCLUDE_DIRS` arguments are passed to the new function [ament_export_include_directories](https://github.com/ament/ament_cmake/blob/master/ament_cmake_export_include_directories/cmake/ament_export_include_directories.cmake).
+  * The `LIBRARIES` arguments are passed to the new functions [ament_export_libraries](https://github.com/ament/ament_cmake/blob/master/ament_cmake_export_libraries/cmake/ament_export_libraries.cmake) if the libraries are targets or [ament_export_library_names](https://github.com/ament/ament_cmake/blob/master/ament_cmake_export_libraries/cmake/ament_export_library_names.cmake) if the library is only identified by the name.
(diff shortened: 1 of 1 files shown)


articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+#### ROS client library
+
+<div class="alert alert-warning" markdown="1">
+NOTE: to be writtten
+</div>
+
+
+## Launch files
+
+While launch files in ROS 1 are specified using [.xml](http://wiki.ros.org/roslaunch/XML) files ROS 2 uses Python scripts to enable more flexibility (see [launch package](https://github.com/ros2/launch/tree/master/launch)).
```
jackieokay, line 158: Maybe give an example here comparing the synatx of xml `roslaunch` and Python `launch`.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+
+Packages containing messages and / or services which should be supported by the [ROS bridge](https://github.com/ros2/ros1_bridge/) need to have a different name in ROS 1 and ROS 2.
+Therefore the suffixes `_msgs` and `_srvs` are replaced (and merged in a single package) with the suffix `_interfaces`.
+
+
+### Package manifests
+
+ROS 2 only support the format 2 of the package specification which is defined in [REP 140](http://www.ros.org/reps/rep-0140.html).
+Therefore the `package.xml` file must be update to format 2 if it uses format 1.
+Since ROS 1 support both formats (1 as well as 2) it is save to perform that conversion in the ROS 1 package.
```
jackieokay: "safe" (not save)

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+### Messages and services
+
+Message files ending in `.msg` must be located in the subfolder `msg`.
+Service files ending in `.srv` must be located in the subfolder `srv`.
+
+These files might need to be updated to comply with the [ROS Interface definition](http://design.ros2.org/articles/interface_definition.html).
+Commonly that involves updating the definition of `Duration` and `Time` fields to the new builtin messages.
+
```
jackieokay, line 54: Should there be a summary of other interface file changes we have recently merged or discussed, like stricter capitalization rules and removal of sequence number?

jackieokay:
Besides some small comments, lgtm.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Build system
+
+#### Build tool
+
+Instead of using `catkin_make`, `catkin_make_isolated` or `catkin build` ROS 2 uses the command line tool [ament build](https://github.com/ament/ament_tools) to build and install a set of packages.
```
tfoote, line 63: We should link here: http://design.ros2.org/articles/ament.html

articles/200_migration_guide_from_ros1.md
tfoote, line 54: We should add that here or in the linked design document.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+#### Messages and services
+
+The namespace of ROS 2 messages and services uses a subnamespace (`msg` or `srv`) after the package name.
+Therefore an include looks like: `#include <my_interfaces/msg/MyMessage.hpp>`.
+The C++ type is then named: `my_interfaces::msg::MyMessage`.
+
+Shared pointer types are provided as typedefs within the message structs: `my_interfaces::msg::MyMessage::SharedPtr` as well as `my_interfaces::msg::MyMessage::ConstSharedPtr`.
+
+For more details please see the article about the [generated C++ interfaces](http://design.ros2.org/articles/generated_interfaces_cpp.html).
+
```
tfoote, line 147: We should add a little here showing examples of changing usages. Maybe from some of the tutorial code.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+ROS 2 only support the format 2 of the package specification which is defined in [REP 140](http://www.ros.org/reps/rep-0140.html).
+Therefore the `package.xml` file must be update to format 2 if it uses format 1.
+Since ROS 1 support both formats (1 as well as 2) it is save to perform that conversion in the ROS 1 package.
+
+Some packages might have different names in ROS 2 so the dependencies might need to be updated accordingly.
+
+
+### Messages and services
+
+Message files ending in `.msg` must be located in the subfolder `msg`.
```
adolfo-rt: This line and the one below need a reword, as there's some room for confusion. I interpret as either:
- Message (service) files **must end** in `.msg` (`.srv`) **and** must be  located in...
- Does the convention allow message and service files to have other extensions than `.msg` and `.srv`?. If so, should they also live in the `msg` and `srv` folders, or elsewhere?.

dirk-thomas:
I have updated the abstract to clarify the goal of the article:

```
This article describes the high-level steps to migrate a ROS 1 package to ROS 2.
It does not aim to be a step-by-step migration instruction.
```

articles/200_migration_guide_from_ros1.md
dirk-thomas, line 158: I don't think this article should go into such detail. In the future that should be covered by the `launch` documentation.

articles/200_migration_guide_from_ros1.md
dirk-thomas: Fixed.

articles/200_migration_guide_from_ros1.md
dirk-thomas, line 54: I have added the phrase that it might necessary to "making the field names comply to stricter naming conventions".

The header sequence is not part of the message / service migration since it does not affect users updating their message / service definitions.

articles/200_migration_guide_from_ros1.md
dirk-thomas, line 63: Fixed (added a separate sentence above).

articles/200_migration_guide_from_ros1.md
dirk-thomas, line 147: I don't think this article should go into such detail. In the future that should be covered by the `rclcpp` documentation.

articles/200_migration_guide_from_ros1.md
dirk-thomas: Good point. I clarified that the file extensions and subfolders are a **must**.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+Service files ending in `.srv` must be located in the subfolder `srv`.
+
+These files might need to be updated to comply with the [ROS Interface definition](http://design.ros2.org/articles/interface_definition.html).
+Commonly that involves updating the definition of `Duration` and `Time` fields to the new builtin messages.
```
wjwwood: I think this could be clearer, maybe: "... involves updating your files to account for the fact that in ROS 1 `time` and `duration` are built-in primitive types but in ROS 2 they are the normal messages `Time` and `Duration`."

I agree with Jackie, and I would suggest that maybe we should set this up as a list of known changes that are commonly made, rather than a single statement, and extend it as we go. Linking to the above article is good, but that states what the rules are in ROS 2, not what is different from ROS 1.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,153 @@
...
+### Build system
+
+#### Build tool
+
+Instead of using `catkin_make`, `catkin_make_isolated` or `catkin build` ROS 2 uses the command line tool [ament build](https://github.com/ament/ament_tools) to build and install a set of packages.
+
+
+#### Pure Python package
+
+If the ROS 1 package uses CMake only to invoke the `setup.py` file and does not contain anything beside Python code it should be converted into a pure Python package in ROS 2:
```
wjwwood: And does not do message generation. This is sort of implied by the "uses CMake only to invoke the `setup.py` file", but might be clearer if stated explicitly.

articles/200_migration_guide_from_ros1.md
```diff
@@ -0,0 +1,157 @@
...
+Due to the amount of changes related to dependencies it has not yet been decided if this compatibility API is useful enough to justify the effort.
+</div>
+
+
+### Update source code
+
+#### Messages and services
+
+The namespace of ROS 2 messages and services uses a subnamespace (`msg` or `srv`) after the package name.
+Therefore an include looks like: `#include <my_interfaces/msg/MyMessage.hpp>`.
```
wjwwood: `#include <my_interfaces/msg/my_message.hpp>`

articles/200_migration_guide_from_ros1.md
wjwwood, line 158: I'm not sure we should fill this part out yet. We don't even have a design document which states what the launch solution will be like or why it will be like that. I don't think it's out of the realm of possibility to also have a static file format similar to roslaunch in addition to a Python interface. I'm not advocating for one thing or the other, I just want to point out that, imo, there is a lot of uncertainty around this part of ROS 2 right now.

wjwwood:
+1, with some comments.

---

On another note, I feel like we might be starting to abuse the design website's purpose with some of these articles.

This article seems to me like it would be better suited as a less formal document (without an author and abstract and all that) maybe even a just a wiki page. And then we have some articles (linked to in this document) which are just declarative, stating what is but largely excluding the why, and I think those might be better suited as something like a REP. My original intention for the articles on this website were that they would be a place for us to capture the rational behind design decisions, outlining the actual design only enough to talk about the rational and to demonstrate the evolution of the design as our understanding and discussions evolved.

I think that as long as this "abuse" of the original idea continues to work, then there is nothing to be done, but it's something for us to keep in mind as we progress. At some point I think we need a wiki and we need to circle back to a process more like the REP's.

articles/200_migration_guide_from_ros1.md
dirk-thomas: I added a comment in parenthesis.

articles/200_migration_guide_from_ros1.md
dirk-thomas: Fixed.

articles/200_migration_guide_from_ros1.md
dirk-thomas: I have update he paragraph but used different words since I find "... to account for the fact that ..." a bit "bulky".

dirk-thomas:
If there are no further I comments I will merge the PR soon.

Issue #26, opened 2015-03-03
Write ament design document


wjwwood:
AC:
- [x] Whitepaper describing build tool design
  - [x] appendix describing changes from catkin

jackieokay:
Dirk started an effort in https://github.com/ros2/design/pull/38. Would you consider this issue closed or are you planning on expanding it?

http://design.ros2.org/articles/ament.html

dirk-thomas:
The article does not yet cover the design but focuses on the differences to catkin.

dirk-thomas:
With the recent merge of #44 I think ament is described well enough for now. If nobody disagrees I would close this ticket soon.

wjwwood:
The only thing that I think we could describe in more detail is how we changed setup files in ament to have local and "recursive" setup files and that we use the tool to create top level setup files to avoid installation collisions where ever possible.

wjwwood:
+1 to close with https://github.com/ros2/design/issues/45

PR #45, opened 2015-07-27
add more information about the setup files

dirk-thomas:
Connects to ros2/design#26


wjwwood:
+1 thanks for putting that in there.

PR #48, opened 2015-08-05
Idea for easing migration to std::*_ptr types

scpeters:
It won't provide complete backwards compatibility, unless we also typedef the pointer casting functions (`std::dynamic_pointer_cast` vs `boost::dynamic_pointer_cast` and so forth for weak and unique pointers as well). But it's a step towards easing the migration.

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -101,6 +101,8 @@ For each pointer type there a non-const and a const `typedef`:
 * `UniquePtr` and `ConstUniquePtr`
 * `WeakPtr` and `ConstWeakPtr`
 
+For partial backwards compatibility with ROS 1, these `typedefs` will also be added to ROS 1 packages using `boost::*_ptr` types. Migrating ROS 1 code to use the `typedefs` will ease migration to ROS 2.
+
 
 ## Services
 


dirk-thomas:
Can you please describe what names those typedefs should have and how they are supposed to be used?

They obviously can't have the same name so I assume you want to keep the previous typedefs and only introduce new once with different names? 

How does that help in migration? Obviously the code can not use a mixture of both at any time.

scpeters:
I think I made the suggestion without doing enough background reading, and for that I apologize.

My intent is to suggest that if the typedef's have the same name between ros 1 and 2, then it will be easier to migrate. Maybe that is already the case in this proposal. For example, I see `geometry_msgs::Twist::ConstPtr` used in [gazebo_ros_diff_drive.h](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_plugins/include/gazebo_plugins/gazebo_ros_diff_drive.h#L134). Ignoring the `_msgs` vs. `_interfaces` suffix, if the `::Twist::ConstPtr` is the same, then that will ease migration.

jack-oquin:
Yes. I think ROS 1 code using something like `std_msgs::String::ConstPtr` for the subscriber callback should work:
- http://wiki.ros.org/ROS/Tutorials/WritingPublisherSubscriber(c%2B%2B)#roscpp_tutorials.2BAC8-Tutorials.2BAC8-WritingPublisherSubscriber.CA-b00794bcdcacff4e742e844261078a50000d0a69_34

Some verification experiments would be helpful.

Most ROS 1 code probably uses the typedefs and not Boost shared pointers directly. We should be able to fix manually the few places that do.

jackieokay:
Here's a short summary of pointer typedefs for messages in ROS 1, for an arbitrary message type `Msg`:

`boost::shared_ptr<Msg<ContainerAllocator>>` --> `Msg::Ptr`.
`boost::shared_ptr<Msg<ContainerAllocator> const>` --> `Msg::ConstPtr`.

`boost::shared_ptr<Msg>` --> `MsgPtr`.
`boost::shared_ptr<Msg const>` --> `MsgConstPtr`.

Source: msg.h template in package genmsg https://github.com/ros/gencpp/blob/indigo-devel/scripts/msg.h.template#L95

This is what the document currently proposes for ROS 2 (and what's currently implemented in the analogous ROS 2 package, `rosidl_generator_cpp`):

`std::shared_ptr<Msg>` --> `Msg::SharedPtr`
`std::shared_ptr<Msg const>` --> `Msg::ConstSharedPtr`

`Msg *` --> `Msg::Ptr`
`const Msg *` --> `Msg::ConstPtr`

source: https://github.com/ros2/rosidl/blob/master/rosidl_generator_cpp/resource/msg__struct.hpp.template#L127

Changing the pointer typedefs in ROS 1 to agree with ROS 2 would break a lot of code because it would mean changing almost every instance of a `boost::shared_ptr<Msg>` to a plain pointer.

However, I think the change in pointer definitions in ROS 2 to be more comprehensive and specific is better, and the pointer definitions should not be changed in ROS 2 to agree with ROS 1.

jack-oquin:
@jacquelinekay: are you saying existing ROS1 code using the Ptr or ConstPtr typedefs will _not_ work with the new C++11 ROS 2 types? 

Specifically, I am concerned about the listener tutorial example:

``` c++
void chatterCallback(const std_msgs::String::ConstPtr& msg)
  {
    ROS_INFO("I heard: [%s]", msg->data.c_str());
  }
```

Why won't that work?

jackieokay:
@jack-oquin: the tutorial example you posted will compile and run. But I think changing the pointer typedefs in ROS 1 to match the typdefs in ROS 2 is a bad idea, and here's why:

The ROS 1 typedefs define `Ptr` as a shared pointer, while the ROS 2 typedefs define `Ptr` as a raw pointer (e.g. `Msg *`).

Let's say we changed the pointer typedefs for message types in ROS 1 so they matched the ROS 2 typedefs. Thus `Msg::Ptr` is now a raw pointer, and `Msg::SharedPtr` is a shared pointer.

Then existing user code that treats a `Msg::Ptr` as a shared_ptr might not compile, because `shared_ptr` has a different API from raw pointers (see [cpp reference](http://en.cppreference.com/w/cpp/memory/shared_ptr) and [boost reference](http://www.boost.org/doc/libs/1_58_0/libs/smart_ptr/shared_ptr.htm#Synopsis)). The raw pointer API is basically a subset of the shared pointer API, since the shared pointer API provides things like `use_count` and `reset`. Even if it compiles, the user code will have unexpected behavior and poor performance because of the different behavior of these two types. `shared_ptr`s deallocate themselves when all owners of that shared pointer are deleted, so changing shared pointers to raw pointers without changing any other code will result in memory leaks.

scpeters:
What if we call raw pointers `Msg::RawPtr` and leave `Msg::Ptr` as an alias to `Msg::SharedPtr` for compatibility?

jackieokay:
Do you mean changing the name of raw pointers in ROS 2 to `Msg::RawPtr` and providing the same typedef for raw pointers in ROS 1?

I think the `RawPtr` name was discussed in an earlier review of this document and turned down basically because it sounds bad/redundant.

wjwwood:
@scpeters I suggested something similar at the time:

https://github.com/ros2/design/pull/33#discussion_r32594616

And though I would still prefer SharedPtr since that is more obvious when reading code, I think what you suggested is probably required.

scpeters:
Without opining on the relative value of each name, I think compatibility could be improved by using `msg::RawPtr`, while still adding `msg::SharedPtr` but keeping `msg::Ptr` as an alias to `msg::SharedPtr`.

jack-oquin:
@jacquelinekay: thanks for explaining the difference between `Ptr` and `SharedPtr` in the document. I didn't realize the meanings had shifted.

+1 to making `Ptr` and `ConstPtr` both shared pointers, like in ROS 1 (only without Boost). That's assuming we fix all the other obstacles like changing package names, etc. I agree with @wjwwood and @scpeters that it's not worth the cumulative effort of changing thousands of callbacks just for the extra clarity of `SharedPtr`.

I don't expect `RawPtr` to get used all that much, so the fact that the name draws attention to its uncommon meaning seems beneficial.

jack-oquin:
Assuming we are careful about maintaining semantically consistent typedefs, I do not yet see any reason to modify the ROS 1 definitions.

Is there something else I am missing?

wjwwood:
> it's not worth the cumulative effort of changing thousands of callbacks just for the extra clarity of SharedPtr.

The callback or the code around the callback is likely going to have to change anyways for one reason or another, so this change is not the difference between having compatibility or not. I was just saying that it will be more familiar / less confusing to ROS 1 developers when writing new code.

jack-oquin:
-0 to making `SharedPtr` and `Ptr` synonyms, but I don't feel strongly about that.

While it aids compatibility and allows new code to declare the pointer type more clearly, the fact that they have the same meaning will be missed by many users, causing minor confusion. Existing ROS programmers already know that `Ptr` means shared. That's why I missed the point entirely until Jackie set me straight.

jack-oquin:
> The callback or the code around the callback is likely going to have to change anyways for one reason or another, so this change is not the difference between having compatibility or not. I was just saying that it will be more familiar / less confusing to ROS 1 developers when writing new code.

I think it's a worthwhile goal to make it compatible. The potential benefits are large.

wjwwood:
> I think it's a worthwhile goal to make it compatible. The potential benefits are large.

I didn't say it was not worthwhile, I just don't think it is possible.

jack-oquin:
If we give up on it too quickly, it will certainly not be possible.

dirk-thomas:
I propose to replace this with ros2/design#53.

scpeters:
#53 looks better to me. I'll close this once consensus has been reached.

PR #50, opened 2015-08-07
Review comments for "Generated C++ interfaces"

jack-oquin:
This pull request is for reviewing the referenced ROS 2 design document:
- http://design.ros2.org/articles/generated_interfaces_cpp.html
- https://github.com/jack-oquin/design/blob/gh-pages/articles/112_generated_interfaces_cpp.md

Relevant discussions:
- #48 
- [REP, REP](https://groups.google.com/forum/#!topic/ros-sig-ng-ros/RaA617pJZtU)
- [Supporting both ROS versions with common sources](https://groups.google.com/forum/#!topic/ros-sig-ng-ros/61C8GUaBfKE)
- [Keeping same message package name](https://groups.google.com/forum/#!topic/ros-sig-ng-ros/wP__Hu9O8kM)

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -18,6 +18,8 @@ author: '[Dirk Thomas](https://github.com/dirk-thomas)'
 </div>
 
 Original Author: {{ page.author }}
+Status: Draft
+Type: Standards Track
 
 
 ## Scope
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -30,15 +30,24 @@ This article specifies the generated C++ code for ROS interface types defined in
 ## Namespacing
 
 All code of a ROS package should be defined in a namespace named after the package.
-To separate the generated code from other code within the package it is defined in a sub namespace:
+To separate the generated code from other code within the package it is defined in a sub-namespace:
+
 * namespace for ROS messages: `package_name::msg`.
 * namespace for ROS services: `package_name::srv`.
+
+<div class="alert alert-warning" markdown="1">
+  **Note:** Those namespaces are not consistent with ROS 1 usage.
+</div>
 
 
 ## Generated files
 
 Following the C++ style guide of ROS 2 the namespace hierarchy is mapped to a folder structure.
 The filenames use lowercase alphanumeric characters with underscores for separating words and end with either `.hpp` or `.cpp`.
+
+<div class="alert alert-warning" markdown="1">
+  **Note:** The suggested naming is not consistent with ROS 1 usage.
+</div>
 
 
 ## Messages
@@ -103,6 +112,9 @@ For each pointer type there a non-const and a const `typedef`:
 * `UniquePtr` and `ConstUniquePtr`
 * `WeakPtr` and `ConstWeakPtr`
 
+<div class="alert alert-warning" markdown="1">
+  **Note:** `Ptr` and `ConstPtr` are *not* compatible with ROS 1 usage (which is shared). That may change.
+</div>
 
 ## Services
 
@@ -118,5 +130,10 @@ The generated code is split across multiple files the same way as message are.
 
 ### Request and response messages
 
-For the request as well as the response part messages are being generated which have the same name as the service with a suffix being either `_Request` or `_Response`.
-These messages are still defined in the `srv` sub namespace.
+The request and response messages are generated with the service name and a suffix of `Request` or `Response`.
+
+These messages are still defined in the `srv` sub-namespace.
+
+<div class="alert alert-warning" markdown="1">
+  **Note:** That is not consistent with ROS 1 usage.
+</div>


articles/112_generated_interfaces_cpp.md
```diff
@@ -101,6 +112,9 @@ For each pointer type there a non-const and a const `typedef`:
 * `UniquePtr` and `ConstUniquePtr`
 * `WeakPtr` and `ConstWeakPtr`
 
+<div class="alert alert-warning" markdown="1">
+  **Note:** `Ptr` and `ConstPtr` are *not* compatible with ROS 1 usage (which is shared). That may change.
+</div>
```
jack-oquin, line 117: I prefer these typedefs proposed by @scpeters, instead:
- `RawPtr` and `ConstRawPtr`
- `Ptr` and `ConstPtr` (for the shared pointers, consistent with ROS 1 usage)
- `UniquePtr` and `ConstUniquePtr`
- `WeakPtr` and `ConstWeakPtr`

articles/112_generated_interfaces_cpp.md
```diff
@@ -28,16 +30,25 @@ This article specifies the generated C++ code for ROS interface types defined in
...
-To separate the generated code from other code within the package it is defined in a sub namespace:
+To separate the generated code from other code within the package it is defined in a sub-namespace:
+
 * namespace for ROS messages: `package_name::msg`.
 * namespace for ROS services: `package_name::srv`.
 
+<div class="alert alert-warning" markdown="1">
+  **Note:** Those namespaces are not consistent with ROS 1 usage.
+</div>
+
```
jack-oquin, line 41: Let's delete the sub-namespaces. In ROS 1, the messages and service message namespaces are _not_ distinct.

articles/112_generated_interfaces_cpp.md
```diff
@@ -28,16 +30,25 @@ This article specifies the generated C++ code for ROS interface types defined in
...
 
 ## Generated files
 
 Following the C++ style guide of ROS 2 the namespace hierarchy is mapped to a folder structure.
 The filenames use lowercase alphanumeric characters with underscores for separating words and end with either `.hpp` or `.cpp`.
 
+<div class="alert alert-warning" markdown="1">
+  **Note:** The suggested naming is not consistent with ROS 1 usage.
+</div>
+
```
jack-oquin, line 51: I believe the header file names should be consistent with ROS 1:

``` C++
#include <my_msgs/MyMessageName.h>
```

The differences are: there is no intermediate directory name; the message name is CamelCase, like the message definition; and the header suffix is `.h` as recommended by the [ROS C++ Style Guide](http://wiki.ros.org/CppStyleGuide#Files).

The argument for changing it is to make the static ROS 1 bridge easier to write (and possibly other, similar hybrid packages). I don't believe that justifies the cost of updating every message header reference in every existing ROS C++ package.

articles/112_generated_interfaces_cpp.md
```diff
@@ -116,5 +130,10 @@ The generated code is split across multiple files the same way as message are.
...
 
-For the request as well as the response part messages are being generated which have the same name as the service with a suffix being either `_Request` or `_Response`.
-These messages are still defined in the `srv` sub namespace.
+The request and response messages are generated with the service name and a suffix of `Request` or `Response`.
+
+These messages are still defined in the `srv` sub-namespace.
+
+<div class="alert alert-warning" markdown="1">
+  **Note:** That is not consistent with ROS 1 usage.
+</div>
```
jack-oquin, line 139: As mentioned above, I recommend removing the `srv` sub-namespace.

jack-oquin:
My general attitude here is that we should minimize the differences between ROS 1 and ROS 2 message references, because that will make a big difference in the total amount of work the ROS community ends up investing in migration and parallel maintenance. 

There will be some differences, but they should each be justified in comparison to the total amount of work they create.

articles/112_generated_interfaces_cpp.md
dirk-thomas, line 41: Since the typename of ROS 1 message and ROS 2 messages is the same with your proposed change how shall these two types then be used in C++ code?

articles/112_generated_interfaces_cpp.md
dirk-thomas, line 51: With your proposed change how can ROS 1 and ROS 2 header be included at the same time?

How will the C equivalent of the message header be named then?

articles/112_generated_interfaces_cpp.md
wjwwood, line 41: If we don't do this should we change Python to not put them in submodules, e.g. `from std_msgs import String` instead of `from std_msgs.msg import String`?

articles/112_generated_interfaces_cpp.md
jack-oquin, line 51: > With your proposed change how can ROS 1 and ROS 2 header be included at the same time?

Through some other ugly hack that only affects those few programs. The first one that comes to mind is playing around with the include paths returned by pkg_config for the ROS1 tree. Yep, that's ugly.

If you don't come up with a better one before I get back from vacation, I promise to come up with several other ideas you probably won't like, either. :smile: 

> How will the C equivalent of the message header be named then?

However you like. The scope of this document is C++ only. There is little or no existing ROS code in C, so you are free to invent new solutions.

Personally, I would investigate making the C++ headers bilinugal for both C and C++. But, maybe that's not feasible. It's up to you, Dirk.

articles/112_generated_interfaces_cpp.md
jack-oquin, line 41: > If we don't do this should we change Python to not put them in submodules, e.g. from std_msgs import String instead of from std_msgs.msg import String?

The Python representation is outside the scope of this C++ article. I have not seen any similar Python design document, yet. 

When we do have one, I will probably advocate doing it similarly to the way Python works in ROS 1.

jack-oquin, line 41: > Since the typename of ROS 1 message and ROS 2 messages is the same with your proposed change how shall these two types then be used in C++ code?

The intent is for them to be used the same way. When compiled for ROS 1, they will resolve to Boost shared pointers. For ROS 2 they will use C++11 shared pointers. Code that explicitly uses the Boost name rather than the typedef will need to be modified, but will then work with either version.

Or, is this another question about that static ROS 1 bridge prototype?

dirk-thomas:
@jack-oquin A comment to clarify why the C++ types use a different namespace has been added in #101. Can this be closed?

dirk-thomas:
I will the ticket for now. Please feel free to comment on it and it can be reopened if necessary.

jack-oquin:
I just saw this, today.

I still strongly disagree with having the message references be different. The stated justification is particularly weak, as it only works in C++ and not in Python.

However, you seem determined to make this mistake, which we will all pay for in the end. I don't have enough influence to fix it.

dirk-thomas:
> I still strongly disagree with having the message references be different. The stated justification is particularly weak, as it only works in C++ and not in Python.

I am not sure which part you are referring to with "different message references". Maybe you can clarify which of the two topics you mean:

* The pointer types which have been integrated in #53.
* The different namespaces for generated C++ message classes in ROS 1 and ROS 2. (In this case the discussion above mentions that we need separate types in order to allow building hybrid applications. We are still not aware of any approach how to achieve that without separating the types in C++. Maybe you can propose a solutions as mentioned in https://github.com/ros2/design/pull/50#r36570888?)

> However, you seem determined to make this mistake, which we will all pay for in the end. I don't have enough influence to fix it.

Please feel free to state your opinion and raise concerns. I am sure your comments are being heard and your feedback is being considered.

jack-oquin:
This is the same issue I raised almost two years ago, the unnecessary breakage of the C++ #include paths for every existing message used by every existing ROS 1 program written in that language.

The justification for that is for the ease of writing a handful of hybrid applications. Are there any examples yet, beyond the ROS 1 message bridge?

I am happy to raise concerns, but I'm not interested in working on things you've already made up your mind about. This looks like one of those. Of all the incompatibilities between ROS 1 and ROS 2, it seems to have the biggest cost and the weakest justification.

tkruse:
Given a ROS2 beta has been out for a while, just ask the existing maintainers of packages distributed for both ROS editions about their experiences for porting.

dirk-thomas:
> This is the same issue I raised almost two years ago, the unnecessary breakage of the C++ #include paths for every existing message used by every existing ROS 1 program written in that language.

> The justification for that is for the ease of writing a handful of hybrid applications. Are there any examples yet, beyond the ROS 1 message bridge?

Let me summarize the current state and rational (to allow other readers to catch up without needinf to read all previous comments). The goal of having a bridge is a crucial goal (not optional and only nice to have). We think it is essential to have one / being able to implement one in C++ because it allows writing such a bridge in the form of a nodelet / component fo maximum efficiency. Both datatypes (ROS 1 and ROS 2) therefore need to be includable at the same time. We did not see a different way to achieve this. We have asked for anybody to propose a possible solution to this which doesn't come with the downside of changing the namespaces (and doesn't have other major disadvantages like custom code generators only for the bridge). It is a very difficult technical problem and we simply don't have a better solution at the moment. Neither back than when the discussion started nor now. We are more than open to consider any proposal.

> I am happy to raise concerns, but I'm not interested in working on things you've already made up your mind about. This looks like one of those. Of all the incompatibilities between ROS 1 and ROS 2, it seems to have the biggest cost and the weakest justification.

Also even thought we have released a beta 1 a few months ago that doesn't mean any API is set in stone. With good reason the ROS 2 developers are very open to any kind of change - even if it would require touch existing to update it.

jack-oquin:
The ROS 1 bridge *is* nice to have.

The current implementation is dependent on all the bridged messages being available in C++ at compile time, which makes a nice small-scale demo, but is not adequate for large systems. 

> We have asked for anybody to propose a possible solution to this which doesn't come with the downside of changing the namespaces (and doesn't have other major disadvantages like custom code generators only for the bridge). It is a very difficult technical problem and we simply don't have a better solution at the moment. Neither back than when the discussion started nor now. We are more than open to consider any proposal.

How about an extra option on one of the code generators, to add an extra directory to the path name? 

How about some CMake trickery specific to building the message bridge demo? Symlink to the ROS 1 messages using a different path.

I don't really *care* how you fix it, just that it gets done.

dirk-thomas:
> How about an extra option on one of the code generators, to add an extra directory to the path name?

All message packages have already been compiled when the bridge is being build. So for this the bridge would need to rebuild all message packages (with that extra option). I don't think that is a scalable solution.

> How about some CMake trickery specific to building the message bridge demo? Symlink to the ROS 1 messages using a different path.

Symlinks are not available on Windows (if we don't require admin privileges).

jack-oquin:
The idea of compiling both versions of all the messages into a single C++ executable is not scalable, either.

If you're doing that, you could just copy the message headers somewhere convenient.

PR #51, opened 2015-08-07
Review comments for "Interface definition"

jack-oquin:
This pull request is for reviewing the referenced ROS 2 design document:
- http://design.ros2.org/articles/interface_definition.html
- https://github.com/jack-oquin/design/blob/interface_definition/articles/110_interface_definition.md

Relevant discussions:
- #48
- [REP, REP](https://groups.google.com/forum/#!topic/ros-sig-ng-ros/RaA617pJZtU)
- [Supporting both ROS versions with common sources](https://groups.google.com/forum/#!topic/ros-sig-ng-ros/61C8GUaBfKE)
- [Keeping same message package name](https://groups.google.com/forum/#!topic/ros-sig-ng-ros/wP__Hu9O8kM)

### DIFF
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -18,7 +18,8 @@ author: '[Dirk Thomas](https://github.com/dirk-thomas)'
 </div>
 
 Original Author: {{ page.author }}
-
+Status: Draft
+Type: Standards Track
 
 ## Scope
 
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -21,6 +21,7 @@ Original Author: {{ page.author }}
 Status: Draft
 Type: Standards Track
 
+
 ## Scope
 
 This article specifies the file format describing the data structures which are being used to exchange information between components.
@@ -61,8 +62,6 @@ Each of these can optionally be a dynamically or statically sized array.
 The following primitive types are defined:
 
 * `bool`
-* `byte`
-* `char`
 * `float32`, `float64`
 * `int8`, `uint8`
 * `int16`, `uint16`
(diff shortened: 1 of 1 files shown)


articles/110_interface_definition.md
```diff
@@ -70,11 +70,15 @@ The following primitive types are defined:
...
 <div class="alert alert-warning" markdown="1">
+  **Question:** in ROS 1, `string` was ASCII. The Python 3 `str` type is UTF-8. What is the DDS `string` type?
+</div>
+
+<div class="alert alert-warning" markdown="1">
   <b>TODO:</b> consider <code>wchar</code>, <code>wstring</code>, <code>u16string</code>, <code>u32string</code>
 </div>
 
 <div class="alert alert-warning" markdown="1">
-  <b>TODO:</b> consider removing <code>byte</code>, <code>char</code> after specifying the mapping to C++ and Python
```
jack-oquin, line 77: I went ahead with this change. Those types have been deprecated for many years, and the mapping was already clearly specified:
- `char`: deprecated alias for `uint8`
- `byte`: deprecated alias for `int8`

See: http://wiki.ros.org/msg#Fields

articles/110_interface_definition.md
```diff
@@ -70,11 +70,15 @@ The following primitive types are defined:
 * `string`
 
 <div class="alert alert-warning" markdown="1">
+  **Question:** in ROS 1, `string` was ASCII. The Python 3 `str` type is UTF-8. What is the DDS `string` type?
```
dirk-thomas, line 73: In ROS 1 as well as in ROS 2 the string message type is only storing a sequence of bytes. Any encoding is user land responsibility.

jack-oquin, line 73: Yes, I guess that's true, but not very satisfying as a specification.

Should we recommend these bytes contain either ASCII or a UTF-8 encoding?

jack-oquin:
This discussion seems moot at this point.

PR #53, opened 2015-08-11
rename raw pointers and keep deprecated ROS 1 typedefs

dirk-thomas:
The renames `Ptr` to `RawPtr` to not break expectations.

It keeps `Ptr` as a `std::shared_ptr` (as it is in ROS 1) but deprecates its usage.

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -96,10 +96,13 @@ The comparison operators `==` and `!=` perform the comparison on a per member ba
 The struct contains `typedefs` for the four common pointer types `plain pointer`, `std::shared_ptr`, `std::unique_ptr`, `std::weak_ptr`.
 For each pointer type there a non-const and a const `typedef`:
 
-* `Ptr` and `ConstPtr`
+* `RawPtr` and `ConstRawPtr`
 * `SharedPtr` and `ConstSharedPtr`
 * `UniquePtr` and `ConstUniquePtr`
 * `WeakPtr` and `ConstWeakPtr`
+
+For similarity to ROS 1 the `typedefs` `Ptr` and `ConstPtr` still exist but are deprecated.
+In contrast to ROS 1 they use `std::shared_ptr` instead of Boost.
 
 
 ## Services


wjwwood:
I think this is a reasonable compromise.

scpeters:
This looks great to me. +1. I'll close #48 after the dust settles.

jack-oquin:
A reasonable approach, but only if the new SharedPtr, etc. is also added to ROS 1. Otherwise, -1

scpeters:
It wouldn't break anything to add the new `typedef`'s to ROS 1, would it?

wjwwood:
I don't see any problem with adding `SharedPtr` to ROS 1 messages. I think it's a good idea.

@jack-oquin what do you mean by `etc.`? You mean the `UniquePtr`/`WeakPtr` ones? They might not translate directly to boost smart pointer concepts. Unlike the relationship between `boost::shared_ptr` and `std::shared_ptr`, I'm not sure the other boost concepts are as universally translatable as that. I'd support `SharedPtr` and `ConstSharedPtr` being back ported to ROS 1, but the others are more questionable.

scpeters:
I don't see a `boost::unique_ptr` defined, though I do see `boost::weak_ptr`. I couldn't find any references on the difference between `boost` and `std` weak pointers, but I did find a [page about the differences between `boost::shared_ptr` and `std::shared_ptr`](http://stackoverflow.com/questions/1086798/differences-between-different-flavours-of-shared-ptr).

jack-oquin:
> @jack-oquin what do you mean by etc.? You mean the UniquePtr/WeakPtr ones? They might not translate directly to boost smart pointer concepts. Unlike the relationship between boost::shared_ptr and std::shared_ptr, I'm not sure the other boost concepts are as universally translatable as that. I'd support SharedPtr and ConstSharedPtr being back ported to ROS 1, but the others are more questionable

Sorry for excessive terseness. Typing this on a small tablet.

I was only thinking of SharedPtr and ConstSharedPtr, but avoided typing it out fully. Those two would allow most existing ROS 1 code to migrate to the preferred typedefs, which are clearer.

gbiggs:
+1. I don't think we should encourage the use of raw pointers. Naming it RawPtr makes it clear what it is for those who know what they are doing.

scpeters:
I just created a pull request to add the SharedPtr and ConstSharedPtr typedef's to ROS 1:

https://github.com/ros/roscpp_core/pull/36

PR #55, opened 2015-08-13
Extend migration guide to walk through an example of a simple package.

gerkey:
My goal is to provide the simplest possible example of migrating an existing ROS 1 package to use ROS 2.  Along the way, I aim to avoid unnecessary changes, so as not to distract from the required changes.  That's why the ROS 2 C++ code doesn't use `auto` everywhere that it could, and also why I'm directly invoking `include_directories()` and `target_link_libraries()` instead of the fancier `ament_target_dependencies()`.

I'll extend the article (or add new ones) to start to cover more complex use cases, like defining messages in a package.  But I'd like to go ahead and get feedback on this PR and if possible merge it for broader community feedback.

The example code discussed in the text is also included in the PR.

---

To read the new content locally, pull this branch, then in the top of the repo, run:

```
jekyll serve --watch --baseurl=''
```

Then point your browser to the [new content](http://localhost:4000/articles/migration_guide_from_ros1.html#example-converting-an-existing-ros-1-package-to-use-ros-2).


dirk-thomas:
Is there a way to avoid the duplication of the content of all the files? Somehow reference the files from the article?

_config.yml
```diff
@@ -18,3 +18,4 @@ exclude:
  - LICENSE
  - README.md
  - vendor
+ - examples
```
dirk-thomas: Please insert in alphabetical order.

articles/200_migration_guide_from_ros1.md
```diff
@@ -162,3 +162,483 @@ NOTE: to be written
 ## Launch files
 
 While launch files in ROS 1 are specified using [.xml](http://wiki.ros.org/roslaunch/XML) files ROS 2 uses Python scripts to enable more flexibility (see [launch package](https://github.com/ros2/launch/tree/master/launch)).
+
+## Example: Converting an existing ROS 1 package to use ROS 2
+
+Let's say that we have simple ROS 1 package called `talker` that uses `roscpp`
+in one node, called `talker`.  This package is in a catkin workspace, located
```
dirk-thomas: Please start each sentence on a new line.

Same below.

articles/200_migration_guide_from_ros1.md
```diff
@@ -162,3 +162,483 @@ NOTE: to be written
...
+~~~
+//  ros::Publisher chatter_pub = n.advertise<std_msgs::String>("chatter", 1000);
+//  ros::Rate loop_rate(10);
+  auto chatter_pub = node->create_publisher<std_msgs::msg::String>("chatter",
+    rmw_qos_profile_default);
+  rclcpp::rate::Rate loop_rate(10);
+~~~
+
+The creation of the outgoing message is different in both the namespace and the
+fact that we go ahead and create a shared pointer:
```
dirk-thomas: Should we mention that this is temporary too? Assuming that we will support publishing a const reference soon.

articles/200_migration_guide_from_ros1.md
```diff
@@ -162,3 +162,483 @@ NOTE: to be written
...
+start by specifying the format version in the `package` tag:
+
+~~~
+<!-- <package> -->
+<package format="2">
+~~~
+
+ROS 2 uses a newer version of `catkin`, called `ament`, which we specify in the
+`buildtool_depend` tag (more specifically, we specify that we want to use
+`ament` to build our package using CMake):
```
dirk-thomas: `buildtool` does not mean that "we want to use ament to build our package". It is simply a build dependency which must be available for the platform the build is being done on in contrast to a "normal" build dependency which must be available for the platform the build is targeted (when cross compiling).

articles/200_migration_guide_from_ros1.md
```diff
@@ -162,3 +162,483 @@ NOTE: to be written
...
+
+~~~
+#catkin_package()
+# At the bottom of the file:
+ament_package()
+~~~
+
+Similarly to how we found each dependent package separately, instead of finding
+them as parts of catkin, we also need to add their include directories
+separately:
```
dirk-thomas: Maybe mention that for this example it doesn't make a different but for more complex example and `ament_auto` it does?

articles/200_migration_guide_from_ros1.md
```diff
@@ -162,3 +162,483 @@ NOTE: to be written
...
+                      ${std_msgs_LIBRARIES})
+~~~
+
+*TODO: explain how `ament_target_dependencies()` simplifies the above steps and
+is also better (also handling `*_DEFINITIONS`, doing target-specific include
+directories, etc.).*
+
+For installation, `catkin` defines variables like
+`CATKIN_PACKAGE_BIN_DESTINATION`.  With `ament`, we just give a path relative to
+the installation root, like `bin` for executables:
```
dirk-thomas: Should we mention here that we install to `bin` for now since we don't have a `rosrun` equivalent yet? I saw it is mentioned below but here might be where the question comes up in the first place.

examples/ros1/src/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+/opt/ros/jade/share/catkin/cmake/toplevel.cmake
```
dirk-thomas: We shouldn't mention this file. First it also does not appear in the folder listing above. Second when building with catkin_make_isolated it is not even there.

jackieokay:
@dirk-thomas With regards to your question about duplication, perhaps we should use `{% include <filename> %}`?

http://jekyllrb.com/docs/templates/#includes

tfoote:
With the include syntax you can also create a directory structure inside include.

articles/200_migration_guide_from_ros1.md
```diff
@@ -162,3 +162,483 @@ NOTE: to be written
...
+
+~~~
+<!--  <buildtool_depend>catkin</buildtool_depend> -->
+  <buildtool_depend>ament_cmake</buildtool_depend>
+~~~
+
+In our build dependencies, instead of `roscpp` we use `rclcpp`, which provides
+the C++ API that we use.   We additionally depend on `rmw_implementation`, which
+pulls in the default implementation of the `rmw` abstraction layer that allows
+us to support multiple DDS implementaitons (we should consider restructuring /
```
jackieokay: "implementaitons"

_config.yml
gerkey: done

articles/200_migration_guide_from_ros1.md
gerkey: done

While I understand the reason for doing this (more concise diffing), I find that style incredibly tedious and a poor tradeoff (because it makes the experience of reading, writing, or editing the underlying text in an editor much worse).

articles/200_migration_guide_from_ros1.md
gerkey: done

articles/200_migration_guide_from_ros1.md
gerkey: I removed the parenthetical phrase.  I'd rather not explain what's actually going on under the hood, because almost nobody should need to understand it.

articles/200_migration_guide_from_ros1.md
gerkey: done

articles/200_migration_guide_from_ros1.md
gerkey: done

examples/ros1/src/CMakeLists.txt
gerkey: Accidentally included; removed.

articles/200_migration_guide_from_ros1.md
gerkey: fixed (also otherwise spell-checked)

gerkey:
I agree that using an include syntax to bring in parts of the source files would be an improvement, but it can always be done later (and in the meantime doesn't have an impact on the value of the article to a reader).

dirk-thomas:
I don't think delaying to use the include syntax buys us anything but overhead. Therefore I took the freedom to commit 13ab40e9b084a29edb4e0744c5f39b02aee2fa60.

gerkey:
Thanks, @dirk-thomas !

I didn't realize that you only wanted to include the entire files, as opposed to the chunks.  I was delaying because I didn't want to have figure out how to insert and operate on tags within the files to be able to show chunks (assuming that this is possible).

wjwwood:
+1

I assume we should merge this and then open pull requests which suggest how we could make them even closer? Things like have a version of the functions which take `queue_size`, considering changing the names to advertise and subscribe again, allowing the user to publish and receive const reference, etc.

tfoote:
+1 for merge and ongoing prs for more iteration

dirk-thomas:
Please squash future PRs before merging.

gerkey:
Oh, right, I always forget about squashing.

Issue #57, opened 2015-09-14
Write parameters design document


tfoote:
Create a design document for the parameter system.

vooon:
Not sure where better to ask.

Does new param system allow me to expose parameters from connected device?
e.g. following schematic:

![param](https://cloud.githubusercontent.com/assets/1023568/10804378/3d33e300-7dd8-11e5-9091-fd1b36160a8b.png)

Where node cache stores additional device-specific information (SET operation require that).
Or i can not partially rebuild node param server?

tfoote:
@vooon I'm not sure I understand your question enough to answer it. Could you explain what you're calling out as a device, ros, and node cache?

vooon:
- _Device_ some external thing connected to computer (with it's own parameters).
  In my case it is flight computer.
- _ROS_ bunch of nodes that uses unified protocol.
- _Node cache_ - place where node that handles communication with device store parameter values

I want to see following sequence:
1. ROS node request to set some parameter
2. driver node reencode and send that request to device
3. device responds with applied value
4. driver node update cache & fire event for other nodes

tfoote:
@vooon Supporting setting the parameter on an external device as you mention would work fine. In your drawing though both the setting and getting from the main ROS system would go to the node cache, and it would talk to the device. I'd suggest considering it more of a proxy for the device, in the realm of a device driver rather than a peer.

vooon:
But if set first changes cache it may inconsistent with actual device value. Device e.g. may reject new value. So for time between driver request and reply cache will store wrong data.
But really that not big deal, now that works similar with roscore param storage.

But one thing not clear to me: would it be possible to reimplement parameter storage & request handlers? Last time i saw rclcpp sources that part is private, so i can't.
Thing that i don't like in my current implementation is double cache: one for additional metadata (and value in device format) and other for value (in roscore, but same may happen with ros2 if i can't extend cache storage type).

jack-oquin:
I have device response issues with [camera1394](http://wiki.ros.org/camera1394) similar to what @vooon describes. Various ones of the thousands of IIDC-compliant cameras support particular modes and option settings. There is no reasonable way for the driver to tell which ones without querying the device and waiting for a response. That works with dynamic_reconfigure, but it takes time. 

I can't tell whether the discussion about immediately responding to parameter updates is consistent with lengthy validations of that nature.

tfoote:
[...]

PR #58, opened 2015-09-15
first draft of parameter design doc

tfoote:
Connects to #57

### DIFF
--- a/articles/055_ros_parameter_design.md
+++ b/articles/055_ros_parameter_design.md
@@ -32,11 +32,13 @@ It provided a service based interface to interact with parameters of other nodes
 
 Other resources related to the parameter design process for ROS 2.0 include:
 
-- Gonzalo's research on parameters.
-  - Discussion: https://groups.google.com/forum/#!topic/ros-sig-ng-ros/YzCmoIsN0o8 and https://groups.google.com/forum/#!searchin/ros-sig-ng-ros/parameter/ros-sig-ng-ros/fwDBcei5Ths/L6ORPfjUDXYJ
-  - Protoype: https://github.com/abellagonzalo/dynamic_config
-  - Final Notes:  [http://wiki.ros.org/sig/NextGenerationROS/Parameters](http://wiki.ros.org/sig/NextGenerationROS/Parameters)
-- Thibault's nodeparam draft REP: https://github.com/tkruse/rep/blob/nodeparam/nodeparam-REP.rst
+* Gonzalo's research on parameters.
+
+  * Discussion: <https://groups.google.com/forum/#!topic/ros-sig-ng-ros/YzCmoIsN0o8> and <https://groups.google.com/forum/#!searchin/ros-sig-ng-ros/parameter/ros-sig-ng-ros/fwDBcei5Ths/L6ORPfjUDXYJ>
+  * Prototype: <https://github.com/abellagonzalo/dynamic_config>
+  * Final Notes:  [http://wiki.ros.org/sig/NextGenerationROS/Parameters](http://wiki.ros.org/sig/NextGenerationROS/Parameters)
+
+* Thibault's nodeparam draft REP: <https://github.com/tkruse/rep/blob/nodeparam/nodeparam-REP.rst>
 
 ## Ideal System
 
@@ -44,47 +46,47 @@ It is useful to consider an ideal system to understand how it would work and how
 An ideal system would support for the combined use cases of ROS 1.0's built-in parameters as well as ROS 1.0's dynamic parameters system.
 Based on that criteria an ideal system would be able to:
 
-- **Set parameter values**
+* **Set parameter values**
 
-    This includes setting groups of parameter values atomically.
+  This includes setting groups of parameter values atomically.
 
-- **Get parameter values**
+* **Get parameter values**
 
-    This includes getting groups of parameter values atomically.
+  This includes getting groups of parameter values atomically.
 
-- **Unset parameter values**
+* **Unset parameter values**
 
-    This includes unsetting groups of parameter values atomically, but may be a special case of setting groups of parameters atomically.
+  This includes unsetting groups of parameter values atomically, but may be a special case of setting groups of parameters atomically.
 
-- **List currently set parameters**
+* **List currently set parameters**
 
-    Since the number of parameters might be large this needs to be able to provide subsets of the parameters.
-    E.g. the parameters could be queried incrementally for a tree-like GUI.
+  Since the number of parameters might be large this needs to be able to provide subsets of the parameters.
+  E.g. the parameters could be queried incrementally for a tree-like GUI.
 
-- **Provide notifications when parameters are added and removed or theie value has been changed**
+* **Provide notifications when parameters are added and removed or theie value has been changed**
 
-- **Reject changes to parameter values**
+* **Reject changes to parameter values**
 
-    This implies that some entity has the authority to reject or accept a change based on arbitrary criteria.
-    This would also include the ability to convey at least part of the criteria for the acceptance of a change to external actors.
-    For example, communicating the range for an integer or a few choices for a string.
-    This type of information would be used to generate generic user interfaces, but might not capture all criteria.
-    Since the validation criteria can be arbitrary complex and can potentially not be communicated to a client the parameter server could offer to validate an atomic set of parameters and respond with a boolean flag if the values would be accepted (based on the current criteria).
-    Obviously the result might be different when the values are set shortly after but it would allow to implement validators in e.g. a GUI generically.
+  This implies that some entity has the authority to reject or accept a change based on arbitrary criteria.
+  This would also include the ability to convey at least part of the criteria for the acceptance of a change to external actors.
+  For example, communicating the range for an integer or a few choices for a string.
+  This type of information would be used to generate generic user interfaces, but might not capture all criteria.
+  Since the validation criteria can be arbitrary complex and can potentially not be communicated to a client the parameter server could offer to validate an atomic set of parameters and respond with a boolean flag if the values would be accepted (based on the current criteria).
+  Obviously the result might be different when the values are set shortly after but it would allow to implement validators in e.g. a GUI generically.
 
-- **Provide visibility into what parameters are expected to pass validation vs be rejected**
+* **Provide visibility into what parameters are expected to pass validation vs be rejected**
 
-    When updating a value it can be valuable to know if the parameter update would be accepted without actually requesting the change to happen.
+  When updating a value it can be valuable to know if the parameter update would be accepted without actually requesting the change to happen.
 
-- **Provide clear rules on the lifetime of a parameter**
+* **Provide clear rules on the lifetime of a parameter**
 
-    These rules would define what the lifetime of the parameter will be and what conditions will clear its value.
+  These rules would define what the lifetime of the parameter will be and what conditions will clear its value.
 
-- **Address all parameters without ambiguity in the names**
+* **Address all parameters without ambiguity in the names**
 
-    For example, one of the challenges of the current system is that there is a naming ambiguity between nodes and parameters `/foo/bar/baz` could be a node `/foo/bar/baz` or a private parameter `baz` on node `/foo/bar`.
+  For example, one of the challenges of the current system is that there is a naming ambiguity between nodes and parameters `/foo/bar/baz` could be a node `/foo/bar/baz` or a private parameter `baz` on node `/foo/bar`.
 
-- **Be logged for playback and analysis**
+* **Be logged for playback and analysis**
 
 ## Proposed Approach
 
@@ -107,11 +109,11 @@ Each parameter consists of a key and a value.
 The key is a string value.
 The value can be one of the following datatypes:
 
-- `float64`
-- `int64`
-- `string`
-- `bool`
-- `bytes[]`
+* `float64`
+* `int64`
+* `string`
+* `bool`
+* `bytes[]`
 
 The datatypes are chosen as non-complex datatypes, as defined in the [interface definitions article](articles/interface_definition.html)
 The full complement of datatypes of different bitdepth and unsigned types are avoided to allow interpretation from text based configuration files.
@@ -123,18 +125,21 @@ It's use is not recommended but can be very convenient, and explicitly supportin
 
 Each node is responsible for providing the following functionality.
 
-- **Get Parameters**
+* **Get Parameters**
   Given a list of parameter names it will return the parameter values.
-- **Set Parameters**
+
+* **Set Parameters**
   Given a list of parameter names, it will request an update of the values subject to validation of the values.
   The updated values can include unsetting the value.
   It will provide an API that can atomically update a set of values such that if any of the values fail validation, none of the values are set.
   The success or failure of this call will be available to the client for each update unit.
   Validation of the values is expected to return as quickly as possible and only be related to accepting or rejecting the set request.
   It should not take into account how the changed value may or may not affect ongoing system performance of the node or the greater system.
-- **List Parameters**
+
+* **List Parameters**
   Provide a list of parameter names which are currently set.
-- **Describe Parameters**
+
+* **Describe Parameters**
   Given a list of parameter names, return their datatype.
 
 This functionality will be exposed through a user API which will support both local API calls as well as invocations on remote nodes via a ROS Service API.
@@ -155,7 +160,6 @@ Specific instances could be launched in different namespaces to support differen
 
 A pattern developed in ROS 1.0 was the `searchParam` mode where a parameter could be set in a namespace and the parameter query would walk up the namespace to search for the parameter.
 A similar behavior can be implemented by allowing the search parameter implementation to walk across the different nodes in hierarchical order.
-
 
 ### Parameter API
 
@@ -181,17 +185,21 @@ The definition of the services to use for interacting remotely are contained in 
 
 Currently there a few parts of the specification unimplemented.
 
-- No parameter subscription registration.
+* No parameter subscription registration.
   The events are published, but there is not way to register a callback for changes to a specific parameter.
   You can currently register for a callback on all changes for parameters of a node.
-- The ability to register callback to validate parameter updates prior to them being updated is not available.
-- There has been no work on logging and playback of logged parameter changes.
-- The ability to list and get expected validation policy has not been implemented.
+
+* The ability to register callback to validate parameter updates prior to them being updated is not available.
+
+* There has been no work on logging and playback of logged parameter changes.
+
+* The ability to list and get expected validation policy has not been implemented.
   It is expected to operate at a slightly higher level than parameters, and it possibly will be related to the component life cycle.
 
 ## Topics not covered at the moment
 
-Going forward, there are still topics to discuss and flesh out either in this document or others. A few to highlight.
+Going forward, there are still topics to discuss and flesh out either in this document or others.
+A few to highlight:
 
 ### Parameter initialization
 
@@ -205,7 +213,7 @@ Adding support for arrays in the interface is relatively straight forward.
 It slightly increases the complexity of the API for users, but can support several use cases.
 A use case for arrays of numbers is expressing a matrix or vector, addressing each position in a matrix by some sort of row-column naming scheme can get very cumbersome.
 
-### Predeclared interface to support static checking/validation.
+### Predeclared interface to support static checking/validation
 
 The ability to declare an API which can help with static checks and prevent logical errors which arrise from setting the wrong parameter based on a typo.
 The node could enforce this by rejecting unexpected names, but there are some cases where knowing the expected parameter names would be useful for developer tools.
--- a/articles/055_ros_parameter_design.md
+++ b/articles/055_ros_parameter_design.md
@@ -40,57 +40,47 @@ Other resources related to the parameter design process for ROS 2.0 include:
 
 ## Ideal System
 
-It is useful to consider the ideal system to understand how it relates to the current system and how a new system could work.
-We'd like to support both the use cases of ROS 1.0 built in parameters as well as dynamic parameters.
-An ideal parameter system would have the qualities laid out in the following paragraphs.
+It is useful to consider an ideal system to understand how it would work and how it would relate to the current system.
+An ideal system would support for the combined use cases of ROS 1.0's built-in parameters as well as ROS 1.0's dynamic parameters system.
+Based on that criteria an ideal system would be able to:
 
-### Set parameter values
+- **Set parameter values**
 
-At the most basic, a parameter system must be able to accept the setting of parameter values.
-An extension is to accept a group of parameters updated atomically.
-This allows you to update coupled parameters such as PID gains without worring about invalid intermediate states.
+    This includes setting groups of parameter values atomically.
 
-### Get parameter values
+- **Get parameter values**
 
-Also at the core functionality return the value of one or more parameters which have previously been set.
-If the value of a parameter is requested which has not been set it will explicitly return `unset`.
-Parameters can be queried individually or in groups atomically so as to get a consistent state.
+    This includes getting groups of parameter values atomically.
 
-### Unset a parameter value
+- **Unset parameter values**
 
-If a parameter has been set it should be able to be unset.
+    This includes unsetting groups of parameter values atomically, but may be a special case of setting groups of parameters atomically.
 
-### Introspection of available parameters
+- **List currently set parameters**
 
-Provide a list of currently set parameters.
+- **Provide notifications when parameter values change or update**
 
-### Notifications of change
+- **Reject changes to parameter values**
 
-When a parameter changes value it should be possible to get a notification.
+    This implies that some entity has the authority to reject or accept a change based on arbitrary criteria.
+    This would also include the ability to convey at least part of the criteria for the acceptance of a change to external actors.
+    For example, communicating the range for an integer or a few choices for a string.
+    This type of information would be used to generate generic user interfaces, but might not capture all criteria.
 
-### Reject parameter changes
+- **Check what parameters are expected to be set for a particular Node**
 
-If a parameter change is requested to an invalid value, the owner of the parameter should be able to reject the change.
-Related to this, it should also be able to communicate the preconditions which are necessary but not sufficient to determine the acceptance of a change.
+    This would help prevent logical errors which arrise from setting the wrong parameter based on a typo.
+    The node could enforce this by rejecting unexpected names, but there are some cases where knowing the expected parameter names would be useful for developer tools.
 
-### Introspection of expected parameters
+- **Provide clear rules on the lifetime of a parameter**
 
-When using parameters, providing a list of expected parameters can prevent accidentally setting parameters with typos in their names etc.
+    These rules would define what the lifetime of the parameter will be and what conditions will clear its value.
 
-### Control parameter lifetime
+- **Address all parameters without ambiguity in the names**
 
-When parameters are set, there should be an understanding of what the lifetime of the parameter value will be and what conditions will clear its value.
+    For example, one of the challenges of the current system is that there is a naming ambiguity between nodes and parameters `/foo/bar/baz` could be a node `/foo/bar/baz` or a private parameter `baz` on node `/foo/bar`.
 
-### Unambigious naming
-
-All parameters should be unambiguously addressable.
-
-For example, one of the challenges of the current system is that there is a naming ambiguity between nodes and parameters `/foo/bar/baz` could be a node `/foo/bar/baz` or a private parameter `baz` on node `/foo/bar`.
-
-### Logging
-
-All changes to parameters should be loggable.
-And when playing back a log file the parameter changes should be able to be re-requested by the log playback.
+- **Be logged for playback and analysis**
 
 ## Proposed Approach
 
--- a/articles/055_ros_parameter_design.md
+++ b/articles/055_ros_parameter_design.md
@@ -70,7 +70,7 @@ When a parameter changes value it should be possible to get a notification.
 
 ### Reject parameter changes
 
-If a parameter change is requested to an invalid value, the parameter server should be able to reject the change.
+If a parameter change is requested to an invalid value, the owner of the parameter should be able to reject the change.
 Related to this, it should also be able to communicate the preconditions which are necessary but not sufficient to determine the acceptance of a change.
 
 ### Introspection of expected parameters
@@ -89,14 +89,14 @@ For example, one of the challenges of the current system is that there is a nami
 
 ### Logging
 
-All changes to parameters should be visible when logging.
-And when playing back a log file the parameter changes should be able to be reapplied.
+All changes to parameters should be loggable.
+And when playing back a log file the parameter changes should be able to be re-requested by the log playback.
 
 ## Proposed Approach
 
 To cover the feature set above, the ROS 2.0 parameter system is proposed as follows.
 
-### Parameters Hosted on nodes
+### Parameters Hosted in Nodes
 
 For the sake of validating parameter lifecycle, all parameters will be hosted on a node.
 Their lifetime will be implicitly tied to the nodes lifetime.
@@ -118,7 +118,7 @@ The value can be one of the following datatypes:
 - `bool`
 - `bytes[]`
 
-The datatypes are chosen as non-complex primatives.
+The datatypes are chosen as non-complex datatypes, as defined in the [interface definitions article](articles/interface_definition.html)
 The full complement of datatypes of different bitdepth and unsigned types are avoided to allow interpretation from text based configuration files.
 
 `bytes` are included to allow the storage of binary blobs.
@@ -148,18 +148,20 @@ The node will allow for the registration of a callback for custom parameter valu
 
 ### Backwards compatibility Parameter Server like behavior
 
-There are use cases where the older behavior with parameters persisting beyond the duration of a specific node are valuable.
-To this end we propose to write a simple node which emulates the policy of the ROS 1.0 parameter server: it runs in namespace `/` and simply accepts all change requests.
+
+There are use cases where the older behavior with parameter server was useful.
+Both persisting beyond the duration of a specific node is valuable as well as having parameters with no specific association to a node which would potentially own or validate the values.
+To this end we propose to write a simple node which emulates the policy of the ROS 1.0 parameter server: it runs in namespace `/` and simply accepts all changes requested.
 
 ### Parameter API
 
 The client libraries will provide an API for interfacing with the Core Parameter API for both local and remote nodes including return codes.
 
-### Parameter Change notification
+### Parameter Events
 
-Each node will provide a topic on which parameter changes will be published.
+Each node will provide a topic on which parameter events will be published.
 This topic is to support monitoring parameters for change.
-It is expect that client libraries will implement the ability to register callbacks for specific parameter change notifications using this topic.
+It is expected that client libraries will implement the ability to register callbacks for specific parameter changes using this topic.
 
 ### Logging and playback
 
--- a/articles/055_ros_parameter_design.md
+++ b/articles/055_ros_parameter_design.md
@@ -41,7 +41,7 @@ Other resources related to the parameter design process for ROS 2.0 include:
 ## Ideal System
 
 It is useful to consider the ideal system to understand how it relates to the current system and how a new system could work.
-We'd like to support both the use cases of standard parameters as well as dynamic parameters.
+We'd like to support both the use cases of ROS 1.0 built in parameters as well as dynamic parameters.
 An ideal parameter system would have the qualities laid out in the following paragraphs.
 
 ### Accept parameter values
@@ -53,6 +53,7 @@ This allows you to update coupled parameters such as PID gains without worring a
 ### Return parameter values
 
 Also at the core functionality return the value of one or more parameters which have previously been set.
+If the value of a parameter is requested which has not been set it will explicitly return `unset`.
 
 ### Unset a parameter value
 
@@ -168,7 +169,8 @@ The defintion of the services to use for interacting remotely are contained in t
 Currently there a few parts of the specification unimplemented.
 
 - No parameter subscription registration.
-  The events are published, but there is not way to register a callback for changes to a parameter.
+  The events are published, but there is not way to register a callback for changes to a specific parameter.
+  You can currently register for a callback on all changes for parameters of a node.
 - The ability to register callback to validate parameter updates prior to them being updated is not available.
 - There has been no work on logging and playback of logged parameter changes.
 - The ability to list and get expected validation policy has not been implemented.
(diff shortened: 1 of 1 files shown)


articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Introspection of available parameters
+
+Provide a list of currently set parameters.
+
+### Notifications of change
+
+When a parameter changes value it should be possible to get a notification.
+
+### Reject parameter changes
```
adolfo-rt: If I understand correctly, this would be the method for enforcing parameter ranges (among other things). And 'be able to communicate the preconditions' would refers to a mechanism that could be used to, for instance, display valid parameter values or ranges, as `rqt_reconfigure` currently does. Is my interpretation correct?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+A tool called dynamic_reconfigure was developed to service this use case.
+It provided a service based interface to interact with parameters of other nodes.
+
+### Other resources
+
+Other resources related to the parameter design process for ROS 2.0 include:
+
+- Gonzalo's research on parameters.
+  - Discussion: https://groups.google.com/forum/#!topic/ros-sig-ng-ros/YzCmoIsN0o8 and https://groups.google.com/forum/#!searchin/ros-sig-ng-ros/parameter/ros-sig-ng-ros/fwDBcei5Ths/L6ORPfjUDXYJ
+  - Protype: https://github.com/abellagonzalo/dynamic_config
```
jackieokay: prototype (not protype)

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Accept parameter values
+
+At the most basic a parameter system must be able to accept the setting of parameter values.
+An extension is to accept a group of parameters updated atomically.
+This allows you to update coupled parameters such as PID gains without worring about invalid intermediate states.
+
+### Return parameter values
+
+Also at the core functionality return the value of one or more parameters which have previously been set.
```
dirk-thomas: _Return_ as well as _unset_ need to support groups of parameters atomically if _setting_ those is supported.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+- Gonzalo's research on parameters.
+  - Discussion: https://groups.google.com/forum/#!topic/ros-sig-ng-ros/YzCmoIsN0o8 and https://groups.google.com/forum/#!searchin/ros-sig-ng-ros/parameter/ros-sig-ng-ros/fwDBcei5Ths/L6ORPfjUDXYJ
+  - Protype: https://github.com/abellagonzalo/dynamic_config
+  - Final Notes:  [http://wiki.ros.org/sig/NextGenerationROS/Parameters](http://wiki.ros.org/sig/NextGenerationROS/Parameters)
+- Thibault's nodeparam draft REP: https://github.com/tkruse/rep/blob/nodeparam/nodeparam-REP.rst
+
+## Ideal System
+
+It is useful to consider the ideal system to understand how it relates to the current system and how a new system could work.
+We'd like to support both the use cases of standard parameters as well as dynamic parameters.
```
jackieokay: Can you define standard parameters in this section? Are standard parameters just ROS 1 parameters?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+
+Each parameter consists of a key and a value.
+The key is a string value.
+The value can be one of the following datatypes:
+
+- `float64`
+- `int64`
+- `string`
+- `bool`
+- `bytes[]`
```
dirk-thomas: This looks like an arbitrary list. Why is there no array of any type beside `bytes`?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+When a parameter changes value it should be possible to get a notification.
+
+### Reject parameter changes
+
+If a parameter change is requested to an invalid value the parameter server should be able to reject the change.
+Related to this it should also be able to communicate the preconditions which are necessary but not sufficient to determine the acceptance of a change.
+
+### Introspection of expected parameters
+
+When using parameters providing a list of expected parameters can prevent accidentally setting parameters with typos in their names etc.
```
jackieokay: Another grammar nitpick: comma after "When using parameters". The sentence is a bit hard to parse in its current state.

actually, don't worry about these comments, I'll address the little things in my own branch.

articles/055_ros_parameter_design.md
tfoote: Exactly.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Parameter Change notification
+
+Each node will provide a topic on which parameter changes will be published.
+This topic is to support monitoring parameters for change.
+It is expect that client libraries will implement the ability to register callbacks for specific parameter change notifications using this topic.
+
+### Logging and playback
+
+When logging an entire system the parameter changes can be logged via standard topic recording of the events channel.
```
jackieokay: What is meant by an "entire system"? The entire ROS 2 graph? Or can it be a subset of the nodes involved in getting/setting parameters?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+- No parameter subscription registration.
+  The events are published but there is not way to register a callback for changes to a parameter.
+- The ability to register callback to validate parameter updates prior to them being updated is not available.
+- There has been no work on logging and playback of logged parameter changes.
+- The ability to list and get expected validation policy has not been implemented.
+  It is expected to operate at a slightly higher level and possibly be more closely related to the component life cycle.
+
+## Topics not covered in at the moment
+
+There are still topics to discuss and flesh out either in this document or others going forward a few to highlight.
```
jackieokay: I don't really understand the last few words of this sentence. Should it be "Going forward, there are still topics to discuss and flesh out either in this document or others. A few to highlight:"

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Parameter initialization
```
jackieokay: Is this supposed to be a subsection of "Topics not covered at the moment"? It should probably have 3 leading `#`s.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,184 @@
...
+
+The above specification has been prototyped; the implementation can be found in [rclcpp](https://github.com/ros2/rclcpp).
+The defintion of the services to use for interacting remotely are contained in the [rcl_interfaces package](https://github.com/ros2/rcl_interfaces)
+
+### Unimplemented
+
+Currently there a few parts of the specification unimplemented.
+
+- No parameter subscription registration.
+  The events are published, but there is not way to register a callback for changes to a parameter.
```
esteve: There is the `on_parameter_event` method.

https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/parameter_client.hpp#L232

articles/055_ros_parameter_design.md
tfoote: Added a line after to clarify it will return `unset` explicitly.

dirk-thomas: My comment was referring to the need to query a set of parameters atomically. The new sentence does not yet cover that.

articles/055_ros_parameter_design.md
tfoote: We chose non-complex non-custom types. And do not provide the full complement of things like unsigned to support scripting.  I'd prefer not to support `bytes` but when we reviewed the API it was deemed important to support binary blobs of which bytes[] was the default choice. 

In our discussions support for arrays was discussed and deferred deciding on adding support for later discussions. I'll add this to the open questions below.

dirk-thomas: I think the support of arrays of e.g. `int64` is crucial to consider in the article. A lot of existing code sets array of primitive data on the parameter server.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+  - Final Notes:  [http://wiki.ros.org/sig/NextGenerationROS/Parameters](http://wiki.ros.org/sig/NextGenerationROS/Parameters)
+- Thibault's nodeparam draft REP: https://github.com/tkruse/rep/blob/nodeparam/nodeparam-REP.rst
+
+## Ideal System
+
+It is useful to consider the ideal system to understand how it relates to the current system and how a new system could work.
+We'd like to support both the use cases of ROS 1.0 built in parameters as well as dynamic parameters.
+An ideal parameter system would have the qualities laid out in the following paragraphs.
+
+### Accept parameter values
```
wjwwood: I'd call this section "set" and the next "get" parameter values. This makes it line up with "unset" and it avoids "accept" which implies that it could have not been accepted ("reject"), which is not covered in the following paragraph.

articles/055_ros_parameter_design.md
tfoote: Added paragraph on arrays at bottom of the article for things to address.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+
+Provide a list of currently set parameters.
+
+### Notifications of change
+
+When a parameter changes value it should be possible to get a notification.
+
+### Reject parameter changes
+
+If a parameter change is requested to an invalid value, the parameter server should be able to reject the change.
```
wjwwood: I'd say "the owner of the parameter should be able to reject ...", using "parameter server" implies how it should be implemented since server invokes the idea of a centralized client-server relationship.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Unambigious naming
+
+All parameters should be unambiguously addressable.
+
+For example, one of the challenges of the current system is that there is a naming ambiguity between nodes and parameters `/foo/bar/baz` could be a node `/foo/bar/baz` or a private parameter `baz` on node `/foo/bar`.
+
+### Logging
+
+All changes to parameters should be visible when logging.
```
wjwwood: "... should be loggable."?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+And when playing back a log file the parameter changes should be able to be reapplied.
```
wjwwood: I'm not sure about this line, since it's possible that any change that is played back could be rejected. Maybe it would be more appropriately scoped to say that it should be possible to playback parameter events.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Proposed Approach
+
+To cover the feature set above, the ROS 2.0 parameter system is proposed as follows.
+
+### Parameters Hosted on nodes
```
wjwwood: "Parameters hosted by nodes" or "Parameters Hosted by Nodes"? Maybe "in" instead of "by" or "on".

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+The key is a string value.
+The value can be one of the following datatypes:
+
+- `float64`
+- `int64`
+- `string`
+- `bool`
+- `bytes[]`
+
+The datatypes are chosen as non-complex primatives.
```
wjwwood: `primitives`, and `non-complex` seems poorly defined to me, i.e. what does that mean?

dirk-thomas: The terms are used in the interface article: http://design.ros2.org/articles/interface_definition.html

Maybe we should reference the article somewhere here?

wjwwood: Based on http://design.ros2.org/articles/interface_definition.html I'd say that `non-complex primitive` is redundant? Since that article describes a `non-primitive` message as complex?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+
+`bytes` are included to allow the storage of binary blobs.
+It's use is not recommended but can be very convenient, and explicitly supporting it is better than having people try to abuse other datatypes such as strings.
+
+### Core API
+
+Each node is responsible for providing the following functionality.
+
+- **Get Parameters**
+  Given a list of parameter names it will return the parameter values.
```
wjwwood, line 129: Should this be "given a node and a list of parameter names" or "a list of parameter addresses"?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+- **Set Parameters**
+  Given a list of parameter names, it will request an update of the values subject to validation of the values.
+  The updated values can include unsetting the value.
+  It will provide an API that can atomically update a set of values such that if any of the values fail validation, none of the values are set.
```
wjwwood, line 134: I think it would be clearer to have `get`, `get atomic`, `set`, and `set atomic` all called out separately, rather than making notes about atomic versions in the description.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+  The success or failure of this call will be available to the client.
```
wjwwood: Should this mention that the non-atomic set will return a success/failure for each value and that the atomic will only return one success/failure? Also should we mention the failure reason is included or is that too much of an implementation detail?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+- **List Parameters**
+  Provide a list of parameter names which are currently set.
+- **Describe Parameters**
+  Given a list of parameter names, return their datatype.
```
wjwwood, line 143: Doesn't this return more than the data type? I believe it should also include meta information like bounds and stuff that would be used programmatically by a tool like the `dynamic_reconfigure_gui`.

articles/055_ros_parameter_design.md
dirk-thomas: I would agree to that.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+This functionality will be exposed through local API calls as well as a ROS Service call API for remote operations.
+
+### Parameter update validation
+
+The node will allow for the registration of a callback for custom parameter value validation.
+
+### Backwards compatibility Parameter Server like behavior
+
+There are use cases where the older behavior with parameters persisting beyond the duration of a specific node are valuable.
+To this end we propose to write a simple node which emulates the policy of the ROS 1.0 parameter server: it runs in namespace `/` and simply accepts all change requests.
```
wjwwood: I think this might be mixing two issues. One is the desire to have a global parameter space where no one owns the parameters, i.e. they are not validated by anyone. The other is to have parameters persist when the owner is shutdown and restarted, which can apply to either parameters owned and validated by a node or to a global parameter server.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Parameter API
+
+The client libraries will provide an API for interfacing with the Core Parameter API for both local and remote nodes including return codes.
+
+### Parameter Change notification
+
+Each node will provide a topic on which parameter changes will be published.
+This topic is to support monitoring parameters for change.
+It is expect that client libraries will implement the ability to register callbacks for specific parameter change notifications using this topic.
```
wjwwood: Should this be "parameter events" rather than "parameter changes"? Events would encompass: accept (atomic?) change, reject (atomic?) change, request (atomic?) change?, get value?, request unset (might be covered by request change), and reject unset. I'm not sure how thorough this document should be.

wjwwood:
I had several comments and suggestions, but I don't think any are controversial or blockers, so +1 after they are addressed.

articles/055_ros_parameter_design.md
tfoote, line 143: Right now it only returns the datatype. Early prototypes had a lot more details in this descriptor but the problem is that requires the core api to be updated any time the validation features change or update. I'd rather see the validation preconditions etc be done via a separate channel.

articles/055_ros_parameter_design.md
tfoote: I think that's getting too detailed for the current form of this article. Once we have this we can then drive down to the API specifics.

articles/055_ros_parameter_design.md
tfoote, line 134: This is pushing toward API level choices like below. I don't think it's necessary to require that there's both a get and get atomic. For example the regular get can be atomic.

articles/055_ros_parameter_design.md
tfoote, line 129: This is the functionality the "node" is providing. I was assuming this had already been addressed to the node, vs the client API which requires the node name and parameter name to first address the right node.

wjwwood, line 129: Well, then maybe a sentence just before `- **Get Parameters**` which states something to this affect. I think that would help since we don't differentiate here between a "core api" and a "client api". It's not clear to me whether or not what's described here is for the local API or the remote, ROS based, API. And if the descriptions are agnostic, then it shouldn't assume the node name is implied, since both ways don't have it implicitly.

articles/055_ros_parameter_design.md
tfoote: I'm updating it to use parameter events nomenclature. The current implementation only publishes events of the changes. If you want the raw requests and potential rejections hopefully you can log the actual requests themselves by logging the RPC calls.

articles/055_ros_parameter_design.md
BobDean, line 134: this is something i did not understand when reading the code, why there is an atomic and a non-atomic api when just atomic would suffice.

wjwwood, line 134: @BobDean It's more like atomic and batch. With atomic, no values are set if any are rejected. With non-atomic some values may be set while others are rejected. Basically non-atomic could be replaced with many calls to the atomic version, one for each value, but that would be less efficient.

wjwwood, line 134: After reading my comment again, `get atomic` may not make sense. From my perspective, `get atomic` just means give me the current state of several values without worrying about part of them getting updated while I'm getting them.

BobDean:
what does a parameter being 'set' by the service mean (i.e. i get back a SetParameterResult of true)? that the value is now in the local cache? Or that the behavior of the node has changed as a result? 
(splitting into multiple posts for tracking)

wjwwood:
> what does a parameter being 'set' by the service mean (i.e. i get back a SetParameterResult of true)? that the value is now in the local cache? Or that the behavior of the node has changed as a result? 
> (splitting into multiple posts for tracking)

I think that must mean that the parameter has been set. Whether or not the parameter has changed the behavior of the node depends on how the user write their code. But it's an important distinction.

I could imagine that you have a few possible types of callbacks the user could provide (not exhaustive and not all maybe be needed or a good idea):
- Validate, which gets called for an individual parameter.
  - Validate batch, which gets called when a group of parameters are changed (maybe grouping is done by namespace, I don't know)
- On update, called anytime a parameter is set and accepted.
- On change, called anytime a parameter is set and accepted and is different from the previous value.
  - Both this and on update could have other settings like coalescing or throttling, and both could have batch versions.

The validate and on update could be collapsed into each other, where on update returned true or false for whether or not the change was accepted.

If the user changes the behavior in the validate callback, then a successful set would indicate the node's behavior changed. If they did so in an on change or separate on update callback then a successful set may not indicate node behavior has changed. Since we cannot control when or where the user makes use of the parameter then I think we cannot tie that to the successful result of a set call.

articles/055_ros_parameter_design.md
wjwwood, line 134: I guess my latest comment isn't that clear either, let me try one more time:

What I meant by `get atomic` is it gets the value of several parameters with a guarantee that they're not being updated at the same time. What I mean by "may not make sense" is that you could argue that all gets should be atomic, therefore removing the need for a specifically atomic get.

BobDean:
From the cpp code, it seems to me that you are thinking of a parameter as a being set of separate entities, vs an entity with multiple parts (struct/class). We had a similar split api as the proposed.  After using it for a while, the realization was that in most cases we were usually making all of calls to gather all of the metadata.  We then switched to an api which worked on a single ConfigValue type for the network aspects. 

So for rframe the equilvalent of ParameterValue, ParameterType, ParameterDescriptor etc is a single ConfigValue data type:

PrimitiveEnum type // bool, string, unit, int etc
string value // value are stored as strings, arguable w.r.t. efficiency
string comment // a descriptive comment, required in our api. 
string  source // the name of whoever or whatever set the value last, source code, config file, gui, great for debugging
timestamp ts  // change timestamp, ns since epoch

the basic api then is:
get(paramName,param,defaultValue,comment)  // sets the value if it does not exist, retrieves it if id does (i.e. loaded from a config file)
get(paramName,value)  // retrieve a value which _should_ exist. returns NOT_FOUND (unset), CHANGED or NOT_CHANGED
set(paramName,value)  // set existing param, fails if not found, side effect: changes the source and timestamp values
keyMap() // returns the parameter set as a map<paramName,configValue>
count(paramName)  // number of duplicate keys, we allow them as xml and other markup formats allow them.

get/set then have versions for handing vector<type> to get multiple instances, why we support that is another discussion.

our executor then has a callback for a ConfigurationChange, which sets the local value from an external source.  we do not currently have "sets" of parameters, but i am thinking about it after looking at the rclcpp api. if there is a race condition related to a parameter change, we handle using other techniques (currently).

BobDean:
i agree that what occurs within the set is an important distinction. in my opinion the set should return
very quickly, which implies that it should not result in a change to node
behavior. What is 'ok' for that change to be is a wide range of
possibilities connected to LifeCycle and safe operation of the robot, which
may take a while to process and not immediately return.

Validation is a solid point. If there was a way to handle it within the
parameter api, that would be my preference (i think someone mentioned this
above).  maybe when the value is initialized a valid parameter range is
passed in. or a lambda.  in other words, using a callback for validation is
fine, but i do not see a need to include parameter validation in a callback
group.

(also, why can i not get my comment to trail the post i am responding to?)

tfoote:
A parameter being set means that it's value has been updated in the node's parameter storage and that all callbacks registered related to the parameter update validation have returned true. We cannot make guarentees about the behavior of the node but if you query the parameter from anywhere, local or remotely it will be the new value (assuming you are querying before another change has made).

As @wjwwood mentioned it is up to the implementer to use the parameter values. If the code has latency in taking in the parameter, such as it will reprocess parameters on the next cycle for a looping process, the parameter is still accepted and set. I don't think it would be a good idea to recommend that the parameter setting block until the next cycle once the parameter has begun to be acted upon. You can take it a step further and not just ask has the node changed behavior but has the hardware changed behavior. Lets take the case that you're setting the desired velocity for a spinning sensor which has a large inertia. If it's spinning at 10 Hz and you ask it to go to 15Hz. The main thread is running in isolation, reading the target velocity once per revolution. Accelerating from 10Hz to 15Hz will take 1-2 seconds. With this setup there are three different times you could report that the parameter has been successfully set. You could report it once the value is validated and recorded, once the value has been passed to the other thread or once the system has completed it's change to the requested value. 

To be able to track whether a goal point has been achieved I think it's improper to try to use the return code from the parameter set call as a way to measure success. The feedback on the state of the system should come from a separate channel such as the measured velocity from an encoder being streamed on a topic. Similarly it's quite possible that if there's a low battery voltage or other side effect out of the purview of the node which makes a perfectly valid target RPM is unachievable. In this case it would be a bad idea for the set parameter call to blocked until achievement of the goal. 

Since we've determined that waiting a potentially indeterminate amount of time for the hardware to react, I think that the same can be said for the software. In the case that I outlined it would be possible for the callback to block until the active thread notified it that it had read the new value and was operating upon it. However that's still a very indeterminate value as that thread may be blocked or do something odd or improper and never actually send the command to the hardware. Thus giving a false sense of "doing something". Thus I would suggest that if introspection of the current set point is important that it should be instrumented for streaming feedback via topic.

Another reason for not relying on the callbacks to imply that the "node is acting on my request" is that a successful callback does not guarantee the state of the remote node either as anther process might have made a different successful request immediately following your request. You could also register for callback on change of the parameters. 

Similarly if the set parameter calls block pending execution or partial execution you also need to start getting into determining queuing behavior and potentially preemption such that two requests coming in almost simultaneously will cause the first one to "fail" as it's value was overriden before the next cycle, or else it queues up and then gets applied in the following cycle as the first callback was still waiting for the full cycle. As I describe this it sounds more like we're getting into the territory of actions instead of parameters. It seems that we should set the policy to returning quickly once the input has been validated. 

I see you replied while I was composing this, and you agree. Since you asked it's clearly not obvious. How much should be added to the document to clarify?

articles/055_ros_parameter_design.md
BobDean: what is the reasoning behind not supporting 16 and 32 bit values?  Is it assumed that the host language or client access api will take care of the conversion?

articles/055_ros_parameter_design.md
BobDean: i recommend:
-   the response of the set operation (local or remote) will only be related to the node accepting or rejecting the set request. It will not include a concept of how the change may or may not have affected the behavior of the system. +   Verification callbacks should be designed to execute as quickly as possible, and not include information external to the parameter metadata, with the exception of managed lifecycle state.

"external to"  is meant to imply that a node may reject a parameter due to being in the wrong lifecycle state. ideallly the "metadata" would include enough information to validate the value.   Other than these two cases, what are valid reasons to reject the set request?  If there are edge cases, i believe they would be the responsibility of system architecture, and not a parameter api.

articles/055_ros_parameter_design.md
BobDean, line 143: recommendation that there is metadata that is part of the description, but is not required for validation. example: comment describing what the parameter does (so parameters are self documenting), the source of the parameter (config file, another node, command line...). 
- Given a list of parameter names, return their descriptive metadata including datatype, description, last update time, and source of last update.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Topics not covered at the moment
+
+Going forward, there are still topics to discuss and flesh out either in this document or others. A few to highlight.
+
+### Parameter initialization
+
+There are several ways to load parameters at startup including command line arguments, roslaunch arguments, and potentially parameter files.
+This is something which should be addressed in conjunction with the new launch system.
+
```
BobDean, line 208: the more 'required' bits of metadata at initialization the better, as it helps with self documentation. i recommend that initialization occur at a single location in the node.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,202 @@
...
+- **List Parameters**
+  Provide a list of parameter names which are currently set.
+- **Describe Parameters**
+  Given a list of parameter names, return their datatype.
+
+This functionality will be exposed through local API calls as well as a ROS Service call API for remote operations.
+
+### Parameter update validation
+
+The node will allow for the registration of a callback for custom parameter value validation.
```
BobDean: has thought to how to ship the validation around occurred? or if it is needed? The valdiation parameters could be bound to the callback.  Given that parameter updates are rare in the system messaging lifetime, having a "validateParameter" api call which validates the value on the node without setting it could enable a dynamic configure like api to work without needing to ship around the validation metadata, or require that validation meet a well defined interface.

wjwwood: You've sort of asked this in more than one place, but validation happens in the node which owns it and can be arbitrarily complex. The metadata is only provided so that simple bounds can be placed on input so that graphical tools can render widgets for the user to use. For example, the Kinect driver might have a parameter for resolution and frequency, which are type int. The metadata would indicate that only values 0-2 are valid for each of those parameters and that each value corresponds to some helper text (like the corresponding resolution or frequency). However, the validation maybe more sophisticated, because the user might select two independently valid selections like high resolution and high rate, but the node can still reject them because that combination of parameters is not valid. This part of the validation is not captured in a machine readable format in the meta data, so it isn't used in the parameter setting UI, but is checked when trying to set the parameters.

tfoote: Allowing a query for in-depth validation without applying is something we could consider. But I don't think that replaces communicating the validation outward. It's very valuable for the guis to be able to have a concept of valid values to be able to render things like slider bars or drop down lists. There are certainly things which cannot be validated remotely. For example checking on local assets. However we could consider exposing a service where the gui could query for prevalidated ranges/values. So that when a gui is pointed at parameter foo. It will call the parameter server and ask get_foo_valid_inputs and it returns something expressing the range [-1,1] for a float to be rendered as a slider or a list of detected serial ports for a drop down device configuration in the GUI. I think this would also be a better place to consider including the description as the description is also something which is useful to users but not to other nodes. It could be rendered next to the slider or other visual element.

articles/055_ros_parameter_design.md
tfoote: I think limiting the verification to the specific metadata is overly limiting. It's not the best example, but you could say that a request to change a serial driver to a non-extant device could be rejected. Of course that depends on if your driver can support running and polling for a non-existant device to become available, in which case it might be ok to configure it for a non-extant device. 

There are other things on the system which are not possible to become available that parameters might reflect such as the number of cores, or a core to which you want affinity for. Or the GPU device id. 

Many of these things can be covered by what I talked about in the other comment with querying for a set of valid inputs, instead of just the conditions. And thinking about this more I see it as generalizing that we want to be able to get a `usage:` style description for any parameter. And more generally this will apply to the whole Node API with services and topics with descriptions and potentially dynamic validation ranges.

I'm going to update this section to try to clarify from @BobDean and @wjwwood.

BobDean:
There are two implications that I see from the design which I do not see directly discussed:

1) There is an implication that parameters may be set while the robot is in an unsafe state. Where unsafe is described as one where changes may cause unexpected results, or cause a failure which may result in a safety problem for nearby humans and objects. Versus putting the robot into a safe state, changing parameters, and resuming the behavior. An example of where safety procedures overlap with aspects of the api and could replace parts of the api. If agreed, then text should be added that parameters may be set in an unsafe state, and it is then the responsibility of the node to deal with that change in a safe manner. 

2) There is an implication that multiple controllers may be changing parameter state at the same time. Use cases describing how this may occur would be useful to further discussions on how to protect against such a case.

wjwwood:
> 1) There is an implication that parameters may be set while the robot is in an unsafe state. Where unsafe is described as one where changes may cause unexpected results, or cause a failure which may result in a safety problem for nearby humans and objects. Versus putting the robot into a safe state, changing parameters, and resuming the behavior. An example of where safety procedures overlap with aspects of the api and could replace parts of the api. If agreed, then text should be added that parameters may be set in an unsafe state, and it is then the responsibility of the node to deal with that change in a safe manner.

+1 to more clearly state this in the design doc. From my perspective this is a more advanced behavior for parameters. So just plain parameters shouldn't have any constraint on when they can be set, but the node can always validate it. When a node has life cycle imposed on it (we've been calling this a managed node, but that's up for discussion), then we can also enforce that only a "managed parameter" can be used with the "managed node". This "managed parameter" can hook in before the user's validation callback for all parameter sets and ensure that the parameter is not changed when in the wrong state. Therefore these nodes will require you to bring them out of the active state before changing a parameter. Regardless of the details of when or why parameter changes will be automatically rejected, I think it should be built on top of basic parameters, not into them.

BobDean:
My concern was primarially with node behavior safety. The "setting in an
unsafe state" is more tied to how the parameter set interacts with the
running behavior, which I feel is separate from lifecycle.  It is important
to lifecycle, but it should be possible for someone to write a "safe" node
without it.  In other words, the set parameter call is a "request" to the
node. The node should respond to it as the designer feels is appropriate.
(which I believe we all agree on, yes?)

tfoote:
Yes I agree that any chance of parameter is a request and the node is allowed to reject it for "safety" reasons or any other policy from the designer. 

Sort of along the same lines as William was suggesting I think it would be good for us to have some recommended policies/patterns which relate to the managed nodes and lifecycles potentially with wrappers which implement that for you.

BobDean:
I may have missed it, but it there an ability to share a "global" or
"common" parameter? such as a common url for a resource, or operational
policy which effects multiple nodes.

tfoote:
@BobDean that's in the paragraph [Backwards compatibility Parameter Server like behavior](https://github.com/ros2/design/pull/58/files#diff-4075c1e78328ec132523517becd96e1bR151)

jack-oquin:
Aren't global parameters more than just a "backwards compatibility" concern?

The concept seems useful in its own right. Some parameters like `use_sim_time` and the old `tf_prefix` are inherently global, they do not apply to just one node.

mikeferguson:
> The concept seems useful in its own right. Some parameters like use_sim_time and the old tf_prefix are inherently global, they do not apply to just one node.

Or, even more common: the robot_description URDF parameter.

tfoote:
I think that the unowned parameters are a pattern that generally should be discouraged. The example of the `robot_description` is a good example where it should have a more clearly defined interface to deal with changing values. https://github.com/ros/robot_state_publisher/issues/29

There are very few pieces of data which are truely global and unchaning like `use_sim_time`.

jack-oquin:
How will unowned parameters be discouraged?

Fairly often, I find it useful to define parameters at a higher level in the namespace for use by more than one node. Isn't that what `searchParam()` is for?

In most cases, there can be a node that "owns" each of those parameters. But, what happens if that node should terminate unexpectedly? Will the parameter persist across a respawn? It would in ROS 1.

Instead of dismissing them as a backwards compatibility option, I think global parameters deserve some serious design treatment.

wjwwood:
> Fairly often, I find it useful to define parameters at a higher level in the namespace for use by more than one node. Isn't that what searchParam() is for?

@jack-oquin It would be fine to have something like `searchParam()` which works in combination with the "list params" function we described here. One of the things we've avoided discussing so far is addressing in ROS 2. I think there is a need for a more formal addressing scheme in ROS, but that's a topic for another article. The only reason I bring it up is that it should be possible to have hierarchies of nodes and parameters within namespaces and a tool like `searchParam()` will still be useful.

> In most cases, there can be a node that "owns" each of those parameters. But, what happens if that node should terminate unexpectedly? Will the parameter persist across a respawn? It would in ROS 1.

I know that's a useful feature that a lot of people use and we've talked about how to preserve that behavior if it is desired, but it's not mentioned explicitly here either as a use case or as something that needs addressing.

@tfoote we should add that point, or I can if you don't have time.

My personal opinion is that for ROS 2 it should be done by a tool which is actually a normal node. This node would act similar to how it's done in ROS 1, but would not use any special interfaces or features if possible. In ROS 1 a node starts up, then contacts the parameter server and gets the current values (either defaults or what was set from a previous run). The generic parameter server node in ROS 2 could provide the same service, which nodes could optionally query before taking the hard coded defaults in their code. Like the ROS 1 parameter server, this generic parameter server node in ROS 2 could provide a global namespace which all nodes could reference and search through and could hold common global parameters like `robot_description` (though I've long thought that should be a latched topic in ROS 1 so it could be captured by a bag file). In both cases parameters are lost if the parameter server is shutdown or crashes and in both cases nodes could store parameters in it and recover them after the node crashes or restarts.

So I think that even if all parameters must be owned by a node, i.e. no true global parameters, you can model this pattern using nodes. We just need to provide the right tools and we could include an instance of this parameter server node in something like the `roscore` command.

> Instead of dismissing them as a backwards compatibility option, I think global parameters deserve some serious design treatment.

I agree, we should describe the motivation for this feature and describe how it can be accomplished with the proposed solution and let people comment on that.

articles/055_ros_parameter_design.md
gbiggs: Doing things like checking for hardware presence when validating a parameter is introducing complex functionality into the handling of parameters. I'm not arguing for whether this is right or wrong - it's a system design issue, in my opinion, so the capability should probably be there. However, at least for managed nodes, it needs to be restricted. If the node is active and allows parameter changes while active, then validation probably should be restricted as @BobDean suggests to avoid impacting execution schedules. Changes that require more involved validation would then be restricted to being performed while the node is inactive.

I guess what this typically poorly-worded comment is saying is that there may need to be a distinction between parameters that can be deterministically (in time) set and parameters that cannot, with a restriction on when the latter can be changed.

tfoote:
I've added some more details about how to use a parameter server node in parent namespaces and added a section on `searchParam` to address the above comments.

articles/055_ros_parameter_design.md
tfoote: In general shorter bit values are important for space and speed optimization. However supporting those optimizations increases the complexity of the interface. In this case it would close to double the number of datatypes supported. People would then have to be careful about up and down conversions and range checks which becomes noteably harder when interacting with loosely typed languages.

dirk-thomas:
The bullet `Check what parameters are expected to be set for a particular Node` listed under `Ideal System` doen't comes from the capabilities in ROS 1. I don't see a reason to consider this necessary for an "ideal system". Having a node prepopulate all its valid parameters and reject others seems to be sufficient. The valid parameter names can also be exposed as "constants" to other entities (if they don't want to hard code the values into their code). Therefore I would propose to remove that specific bullet (and the later remark that it hasn't been implemented anyway).

If the part should stay in the article (which I think it shouldn't) it would be better suited in the "Topics not covered at the moment" section at the end.

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,207 @@
...
+- **List Parameters**
+  Provide a list of parameter names which are currently set.
+- **Describe Parameters**
+  Given a list of parameter names, return their datatype.
+
+This functionality will be exposed through a user API which will support both local API calls as well as invocations on remote nodes via a ROS Service API.
+
+### Parameter update validation
+
+The node will allow for the registration of a callback for custom parameter value validation.
```
dirk-thomas: The wording of this sentence sounds very fuzzy to me. It could be interpreted that external entities can register these validation callbacks which I think is not the case?

articles/055_ros_parameter_design.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Search parameter behavior
+
+A pattern developed in ROS 1.0 was the `searchParam` mode where a parameter could be set in a namespace and the parameter query would walk up the namespace to search for the parameter.
+A similar behavior can be implemented by allowing the search parameter implementation to walk across the different nodes in hierarchical order.
+
+
+### Parameter API
+
+The client libraries will provide an API for interfacing with the Core Parameter API for both local and remote nodes including return codes.
```
dirk-thomas: This sentence seems to say almost the same as the last paragraph of "Required functionality" (https://github.com/ros2/design/pull/58/files#diff-4075c1e78328ec132523517becd96e1bR141).

dirk-thomas:
Beside my comments I added a few commits to the PR rephrasing little things and adding a some additional information.

tfoote:
@BobDean 

> (also, why can i not get my comment to trail the post i am responding to?)

Github doesn't support branching "threading" of replies. If you want to reply to something specific the best thing to do is to quote it and then reply. You get specific sub-threads from line comments, but they're kind of volatile as if the line of code is updated the whole sub-thread can vanish.

tfoote:
I've addressed @dirk-thomas 's comments. I'd like to suggest we merge this now and get it posted on the site for others to find. I'm sure there will be more updates to come but it's at a relatively self consistent state at the moment.

tfoote:
Thanks everyone for the feedback. It's now on the site at: http://design.ros2.org/articles/ros_parameters.html

PR #64, opened 2015-09-28
Update 030_ros_with_zeromq.md

scpeters:
I think `without serialization` was a typo.
It seemed grammatically incorrect as well.
Please propose a different fix if my interpretation is wrong.


wjwwood:
:+1:

PR #65, opened 2015-09-30
add article listing changes of ROS 2 over ROS 1

dirk-thomas:
In the ongoing discussion many posts reduce ROS 2 to a single feature, e.g. DDS. In order to not having to reiterate all the other aspects which ROS 2 aims to address this article lists the conceptional changes currently proposed.

I am sure I missed several important ones in this first draft. Please comment with anything missing (or directly add additional commits to this branch).

### DIFF
--- a/articles/changes.md
+++ b/articles/changes.md
@@ -1,0 +1,257 @@
+---
+layout: default
+title: Changes between ROS 1 and ROS 2
+permalink: articles/changes.html
+abstract:
+  This article provides an overview about the changes being made in ROS 2 compared to ROS 1.
+published: true
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+---
+
+* This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+
+## Preface
+
+Each change is described as briefly as possible but giving enough context and rational for a reader familiar with ROS 1.
+If further external information is available (e.g. other articles) it should be linked to.
+
+
+## Platforms and dependencies
+
+### Platforms
+
+ROS 1 is only being CI tested on Ubuntu.
+It is actively supported by the community on other Linux flavors as well as OS X.
+
+ROS 2 will be CI tested and supported on Ubuntu, OS X as well as Windows (see [ci.ros2.org](http://ci.ros2.org/)).
+
+
+### Languages
+
+#### C++ standard
+
+ROS 1 is targeting C++03.
+ROS 2 will use C++11 (and newer) when supported on all major platforms.
+
+
+#### Python
+
+ROS 1 is targeting Python 2.
+ROS 2 will use 3.4+.
+
+
+### Reusing existing middleware
+
+ROS 1 uses a custom serialization format, a custom transport protocol as well as a custom central discovery mechanism.
+ROS 2 will have an [abstract middleware interface](http://design.ros2.org/articles/ros_middleware_interface.html), through which serialization, transport, and discovery will be provided by one of several existing DDS implementations.
+
+DDS will enable ROS 2 to provide various Quality of Service parameters which will improve communication over different networks.
+
+
+## Build system
+
+For more information about the build system please see the [ament](http://design.ros2.org/articles/ament.html) article.
+
+
+### Support other build systems beside CMake
+
+Every ROS package is a CMake project.
+In ROS 2 other build systems will be supported, e.g. plain Python packages.
+
+
+### Python packages
+
+In ROS 1 a package with Python code can only use a small subset of the features available in setup.py files since the setup.py file is being processed by custom logic from within CMake.
+In ROS 2 a Python package can use anything in setup.py files, e.g. entry points since they are being invoked with `python3 setup.py install`.
+
+
+### Environment setup
+
+In ROS 1 the build tool generates scripts which must be sourced in order to setup the environment before being able to use the built ROS packages.
+This approach only works when the ROS packages are being built with ROS specific build tool.
+
+In ROS 2 the environment setup is separated into package-specific scripts and workspace-specific scripts.
+Each package provides the necessary scripts to make itself usable after being built.
+The build tool only invokes the workspace-specific scripts which then call the package-specific scripts.
+
+
+### No non-isolated build
+
+In ROS 1 multiple packages can be built in a single CMake context.
+While this speeds up the build step, every package needs to ensure that cross package target dependencies are defined correctly.
+Additionally all packages share the same namespace which leads to colliding target names, etc.
+
+In ROS 2 only isolated builds are supported, i.e. every package is built separately on its own.
+
+
+### No devel space
+
+In ROS 1 packages can be built without installing them.
+From the devel space in combination with the source space the system is already usable.
+But every package has to actively support the devel space, e.g. in environment hooks and CMake code.
+
+In ROS 2 a package must be installed after building it before it can be used.
+
+One reason for the devel space in ROS 1 is to enable the developer to change files, e.g. Python code or launch files, and use the modified code directly without the need to rebuild the package.
+This benefit is preserved in ROS 2 by optionally replacing copy operations in the install steps with symlinks.
+
+
+### Support catkin_simple use case
+
+In ROS 1 the package catkin_simple is aiming to make writing the CMake code of ROS packages easier.
+In many cases it is not achieving this goal which is often due to restrictions of the design necessary for support features like the devel space.
+
+In ROS 2 the CMake API was restructured to support this use case.
+
+
+### Minimal support for packages without manifest
+
+In ROS 1 only packages with a manifest file are considered by the build system.
+In ROS 2 it is possible to detect packages with supported build system in folders without a manifest file.
+If the package follows common practice it might even be possible to detect some of the missing meta information (like dependencies).
+
+
+## Messages, Services
+
+For more information please see the [ROS interface definition](http://design.ros2.org/articles/interface_definition.html) article.
+
+
+### Separated namespaces
+
+In ROS 1 .msg and .srv files can have the same name but the generated code collides.
+The same is the case for the request and response parts of services.
+
+In ROS 2 the generated code uses separate namespaces to guarantee it is collision-free.
+
+
+### Optional default values in message definitions
+
+In ROS 2 primitive values in messages can now have default values, set when the message is constructed.
+Default values for non-primitive fields, i.e. nesting messages, is not yet possible.
+
+
+### Optional upper bounds for arrays and strings
+
+This is necessary in order to calculate the maximum size of a message in memory, which allows for preallocation of messages with a dynamic sizes.
+This is useful for performance and for use cases like real-time.
+
+
+### Unify duration and time types
+
+In ROS 1 the duration and time types are defined in the client libraries.
+The member names of the data structures are different in C++ (sec, nsec) and Python (secs, nsecs).
+
+In ROS 2 these types will be defined as messages and therefore be consistent across languages.
+
+
+### Remove sequence field from Header message
+
+The field has been deprecated for a long time and was not set consistently in ROS 1.
+
+
+## Client libraries
+
+### Across languages
+
+#### Notifications
+
+In ROS 1 all information about the ROS graph must be polled from the master.
+In ROS 2 changes will be published instead.
+
+
+#### Components with life cycle
+
+In ROS 1 every node usually has its own main function.
+In ROS 2 it will be recommended to subclass from a component which has a life cycle.
+
+The life cycle can be used by tools like roslaunch to start a system composed of many components in a deterministic way.
+
+For more information please see the [node life cycle](http://design.ros2.org/articles/node_lifecycle.html) article.
+
+
+#### Parameters and Dynamic Reconfigure
+
+In ROS 1 global parameters and node-specific dynamic reconfigure parameters are two separate concepts.
+In ROS 2 a unified approach is being used.
+It will be similar to dynamic reconfigure and a node named "global parameter server" will accept requests to set values unconditionally.
+In ROS 1 all this information needs to be polled for changes in ROS 2 changes will be published to notify other entities.
+
+For more information please see the [parameter design](http://design.ros2.org/articles/ros_parameters.html.html) article.
+
+
+#### Threading model
+
+In ROS 1 the developer can only choose between single-threaded execution or multi-threaded execution.
+In ROS 2 more granular execution models are available (e.g. across multiple nodes) and custom executors can be implemented easily.
+
+
+#### Graph API
+
+In ROS 1 nodes and topics can be remapped at startup time only.
+In ROS 2 it will be possible to remap those at runtime as well as create aliases.
+
+
+### C and C++
+
+#### Support for real-time
+
+ROS 1 does not support writing real-time code but relies on external frameworks like Orocos.
+In ROS 2 it will be possible to write real-time nodes when using a proper RTOS and with carefully written user code.
+
+
+### C++
+
+#### Node vs. Nodelet
+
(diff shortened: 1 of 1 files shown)


tfoote:
looks good on a quick read between flights. I agree there's more to be added, not yet covered. And there's a question of how to scope some of the higher level things. Presumably if they might have their own design article they don't go into this?

dirk-thomas:
I think that they should go into this document with a brief description and a link to the article.

Can you enumerate specific items which should be added?

tfoote:
The first one off the top of my head is component life cycle.

tfoote:
connected to that is support for deterministic startup, and lifecycle aware launching, and scriptable launching. 

Extending the graph API is also support for dynamic remapping. As well as support for aliasing topics. 

I'll try to think of more on my next flight.

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+In ROS 2 it will be possible to write real-time nodes when using a proper RTOS and with carefully written user code.
+
+
+### C++
+
+#### Node vs. Nodelet
+
+In ROS 1 the API for nodes and nodelets is different and requires a decision at programming time.
+In ROS 2 it will be possible to made this a deploy-time decision.
+It will be recommended to compile each component into a shared library which can then be loaded in a separate process or shared the same process with other components.
```
dogoepp: What does the sentence mean after "or" ?

dogoepp:
It's great for a non-expert like me to have this kind of comparisons. Plus, it could help some people understand why ROS 2 is built from scratch and using DDS.

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+One reason for the devel space in ROS 1 is to enable the developer to change e.g. Python code and use the modified code directly without the need to rebuild the package.
+This benefit is preserved in ROS 2 by optionally replacing copy operations in the install steps with symlinks.
+
+
+### Support catkin_simple use case
+
+In ROS 1 the package catkin_simple is aiming to make writing the CMake code of ROS packages easier.
+In many cases it is not achieving this goal which is often due to restrictions of the design necessary for support features like the devel space.
+
+In ROS 2 the CMake API has restructured to support this use case.
```
dogoepp: has -> was

articles/changes.md
dirk-thomas: I have updated the sentence. Please check if it is clear enough now or needs further information.

articles/changes.md
dirk-thomas: Thx. Fixed.

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+#### Python
+
+ROS 1 is targeting Python 2.
+ROS 2 will use 3.4+.
+
+
+### Reusing existing middleware
+
+ROS 1 uses a custom serialization format, a custom transport protocol as well as a custom central discovery mechanism.
+ROS 2 will use an abstract interface and provide multiple implementations of that interface based using existing DDS implementations.
```
wjwwood: This sentence is confusing, maybe `ROS 2 will have an abstract middleware interface, through which serialization, transport, and discovery will be provided by one of several existing DDS or RTPS implementations.`?

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+### Support other build systems beside CMake
+
+Every ROS package is a CMake project.
+In ROS 2 other build systems will be supported, e.g. plain Python packages.
+
+
+### Python packages
+
+In ROS 1 a package with Python code can only use a small subset of the features available in setup.py files.
+In ROS 2 a Python package can use anything in setup.py files, e.g. entry points.
```
wjwwood: It's probably worth noting that it can use a `setup.py` without also having a `CMakeLists.txt`.

articles/changes.md
dirk-thomas: Sounds better. I updated it except mentioning RTPS since it don't think the protocol itself without any semantic could fill the interface.

articles/changes.md
dirk-thomas: Isn't that covered by the block above?

wjwwood: Technically, yeah. It's just not very clear why you get to use all of the features in `setup.py` with ROS 2 but not with ROS 1.

dirk-thomas: Ok, I added some clarification.

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+### Environment setup
+
+In ROS 1 the build tool generates scripts which can be sourced to setup the environment before being able to use the built ROS packages.
```
wjwwood: `... which can be sourced ...` makes it sound optional and then later in the sentence it is described as a prerequisite for using the built packages. Maybe this would be more consistent: `In ROS 1 the build tool generates scripts which must be sourced in order to setup the environment before being able to use the built ROS packages.

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+This approach only works when the ROS packages are being built with ROS specific build tool.
+
+In ROS 2 the environment setup is separated into package-specific scripts and workspace-specific scripts.
+Each package provides the necessary script to make it usable after being built.
```
wjwwood: `itself` rather than `it`; `scripts` or `script`?

articles/changes.md
```diff
@@ -0,0 +1,241 @@
...
+The ROS specific tool only handles the workspace-specfic scripts which only call the package-specific scripts.
```
wjwwood: The tool isn't really ROS specific is it? Doesn't it also provides package specific scripts for build types which are not "ament", like plain CMake?

articles/changes.md
```diff
@@ -0,0 +1,245 @@
...
+In many cases it is not achieving this goal which is often due to restrictions of the design necessary for support features like the devel space.
+
+In ROS 2 the CMake API was restructured to support this use case.
+
+
+### Minimal support for packages without manifest
+
+In ROS 1 only packages with a manifest file are considered by the build system.
+In ROS 2 it is possible to detect packages with supported build system in folders without a manifest file.
+If the package follows common practice it might even be possible to detect some of the missing meta information (like dependencies).
```
wjwwood, line 121: I'm not sure if it's worth pointing out which things could be trivially done in ROS 1. For example this section as well as the support of other build systems (other than CMake) natively could be done and has been proposed in ROS 1.

articles/changes.md
```diff
@@ -0,0 +1,245 @@
...
+
+The life cycle can be used by tools like roslaunch to start a system composed of many components in a deterministic way.
+
+
+#### Parameters and Dynamic Reconfigure
+
+In ROS 1 global parameters and node-specific dynamic reconfigure parameters are two separate concepts.
+In ROS 2 a unified approach is being used.
+It will be similar to dynamic reconfigure and a node named "global parameter server" will accept requests to set values unconditionally.
+In ROS 1 all this information needs to be polled for changes in ROS 2 changes will be published to notify other entities.
```
wjwwood, line 187: I think the master can push notification of changes to the parameters to the nodes, so the nodes don't have to poll, but that could just be the provided interface where as the implementation actually uses polling.

articles/changes.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+#### Threading model
+
+In ROS 1 the developer can only choose between single-threaded execution or multi-threaded execution.
+In ROS 2 more granular execution models are available.
```
wjwwood: I don't know if this is accurate or maybe it's not what we should be emphasizing. Perhaps instead point out that it separates execution from nodes, allows execution over more than one node at a time (similar to what you said here, but not exactly the same), or that it is extensible whereas `ros::spin()`/`ros::AsynSpinner` wasn't.

articles/changes.md
```diff
@@ -0,0 +1,245 @@
...
+
+In ROS 2 it will be possible to create multiple nodes in a process.
+
+
+## Tools
+
+### roslaunch
+
+In ROS 1 roslaunch files are defined in xml with very limited capabilities.
+In ROS 2 launch file can be written in Python which enables to use more complex logic like conditionals etc.
```
wjwwood, line 234: This topic is still poorly defined at this point. We don't have a design doc or a prototype, or even (in my opinion) a clear vision for how this will replace roslaunch (in terms of API, matching or dropping features of roslaunch). I think it would be good to mention that roslaunch is still under design and development as not to give people the idea that we've come to some concrete conclusion on the topic.

articles/changes.md
```diff
@@ -0,0 +1,245 @@
...
+### ABI versioning
+
+ROS 1 rebuilds all downstream packages since it assumes ABI incompatibility.
+To avoid this significant overhead a ROS 2 package should be able to declare its ABI to avoid rebuilding downstream packages whenever possible.
+
+
+### Binary packages for Windows
+
+ROS 1 can only be built from source on Windows (which also only works for a few ROS packages and is not supported).
+ROS 2 will provide binary package based on Chocolatey.
```
wjwwood, line 257: Is this already covered in the "Platforms" section? Maybe just a comment that binaries are going to provided for Windows in the other section would be enough.

wjwwood:
Other than the comments, lgtm.

articles/changes.md
```diff
@@ -0,0 +1,245 @@
...
+
+ROS 2 will be CI tested and supported on Ubuntu, OS X as well as Windows.
+
+
+### Languages
+
+#### C++ standard
+
+ROS 1 is targeting C++03.
+ROS 2 will use C++11.
```
esteve: Many of the features in C++14 are supported by GCC 4.9:

https://gcc.gnu.org/projects/cxx1y.html
https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2014

by the time ROS 2 is released, C++14 will be fully supported by all three major compilers (GCC, Clang, VS2015)

esteve:
@dirk-thomas FYI I committed a fix for a typo: https://github.com/ros2/design/commit/33b12a5f0adc0d5676fa6b2e5f126ae8997dc5b7

articles/changes.md
dirk-thomas: I don't think this is the case for VS 2015: http://blogs.msdn.com/b/vcblog/archive/2015/06/19/c-11-14-17-features-in-vs-2015-rtm.aspx

We should probable mention here: `C++11 (and newer) when supported on all major platforms`.

dirk-thomas:
In the article we will focus on what is different between ROS 1 and ROS 2. We don't consider what could be backported. When that is the case the article can be updated. The article can also be updated with a note for each of these changes if it would be feasible to backport or what the effort would be.

articles/changes.md
dirk-thomas, line 257: The platforms section only mentions the supported platforms. Here I wanted to emphasize that binary packages will be available. I think this "justifies" a separate subsection.

dirk-thomas:
Thanks for the feedback. I have tried to update the article accordingly. I have also added several external links where I thought it would make sense: aaa6e60d271bd93661fdf74fac5dac2964aab145

If there is no further feedback I would squash and merge this soon.

articles/changes.md
```diff
@@ -0,0 +1,257 @@
...
+
+For more information please see the [ROS interface definition](http://design.ros2.org/articles/interface_definition.html) article.
+
+
+### Separated namespaces
+
+In ROS 1 .msg and .srv files can have the same name but the generated code collides.
+The same is the case for the request and response parts of services.
+
+In ROS 2 the generated code uses separate namespaces to guarantee it is collision-free.
```
jack-oquin, line 134: These separate namespaces are a bad idea. They introduce gratuitous differences between ROS 1 and ROS 2 in order to solve a merely theoretical "problem". Allowing a service to have the same name as a message would just confuse people, anyway.

dirk-thomas, line 134: We have heard your concern. The current ROS 1 <-> ROS 2 bridge is a pretty practical example though where this is necessary.

This article only describes what the current changes are. We will for sure revisit this when looking closer into migration and compatibility layers.

wjwwood, line 134: > They introduce gratuitous differences between ROS 1 and ROS 2 in order to solve a merely theoretical "problem".

I don't think it is a theoretical problem, since we were forced to create the `std_srvs` package just for this case in ROS 1, i.e. there is an `std_msgs/Empty.msg` as well as an `std_srvs/Empty.srv`. Both of which are valid in my opinion and being forced to put them in separate packages is an artificial constraint from my perspective.

> The current ROS 1 <-> ROS 2 bridge is a pretty practical example though where this is necessary.

To be fair this is a side effect of this change, not really a rational, but it is a convenient solution. The goal of having different include paths and namespaces for messages for the bridge could be solved some other way, just not nearly as elegantly.

---

I was thinking about a compromise for this issue and I think I have a proposal.

In ROS 2 we could generate a file for each service and messages that is like the ROS 1 include, e.g. `#include <std_msgs/String.h>`, the contents of which would look something like this:

``` c++
#include <std_msgs/msg/string.hpp>

namespace std_msgs
{
using std_msgs::msg::String;
using std_msgs::msg::String_;
// ... and other symbols in message headers.
}
```

That way you can continue to include messages in ROS 2 as you did in ROS 1 and you get the same symbols (mostly). For the case of `Empty.msg` and `Empty.srv` the `<std_msgs/Empty.h>` could contain this:

``` c++
#error Including <std_msgs/Empty.h> is ambiguous since there is a Message and a Service by that name, use either <std_msgs/msg/empty.hpp> or <std_msgs/srv/empty.hpp>.
```

Some points about the approach:
- We can allow opting in/out of this header by installing them to a place like `<prefix>/include/ros1_compat/std_msgs/Empty.h` and conditionally adding `<prefix>/ros1_compat` to the include directories.
- The bridge can explicitly opt out of this, allowing it to have different include paths for ROS 1 and ROS 2.
- We can choose whether or not to put a deprecation message in the compat header to encourage people to migrate to the new format, which I still believe is better because it is more conceptually consistent and provides more information to people reading unfamiliar code.
- We can provide similar symbols to what is in ROS 1, but somethings will likely end up being different, so I'm still not optimistic that we can achieve a zero change migration w.r.t. messages.

jack-oquin, line 134: +1 Good idea, @wjwwood 

I didn't know about the `std_svcs/Empty.h` issue, but it supports my point that there is _no_ _way_ to have a name collision between a ROS 1 message and a service. As a purely architectural issue, I don't think we should introduce that as a possibility.

Nor do I see a need for `std_msgs/Empty.h` to throw a compile error. That name unambiguously refers to the message. If the service had been desired, the program would have included `std_svcs/Empty.h` instead.

> We can choose whether or not to put a deprecation message in the compat header to encourage people to migrate to the new format, which I still believe is better because it is more conceptually consistent and provides more information to people reading unfamiliar code.

Please don't deprecate those references any time in the first few years. Maybe later, after most code has finally been migrated.

> We can provide similar symbols to what is in ROS 1, but somethings will likely end up being different, so I'm still not optimistic that we can achieve a zero change migration w.r.t. messages.

I don't expect zero change. The reason I keep harping on these message incompatibilities is because most ROS 1 code I know about uses messages all over the place as fundamental data structures. It's very convenient to do it that way.

I want to eliminate the bulk of these message incompatibilities, so we'll be looking at 50-line diffs for a package migration, and not 500-line diffs. That order of magnitude will save engineer-centuries of work for the entire ROS community.

wjwwood, line 134: > I didn't know about the std_svcs/Empty.h issue, but it supports my point that there is no way to have a name collision between a ROS 1 message and a service.

I think it does not support your point, in fact it demonstrates that we had to create a new package to work around the collision where we otherwise would not have done so. So, I was saying that `Empty.msg` and `Empty.srv` shouldn't be in separate packages (as they are now), in which case you would have a collision and therefore the header `std_msgs/Empty.h` would be ambiguous.

jack-oquin, line 134: It does support _my_ point: that there is no ambiguity with any ROS 1 message or service. 

Your point is that you would like to have had the ability to define an Empty message and and Empty service in the same package. Maybe so. But, there would have been no great harm in defining `std_msgs/EmptyService`, instead. 

It's not that important, and not worth introducing significant migration effort. If you can fix that problem, then OK.

PR #66, opened 2015-10-21
reworked the ideal system section to be more concise, have less of an impact on the TOC

wjwwood:
@tfoote I reworded some things while reorganizing the content, so I think you should review the changes before taking them.

Also I clarified that it should be possible to unset groups of parameters atomically.

### DIFF
--- a/articles/055_ros_parameter_design.md
+++ b/articles/055_ros_parameter_design.md
@@ -40,57 +40,47 @@ Other resources related to the parameter design process for ROS 2.0 include:
 
 ## Ideal System
 
-It is useful to consider the ideal system to understand how it relates to the current system and how a new system could work.
-We'd like to support both the use cases of ROS 1.0 built in parameters as well as dynamic parameters.
-An ideal parameter system would have the qualities laid out in the following paragraphs.
+It is useful to consider an ideal system to understand how it would work and how it would relate to the current system.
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks lgtm

articles/055_ros_parameter_design.md
```diff
@@ -40,57 +40,47 @@ Other resources related to the parameter design process for ROS 2.0 include:
...
+- **Provide notifications when parameter values change or update**
 
-### Notifications of change
+- **Reject changes to parameter values**
 
-When a parameter changes value it should be possible to get a notification.
+    This implies that some entity has the authority to reject or accept a change based on arbitrary criteria.
+    This would also include the ability to convey at least part of the criteria for the acceptance of a change to external actors.
+    For example, communicating the range for an integer or a few choices for a string.
+    This type of information would be used to generate generic user interfaces, but might not capture all criteria.
```
dirk-thomas, line 68: I would propose to add the following:

```
Since the validation criteria can be arbitrary complex and can potentially not be communicated to a client the parameter server could offer to validate an atomic set of parameters and respond with a boolean flag if the values would be accepted (based on the current criteria). Obviously the result might be different when the values are set shortly after but it would allow to implement validators in e.g. a GUI generically.
```

tfoote, line 68: Thanks added.

PR #70, opened 2015-12-27
Request for information on the extensibility of ROS 2.0 messages

andreasBihlmaier:


wjwwood:
@andreasBihlmaier Thanks for the question. Currently we don't have a plan for this yet. I hope to spend time on it in the near future, but it might depend on using DDS's X-Types interface: http://www.omg.org/spec/DDS-XTypes/1.1/

I'll leave this open until we can properly address it.

clalancette:
I don't think this properly qualifies as a PR (it doesn't look like this was ever intended to be merged), so instead I've made a new issue: https://github.com/ros2/design/issues/166 .  I'm going to close this out.

PR #72, opened 2015-12-31
Add a document for ros_time

tfoote:
This adds a document for the ros time abstraction. 

This also has a few cleanups of the config, and supports local testing using the official jekyll docker image.

### DIFF
--- a/articles/113_ros_time.md
+++ b/articles/113_ros_time.md
@@ -1,0 +1,159 @@
+---
+layout: default
+title: Clock and Time
+permalink: articles/clock_and_time.html
+abstract:
+  This article describes the ROS primitives to support programming which can run both in real time as well as simulated time which may be faster or slower.
+published: true
+author: '[Tully Foote](https://github.com/tfoote)'
+---
+
+- This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Background
+
+Many robotics algorithms inherently rely on timing as well as synchronization.
+To this end we require that nodes running in the ROS network have a synchronized system clock such that they can accurately report timestamps for events.
+
+There are however several use cases where being able to control the progress of the system is important.
+
+## Use cases requiring time abstraction
+
+When playing back logged data it is often very valuable to support accelerated, slowed, or stepped control over the progress of time.
+This control can allow you to get to a specific time and pause the system so that you can debug it in depth.
+It is possible to do this with a log of the sensor data, however if the sensor data is out of synchronization with the rest of the system it will break many algorithms.
+
+Another important use case for using an abstracted time source is when you are running logged data against a simulated robot instead of a real robot.
+Depending on the simulation characteristics, the simulator may be able to run much faster than real time or it may need to run much slower.
+Running faster than real time can be valuable for high level testing as well allowing for repeated system tests.
+Slower than real time simulation is necessary for complicated systems where accuracy is more important than speed.
+Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
+Additionally if the simulation is paused the system can also pause using the same mechanism.
+
+### Challenges in using abstracted time
+
+There are many algorithms for synchronization and they can typically achieve accuracies which are better than the latency of the network communications between devices on the network.
+However, these algorithms take advantage of assumptions about the constant and continuous nature of time.
+
+An important aspect of using an abstracted time is to be able to manipulate time.
+In some cases, speeding up, slowing down, or pausing time entirely is important for debugging.
+
+The ability to support pausing time requires that we not assume that the time values are always increasing.
+
+When communicating the changes in time propagation, the latencies in the communication network becomes a challenge.
+Any change in the time abstraction must be communicated to the other nodes in the graph, but will be subject to normal network communication latency.
+This inaccuracy is proportional to the latency of communications and also proportional to the increase in the real time factor.
+If very accurate timestamping is required when using the time abstraction, it can be achieved by slowing down the real time factor such that the communication latency is comparatively small.
+
+The final challenge is that the time abstraction must be able to jump backwards in time, a feature that is useful for log file playback.
+This behavior is similar to a system clock after a negative date change, and requires developers using the time abstraction to make sure their algorithm can deal with the discontinuity.
+Appropriate APIs must be provided for to the developer API to enable notifications of jumps in time, both forward and backwards.
+
+### Time Abstractions
+
+There will be at least three versions of these abstractions with the following types, `SystemTime`, `SteadyTime` and `ROSTime`.
+These choices are designed to parallel the [`std::chrono`][] [`system_clock`][] and [`steady_clock`][].
+It is expected that the default choice of time will be to use the `ROSTime` source, however the parallel implementations supporting `steady_clock` and `system_clock` will be maintained for use cases where the alternate time source is required.
+
+#### System Time
+
+For convenience in these cases we will also provide the same API as above, but use the name `SystemTime`.
+
+`SystemTime` will be directly tied to the system clock.
+
+#### Steady Time
+
+Example use cases for this include hardware drivers which are interacting with peripherals with hardware timeouts.
+
+In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
+
+#### ROS Time
+
+The `ROSTime` will report the same as `SystemTime` when an ROS Time Source is not active.
+When the ROS time source is active `ROSTime` will return the latest value reported by the Time Source.
+
+### ROS Time Sources
+
+#### Default Time Source
+
+To implement the time abstraction the following approach will be used.
+
+The time abstraction can be published by one source on the `/clock` topic.
+The topic will contain the most up to date time for the ROS system.
+If a publisher exists for the topic, it will override the system time when using the ROS time abstraction.
+If `/clock` is being published, calls to the ROS time abstraction will return the latest time received from the `/clock` topic.
+If time has not been set it will return zero if nothing has been received.
+A time value of zero should be considered an error meaning that time is uninitialized.
+
+If the time on the clock jumps backwards, a callback handler will be invoked and be required to complete before any calls to the ROS time abstraction report the new time.
+Calls that come in before that must block.
+The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
+
+The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific.
+
+##### No Advanced Estimating Clock
+
+There are more advanced techniques which could be included to attempt to estimate the propagation properties and extrapolate between time ticks.
+However all of these techniques will require making assumptions about the future behavior of the time abstraction.
+And in the case that playback or simulation is instantaneously paused, it will break any of these assumptions.
+There are techniques which would allow potential interpolation, however to make these possible it would require providing guarantees about the continuity of time into the future.
+For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
+Tuning the parameters for the `/clock` topic lets you trade off time for computational effort and/or bandwidth.
+
+#### Custom Time Source
+
+It is possible that the user may have access to an out of band time source which can provide better performance than the default source the `/clock` topic.
+It might be possible that for their use case a more advanced algorithm would be needed to propagate the simulated time with adequate precision or latency with restricted bandwidth or connectivity.
+The user will be able to switch out the time source for either each instance of their Time object as well as have the ability to override the default for the process.
+
+It is possible to use an external time source such as GPS in as a ROSTime source, but it is recommended to integrate a time source like that using standard ntp integrations with the system clock since that is already an established mechanism and will not need to deal with more complicated changes such as time jumps.
+
+## Implementation
+
+The `SystemTime`, `SteadyTime`, and `ROSTime` API's will be provided by each client library in an idiomatic way, but they may share a common implementation, e.g. provided by `rcl`.
+However, if a client library chooses to not use the shared implementation then it must implement the functionality itself.
+
+### Public API
+
+In each implementation will provide `Time`, `Duration`, and `Rate` datatypes, for all three time source abstraction..
+The `Duration` will support a `sleep_for` function as well as a `sleep_until` method.
+The implementation will also provide a `Timer` object which will provide periodic callback functionality for all the abstractions.
+
+### RCL implementation
+
+In `rcl` there will be datatypes and methods to implement each of the three time abstractions for each of the core datatypes.
+However at the `rcl` level the implementation will be incomplete as it will not have a threading model and will rely on the higher level implementation to provide any threading functionality such as is required by sleep methods.
+
+It will provide implementations parallel to the public datastructures and storage which the client library can depend upon and to which it can delegate.
+The underlying datatypes will also provide ways to register notifications, however it is the responsibility of the client library implementation to collect and dispatch user callbacks.
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> Enumerate the <code>rcl</code> datastructures and methods here.
+</div>
+
+## References
+
+The default time source is modeled on the ROS Clock and ROS Time system used in ROS 1.0.
+For more information on the implementation in ROS 1.0 see:
+
+- [ROS Clock Documentation](http://wiki.ros.org/Clock)
+- [rospy Time Documentation](http://wiki.ros.org/rospy/Overview/Time)
+- [roscpp Time Documentation](http://wiki.ros.org/roscpp/Overview/Time)
+
+### Real Time
+
+In this article the term 'real time' is used to express the true rate of progression of time.
+This is not connected to 'real-time' computing with deterministic deadlines.
+
+[`std::chrono`]: http://en.cppreference.com/w/cpp/chrono
+[`steady_clock`]: http://en.cppreference.com/w/cpp/chrono/steady_clock
+[`system_clock`]: http://en.cppreference.com/w/cpp/chrono/system_clock


_config.yml
```diff
@@ -1,7 +1,8 @@
 name: ROS2 Design
 description: "Distilled design documents related to the ROS 2.0 effort"
 
-baseurl: http://design.ros2.org
+baseurl: ""
+url: http://design.ros2.org
```
wjwwood, line 4: Why?

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+Original Author: {{ page.author }}
+
+
+## Background
+
+Many robotics algorithms inherently rely on timing as well as synchronization.
+To this end we require that nodes running in the ROS network have a synchronized system clock such that they can accurately report timestamps for events.
+
+There are however several use cases where being able to control the progress of the system are important.
```
wjwwood: It's not clear how this relates to time. Should it be `cases where being able to control the flow of the system time are important.` or `cases where being able to control the progress of the system by changing the flow of time are important.`?

_config.yml
tfoote, line 4: That's how jekyll is designed to be used, it's mostly to allow github-pages to render properly for project prefixes. 

See: https://jekyllrb.com/docs/github-pages/ and
https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/

wjwwood, line 4: That must have changed (again), because that wasn't the case before. Just make sure double check that the live site is working correctly after merging.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+An important aspect of using an abstracted time is to be able to manipulate time.
+In some cases, changing the speed dynamically for example speeding up or slowing down all the way to pause is important for debugging.
+
+The ability to support pausing time requires that we not assume that the time values are always increasing.
+
+When communicating the changes in time propagation the latencies in the communication network becomes a challenge.
+Any change in the time abstraction must be communicated to the other nodes in the graph, but will be subject to normal network communication latency.
+This inaccuracy is proportional to the latency of communications and also proportional to the increase in the realtime factor.
+As such if very accurate timestamping is required using the time abstraction is required it can be achieved by slowing down the realtime factor such that the communication latency is comparatively small.
```
wjwwood: Avoid using `is required` twice in the sentence, maybe `if very accurate timestamping is required when using the time abstraction, it can be achieved by slowing down the realtime factor`

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+The final challenge is that to be useful for log file playback the time abstraction must be able to jump backwards in time.
+This is a very different behavior from most clocks, and requires developers using the time abstraction to make sure their algorithm can deal with the discontinuity.
```
wjwwood: The system clock can do this, so maybe instead of saying "This is a very different behavior from most clocks", you could say "This behavior is similar to a system clock after a negative date change".

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+## System Behavior
+
+To implement the time abstraction the following approach will be used.
+
+The time abstraction can be published by one source on the `/clock` topic.
+The topic will contain the latest abstracted time for the ROS system.
```
wjwwood: Is abstracted the right word? Is simulated time more appropriate? Abstracted makes sense when you're talking about the time interface which might be a system clock or a simulated clock behind the scene, but here we're publishing the simulated clock or nothing at all.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+If a publisher exists for the topic it will override the system time when using the ROS time abstraction.
+If `/clock` is being published calls to the ROS time abstraction will return the latest time received from the `/clock` topic with a default of zero if nothing has been received.
```
wjwwood: This might be clearer in two sentences:

```
If `/clock` is being published, calls to the ROS time abstraction will return the latest time received from the `/clock` topic.
If the current time is requested before any messages have been received on the `/clock` topic then a default value of zero is returned.
```

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+If the time on the clock jumps backwards a callback handler will be invoked and be required to complete before any calls to the ROS time abstraction report the new time.
+Calls that come in before that must block.
+The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
```
wjwwood, line 99: I think there needs to be more precise language here. It's pretty confusing/unclear as to what calls are coming in and which ones must block. It might benefit from some examples, which might seem redundant, but I think would help to be clearer.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific.
+
+### No Advanced Estimating Clock
+
+There are more advanced techniques which could be included to attempt to estimate the propagation properties and extrapolate between time ticks.
+However all of these techniques will require making assumptions about the future behavior of the time abstraction.
+And in the case that playback or simulation is instantaneously paused it will break any of these assumptions.
+There are techniques which would allow potential interpolation, however to make these possible it would require providing guarantees about the continuity of time into the future.
+For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
+These are both
```
wjwwood: ...

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+However all of these techniques will require making assumptions about the future behavior of the time abstraction.
+And in the case that playback or simulation is instantaneously paused it will break any of these assumptions.
+There are techniques which would allow potential interpolation, however to make these possible it would require providing guarantees about the continuity of time into the future.
+For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
+These are both 
+
+## Public API
+
+In each client library there will be provided an abstraction of `Time`, `Duration`, and `Rate`.
+The duration abstraction will support a `sleep` function which
```
wjwwood: which ...

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Walltime
+
+There are several use cases where using the ROS time abstraction are inappropriate.
+Example use cases for this include hardware drivers which are interacting with peripherals with hardware timeouts.
+For convenience in these cases we will also provide the same API as above, but use the name `WallTime`.
+
+`WallTime` will be directly tied to the system clock.
+
+In nodes which require the use of `WallTime` for interacting with hardware or other peripherals it is expected that they do a best effort to isolate any `WallTime` information inside their implementation and translate external interfaces to use the ros time abstraction.
```
wjwwood: We should probably have a `SteadyWallClock` which is tied to `chrono::steady_clock` as well?

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Implementation
+
+At the `rcl` level we will provide a time abstraction which will return the latest 'ROS time' as provided by the `/clock` topic.
+It will provide the callback when time changes.
+
+It is expected that most client libraries will provide a natural implementation of the time primitives, `Time`, `Duration`, `Rate`, and `Timer`.
```
wjwwood: This is mentioned in the `## Public API` section, should it mention that here or be consolidated?

wjwwood:
Thanks for taking some time to write this out.

Maybe we need to have a section about pausing? Was that supported in ROS 1's simulated time? How should it work, i.e. will it rely on publishes to `/clock` to stop or is there a separate mechanism for pausing? Are there corner cases like what happens if you pause and then the next time update is in the past?

Also I think we could spend some time thinking about blocking of the time function. For example, should it block while time going backwards callbacks are being executed or just return the latest time? I think it might be undesirable to have the get current time function block under some (asynchronous) conditions.

articles/113_ros_time.md
jackieokay, line 99: I agree, this sentence was unclear (I think it needs to be expressed in 2 or 3 sentences). "before time will be in the past" is confusing. Perhaps it would be useful to provide a high-level example that describes the chain of events after a developer registers a callback that does something interesting and starts publishing abstracted time.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+As such if very accurate timestamping is required using the time abstraction is required it can be achieved by slowing down the realtime factor such that the communication latency is comparatively small.
+
+The final challenge is that to be useful for log file playback the time abstraction must be able to jump backwards in time.
+This is a very different behavior from most clocks, and requires developers using the time abstraction to make sure their algorithm can deal with the discontinuity.
+
+## System Behavior
+
+To implement the time abstraction the following approach will be used.
+
+The time abstraction can be published by one source on the `/clock` topic.
```
jackieokay, line 90: What is expected to happen when I have two publishers to the `/clock` topic that are interleaved (same frequency, different start times)? I've never tried this in ROS 1. What can be done to prevent unexpected behavior here, or at least warn the user?

jackieokay:
I tweaked some spelling and grammar in branch `ros_time_jackie`: https://github.com/ros2/design/tree/ros_time_jackie

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+
+## Background
+
+Many robotics algorithms inherently rely on timing as well as synchronization.
+To this end we require that nodes running in the ROS network have a synchronized system clock such that they can accurately report timestamps for events.
```
dirk-thomas, line 25: The whole aspect of time synchronization seems not relevant to me in this article. The main goal is to allow the time to go slower / faster / be paused. Therefore I would propose to remove this reference here as well as in the next section.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
+These are both 
+
+## Public API
+
+In each client library there will be provided an abstraction of `Time`, `Duration`, and `Rate`.
+The duration abstraction will support a `sleep` function which 
+It will also provide a `Timer` object which will provide periodic callback functionality using the ROS time abstraction.
+
+### Walltime
```
dirk-thomas: Should we add a reference why these are called `Wall*`? We might know but new users might have no clue where the term comes from.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+## System Behavior
+
+To implement the time abstraction the following approach will be used.
+
+The time abstraction can be published by one source on the `/clock` topic.
+The topic will contain the latest abstracted time for the ROS system.
+If a publisher exists for the topic it will override the system time when using the ROS time abstraction.
+If `/clock` is being published calls to the ROS time abstraction will return the latest time received from the `/clock` topic with a default of zero if nothing has been received.
+
+If the time on the clock jumps backwards a callback handler will be invoked and be required to complete before any calls to the ROS time abstraction report the new time.
```
dirk-thomas: A callback handler is a very specific implementation choice. Depending on the client library I think other choices might be selected instead. Therefore I would propose to not specify this implementation detail in the article but only describe the concept: "entities can be notified in case of clock jumps".

dirk-thomas:
Please rebase to validate against the latest linters.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
+---
+layout: default
+title: Clock and Time
+permalink: articles/clock_and_time.html
+abstract:
+  This article describes the ROS primatives to support programming which can run both in realtime as well as simulated time.
```
gbiggs: You need to define "real time" here. I assume that you're talking about running in "world time" (I don't actually know if there's an agreed-upon phrase to describe that), not meeting deadlines.

articles/113_ros_time.md
gbiggs, line 25: Doesn't the idea of allowing the rate of change of time to be altered inherently introduce the idea of a synchronised clock across each node? At least for nodes that do calculations involving time.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+There are however several use cases where being able to control the progress of the system is important.
+
+## Use cases requiring time abstraction
+
+When playing back logged data it is often very valuable to support accelerated, slowed, or stepped control over the progress of time.
+This control can allow you to get to a specific time and pause the system so that you can debug it in depth.
+It is possible to do this with just the playback of the sensor data, however if the sensor data is out of synchronization with the rest of the system it will break many algorithms.
+
+Another important use case for using an abstracted time source is when you are running against a simulated robot instead of a real robot.
+Depending on the simulation characteristics the simulator may be able to run much faster than realtime or much slower.
```
gbiggs: Definition of real time needed here, too.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+Running faster than realtime can be valuable for high level testing as well allowing for repeated system tests.
+Slower than realtime simulation is necessary for complicated systems where accuracy is more important than speed.
+Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
+And if the simulation is paused the system can also pause using the same mechanism.
```
gbiggs: Pedantic comment: "And" -> "Additionally,"

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Challenges in using abstracted time
+
+There are many algorithms for synchronization and they can typically achieve accuracies which are better than the latency of the network communications between devices on the network.
+However, these algorithms take advantage of assumptions about the constant and continuous nature of time.
+
+An important aspect of using an abstracted time is to be able to manipulate time.
+In some cases, changing the speed dynamically for example speeding up or slowing down all the way to pause is important for debugging.
```
gbiggs: There's a bit of repitition here from the use cases section.

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+
+The ability to support pausing time requires that we not assume that the time values are always increasing.
+
+When communicating the changes in time propagation the latencies in the communication network becomes a challenge.
+Any change in the time abstraction must be communicated to the other nodes in the graph, but will be subject to normal network communication latency.
+This inaccuracy is proportional to the latency of communications and also proportional to the increase in the realtime factor.
```
gbiggs: This sentence is mixing discussion about network latency with the term "realtime", so I'm honestly not sure which definition of "realtime" you mean.

articles/113_ros_time.md
gbiggs, line 90: If there is a way to gather information about the topology of the system, specifically who is publishing on what topics, then at the very least it should be possible to add something to roswtf to catch this problem.

articles/113_ros_time.md
gbiggs: I don't think that "simulated" is the right word either; that implies that these concepts only apply when using time from a simulation source.

What about "virtual time"? "ROS time?" There must be a decent phrase that describes what is being done.

articles/113_ros_time.md
gbiggs: I'm not sure a default value of zero is a good idea. Wouldn't not having the current time available be an error condition if I'm doing something time-relevant? Otherwise I might find I suddenly get a massive jump in time when the first clock message comes in.

articles/113_ros_time.md
gbiggs: +1

articles/113_ros_time.md
```diff
@@ -0,0 +1,115 @@
...
+And in the case that playback or simulation is instantaneously paused it will break any of these assumptions.
+There are techniques which would allow potential interpolation, however to make these possible it would require providing guarantees about the continuity of time into the future.
+For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
+These are both 
+
+## Public API
+
+In each client library there will be provided an abstraction of `Time`, `Duration`, and `Rate`.
+The duration abstraction will support a `sleep` function which 
+It will also provide a `Timer` object which will provide periodic callback functionality using the ROS time abstraction.
```
gbiggs: What sort of API will `Rate` supply? Will there be a "sleep until next tick" sort of call?

articles/113_ros_time.md
gbiggs: +1

articles/113_ros_time.md
tfoote, line 99: This will be clarified when I update the design from the meeting with @wjwwood

articles/113_ros_time.md
tfoote: I'm going to update it to use 
system, steady, and ros, instead of the old rostime and walltime.

articles/113_ros_time.md
tfoote, line 90: Two conflicting publishers is going to lead to bad results. It certainly can be rolled into roswtf, and the subscribers should probably warn as well.

articles/113_ros_time.md
tfoote, line 25: Our default behavior of falling back to the system time requires them to be in sync. If we did not want to require that we would have to publish `/clock` all the time. This is here to remind people that in the default case synchronization is necessary.

articles/113_ros_time.md
tfoote: See other note about switching to use system and steady, with ros as the 3rd abstraction.

articles/113_ros_time.md
tfoote: I changed all usages to be 'real time' and added a note to differentiate 'real time' from requirements for deterministic computing requirements.

tfoote:
Thanks for the reviews, I've tried to resolve most of the comments. I had a meeting with @wjwwood and will be updating the document with more details from that shortly.

tfoote:
I've updated this to have more of the context from the design meeting with @wjwwood. Please take a look.

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+There will be at least three versions of these abstractions with the following types.
+There are several use cases where using the ROS time abstraction are inappropriate.
+
+- SystemTime: which will follow the system clock.
+- SteadyTime: which will follow the steady clock on the system.
+- ROSTime: which will follow the system clock unless it's designated Time Source is active.
+
+### System Time
+
+For convenience in these cases we will also provide the same API as above, but use the name `SystemTime`.
```
wjwwood, line 69: The layout of the text makes "these cases" sort of ambiguous. I'd suggest describing which cases directly here instead.

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+### Steady Time
+
+Example use cases for this include hardware drivers which are interacting with peripherals with hardware timeouts.
+
+In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
+
+### ROS Time
+
+The `ROSTime` will report the same as `SystemTime` when an external Time Source is not active.
+When the external time source is active `ROSTime` will return the latest value reported by the Time Source.
```
wjwwood: These three sections could be a little more specific in my opinion. It might even be good to provide examples of how each behaves when time jumps backwards and why (system time changes versus ROS Time Source changes).

Alternatively you could use terminology that is not specific to our system (we're already doing that if you liken it to `std::chrono`) and link to its definition on wikipedia or similar.

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+However all of these techniques will require making assumptions about the future behavior of the time abstraction.
+And in the case that playback or simulation is instantaneously paused, it will break any of these assumptions.
+There are techniques which would allow potential interpolation, however to make these possible it would require providing guarantees about the continuity of time into the future.
+For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
+Tuning the parameters for the `/clock` topic lets you trade off time for computational effort and/or bandwidth.
+
+#### Custom Time Source
+
+It is possible that the user may have access to an out of band time source which can provide better performance than the default source the `/clock` topic.
+A potential alternative source of time might be a GPS.
```
wjwwood: I'm not sure this is worth mentioning since the preferred way to integrate a GPS would be to integrate it into a NTP daemon. A better example would be to use a time synchronization algorithm (like NTP's) with simulated time. That's a bit complicated to make clear here, so I'd suggest just leaving it at you can replace the time source for ROSTime with any arbitrary time source or algorithm.

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+The user will be able to switch out the time source for either each instance of their Time object as well as the ability to override the default for the process.
+
+## Implementation
+
+The public API will be provided by the client libraries.
+The client libraries will be able to rely on a base implementation in rcl which provides a common behavior.
+If the client library implementer chooses they my also reimplement the core behavior, however it is expected that the rcl implementation should be adaquate to support any language with a foreign function interface for C code.
```
wjwwood: The first sentence is a bit ambiguous to me, maybe instead replace the first two sentences with:

> The SystemTime, SteadyTime, and ROSTime API's will be provided by each client library in an idiomatic way, but they may share a common implementation, e.g. provided by `rcl`.

Then shorten the last sentence just to make it clear that it is not optional:

> However, if a client library chooses to not use the shared implementation then it must implement the functionality itself.

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Public API
+
+Each client library is expected to provide parallel interfaces for each of the three abstract types of time.
```
wjwwood: Already said this in the previous section.

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+It is up to the implementation to choose the most relevant appropriate way to select the parallel implementations.
```
wjwwood: This is confusing to me. By "implementation" do you mean "client library", "relevant appropriate" sounds redundant, and what parallel implementations are the client libraries choosing from?

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+
+In each client library will provide `Time`, `Duration`, and `Rate` implementations.
```
wjwwood: "Each client library"? or "In each client library there will be ... implementations."? Maybe this:

> Each client library will also provide `Time`, `Duration`, and `Rate` concepts which operate on one or more of the three time concepts described above.

Also it's not clear which of those three will support which of the aforementioned time types, e.g. will `Duration` and `Rate` support all three time types?

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+The duration abstraction will support a `sleep_for` function as well as a `sleep_until` method.
+It will also provide a `Timer` object which will provide periodic callback functionality using the ROS time abstraction.
+
+### RCL implementation
+
+In `rcl` there will be representatative datatypes and methods to implement each of the three time abstractions for each of the core datatypes.
```
wjwwood: I'd drop "representative".

articles/113_ros_time.md
```diff
@@ -0,0 +1,154 @@
...
+However at the `rcl` level the time API will not have a threading model and will rely on the client library implementation to do any threading such as is required by sleep methods.
+
+It will provide implementations parallel to the public datastructures and storage which the client library can depend upon and to which it can delegate.
+The underlying datatypes will also provide ways to register notifications, however it is the responsibility of the client library implementation to collect and dispatch user callbacks.
```
wjwwood, line 137: For the previous three sentences I understand what you're saying, but I think others will find it harder to follow. Consider either removing these technical details of the implementation or describe them more thoroughly, perhaps with examples.

wjwwood:
I had some comments, but they're mostly questions clarifying the language. This is looking pretty good. Thanks for working on it @tfoote.

tfoote:
@wjwwood I added some clarifications and a todo

ahundt:
You may want to consider reading papers on time synchronization, particularly if you want to make ros2 usable for those with realtime constraints. One example use case is dealing with data from a pair of stereo cameras which need to be sent signals to remain synchronized to an accuracy within nanoseconds for the most accurate depth data possible. In particular, it would be ideal if you can expose time uncertainty in addition to the time itself.

Another useful use case is [google's global time synchronization database paper](https://static.googleusercontent.com/external_content/untrusted_dlcp/research.google.com/en//archive/spanner-osdi2012.pdf).

Finally, here are general time synchronization google scholar search results.
https://scholar.google.com/scholar?hl=en&q=time+synchronization&btnG=&as_sdt=1%2C21&as_sdtp=&search_plus_one=form

wjwwood:
> You may want to consider reading papers on time synchronization,

I'm not sure what leads you to believe we haven't, but for those who are interested, we looked at many things in addition to the Google paper you linked to (in case anyone else is interested in the subject):
- The NTP RFC's:
  - https://www.ietf.org/rfc/rfc5905.txt
  - http://www.ietf.org/rfc/rfc5906.txt
  - http://www.ietf.org/rfc/rfc5907.txt
  - http://www.ietf.org/rfc/rfc5908.txt
- A new algorithm called TICSync:
  - http://www.robots.ox.ac.uk/~mobile/Papers/2011ICRA_arh.pdf
- The chrony documentation:
  - http://chrony.tuxfamily.org/comparison.html
  - http://tldp.org/HOWTO/Clock.html
- Clock Synchronization for Real-Time Industrial systems:
  - http://cdn.intechopen.com/pdfs-wm/12259.pdf

I think our perspective, which we're trying to convey with this document, is that the ROS 1 time interface is pretty good, in that it has been used successfully in many use cases thus far. However, we found a few places where it could be improved, like exposing the update mechanism through a programmatic extension point (whereas in ROS 1 you could only update it with the `/clock` topic). The other thing we're thinking of adding is the ability for the user to get a callback when the simulated time is changed or goes backwards.

In my opinion providing an interface similar to the interface provided by the POSIX API, or the C++ chrono API, is a good goal for the simulated time interface that way it can be a drop in replacement when using a normal clock in many situations. Any other extensions to control or slew time, or measure uncertainty in time, should in my opinion either be a transparent side-effect (like NTP or chrony) or through a side channel separate from the normal interface.

If you've got experience with time synchronization, then please help us by making concrete suggestions as to how we might change the interface and provide more functionality or flexibility without compromising the existing interface's usability and familiarity.

As to your comments you've already made:

> particularly if you want to make ros2 usable for those with realtime constraints.

I don't see any way the time interface needs to be changed in order to support realtime scenarios, but if you have a suggestion, please make it. We've given a lot of thought to this, so for example, getting the time with the `rcl` interface (our C API for ROS 2), it is accessed with by an atomic int64, so on modern processors where `ATOMIC_LLONG_LOCK_FREE` is true the access of the current time, simulated or not, will be lock free and thread safe. Also we provide a portable steady clock and a system clock in addition to the simulated clock (assumed to behave like the system clock but flow differently).

> One example use case is dealing with data from a pair of stereo cameras which need to be sent signals to remain synchronized to an accuracy within nanoseconds for the most accurate depth data possible.

I believe that is best solved by triggering the camera's electronically (with a GPS PPS for example) and using sequence numbers to synchronize them, but that's up to the person integrating the cameras and not the role of the framework, I think. In all my experience and research the only thing I feel comfortable saying is that there is no single system that meets everyone's timing needs. So I think the appropriate role of ROS is to provide a single decent way of simulating time (for common cases like playing back log data and integrating with simulators) and let others use a different method if they need. I'm not aware of any proposed time system that cannot be used in combination with ROS 1 or the proposed design for ROS 2. If you're aware of a system that could not be used with ROS 1 or used with the proposed clock simulation mechanism for ROS 2, then please let us know by describing it and what about the current design is in conflict.

> In particular, it would be ideal if you can expose time uncertainty in addition to the time itself.

Can this not already be exposed with a side channel, separate from the normal "get time" mechanisms? Can you propose a mechanism that better supports this use case while also supporting use cases without this metric (I'd argue far more common)? Or a more fundamental question: Does the framework really need to solve this problem?

ahundt:
Thanks for the detailed response! I'm just trying to contribute based on reading the document and request for contributions on the ros2 website, I wasn't trying to assume what you may/may not know. Sorry if it came across in another way! :-)

could `ros2::Duration duration = ros2::time::now()` internally store a time and an uncertainty parameter and provide access to it without disrupting current APIs? For example in addition to `duration.toSec()` one could also call `duration.uncertainty()`.

> Can this not already be exposed with a side channel, separate from the normal "get time" mechanisms? 

The difficulty with a side channel is once you have hundreds of packages using one time function it is incredibly difficult to change it all out to use another! That's why I made my suggestion.

> http://www.robots.ox.ac.uk/~mobile/Papers/2011ICRA_arh.pdf

I was actually looking for that TICSync paper while I was writing this post but couldn't recall the name! I'm glad you already knew about it. 

> Or a more fundamental question: Does the framework really need to solve this problem?

In absolute terms, of course not. But if it did at such a low level of the system from the beginning? Boy would that make a huge difference when engineering real systems!

> I believe that is best solved by triggering the camera's electronically (with a GPS PPS for example) and using sequence numbers to synchronize them, but that's up to the person integrating the cameras and not the role of the framework, I think.

Yes, that's definitely the way to sync the frames when you have people with the hardware and expertise to do so! If you've not got that... the next best thing would be if the error in your measurements could be easily quantified by ROS2 Time. :-) 

Thanks for your consideration!

wjwwood:
> Thanks for the detailed response! I'm just trying to contribute based on reading the document and request for contributions on the ros2 website, I wasn't trying to assume what you may/may not know. Sorry if it came across in another way! :-)

My comments were meant to come across neutral, not negatively. I also apologize if they seemed to be. We do appreciate your interest and questions.

> could ros2::Duration duration = ros2::time::now() internally store a time and an uncertainty parameter and provide access to it without disrupting current APIs? For example in addition to duration.toSec() one could also call duration.uncertainty().
> 
> > Can this not already be exposed with a side channel, separate from the normal "get time" mechanisms?
> 
> The difficulty with a side channel is once you have hundreds of packages using one time function it is incredibly difficult to change it all out to use another! That's why I made my suggestion.

From a logistical perspective, I think it would be better to not integrate a query like this into the core Time type. For one reason, who's to say that this is the only metadata about time we'd want to store and query (which goes back to my argument about there not being a silver bullet solution that would work for everyone). Another issue is that I don't know if we could always deliver this information to the user (how do you calculate or get the uncertainty for the system clock). I'd rather see an interface like (taking TICSync as an example) `auto uncertainty = ticsync::get_uncertainty()`, or if necessary a function that gets both the uncertainty and the time at the same moment:

``` c++
ticsync_ros::ROSTimeWithUncertainty ticsync_time = ticsync_ros::now();
float uncertainty = ticsync_time.uncertainty();
float now = ticsync_time.ros_time.to_secs();
```

To your point that such an API as I just suggested would not work because you'd need to change all the code that already uses plain "ROS time", I'd say that I don't believe that is a limitation because even if uncertainty were in the core API we cannot force people to use it (I don't think anyways). I'm also not convinced that everyone would need to use the API which contain extra metadata like `uncertainty`.

Now, I don't even know if `uncertainty` is something TICSync could give you, but procedurally, I think having layered API's like I demonstrated above is the way to go until one API becomes the obvious solution for pretty much everyone.

That being said, I'm open to discussing individual extensions to the Time API if you can argue why it needs to be in the core API and cannot not be easily layered on top and used by those who need it. A good example of this is simulated time. The ROS Time type which can either give simulated time or time from the system clock is needed to facilitate simulation and log playback situations. It's possible a similar set of use cases surrounding data synchronization or interpolation would need something like uncertainty. But to be elevated to the core API I think it needs to two qualities: it needs to affect the behavior of the ROS time constructs for all code that uses it and it needs to have a reasonable fallback behavior when not in use. The simulated time object (ROS Time) both needs to affect many existing calls to the time API (but not all which is what "wall time" is for) and it behaves reasonably when simulated time is not being used.

gbiggs:
> On 2016/04/17, at 13:54, William Woodall notifications@github.com wrote:
> 
> Thanks for the detailed response! I'm just trying to contribute based on reading the document and request for contributions on the ros2 website, I wasn't trying to assume what you may/may not know. Sorry if it came across in another way! :-)
> 
> My comments were meant to come across neutral, not negatively. I also apologize if they seemed to be. We do appreciate your interest and questions.
> 
> could ros2::Duration duration = ros2::time::now() internally store a time and an uncertainty parameter and provide access to it without disrupting current APIs? For example in addition to duration.toSec() one could also call duration.uncertainty().
> 
> Can this not already be exposed with a side channel, separate from the normal "get time" mechanisms?
> 
> The difficulty with a side channel is once you have hundreds of packages using one time function it is incredibly difficult to change it all out to use another! That's why I made my suggestion.
> 
> From a logistical perspective, I think it would be better to not integrate a query like this into the core Time type. For one reason, who's to say that this is the only metadata about time we'd want to store and query (which goes back to my argument about there not being a silver bullet solution that would work for everyone). Another issue is that I don't know if we could always deliver this information to the user (how do you calculate or get the uncertainty for the system clock). I'd rather see an interface like (taking TICSync as an example) auto uncertainty = ticsync::get_uncertainty(), or if necessary a function that gets both the uncertainty and the time at the same moment:
> 
> ticsync_ros::ROSTimeWithUncertainty ticsync_time = ticsync_ros::now();
> float uncertainty = ticsync_time.uncertainty();
> float now = ticsync_time.ros_time.to_secs();
> To your point that such an API as I just suggested would not work because you'd need to change all the code that already uses plain "ROS time", I'd say that I don't believe that is a limitation because even if uncertainty were in the core API we cannot force people to use it (I don't think anyways). I'm also not convinced that everyone would need to use the API which contain extra metadata like uncertainty.
> 
> Now, I don't even know if uncertainty is something TICSync could give you, but procedurally, I think having layered API's like I demonstrated above is the way to go until one API becomes the obvious solution for pretty much everyone.
> 
> That being said, I'm open to discussing individual extensions to the Time API if you can argue why it needs to be in the core API and cannot not be easily layered on top and used by those who need it. A good example of this is simulated time. The ROS Time type which can either give simulated time or time from the system clock is needed to facilitate simulation and log playback situations. It's possible a similar set of use cases surrounding data synchronization or interpolation would need something like uncertainty. But to be elevated to the core API I think it needs to two qualities: it needs to affect the behavior of the ROS time constructs for all code that uses it and it needs to have a reasonable fallback behavior when not in use. The simulated time object (ROS Time) both needs to affect many existing calls to the time API (but not all which is what "wall time" is for) and it behaves reasonably when simulated time is not being used.

I think that the layered API approach suggested by Will is a good solution to this problem.

I am curious, though, how often does uncertainty information about a time source change?

Geoff

ahundt:
@gbiggs it can change from one second to another but can be dramatic in real use cases. For example, ping a server on the other side the world for a year straight from a cell phone and check out the resulting stats to see how much it could vary for a similarly connected robot driving around various continents. You don't really have to do it, just thinking about everywhere a heavy traveler goes in a year and the reliability of cellphone networks will likely be convincing. :-)

wjwwood:
> For example, ping a server on the other side the world for a year straight from a cell phone and check out the resulting stats to see how much it could vary for a similarly connected robot driving around various continents.

But that's travel time, not time differences between the machines. Those two machines would need to have their time sources synchronized through a different mechanism like NTP or with the PPS of GPS (making them each a "stratum 0" time sources https://en.wikipedia.org/wiki/Network_Time_Protocol#Clock_strata). I see the measuring of the message travel time (or of ping) is a different issue altogether... Even if you're talking about the average ping time over a year, that's not got anything to do with the time on the local system versus some relative other time source.

ahundt:
> But that's travel time, not time differences between the machines. Those two machines would need to have their time sources synchronized through a different mechanism like NTP or with the PPS of GPS (making them each a "stratum 0" time sources https://en.wikipedia.org/wiki/Network_Time_Protocol#Clock_strata). I see the measuring of the message travel time (or of ping) is a different issue altogether... Even if you're talking about the average ping time over a year, that's not got anything to do with the time on the local system versus some relative other time source.

I agree ping != time difference. Sorry for the confusion, I was merely trying to illustrate some source of uncertainty. 

In other words, while the number may be very small, even the machine next to the atomic clock on stratum 0 has an uncertainty value that can be estimated. That uncertainty will also change by some small amount with every measurement. Even the uncertainty measurement isn't 100% guaranteed to be certain or fixed. I could suddenly unplug and replug the connection to the atomic clock, or more likely my GPS signal could drop out, and that would cause a change (increase) in uncertainty. That uncertainty would decrease when the connection is re-established to the GPS signal. Correct?

tfoote:
Indeed uncertainty can change and it can be valuable to know about it for certain algorithms. However in most cases such as realtime mentioned at the top of the thread it's mostly about guaranteeing a minimum performance bound. 

Applications requiring very high precision such as the triggering between stereo pairs usually do that with a dedicated electrical circuit in hardware. There's a level that that is unachievable using standard networking protocols, especially if we need our system to work over arbitrary wireless networks.

Building in the ability to the core of the system to measure, estimate, and communicate uncertainty will add overhead. And I think that this ability is separable from the core implementation. I think that it would be great if there was an implementation of TICSync which could be put in as a source of system time, and can be hooked in as a custom time source.

One of the major challenges is simply picking a way of representing uncertainty. Depending on your application there are many different ways to represent uncertainty and depending on your application you'll use different approximations. Many use cases use the first order Gaussian model. But there are many more models with higher fidelity which can provide more information, such as a probability distribution function, which can range in representation from discrete values histogram to a continuous higher order function. 

Depending on your representation there is completely different math for operating on the uncertainty, especially propagating it and using it's results making it impossible for us to pick the one implementation for everyone to use for their application. 

As we talk about representing the uncertainty of the measurements I would suggest that for the areas where the uncertainty is important that a compound datatypes be created. Like we have PoseStamped and PoseWithCovarianceStamped we could have a TimeWithGuassianUncertainty message. It would contain the time representation and the uncertainty representation. And then methods can be defined to operate on those messages. If these messages are picked up and used in more packages it will become a standard. For other applications maybe a min and max range or bounds or more complicated uncertainty representation will be defined and become adopted by the system. 

To support this we would just need a system time source which can be queried for it's uncertainty and extending the clock API to be able to query it. With that any data producer which wants to publish the clock uncertainty datatype can query the clock time and uncertainty and then publish the datatype. 

But until we have a clock implementation that can provide an uncertainty measurement I don't think it's worth trying to build out this infrastructure. Once we have an implementation which can provide an estimate of the uncertainty we can extend the Clock api to query it and define messages to contain the datatype. 

PS note that this is a closed issue so you're only getting people who were subscribed and get the notifications. Unless we want to reopen(which we can't since it's already merged) this it should probably be on new thread, optimally a concrete proposal that can be discussed.

ahundt:
sure, where do you advise?

jackieokay:
@ahundt you could move the conversation to the sig-ng mailing list, or if you want to start a new design proposal incorporating some of the ideas from this discussion, you could make a pull request to this repository with a new article.

PR #73, opened 2016-01-07
added new realtime proposal

jackieokay:
Connects to ros2/realtime_support#17

Proposal for a more test-driven approach to real-time requirements in ROS 2.

### DIFF
--- a/articles/121_realtime_proposal.md
+++ b/articles/121_realtime_proposal.md
@@ -1,0 +1,248 @@
+---
+layout: default
+title: Proposal for Implementation of Real-time Systems in ROS 2
+permalink: articles/realtime_proposal.html
+abstract: Proposal for a test-driven approach to the real-time performance requirement in ROS 2.
+published: true 
+author: Jackie Kay
+---
+
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Requirements and Implementation of Real-Time Systems
+
+### System requirements
+
+The requirements of a real-time system vary depending on the use case.
+At its core, the real-time requirement of a system has two components:
+
+- Latency
+
+  - Update period (also known as deadline)
+  - Predictability
+
+- Failure mode
+
+  - How to react to a missed deadline
+
+Language referring to hard/soft/firm real-time systems generally refers to the failure mode.
+A hard real-time system treats a missed deadline as a system failure.
+A soft real-time system tries to meet deadlines but does not fail when a deadline is missed.
+A firm real-time system discards computations made by missed deadlines and may degrade its performance requirement in order to adapt to a missed deadline.
+
+Though in this document we consider the failure mode orthogonal to latency characteristics, a hard real-time failure mode is often associated with high predictability (low jitter).
+Safety-critical systems often require a hard real-time system with high predictability.
+
+### Implementation choices: computer architecture and OS
+
+Processors vary in capability, number of cores, memory size, and instruction set architectures, some of which are better suited to real-time performance than others.
+The selection of computer architecture may be due to cost, size, or energy efficiency, which could set a constraint for the rest of the system.
+A single robot may contain multiple computers with a variety of processor types, from the main control board to sensors.
+
+Another major consideration when designing a real-time system is the choice of operating system.
+A "hard" real-time operating system is truly deterministic.
+"Soft" real-time operating systems are designed to meet most deadlines with low jitter, but are nondeterministic.
+Most conventional user-facing operating systems are not designed with fairness and predictability as the first priorities, often because they have to support graphical displays and interactivity.
+Here are some features critical to predictable performance that are often exposed to the user in an RTOS, but not in a conventional OS (see Wikipedia, [Real-time operating system](https://en.wikipedia.org/wiki/Real-time_operating_system)).
+For more information about why these features matter, refer to the [introductory article on real-time systems](http://design.ros2.org/articles/realtime_background.html).
+
+- Task scheduling
+
+  - Expose thread priority and deterministic schedulers that incorporate priority
+  - Allow full thread preemption
+
+- Concurrency and synchronization primitives
+
+  - Floored semaphores (allow a higher priority thread to take control of locked semaphore)
+
+- Interrupt handling
+
+  - Disable interrupts
+  - Mask interrupts (temporarily suspend interrupts in a critical section)
+
+- Memory allocation
+
+  - Avoid pagefaults
+  - Avoid nondeterministic heap allocation algorithms
+
+Some real-time systems have a non real-time component, requiring inter-communication between the real-time and non-real-time components.
+An example from the ROS 1 world is real-time Orocos components communicating with ROS nodes.
+Another example is real-time code with a graphical or user interaction component.
+
+Scheduling non-real-time threads alongside real-time threads on the same core can introduce nondeterminism into a system.
+The non-real-time threads must have a lower priority than real-time threads in order to not interfere with real-time deadlines.
+The real-time threads should never be blocked on the non-real-time threads and should never be pre-empted by non-real-time threads.
+
+Real-time operating systems have a variety of methods for enforcing these constraints.
+For example, a Linux RT_PREEMPT system could use the [SCHED_IDLE](http://man7.org/linux/man-pages/man2/sched_setscheduler.2.html) scheduling policy for non-real-time threads.
+Systems like QNX and Xenomai give the option to partition real-time and non-real-time components onto different cores, or even different kernels.
+Partitioning is the safest approach to real-time and non-real-time intercommunication because it minimizes the change of interference from non-real-time threads, with the added cost of latency due to coordinating data between the partitions.
+
+Another important constraint is which processor architectures are supported by which operating systems, and which RTOS is an appropriate choice due to performance constraints of the architecture.
+For example, an architecture that doesn't support SIMD instructions cannot take full advantage of parallelism, but if the system doesn't run an OS that supports threading, or doesn't have multiple cores, this constraint doesn't matter.
+
+## Goal for ROS 2
+
+In order to support real-time systems in ROS 2, we propose an iterative approach for identifying limitations in the API that force the code path into nondeterministic/unpredictable behavior and performance bottlenecks.
+
+1. Select two use cases with real-time requirements that represent two extremes of the space described above, for example:
+
+   - An STM32 embedded board running no operating system or a hard real-time OS with 1 kHZ update frequency with 3% allowable jitter.
+   - An x86 computer running Xenomai or RT_PREEMPT that receives the sensor data over Ethernet or USB and processes the output in soft or firm real-time, but also runs separate non-RT processes for user interaction.
+
+2. Implement the system with ROS 2.
+
+3. Profile the performance of the system.
+
+4. Identify solutions to improve the performance of the system, e.g. by exposing more concurrency primitives through the appropriate abstraction.
+
+5. Adapt the ROS 2 API or underlying libraries to implement these solutions and to improve the experience of writing real-time code.
+
+6. Refactor the system for the new API if necessary and repeat.
+
+## Software architecture challenges and solutions
+
+Wherever ROS 2 libraries expose sources of nondeterminism, there should be a path that exposes an alternative.
+This section evaluates the current state of ROS 2 and rclcpp, predicts challenges to fulfilling real-time requirements, and presents possible solutions.
+
+### Memory Allocation
+
+Poor memory management can create performance issues due to page faults and nondeterministic algorithms for dynamically allocating memory.
+Although C++ has an advantage over many programming languages because memory management is exposed, the default memory allocator for operator `new` and dynamically sized data structures is not well-suited to real-time applications (see [Composing High-Performance Memory Allocators](https://people.cs.umass.edu/~emery/pubs/berger-pldi2001.pdf)).
+Dynamically-sized data structures in the C++ standard library offer an allocator interface, such as `std::vector` (see [cppreference.com](http://en.cppreference.com/w/cpp/concept/Allocator)).
+C++11 greatly improved this interface by introducing [`std::allocator_traits`](http://en.cppreference.com/w/cpp/memory/allocator_traits), which constructs an allocator compatible with the standard allocator interface from a minimal set of requirements
+To allow for optimal performance in `rclcpp`, the client library should expose custom allocators for publishers, subscribers, etc. that are propagated to standard library structures used by these entities.
+
+As of the writing of this document (January 2016), selecting custom allocators in `rclcpp` is almost fully implemented.
+The [`realtime_support`](https://github.com/ros2/realtime_support) package will offer real-time memory allocators and example code for real-time ROS 2.
+An example of the ROS 2 allocator API using the Two Level Segregate Fit allocator is currently available [in realtime_support](https://github.com/ros2/realtime_support/blob/master/tlsf_cpp/example/allocator_example.cpp).
+
+The following work remains to be done:
+
+- Improve the interface in rclcpp such that the user need only provide an "allocate" function, a "deallocate" function, and an optional pointer to the allocator state (this interface is currently outlined in the rcl API).
+- Custom allocators for services/clients.
+- Custom allocators for parameters.
+- Fully implement and test the allocator interface in rcl, rclc, and rmw.
+
+### Scheduling and synchronization
+
+Parallelism is valuable for real-time systems because it can reduce latency.
+
+Even in single-core architectures, concurrency is important to us because asynchronous patterns are common in ROS and ROS 2.
+
+The typical approaches to synchronizing concurrent access to data tend to decrease the predictability of a system and increase the complexity.
+For example, mutexes introduce the possibility of deadlocks when used improperly.
+As mentioned above, many real-time operating systems offer alternative synchronization primitives for real-time safety.
+
+Guidelines for synchronization in ROS 2:
+
+- Prefer atomics to mutexes when possible.
+
+  - Many atomic types are not implemented as lock-free due to architectural limitations, can check in C++ with [`std::atomic_is_lock_free`](http://en.cppreference.com/w/cpp/atomic/atomic_is_lock_free)
+
+- Thoroughly test multithreaded code for deadlock and livelock at the earliest opportunity.
+
+  - Consider compiling multithreaded regression tests with the [Clang ThreadSanitizer](http://clang.llvm.org/docs/ThreadSanitizer.html).
+
+- Hide synchronization primitives behind a layer of abstraction that wraps standard library classes such as `std::mutex` and `std::condition_variable` as the default behavior.
+
+  - Inject alternative synchronization primitives from RTOS-specific APIs where applicable.
+
+As of the time of writing, the Executor and the IntraProcessManager in rclcpp are designed to partially follow the last guideline.
+However, the abstractions are messy (particularly for the IntraProcessManager).
+There is no real-time multithreaded Executor that uses "real-time safe" synchronization primitives, similarly there is no such IntraProcessManager--there is only the abstraction path that would allow such alternatives to be implemented.
+There are also synchronization primitives that are not hidden behind an abstraction layer for the global interrupt handler and for parameters.
+
+### Middleware selection and QoS
+
+If the underlying middleware used in ROS 2 is not suitable for real-time performance, the entire system will fail to meet the real-time requirement.
+Network communication and discovery, particularly asynchronous communication, introduces a nondeterministic element into software.
+Though some real-time operating systems partially mitigate this (e.g. disabling interrupts), careful selection and use of a middleware is important.
+
+The embedded component of the system proposed will likely use `freertps` as the underlying middleware.
+The "desktop" system can be matrix-tested using the various DDS implementations supported by ROS 2.
+This allows us to compare the real-time performance of the different DDS implementations.
+
+Quality of service is always an important consideration with DDS.
+Some DDS QoS options were designed with real-time performance in mind, such as the DEADLINE policy (see opendds.org, [QoS Policy Usages](http://www.opendds.org/qosusages.html)).
+It may be convenient to expose these policies and provide a "real-time" QoS profile in rmw as a suggestion to ROS 2 users.
+Benchmarking and profiling may provide data to support the introduction of more QoS options into the rmw API.
+
+### Component lifecycle
+
+It should be noted that the planned component lifecycle features are useful for real-time programming, since it is common that real-time systems have a initialization phase in which the real-time constraint is relaxed, in which it is safe to preallocate memory, change thread priority, etc.
+Distinguishing between initialization and execution could lead to better self-verification of real-time constraints during the execution phase (see the ROS 2 design article on [managed nodes](http://design.ros2.org/articles/node_lifecycle.html)).
+
+## Testing
+
+The proposed system must be tested for correctness and expected behavior.
+The performance requirements of the system, such as latency, must also be benchmarked, and the inner workings of the system must be accessible, in order to identify bottlenecks.
+Therefore we propose a mix of automated unit tests, performance benchmarking, and profiling/introspection tools.
+
+### Benchmarking
+
+The system will be subjected to varying levels of stress to benchmark "best" and "worst"-case performance statistics.
+
+Stress can be applied "externally" via polluting the cache and running many garbage processes alongside the test system (if on an OS that supports processes).
+It could also be added "internally" (to ROS 2) by arbitrarily adding more publishers, subscribers, clients, services, etc.
+
+Possible performance metrics include:
+
+- Minimum, maximum, and average latency
+
+  - Latency of the update loop
+  - Message round trip time over the middleware
+
+- Standard deviation of latency
+
+- Number of missed deadlines/overruns
+
+- Number of messages received vs. number of messages published
+
+`rttest` is an existing library developed to instrument the update loop of a real-time process and report on various metrics described here.
+However, it contains code specific to Linux pthreads in its current state.
+If it is to be used in this work, the library needs to be generalized to support other operating systems.
+
+### Profiling Tools
+
+In addition to gathering performance statistics, it is useful to introspect the conditions under which the statistics were gathered to pinpoint areas where the code needs improvement.
+
+Two well-established open source command line profiling tools with orthogonal methodologies are:
+
+- `perf`: counts system-wide hardware events, e.g. instructions executed, cache misses, etc. (see [perf wiki](https://perf.wiki.kernel.org/index.php/Main_Page))
+
+  - Linux only.
+
+- `gprof`: recompiles source code with instrumentation to count time spent in each function for a particular program, with option to format output in a call graph (see [GNU gprof documentation](https://sourceware.org/binutils/docs/gprof/)).
+
+  - GNU and C/C++ only.
+
+However, these particular tools may not be well suited to an RTOS or a bare metal embedded system: `perf` is specialized for the Linux kernel and `gprof` requires code to be compiled and linked with `gcc`.
+These tools could be used to profile ROS 2 code on a patched Linux kernel like RT_PREEMPT, but there may be performance quirks in another RTOS that are not captured on Linux.
+When this proposal is implemented, if changes are made to improve performance is made based on profiling data from Linux but no overall improvement is seen in the embedded system, then an alternative approach to profiling should be investigated.
+For example, the [RTEMS](https://devel.rtems.org/wiki/Developer/Tracing) real-time operating system offers a tracing tool that could be used for profiling (8).
+
+### Verification
+
+In addition to performance testing, which can be thought of as a kind of regression testing, any change made to ROS 2 with the intent to improve performance in the real-time test system should be verified with an automated test.
+For example, the allocator pipeline has a test case for publishers/subscribers with custom allocators that fails if calls are made to the default system allocator during the real-time execution phase.
+A similar test could be written for the proposed goal of abstracting synchronization primitives, though the implementation may be more challenging since the allocator pipeline was helped by the ability to override global new in C++.
+
+The expected response of a hard or firm real-time system to missed deadlines should also be thoroughly tested.
+
+## Timeline
+
+This proposal represents a short burst of work to develop the initial test system, followed by extended maintenance as testing identifies performance bottlenecks.
+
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please rebase to validate against the latest linters.

jackieokay:
Rebased, Travis passed.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+- Failure mode
+
+  - How to react to a missed deadline
+
+Language referring to hard/soft/firm real-time systems generally refers to the failure mode.
+A hard real-time system treats a missed deadline as a system failure.
+A soft real-time system tries to meet deadlines but does not fail when a deadline is missed.
+A firm real-time system discards computations made by missed deadlines and may degrade its performance requirement in order to adapt to a missed deadline.
+
+Though in this document we consider the failure mode orthogonal to latency characteristics, hard real-time systems are usually designed to be safety critical and therefore require high determinism (low jitter).
```
dirk-thomas: `hard real-time systems are usually designed to be safety critical`

I would say it is the other way around:

`safety critical systems are usually designed to be hard real-time`?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Requirements and Implementation of Real-Time Systems
+
+### System requirements
+
+The requirements of a real-time system varies depending on the use case.
```
dirk-thomas: `requirements .. varies`

Isn't it either `requirements vary` or `requirement varies` (I think you want the first one)?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+
+Processors vary in capability, number of cores, memory size, and instruction set architectures, some which are better suited to real-time performance than others.
+The selection of computer architecture may be due to cost, size, or energy efficiency, which could set a constraint for the rest of the system.
+A single robot may contain multiple computers with a variety of processor types, from the main control board to sensors.
+
+Another major consideration when designing a real-time system is the choice of operating system.
+A true "hard" real-time operating system is fully deterministic.
+"Soft" real-time operating systems are designed to meet most deadlines deterministically, but may still have an element of nondeterminism.
+Most conventional user-facing operating systems are not designed with determinism or fairness as the first priority, often because they have to support graphical displays and interactivity.
+Here are some features critical to deterministic performance that are often exposed to the user in an RTOS, but not in a conventional OS (1).
```
dirk-thomas: Instead of relying on manual numbering (`(1)`) this should use markdown specific markup. Otherwise it is not linked and will likely get out of sync.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+
+Implement the system with ROS 2.
+
+Profile the performance of the system.
+
+Identify solutions to improve the performance of the system, e.g. by exposing more concurrency primitives through the appropriate abstraction.
+
+Adapt the ROS 2 API or underlying libraries to implement these solutions and to improve the experience of writing real-time code.
+
+Refactor the system for the new API if necessary and repeat.
```
dirk-thomas: Are the last five sentences some kind of bullets? If yes, should the become a list and a sentence before like "Next steps"?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Software architecture challenges and solutions
+
+Wherever ROS 2 libraries expose sources of nondeterminism, there should be a path that exposes an alternative.
+This section evaluates the current state of ROS 2 and rclcpp, predicts challenges to fulfilling real-time requirements, and presents possible solutions.
+
+### Memory Allocation
+
+As of the writing of this document (1/6/15), selecting custom allocators in rclcpp is almost fully implemented.
```
dirk-thomas: You wrote this already a year ago? :eyes:

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+This allows us to compare the real-time performance of the different DDS implementations.
+
+Quality of service is always an important consideration with DDS.
+Some DDS QoS options were designed with real-time performance in mind, such as the DEADLINE policy (4).
+It may be convenient to expose these policies and provide a "real-time" QoS profile in rmw as a suggestion to ROS 2 users.
+Benchmarking and profiling may provide data to support the introduction of more QoS options into the rmw API.
+
+### Component lifecycle
+
+It should be noted that the planned component lifecycle features are useful for real-time programming, since it is common that real-time systems have a initialization phase in which the real-time constraint is relaxed, in which it is safe to preallocate memory, change thread priority, etc.
```
dirk-thomas, line 181: Reference to the article.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+For example, the allocator pipeline has a test case for publishers/subscribers with custom allocators that fails if calls are made to the default system allocator during the real-time execution phase.
+A similar test could be written for the proposed goal of abstracting synchronization primitives, though the implementation may be more challenging since the allocator pipeline was helped by the ability to override global new in C++.
+
+The expected response of a hard or firm real-time system to missed deadlines should also be thoroughly tested.
+
+## Timeline
+
+This proposal represents a short burst of work to develop the initial test system, followed by extended maintenance as testing identifies performance bottlenecks.
+
+Since a key part of this proposal is the development of a hard real-time embedded system, this work depends on the development of `freertps` and `rmw_freertps` and the porting of `rclcpp` to STM32 or another microcontroller type.
```
dirk-thomas: of `rclcpp` or `rclc`?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+
+- Alpha 4 (2/12/15): Create initial system and test suite (2-3 weeks).
+- Every release thereafter: Iterate on ARPI, refactor and expand test suite as new features are added to core ROS 2 libraries (1-2 weeks each release cycle).
```
dirk-thomas: ARPI?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,230 @@
...
+
+This work should share as much labor as possible with the roscar effort (running ROS 2 on a small mobile robot) or perhaps be considered a subset of roscar.
```
dirk-thomas: I wouldn't refer to `roscar` in this article. Also the point should be obvious - we always try to reuse stuff to minimize our effort.

dirk-thomas:
Beside the minor comments LGTM.

articles/121_realtime_proposal.md
jackieokay: I actually don't think we have a current example of markdown specific citation, it's something I struggled with the last time I contributed an article to this repository... I think the conclusion last time was that we need to add a plugin in order to properly sync references.

articles/121_realtime_proposal.md
jackieokay: yes, I had a very long backlog of work before making this pull request :stuck_out_tongue_closed_eyes:

articles/121_realtime_proposal.md
jackieokay, line 181: The reference was added in 928dd5f9153b790b0180d976d1bfde4df4333bda

https://github.com/ros2/design/commit/928dd5f9153b790b0180d976d1bfde4df4333bda#diff-2309eff1b7fac7aaf93df0fde7d4c62eR155

articles/121_realtime_proposal.md
dirk-thomas: We could either use at least a link or don't put the URLs at the bottom but link them directly inline?

jackieokay: Fine with me.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,215 @@
...
+
+Some real-time systems have a non real-time component, e.g. for graphical display or other user interaction.
+This would require an operating system that allows non real-time threads to be scheduled alongside real-time threads.
+Because of the extra nondeterminism introduced by this requirement, this kind of system would be better suited to a soft real-time constraint.
+
+Another important constraint is which processor architectures are supported by which operating systems, and which RTOS is an appropriate choice due to performance constraints of the architecture.
+
+## Goal for ROS 2
+
+In order to support real-time systems in ROS 2, we propose an iterative approach for identifying limitations in the API that force the code path into performance bottlenecks.
```
tkruse: Hi. Not knowing anything about real-time, I stumbled on this sentence, because earlier paragraphs keep talking about determinism only, but this one suddenly jumps to performance bottlenecks only. As in:  "real-time system require determinism. Goal for ROS2 is higher performance." Also in general, I would be curious in how to make sure that later additions / changes to the ROS2 code will not break whatever state is achieved. I.e. is it possible to write regression tests for the performance / determinism guarantees? I don't know what's technically possible there, if no solution exists for automated testing, then maybe a result could be a maintained ros package / library that can be run anytime by anyone on the given boards to check whether ros2 has changed in an undesired way. In other words, don't plan to do this just once, plan to do this on every release (that changed core code), if possible.
The other part I don't quite get is that ROS2 wraps rmw implementations, so in some sense if the rmw implementation guarantees some performance/non-determinism, then you might consider writing tests that test the ros2 glue code with only a dummy rmw, even if additional integration tests can also be valuable and detect wrong assumptions about rmw implementations.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,215 @@
...
+6. Refactor the system for the new API if necessary and repeat.
+
+## Software architecture challenges and solutions
+
+Wherever ROS 2 libraries expose sources of nondeterminism, there should be a path that exposes an alternative.
+This section evaluates the current state of ROS 2 and rclcpp, predicts challenges to fulfilling real-time requirements, and presents possible solutions.
+
+### Memory Allocation
+
+As of the writing of this document (1/6/16), selecting custom allocators in rclcpp is almost fully implemented.
```
tkruse: An introductory sentence about what custom allocators are and why they are relevant here may be nice.

articles/121_realtime_proposal.md
tkruse: Ah, I guess some of this is adressed at the bottom of the document, so ignore the regression test question here (though you might add a reference to later sections).

jackieokay: Using a dummy rmw implementation for testing is a good idea; I believe it's something we've considered before for other reasons but didn't do to focus on implementing features instead.

I tried to clarify the wording of this sentence in a04b406. Let me know if you have more feedback.

articles/121_realtime_proposal.md
jackieokay: Addressed in a04b406.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,225 @@
...
+
+### System requirements
+
+The requirements of a real-time system vary depending on the use case.
+At its core, the real-time requirement of a system has two components:
+
+- Latency
+
+  - Update period
+  - Determinism (measured by jitter)
```
wjwwood: Just as a point of language - is this an appropriate use of "determinism"? This word invokes the idea of a philosophy to me, and the related word deterministic to me is fairly binary: It is either deterministic or not, there is no degree to which something can be deterministic or not. See:
- http://www.merriam-webster.com/dictionary/determinism

> determinism (1) philosophy : the belief that all events are caused by things that happened before them and that people have no real ability to make choices or control what happens
- https://en.wikipedia.org/wiki/Deterministic_algorithm

> ... given a particular input, will **always** produce the same output ...

I'm not sure what a better term would be, but predictable (predictability) seems like a candidate to me. The idea being:
- A hard realtime system's jitter is absolutely predictable therefore it is deterministic (guaranteed or proven).
- A soft realtime system's jitter is reasonably predictable (within a statistical bound).
  - At this point it is more of a variable quality of the system rather than a characteristic.
- A firm realtime system can adjust its prediction (requirement) for jitter based on failures?

Sorry to be so pedantic, but I think it will help us to use precise language where possible.

To be clear, I think your descriptions of those terms are spot on, I'm proposing we not use the word "determinism".

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,225 @@
...
+This would require an operating system that allows non real-time threads to be scheduled alongside real-time threads.
+Because of the extra nondeterminism introduced by this requirement, this kind of system would be better suited to a soft real-time constraint.
+
+Another important constraint is which processor architectures are supported by which operating systems, and which RTOS is an appropriate choice due to performance constraints of the architecture.
+
+## Goal for ROS 2
+
+In order to support real-time systems in ROS 2, we propose an iterative approach for identifying limitations in the API that force the code path into nondeterministic behavior and performance bottlenecks.
+
+1. Select two use cases with real-time requirements that represent two extremes of the space described by diagram, for example:
```
wjwwood: "as described by the diagram"? Also what diagram? Is there one missing or did I over look it?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,225 @@
...
+## Software architecture challenges and solutions
+
+Wherever ROS 2 libraries expose sources of nondeterminism, there should be a path that exposes an alternative.
+This section evaluates the current state of ROS 2 and rclcpp, predicts challenges to fulfilling real-time requirements, and presents possible solutions.
+
+### Memory Allocation
+
+Poor memory management can create performance issues due to page faults and nondeterministic algorithms for dynamically allocating memory.
+Although C++ has an advantage over many programming languages because memory management is exposed, the default memory allocator for operator `new` and dynamically sized data structures is not well-suited to real-time applications (see [Composing High-Performance Memory Allocators](https://people.cs.umass.edu/~emery/pubs/berger-pldi2001.pdf)).
+C++11 implements an allocator interface for injecting a custom allocator into standard library structures such as `std::vector` (see [cppreference.com](http://en.cppreference.com/w/cpp/concept/Allocator)).
```
wjwwood: Is it just as accurate to say "C++ implements an allocator interface ..." or is it truly a feature of C++11?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,225 @@
...
+- Custom allocators for parameters.
+- Fully implement and test the allocator interface in rcl, rclc, and rmw.
+
+### Scheduling and synchronization
+
+Parallelism is valuable for real-time systems because it can reduce latency.
+
+Even in single-core architectures, concurrency is important to us because asynchronous patterns are common in ROS and ROS 2.
+
+The typical approaches to synchronizing concurrent access to data tend to decrease the determinism of a system and increase the complexity.
```
wjwwood: "tend to make the system non-deterministic"? This goes back to my argument that deterministic is not a sliding scale.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,225 @@
...
+For example, mutexes introduce the possibility of deadlocks when used improperly.
+As mentioned above, many real-time operating systems offer alternative synchronization primitives for real-time safety.
+
+Guidelines for synchronization in ROS 2:
+
+- Prefer atomics to mutexes when possible.
```
wjwwood, line 147: Might be worth mentioning that atomics should be checkable to see if they are lock-free. Therefore our code will be lock-free given these other conditions hold, e.g. `atomic_is_lock_free<T>` is `true` for `T`.

wjwwood:
Other than a few comments, this lgtm. I think it's pretty clear and straight forward about what we're trying to achieve, how to know when we've achieved it, and what's involved. Thanks for working on it @jacquelinekay.

articles/121_realtime_proposal.md
jackieokay: Yeah, I agree, I have mostly misused the term "determinism" throughout this document. A system is either deterministic or nondeterministic. I will take a look at improving the precision of the language in this article.

articles/121_realtime_proposal.md
jackieokay: Haha, I was planning on adding diagrams to visualize the space of a hardware platforms vs. commonly chosen realtime-requirements and then I decided my artistic skills weren't up to par. I'll take out this reference.

articles/121_realtime_proposal.md
jackieokay: I'm not sure what you mean. What do you think is a more precise wording of the idea here?

wjwwood: If there is something new in C++11 that allows you to do this then change nothing. If you could do it before C++11, then consider just using C++. That's all I was trying to say.

jackieokay: true, the most correct thing to say would be that allocators for standard library objects have been available for a while, but the C++11 standard makes them more convenient to use by offering `allocator_traits`.

jackieokay:
I'll wait for Travis to pass and squash and merge if it looks good, unless anyone thinks I should solicit more feedback from the community beforehand. I could also solicit feedback after merging.

gbiggs:
This member of the community has actually queued up some ROS design article reviewing as Friday Fun Time work (seriously, compared to the other stuff I have to do, reviewing ROS design articles counts as "fun"). Assuming I have some, I can provide feedback before or after merging as you like, but where would you like it if after merging? The mailing list? It's kinda hard to quote there...

jackieokay:
@gbiggs I'll just leave this PR open so that you can post comments. Have Fun! :)

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+## Requirements and Implementation of Real-Time Systems
+
+### System requirements
+
+The requirements of a real-time system vary depending on the use case.
+At its core, the real-time requirement of a system has two components:
+
+- Latency
+
+  - Update period
```
gbiggs: I'm probably being pedantic, but for consistency with the "failure mode" comment, perhaps include the term "deadline" here, too (as a synonym for update period?).

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+
+  - How to react to a missed deadline
+
+Language referring to hard/soft/firm real-time systems generally refers to the failure mode.
+A hard real-time system treats a missed deadline as a system failure.
+A soft real-time system tries to meet deadlines but does not fail when a deadline is missed.
+A firm real-time system discards computations made by missed deadlines and may degrade its performance requirement in order to adapt to a missed deadline.
+
+Though in this document we consider the failure mode orthogonal to latency characteristics, a hard real-time failure mode is often associated with high predictability (low jitter).
+Safety-critical systems often require a hard real-time with system with high predictability.
```
gbiggs: "real-time with system" -> "real-time system"

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Implementation choices: computer architecture and OS
+
+Processors vary in capability, number of cores, memory size, and instruction set architectures, some which are better suited to real-time performance than others.
```
gbiggs: "some which" -> "some of which"

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+- Concurrency and synchronization primitives
+
+  - Floored semaphores (allow a higher priority thread to take control of locked semaphore)
+
+- Interrupt handling
+
+  - Disable interrupts
+  - Mask interrupts (temporarily suspend interrupts in a critical section)
+
+- Memory allocation
```
gbiggs, line 71: What about memory allocation is important? Perhaps add something about how it must be deterministic, page faults are evil, etc.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+
+Some real-time systems have a non real-time component, e.g. for graphical display or other user interaction.
```
gbiggs: Or for communication with non-real-time parts of a robot's software. ROS1 and Orocos being used together are the perfect example here.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+This would require a soft real-time operating system that allows non real-time threads to be scheduled alongside real-time threads.
```
gbiggs: Is a soft real-time OS required? Or just "an OS with real-time capability [irrespective of it being hard, firm or soft] that allows both real-time and non-real-time threads to be scheduled together without impacting the performance of the real-time threads." e.g. QNX supports hard real-time but can also run non-real-time threads; you can even throw all your non-real-time stuff in a separate partition to make sure it really can't have any impact on the hard real-time parts.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+
+Another important constraint is which processor architectures are supported by which operating systems, and which RTOS is an appropriate choice due to performance constraints of the architecture.
+
+## Goal for ROS 2
+
+In order to support real-time systems in ROS 2, we propose an iterative approach for identifying limitations in the API that force the code path into nondeterministic/unpredictable behavior and performance bottlenecks.
+
+1. Select two use cases with real-time requirements that represent two extremes of the space described above, for example:
+
+   - An STM32 embedded board running no operating system or a hard real-time OS with 1 kHZ update frequency with 3% allowable jitter.
```
gbiggs, line 98: Where did 3% come from?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+   - An x86 computer running Xenomai or RT_PREEMPT that receives the sensor data over Ethernet or USB and processes the output in soft or firm real-time, but also runs separate non-RT processes for user interaction.
```
gbiggs, line 99: I think that Xenomai and RT_PREEMPT present use cases significantly different enough to be worth considering them separately. As both are commonly used in robotics, I also think it is worth doing both.

gbiggs, line 99: What about hard real-time distributed communications? This use case mentions using Ethernet/USB, but it is not specific on whether that is a real-time distributed link to another computer running ROS, or something more like a Hokuyo sensor connected using Ethernet. The ability to communicate over a real-time hardware link may be a significant use case in future robots that contain multiple computers providing a single integrated (possibly hard) real-time control system.

Actually now that I think about it, this may also apply to the embedded use case - think of a robot made up of distributed embedded controllers all running their own parts of the robot, but with real-time response.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+Although C++ has an advantage over many programming languages because memory management is exposed, the default memory allocator for operator `new` and dynamically sized data structures is not well-suited to real-time applications (see [Composing High-Performance Memory Allocators](https://people.cs.umass.edu/~emery/pubs/berger-pldi2001.pdf)).
+Dynamically-sized data structures in the C++ standard library offer an allocator interface, such as `std::vector` (see [cppreference.com](http://en.cppreference.com/w/cpp/concept/Allocator)).
+C++11 greatly improved this interface by introducing [`std::allocator_traits`](http://en.cppreference.com/w/cpp/memory/allocator_traits), which constructs an allocator compatible with the standard allocator interface from a minimal set of requirements
+To allow for optimal performance in `rclcpp`, the client library should expose custom allocators for publishers, subscribers, etc. that are propagated to standard library structures used by these entities.
+
+As of the writing of this document (1/6/16), selecting custom allocators in `rclcpp` is almost fully implemented.
```
gbiggs: Ambiguous American date order! Yuck! ;)

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+An example of the ROS 2 allocator API using the Two Level Segregate Fit allocator is available [on Github](https://github.com/ros2/realtime_support/blob/master/tlsf_cpp/example/allocator_example.cpp).
+
+The following work remains to be done:
+
+- Improve the interface in rclcpp such that the user need only provide an "allocate" function, a "deallocate" function, and an optional pointer to the allocator state (this interface is currently outlined in the rcl API).
+- Custom allocators for services/clients.
+- Custom allocators for parameters.
+- Fully implement and test the allocator interface in rcl, rclc, and rmw.
```
gbiggs, line 133: Is ROS2 going to provide one or a set of allocators suitable for the real-time case in some common situations? Or is the current plan just to provide an API for providing your own allocator?

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Scheduling and synchronization
```
gbiggs, line 135: You're accumulating a lot of knowledge about implementing real-time, so it would be good to preserve it in a ROS-specific fashion so that it is immediately accessible to ROS users. At an appropriate point in the future, it would be worth adding some documentation to the wiki giving an example of correct use of atomics and real-time-safe synchronisation primitives. Some people may find this useful for writing their nodes, but the use case I'm thinking of is people writing custom executors that meet their (assumed to be) unique needs.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+
+Stress can be applied "externally" via polluting the cache and running many garbage processes alongside the test system (if on an OS that supports processes).
+It could also be added "internally" (to ROS 2) by arbitrarily adding more publishers, subscribers, clients, services, etc.
+
+Possible performance metrics include:
+
+- Minimum, maximum, and average latency
+- Standard deviation of latency
+- Number of missed deadlines/overruns
+- Number of messages received vs. number of messages published
```
gbiggs, line 208: Two types of latency should be checked: Update loop, and message publish to message receive.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+It should be noted that the planned component lifecycle features are useful for real-time programming, since it is common that real-time systems have a initialization phase in which the real-time constraint is relaxed, in which it is safe to preallocate memory, change thread priority, etc.
+Distinguishing between initialization and execution could lead to better self-verification of real-time constraints during the execution phase (see the ROS 2 design article on [managed nodes](http://design.ros2.org/articles/node_lifecycle.html)).
+
+## Testing
+
+The proposed system must be tested for correctness and expected behavior.
+The performance requirements of the system, such as latency, must also be benchmarked, and the inner workings of the system must be accessible, in order to identify bottlenecks.
+Therefore we propose a mix of automated unit tests, performance benchmarking, and profiling/introspection tools.
+
+### Benchmarking
```
gbiggs, line 190: The OSRF could work with these people to produce (or use theirs) a real-time QA farm for ROS.

http://www.osadl.org/QA-Farm-Realtime.qa-farm-about.0.html

If resources are available, it might even be useful providing one that package maintainers can access.

articles/121_realtime_proposal.md
```diff
@@ -0,0 +1,227 @@
...
+The expected response of a hard or firm real-time system to missed deadlines should also be thoroughly tested.
+
+## Timeline
+
+This proposal represents a short burst of work to develop the initial test system, followed by extended maintenance as testing identifies performance bottlenecks.
+
+Since a key part of this proposal is the development of a hard real-time embedded system, this work depends on the development of `freertps` and `rmw_freertps` and the porting of `rclc` to STM32 or another microcontroller type.
+
+- Alpha 4 (2/12/15): Create initial system and test suite (2-3 weeks).
+- Every release thereafter: Iterate on ARPI, refactor and expand test suite as new features are added to core ROS 2 libraries (1-2 weeks each release cycle).
```
gbiggs: ARPI?

articles/121_realtime_proposal.md
jackieokay, line 71: I go into this in more detail on line 104 (https://github.com/ros2/design/pull/73#discussion-diff-49821578R104), should I summarize the ideas here and there, move them here, or leave it as is?

articles/121_realtime_proposal.md
jack-oquin: I was momentarily confused, too. I suppose that's 2016-01-06?

jack-oquin:
+1 overall. A solid piece of work, good plan, useful feedback.

articles/121_realtime_proposal.md
jackieokay: I know, everything we do is backwards...

articles/121_realtime_proposal.md
jackieokay, line 190: Yeah, we tried contacting them a while ago and didn't get very far--I might reach out again and link them to this document once it's done. Do you have any contacts there?

articles/121_realtime_proposal.md
jackieokay, line 98: It's an arbitrary choice. I could write a bit more about how these requirements are chosen, but in this case the requirement doesn't matter as much as profiling the jitter and driving it down does. Any advice on this matter?

articles/121_realtime_proposal.md
jackieokay, line 99: True, the ability to test distributed communication between `n` embedded boards would be really useful, where `n` is on the order of 10-20.

Since software is awesome and hardware is icky (biased view here), I wonder if it's possible to emulate many small embedded ARM computers on one beefy x86 server. That way we could arbitrarily scale the size of an emulated embedded distributed system, without having to order and hook up more boards... This idea might be a bit out of scope, however.

articles/121_realtime_proposal.md
jackieokay, line 135: Agreed. Right now I don't think I should write these tutorials and advertise them as the magic bullet to your robotics performance woes because we haven't proven anything concrete about ROS 2's performance yet. But once the client library API has settled a bit and the correct "patterns" have been proven, writing these tutorials will be a high priority so that we can educate those who are interested in these issues.

articles/121_realtime_proposal.md
jackieokay, line 133: 511bfb3 adds the sentence on line 122 about the realtime_support package, which currently provides a starter RT-allocator

articles/121_realtime_proposal.md
gbiggs, line 71: I think a one-line summary would be suitable.

articles/121_realtime_proposal.md
gbiggs, line 99: You probably could but I think you would need really detailed profiling to prove anything that way.

Another option is to get your local embedded expert to set something up for you. ;)

articles/121_realtime_proposal.md
gbiggs, line 190: Unfortunately I don't. I only heard about them through ROS Industrial. If their software is available for download you could probably produce a similar setup but it seems like it would take considerable effort. Still, I think it would be a valuable resource for maintaining the real-time suitability if ROS and nodes in the future. Something to do when more time is available?

gbiggs:
No more comments from me. I think the plan will lead to a solid real-time platform. +1

jackieokay:
Thank you Geoff, Jack and Thibault for your feedback!

PR #76, opened 2016-01-12
define type mapping between ROS IDL and C++ / Python

dirk-thomas:
Since this is a critical piece I would like to collect multiple +1's for this before merging it. Please review carefully.

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -27,8 +27,8 @@ This article specifies the generated C++ code for ROS interface types defined in
 All code of a ROS package should be defined in a namespace named after the package.
 To separate the generated code from other code within the package it is defined in a sub namespace:
 
-- namespace for ROS messages: `package_name::msg`.
-- namespace for ROS services: `package_name::srv`.
+- namespace for ROS messages: `<package_name>::msg`.
+- namespace for ROS services: `<package_name>::srv`.
 
 ## Generated files
 
@@ -47,8 +47,46 @@ For each message two files are being generated:
 - `<my_message_name>.hpp` currently only includes `<my_message_name>__struct.hpp`
 - `<my_message_name>__struct.hpp` containing the definition of the struct
 
-This allow to add addition files beside the one with the suffix `__struct` to provide additional functionality.
+This allows to add additional files besides the one with the suffix `__struct` to provide additional functionality.
 For each additional functionality it can be decided to include it from the first header file.
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> specify content of <code>&lt;my_message_name&gt;__traits.hpp</code> file
+</div>
+
+### Types
+
+#### Mapping of primitive types
+
+| ROS type | C++ type    |
+| -------- | ----------- |
+| bool     | bool        |
+| byte     | uint8_t     |
+| char     | char        |
+| float32  | float       |
+| float64  | double      |
+| int8     | int8_t      |
+| uint8    | uint8_t     |
+| int16    | int16       |
+| uint16   | uint16      |
+| int32    | int32       |
+| uint32   | uint32      |
+| int64    | int64       |
+| uint64   | uint64_t    |
+| string   | std::string |
+
+#### Mapping of arrays and bounded strings
+
+| ROS type                | C++ type         |
+| ----------------------- | ---------------- |
+| static array            | std::array<T, N> |
+| unbounded dynamic array | std::vector<T>   |
+| bounded dynamic array   | std::array<T, N> |
+| bounded string          | std::string      |
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> bounded dynamic array are currently mapped to <code>std::array</code> but that prevents to store less then N values
+</div>
 
 ### Members
 
@@ -67,7 +105,7 @@ The *default constructor* initializes all members with their default value.
 Optionally the constructor can be invoked with an allocator.
 
 The struct has no constructor with positional arguments for the members.
-The short reason for this is that if code would rely on positional arguments to construct data objects changing a message definition would break existing code in various ways.
+The short reason for this is that if code would rely on positional arguments to construct data objects changing a message definition would break existing code in subtle ways.
 Since this would discourage evolution of message definitions the data structures should be populated by setting the members separately, e.g. using the setter methods.
 
 ### Setters
@@ -107,5 +145,6 @@ The generated code is split across multiple files the same way as message are.
 
 ### Request and response messages
 
-For the request as well as the response part messages are being generated which have the same name as the service with a suffix being either `_Request` or `_Response`.
-These messages are still defined in the `srv` sub namespace.
+For the request and response parts of a service separate messages are being generated.
+These messages are named after the service and have either a `_Request` or `_Response` suffix.
+They are are still defined in the `srv` sub namespace.
--- a/articles/114_generated_interfaces_python.md
+++ b/articles/114_generated_interfaces_python.md
@@ -1,0 +1,113 @@
+---
+layout: default
+title: Generated Python interfaces
+permalink: articles/generated_interfaces_python.html
+abstract:
+  This article describes the generated Python code for ROS 2 interfaces.
+published: true
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+---
+
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Scope
+
+This article specifies the generated Python code for ROS interface types defined in the [interface definition article](interface_definition.html).
+
+## Namespacing
+
+All code of a ROS package should be defined in a Python package with the same name.
+To separate the generated code from other code within the package it is defined in a sub module:
+
+- module for ROS messages: `<package_name>.msg`.
+- module for ROS services: `<package_name>.srv`.
+
+## Generated files
+
+Following the Python conventions the namespace hierarchy is mapped to a folder structure.
+The package and module names and therefore the folder and file names use lowercase alphanumeric characters with underscores for separating words (following PEP 8).
+Python files end with `.py`.
+
+## Messages
+
+For a message a Python `class` with the same name as the message is generated in the file `_<my_message_name>.py`.
+The Python module `<package_name>.msg` / `<package_name>.srv` exports all message / service classes without the message module name to shorten import statements, e.g. `import <package_name>.msg.<MyMessageName`
+
+### Types
+
+#### Mapping of primitive types
+
+| ROS type | Python type                  |
+| -------- | ---------------------------- |
+| bool     | builtins.bool                |
+| byte     | builtins.bytes with length 1 |
+| char     | builtins.str with length 1   |
+| float32  | builtins.float               |
+| float64  | builtins.float               |
+| int8     | builtins.int                 |
+| uint8    | builtins.int                 |
+| int16    | builtins.int                 |
+| uint16   | builtins.int                 |
+| int32    | builtins.int                 |
+| uint32   | builtins.int                 |
+| int64    | builtins.int                 |
+| uint64   | builtins.int                 |
+| string   | builtins.str                 |
+
+#### Mapping of arrays and bounded strings
+
+| ROS type                | Python type   |
+| ----------------------- | ------------- |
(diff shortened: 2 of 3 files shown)


articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,108 @@
...
+| bounded string          | builtins.str  |
+
+### Properties
+
+The class has same-named properties for every field of the message.
+The setter of a property will ensure that ROS type constraints beyond the Python type are enforced.
+
+### "Constants"
+
+A constant is are defined as class variables with uppercase name.
```
esteve: s/is are/is/

esteve: Or "Constants are defined as..."

dirk-thomas: FIxed in 8b3f8cfa6803b9cb6b7926c40bb96d86d39ddf2f

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,108 @@
...
+| bounded string          | builtins.str  |
+
+### Properties
+
+The class has same-named properties for every field of the message.
+The setter of a property will ensure that ROS type constraints beyond the Python type are enforced.
+
+### "Constants"
+
+A constant is defined as a class variables with uppercase name.
```
esteve: A constant is defined as a class variable with an uppercase name.

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Constructors
+
+The *__init__* function initializes all members with their default value.
+
+The class constructor supports keyword arguments for all members.
```
esteve: keyword arguments only or in addition to positional arguments?

esteve: I'd prefer only keyword arguments, but it should be noted in the document.

articles/114_generated_interfaces_python.md
dirk-thomas: 969f528c34def3565f3da7f450472c22ab01decb

articles/114_generated_interfaces_python.md
dirk-thomas: 969f528c34def3565f3da7f450472c22ab01decb

articles/112_generated_interfaces_cpp.md
```diff
@@ -50,6 +50,40 @@ For each message two files are being generated:
 This allow to add addition files beside the one with the suffix `__struct` to provide additional functionality.
```
esteve: "This allows to add additional files besides the one with the suffix `__struct` to provide additional functionality."

Though there are too many add and additional in this sentence for my taste.

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Generated files
+
+Following the Python conventions the namespace hierarchy is mapped to a folder structure.
+The package and module names and therefore the folder and file names use lowercase alphanumeric characters with underscores for separating words (following PEP 8).
+Python files end with `.py`.
+
+## Messages
+
+For a message a Python `class` with the same name is generated in the file `<my_message_name>.py`.
```
esteve: Should this follow the same pattern as ROS1, where messages are generated in `_<my_message_name>.py` and there's an `__init__.py` file that imports all of them? For example:

``` python
from package.msg._Structure import Structure
```

and in `_Structure.py`:

``` python
class Structure:
    pass
```

articles/112_generated_interfaces_cpp.md
dirk-thomas: Update in fc387348fe8f233bbfb4f1cb77de4aa075e78df4 - even if the line was not even part of this PR.

articles/114_generated_interfaces_python.md
esteve: Also, is it worth mentioning that, since we're supporting Python 3 only, we can omit `(object)` in the class definition?

dirk-thomas: We could do that.

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,110 @@
...
+| float64  | builtins.float               |
+| int8     | builtins.int                 |
+| uint8    | builtins.int                 |
+| int16    | builtins.int                 |
+| uint16   | builtins.int                 |
+| int32    | builtins.int                 |
+| uint32   | builtins.int                 |
+| int64    | builtins.int                 |
+| uint64   | builtins.int                 |
+| string   | builtins.str                 |
```
esteve, line 63: Could you add a `TODO` or a note about encoding and that Python 3 only uses unicode for strings?

articles/114_generated_interfaces_python.md
dirk-thomas: I think the reference to `object` is way to technical for this article. It is at best an implementation detail.

articles/114_generated_interfaces_python.md
dirk-thomas, line 63: The TODO is already in the ROSIDL article. While it affects all languages I am not sure we need to duplicate the todos.

articles/114_generated_interfaces_python.md
esteve: Yeah, I agree. Do you think we should mention that these types are Python 3 specific? Or is that already implied?

dirk-thomas: That is already mentioned in another design article if I recall correctly.

articles/114_generated_interfaces_python.md
wjwwood, line 63: Until the TODO in the rosidl article is resolved we either have to use unicode or not, so I think there should be a TODO here saying what we're going to do for now and mention that it may change or be clarified once that TODO is resolved. So I think a TODO here is a valuable, but temporary, clarification which isn't redundant.

wjwwood, line 63: Also, I think [`bytearray`](https://docs.python.org/3.1/library/functions.html#bytearray) might be a better match for string in rosidl. This forces the user to create a bytearray by providing their string and the expected encoding. The result can be stored in a `std::string` and `char *` when deserialized on the other side. I think mapping a Python3 string into a rosidl string (and therefore a IDL string) and hoping it doesn't have unicode in it is a bad idea. Put another way I think the problem should come up in Python code on assignment rather than on publish.

wjwwood, line 63: Also, since we're using Python 3.5+ (something that isn't mentioned in the article yet and I think should since it affects the mapping between types) we can use VarType and properties to help enforce this: https://docs.python.org/3/library/typing.html#typing.TypeVar

For example if we add `wstring` to rosidl then we can have that map to `TypeVar('wstring', bytearray, str)`, since `wstring` is a superset of `string`.

This might not end up working, but these are just some thoughts I had.

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,110 @@
...
+| bounded string          | builtins.str  |
+
+### Properties
+
+The class has same-named properties for every field of the message.
+The setter of a property will ensure that ROS type constraints beyond the Python type are enforced.
+
+### "Constants"
+
+A constant is defined as a class variable with uppercase name.
```
wjwwood, line 81: Are they read-only?

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Constructors
+
+The *__init__* function initializes all members with their default value.
+
+The class constructor supports only keyword arguments for all members.
+The short reason for this is that if code would rely on positional arguments to construct data objects changing a message definition would break existing code in various ways.
```
wjwwood: Instead of "various ways" I'd say "subtle ways" or "easily overlooked ways".

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,110 @@
...
+The class contains only two `types`:
+
+- `Request` which is the type of the request part of the service
+- `Response` which is the type of the request part of the service
+
+The generated code is split across multiple files the same way as message are.
+
+### Request and response messages
+
+For the request as well as the response part messages are being generated which have the same name as the service with a suffix being either `_Request` or `_Response`.
```
wjwwood: This sentence was awkward for me to read, how about: "The request and response components for a Service are messages whose names are made up of the Service's name and either a `_Request` or `_Response` suffix respectively."

articles/114_generated_interfaces_python.md
```diff
@@ -0,0 +1,110 @@
...
+
+Original Author: {{ page.author }}
+
+## Scope
+
+This article specifies the generated Python code for ROS interface types defined in the [interface definition article](interface_definition.html).
+
+## Namespacing
+
+All code of a ROS package should be defined in a namespace named after the package.
```
wjwwood: Should we use the term `namespace` here or the Pythonic term `package`? Perhaps we should continue to use `namespace`, but clarify that what we mean is within in a Python package there is a `msg` and/or `srv` sub-package in which there is a Python module for each message or service which contains a Python class.

articles/112_generated_interfaces_cpp.md
```diff
@@ -47,9 +47,43 @@ For each message two files are being generated:
...
+| uint64   | uint64_t    |
+| string   | std::string |
+
+#### Mapping of arrays and bounded strings
+
+| ROS type                | C++ type         |
+| ----------------------- | ---------------- |
+| static array            | std::array<T, N> |
+| unbounded dynamic array | std::vector<T>   |
+| bounded dynamic array   | std::vector<T>   |
```
ernestmc: Currently bounded arrays are mapped as std::array<T, N>

articles/114_generated_interfaces_python.md
dirk-thomas, line 81: Yes, good point: 5f23395054c82b14c433a42ea59e84afefac56f6

articles/114_generated_interfaces_python.md
dirk-thomas: Done: 8af72032a1fbb5db90e79f3010137ed9a6fa686

articles/114_generated_interfaces_python.md
dirk-thomas: I rephrased it: 082b3bdb8331a23319364425fda1e8f84bfc84f6

dirk-thomas: Wrong file: a7b3fe8d4c4248fbc461cd1c0adcb769f337c7cb

articles/112_generated_interfaces_cpp.md
dirk-thomas: Indeed, but I think we have to revisit this. An array always has N elements but we need to be able to store explicitly only M (<N) elements (and are able to get the count M).

I added a TODO for that: 84131cfa908ab465b7ae541612700183014f3843

articles/114_generated_interfaces_python.md
dirk-thomas: I described the underscore / exporting in 7cd87daf60f5abc72d1a0fe68331bb4252577ebd

articles/114_generated_interfaces_python.md
dirk-thomas, line 63: I agree that any encoding check should happen in pure Python code. Simply because the goal is that the Python data structure are usable (and consistent) when being used independently.

We are not using Python 3.5 right now since the current version on Trusty is 3.4. But we should keep this in mind when updating the dependencies with Xenial.

articles/114_generated_interfaces_python.md
dirk-thomas: Rephrased in 696173c9e6eb5f9f896d3138394fc3e29117c828

dirk-thomas:
I have addressed all the feedback. I will rewrite this branch since the number of commits affecting two different articles becomes a bit messy. If you already have a fork of this make sure to pull the latest version.

wjwwood:
I still think a TODO about unicode (and what we're doing in the mean time) is a good idea.

Otherwise, lgtm.

dirk-thomas:
I added a TODO for this in the language agnostic article: d67f9dfd952f33e962815061b3ebb69f4cca8c1e

esteve:
+1

esteve:
Though I agree with @wjwwood that a TODO here about the Python-specific issue of `str` vs `unicode` types is valuable.

esteve:
@dirk-thomas Whoa, that was fast :-)

PR #78, opened 2016-01-28
fix markdown for code snippet

dirk-thomas:


jackieokay:
yep, this is a problem I've had with Markdown before that I didn't notice when I committed this. thanks

+1

PR #79, opened 2016-01-29
enable pygment highlighting

### LINKED ISSUE #77: rendering of C code samples
I'm not sure if this is just my browser, but the triple-backtick on [this page](http://design.ros2.org/articles/realtime_background.html) (e.g., the "Memory Management" section) seems to not quite render the following C code correctly. Maybe it needs a triple-tilde instead? I don't know. (the content is awesome, btw... this is just about its rendering)

codebot:
This PR enables pygments, adds a Pygmentize-generated stylesheet, and changes a few code blocks to use native Jekyll native syntax-highlighting. I don't know if this is the best way to do it, but this seems to work and fixes #77 (at least running locally on my box).

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,6 +8,7 @@ oauth_client_id: efa6dd8eae9106381720
 
 markdown: kramdown
 highlighter: pygments
+pygments: true
 
 relative_permalinks: false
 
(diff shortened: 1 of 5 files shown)


jackieokay:
There was a problem with CI due to our Markdown linter. I can look into it.

https://travis-ci.org/ros2/design/builds/105571635

jackieokay:
Sorry, I misreported why the linter failed, it's actually pretty tricky--the `{% highlight cpp %}` sections are linted as markdown now when they previously were being ignored.

tfoote:
Apparently we should be looking at rouge instead of pygments: https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

PR #80, opened 2016-02-03
code block without language specified

codebot:
As we discussed today, hopefully this will just render as non-highlighted monospaced text in both the github viewer and through the Jekyll pipeline.


dirk-thomas:
Since generic code blocks do work I merged this. See http://design.ros2.org/articles/realtime_background.html for the result.

PR #81, opened 2016-02-03
switch over to use rouge and GFM

tfoote:
Switch to rouge from pygments.

Also switch to upstream provided github-pages docker clone: The pages label is supposed to be as close to github's rendering as possible. https://github.com/jekyll/docker

This correctly labels the code with the language. However the coloring is still not rendered.

PR #82, opened 2016-02-04
remove usage of baseurl

dirk-thomas:
With #81 setting the `baseurl` to be empty there is no use in keeping these.


tfoote:
+1 tested with #81. The existing Dockerfile on trunk no longer works due to the gem version of jekyll needing a newer version of ruby than is available in trusty

PR #83, opened 2016-02-09
Mark sudo as false for travis.

tfoote:
This will get us onto the new container based infrastructure.
https://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 rvm:
 - 2.1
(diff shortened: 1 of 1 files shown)


tfoote:
The build itself is slightly slower, but there was no significant wait for the build to star.

wjwwood:
Lgtm

PR #88, opened 2016-04-12
Move migration guide to wiki

gerkey:
Per mailing list discussion, we're moving the migration guide to the wiki for better fluidity.


dirk-thomas:
Will there be any review step for the wiki page? Will someone watch the RSS feed or will a webhook be setup to notify about changes?

gerkey:
The goal of the migration is to facilitate and encourage contribution. I don't see a need to review edits to a document like this unless we encounter vandalism or bad actors. Let's begin with the assumption that everyone will do the right thing and then react to any divergences if they happen.

We will have to decide how to manage edit permissions on the wiki. As far as I can tell, we can either open the wiki to all github users or we can add people one-by-one as collaborators (https://help.github.com/articles/changing-access-permissions-for-wikis/); presumably we can add somebody as a collaborator on the ros2 repo to have wiki edit access but not grant them push access?

In any case, I'm inclined to start with the all-github-users approach and then restrict it if we encounter problems.

dirk-thomas:
I wasn't suggesting any restrictions to edit the wiki but notifying some people about modifications which is the same we do for wiki.ros.org. Imo it has proven invaluable to take a brief look to changes (after they have been done).

gerkey:
Ah, sorry; I thought that you were proposing to retain the pre-edit review process that we have for design articles.

Notification of wiki edits sounds good to me. Looks like (http://stackoverflow.com/questions/8407917/how-can-you-track-or-be-notified-of-changes-to-github-wikis) you can:
1. Install a service hook that'll be triggered on wiki edit.
2. Poll the github API yourself.
3. Subscribe to the RSS feed: https://github.com/ros2/design/wiki.atom

The RSS feed seems the easiest. Will that work for you?

If not, then I'm inclined to just move this page to wiki.ros.org, where we already have the notification capability that we want; no sense spending time on building more infrastructure around github's wiki.

wjwwood:
> If not, then I'm inclined to just move this page to wiki.ros.org, where we already have the notification capability that we want; no sense spending time on building more infrastructure around github's wiki.

It might be confusing to put it on the ROS 1 wiki, but I suppose that's the best option for now if you think the process is currently too arduous.

jack-oquin:
Let's try the RSS feed for a while, as Brian suggested, and only move it to the ROS 1 wiki if that proves insufficient.

gerkey:
RSS feed, it is. Thanks for the feedback!

tfoote:
Needs fixes for what look to be stricter linter rules.

dirk-thomas:
I created #89 to address the failing tests and retriggered Travis.

wjwwood:
I added the atom feed to my feed reader.

PR #89, opened 2016-04-26
fix linter warnings

dirk-thomas:


jackieokay:
Is there a different way to add emphasis to these lines (that's what the *\* does, correct) without a linter warning?

articles/055_ros_parameter_design.md
```diff
@@ -45,26 +45,26 @@ It is useful to consider an ideal system to understand how it would work and how
 An ideal system would support for the combined use cases of ROS 1.0's built-in parameters as well as ROS 1.0's dynamic parameters system.
 Based on that criteria an ideal system would be able to:
 
-- **Set parameter values**
```
jackieokay, line 48: Looking at the document, it's probably ok to remove the emphasis here and on all the bullet points because they're already offset from the text below them.

articles/030_ros_with_zeromq.md
```diff
@@ -18,7 +18,7 @@ author: '[William Woodall](https://github.com/wjwwood)'
 {{ page.abstract }}
 </div>
 
-> **This document pre-dates the decision to build ROS 2 on top of DDS.**
```
jackieokay, line 21: This warning and the one in serialization.md seem important enough to keep the emphasis/boldface formatting. Although, they do come at the beginning of the document, so I guess they're already emphasized.

dirk-thomas:
I am not aware of a different way. Any suggestions are welcome.

jackieokay:
The warning is "Don’t use emphasis to introduce a section, use a heading"

So maybe you could move the lines with emphasis in `030_ros_with_zeromq.md` and `serialization.md` to below "Original Author" and keep the emphasis, since it will no longer be the beginning of the section.

dirk-thomas:
In the serialization.md file the line in questions is already below the author (https://github.com/ros2/design/pull/89/files#diff-dbe457afabc0bf6088ba95dcad6aef96R20) which was flagged by the linter. So I am not sure what you are suggesting but please feel free to edit the PR in place.

jackieokay:
hmm. Maybe the warning is actually about emphasis at the beginning of a paragraph.

jackieokay:
Doesn't seem like it's worth the workarounds I'm thinking of.

+1

articles/030_ros_with_zeromq.md
tkruse, line 21: The linter warning could be read as: "avoid standalone fully emphasized lines, because they may become rendered too look like section headings. ". You could raise attention like this instead: "_Warning:_ ..."

PR #90, opened 2016-05-18
adding article about topic and service names and how they relate to DDS

wjwwood:
This is an initial draft of an article about the ROS 2 topic and service names.

Things I need to do still:
- [x] talk about capital letter alternative
- [x] summarize differences from ROS 1

Some points of interest or RFC:
- Stuff about URL's
- ~~Changes in the way the `~` is interpreted~~
- ~~Is a single `_` a valid namespace? Like `foo/_/bar`?~~
  - It is not, because `/foo/_/_bar` is ambiguous. It is a special case of "tokens must not end with `_`".

Some other points that require deeper discussion:
- Should services be content oriented like topics or service oriented like websites?
  - Should it be `rosservice://node.name/service/name/space` or just `rosservice:///service/name/space`?

### DIFF
--- a/articles/115_topic_and_service_name_mapping.md
+++ b/articles/115_topic_and_service_name_mapping.md
@@ -246,13 +246,12 @@ Here are some examples of translations between ROS topic names and the correspon
 ### ROS Topic and Service Name Length Limit
 
 Because DDS topic names must be limited to 255 characters, the length of a ROS topic or service name is also limited in length.
-Calculating this length is governed by an unfortunately non-trivial algorithm.
-In the case of a topic:
+In the case of a topic, the length is governed by the following algorithm:
 
-`C + (N * 2) <= 255 - P`
+`C + N + P <= 255`
 
-Where `P` is `8`, the maximum possible prefix length, `C` is the number of non forward slash characters (`[a-z,A-Z,0-9,_]`) in the topic name, and `N` is the number of namespace separators, i.e. forward slashes (`/`), in the name.
-Note that this algorithm must be applied on a fully qualified name, i.e. after expanding all substitutions and the private namespace substitution character (`~`).
+Where `P` is `8`, the maximum possible length of the ROS specific prefix, `C` is the number of characters in the topic name, and `N` is the number of name tokens in the topic name.
+Note that this algorithm must be applied on a fully qualified name, i.e. after expanding all substitutions and the private namespace substitution character (`~`), after removing any URL related syntax (e.g. without the `rostopic://` prefix).
 
 Services are governed by the same algorithm, but in some implementations may require additional characters to be subtracted from the limit depending on how the request and response topics are created by the middleware.
 In the specific case of RTI Connext's Request-Reply implementation, they append the `Request` and `Reply` strings to the topic names.
--- a/articles/115_topic_and_service_name_mapping.md
+++ b/articles/115_topic_and_service_name_mapping.md
@@ -58,15 +58,31 @@ In this section an outline of the proposed constrains for ROS 2 topic and servic
 
 For convenience here is a summary of all rules for topic and service names in ROS 2:
 
-- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), tildes (`~`), or forward slashes (`/`)
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), balanced curly braces (`{}`), or forward slashes (`/`)
+- may start with a tilde (`~`)
 - must not start with a numeric character (`[0-9]`)
 - must not contain any number of repeated underscores (`_`)
 - must not end with an underscore (`_`)
 - must not have an underscore (`_`) followed by a forward slash (`/`), i.e. `_/`
 - must not end with a forward slash (`/`)
 - must not contain any number of repeated forward slashes (`/`)
-- must not have tilde (`~`) adjacent to anything other than forward slashes (`/`)
-- may start with or end with a tilde (`~`)
+- must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`
+- must have balanced curly braces (`{}`), i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`
+
+### Fully Qualified Names
+
+The topic and service name rules allow for some convenience syntax, which in some cases requires additional context to expand to the fully qualified name and then to the DDS equivalent name.
+For example, names may be relative (e.g. `foo` versus the absolute `/foo`), they may contain the private namespace substitution character, or arbitrary substitutions which are within the bracket syntax.
+With context, each of these features can be expanded to some simple string to form the fully qualified name.
+Fully qualified names have these additional restrictions:
+
+- must start with a forward slash (`/`), i.e. they must be absolute
+- must not contain tilde (`~`) or curly braces (`{}`)
+
+Note that expanded substitutions must result in a valid name.
+For example, a name `foo_{bar}` is not valid if `bar` is expanded as `_baz`, because that would result in a fully qualified name of `foo__baz` which contains repeated underscores (`_`).
+
+### Uniform Resource Locators (URL's)
 
 Additionally, topic and service names can be represented in the [Uniform Resource Locator (URL)](https://en.wikipedia.org/wiki/Uniform_Resource_Locator) format to further disambiguate the resource name.
 A topic name may be preceded by a `rostopic://` scheme prefix, and a service name may be preceded by a `rosservice://` scheme prefix.
@@ -78,16 +94,20 @@ A relative name `foo/bar` could would be represented (as a topic) with `rostopic
 
 For example, these are valid names:
 
-| `foo`       | `abc123`            | `_foo`               | `Foo` | `BAR`       |
-| `~`         | `foo/bar`           | `~/foo`              | `/~`  | `foo/~/bar` |
-| `foo/_/bar` | `rosservice:///foo` | `rostopic://foo/bar` |       |             |
+| `foo`       | `abc123`            | `_foo`               | `Foo`        | `BAR`            |
+| `~`         | `foo/bar`           | `~/foo`              | `{foo}_bar`  | `foo/{ping}/bar` |
+| `foo/_/bar` | `rosservice:///foo` | `rostopic://foo/bar` |              |                  |
 
 But these are not valid names:
 
-| `123abc`   | `123`  | `__foo`  | `foo__bar` | `foo bar`  |
-| `foo__`    | ` `    | `foo_`   | `foo//bar` | `foo/`     |
-| `foo_/bar` | `~foo` | `foo~`   | `foo~/bar` | `foo/~bar` |
-| `/_/bar`   | `_`    | `_/_bar` |            |            |
+| `123abc`   | `123`  | `__foo`  | `foo__bar` | `foo bar`   |
+| `foo__`    | ` `    | `foo_`   | `foo//bar` | `foo/`      |
+| `foo_/bar` | `~foo` | `foo~`   | `foo~/bar` | `foo/~bar`  |
+| `/_/bar`   | `_`    | `_/_bar` | `/~`       | `foo/~/bar` |
+
+These are some valid fully qualified names:
+
+| `/foo`     | `/bar/baz` | `rostopic:///ping` | `/_private/thing` | `/public_namespace/_private/thing` |
 
 ### Namespaces
 
@@ -99,10 +119,10 @@ Topic and service names:
 
 - must not end with a forward slash (`/`)
 
-The last token is the topic or service name, and any preceding tokens comprise the namespace of the topic or service.
+The last token is the topic or service basename, and any preceding tokens comprise the namespace of the topic or service.
 
-For example, the topic name `/foo/bar/baz` comprises of a topic or service named `baz` in the `/foo/bar` namespace.
-In another example, the name `/foo` splits into one token, such that it comprises of a topic or service named `foo` in the `/` namespace (the root namespace).
+For example, the topic name `/foo/bar/baz` comprises of a topic or service with the basename `baz` in the `/foo/bar` namespace.
+In another example, the name `/foo` splits into one token, such that it comprises of a topic or service with the basename `foo` in the `/` namespace (the root namespace).
 
 Topic and service names:
 
@@ -126,7 +146,7 @@ Topic and service name tokens:
 
   - rational: it removes the chance for accidental `//` from concatenation
 
-- may use alphanumeric characters (`[0-9|a-z|A-Z]`), or an underscore (`_`)
+- may use alphanumeric characters (`[0-9|a-z|A-Z]`), an underscore (`_`), or balanced curly braces (`{}`)
 
 - must not start with numeric characters (`[0-9]`)
 
@@ -144,14 +164,21 @@ Topic and service name tokens:
 
 - may be a single tilde character (`~`)
 
+### Substitutions
+
+The bracket syntax (`{substitution_name}`) maybe used in non-fully qualified names to substitute useful contextual information into the name.
+The set of substitution keys (names) are not set in this document, but some reasonable examples might be: `{node}` expands to the current node's name or `{ns}` expands to the current node's namespace.
+
+Substitutions are expanded before the private namespace substitution character is expanded.
+Therefore a substitution may contain the private namespace substitution character, i.e. `~`.
+For example, given the name `{private}foo`, that there is a substitution called `{private}` which expands to `~/_`, the current node name is `my_node`, and the current node's namespace is `my_ns`, then the fully qualified name would be `/my_ns/my_node/_foo`.
+
 ### Private Namespace Substitution Character
 
 The special single character token `~` will be replaced with a namespace snippet that is a concatenation of the namespace for the node and the node name.
 For example, for a node `node1` in a namespace `/foo` would result in `~` being replaced with `/foo/node1`.
 For another example, a node `node1` in a namespace `foo/bar` would result in `~` being replaced with `foo/bar/node1`.
-It may be used in any token position.
-Any resulting empty namespaces which result from expanding the `~` will be collapsed.
-For example, when `~` expands to `/foo/node1` and the name `ping/~/pong` is given, it will expand to `ping/foo/node1/pong` and not to `ping//foo/node1/pong`.
+It must be used at the beginning of a non-fully qualified name, if at all.
 Here is a table with some example expansions:
 
 | **Input Name** | Node: `my_node` NS: none | Node: `my_node` NS: `/foo`     |
@@ -160,10 +187,6 @@ Here is a table with some example expansions:
 | `/ping`        | *`/ping`*                | *`/ping`*                      |
 | `~`            | *`/my_node`*             | *`/foo/foo/my_node`*           |
 | `~/ping`       | *`/my_node/ping`*        | *`/foo/foo/my_node/ping`*      |
-| `/~`           | *`/my_node`*             | *`/foo/my_node`*               |
-| `/~/ping`      | *`/my_node/ping`*        | *`/foo/my_node/ping`*          |
-| `ping/~/pong`  | *`/ping/my_node/pong`*   | *`/foo/ping/foo/my_node/pong`* |
-| `/ping/~/pong` | *`/ping/my_node/pong`*   | *`/ping/foo/my_node/pong`*     |
 
 ### Hidden Topic or Service Names
 
@@ -171,8 +194,8 @@ Any topic or service name that contains any tokens (either namespaces or a topic
 
 ## Mapping of ROS 2 Topic and Service Names to DDS Topics
 
-The ROS topic and service name constraints allow more characters than the DDS topic names because ROS allows for the forward slash (`/`) and the tilde (`~`).
-Since ROS 2 topic and service names are expanded to absolute names before being used, there is always a leading forward slash (`/`) and any tildes (`~`) will have been expanded.
+The ROS topic and service name constraints allow more characters than the DDS topic names because ROS allows for the forward slash (`/`) the tilde (`~`), and the balanced curly braces (`{}`).
+Since ROS 2 topic and service names are expanded to fully qualified names (which are absolute) before being used, there is always a leading forward slash (`/`) and any balanced bracket (`{}`) substitutions and tildes (`~`) will have been expanded.
 Additionally any URL related syntax is removed, e.g. the `rostopic://` prefix.
 Therefore only forward slashes have to be substituted when converting to DDS topic names.
 
@@ -215,6 +238,43 @@ Here are some examples of translations between ROS topic names and DDS topic nam
 ## Concerns/Alternatives
 
 This section lists concerns about the proposed design and alternatives that were considered.
+
+### Alternative Name Rules and Concerns About Name Rules
+
+There were some suggested but then rejected alternatives to the rules for topic and service names.
+
+#### More Versatile Private Namespace Substitution Character
+
+Currently the `~` private namespace substitution character may only be a the the beginning of the name, but it was also suggested that it could be placed anywhere within the name and could be substituted in place.
+
+This was rejected because it was complicated to explain and did not behave the same as the `~` when used in filesystem paths on Unix machines.
+Also, it was difficult to justify its existence because all suggested use cases were quite contrived.
+It also behaved differently from how it worked in ROS 1, which was yet another negative for this alternative.
+
+#### Alternative Substitution Syntax
+
+There were some alternative syntaxes proposed for substitutions in the names before the plain balanced curly braces syntax (`{}`) was selected:
+
+- `%{sub}`
+- `${sub}`
+- `$sub`
+
+The most serious alternatives considered were the "bash-like" syntax of `${sub}` and `$sub`.
+The `$sub` style syntax has the downside of being difficult to process and impossible express some kinds of concatenation.
+The `${sub}` was a strong candidate, but ultimately was rejected because it would collide with use in shell scripts.
+For example, you can imagine a shell script that runs a node and remaps a topic name would contain these substitutions, but they would need to be escaped to prevent bash itself from trying to expand them.
+The `{}` syntax avoids this problem but is also easy to parse.
+
+The `{}` syntax will collide with Python String substitution, but since that is an explicit action (unlike shell substitution which will implicitly always occur) it's less of an issue.
+
+#### Concerns About Substitutions
+
+This document does not prescribe what substitutions should be supported by implementations.
+This was done to avoid blocking progress on this document on the need to agree on the set of required substitutions.
+However, this is just delaying the issue.
+For the substitutions to be useful, then all implementations that process topic and service names need to support them.
+
+This compromise was made so that when more work is done on substitutions, it would hopefully not require changes to the name syntax, but instead revolve around what substitutions to support and whether or not they are optional to support.
 
 ### Alternative ROS to DDS Mapping
 
--- a/articles/115_topic_and_service_name_mapping.md
+++ b/articles/115_topic_and_service_name_mapping.md
@@ -210,9 +210,23 @@ Here are some examples of translations between ROS topic names and DDS topic nam
 
 This section lists concerns about the proposed design and alternatives that were considered.
 
-### Alternative ROS Specific Name Prefix Designs
+### Alternative ROS to DDS Mapping
 
 There was some discussion of alternatives and concerns with respect to the ROS -> DDS translation.
+
+#### Capital Letter Substitution Alternative
+
+Since the forward slash (`/`) is replaced with double underscores (`__`) when translating to DDS from ROS topic names, two characters out of the 256 character limit are lost with each additional namespace.
+One proposed alternative was to add a constraint that ROS topic names could not use capital letters, and then capital letters could be used as the stand in for the forward slashes (`/`).
+
+Trade-offs:
+
+- Uses one less character per namespace and makes it easier to calculate the maximum length of a ROS topic or service name.
+- Prevents users from using capital letters in their names, which is constraining and might be a problem for backwards compatibility with ROS 1 topics and services.
+
+Rational:
+
+Preventing users from using capital letters was too constraining for the added benefit.
 
 #### Prefix with Double Underscores
 
--- a/articles/115_topic_and_service_name_mapping.md
+++ b/articles/115_topic_and_service_name_mapping.md
@@ -21,11 +21,11 @@ Original Author: {{ page.author }}
 
 ## Context
 
-Before proposing constraints for ROS 2 topic and service names and a mapping to underlying DDS topics, this article will first summarize the existing guidelines and restrictions for ROS 1 and DDS.
+Before proposing constraints for ROS 2 topic and service names and a mapping to underlying DDS topics, this article will first summarize the existing guidelines and restrictions for both ROS 1 and DDS.
 
 ### ROS 1 Names
 
-In ROS 1, topic and service name guidelines are covered all under the umbrella of "ROS Names" and have these restrictions:
+In ROS 1, topic and service name guidelines are all covered under the umbrella of "ROS Names" and have these restrictions:
 
 *From [http://wiki.ros.org/Names](http://wiki.ros.org/Names):*
 
@@ -58,8 +58,9 @@ In this section an outline of the proposed constrains for ROS 2 topic and servic
 
 For convenience here is a summary of all rules for topic and service names in ROS 2:
 
-- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), balanced curly braces (`{}`), or forward slashes (`/`)
-- may start with a tilde (`~`)
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), or forward slashes (`/`)
+- may use balanced curly braces (`{}`) for substitutions
+- may start with a tilde (`~`), the private namespace substitution character
 - must not start with a numeric character (`[0-9]`)
 - must not contain any number of repeated underscores (`_`)
 - must not end with an underscore (`_`)
@@ -67,12 +68,12 @@ For convenience here is a summary of all rules for topic and service names in RO
 - must not end with a forward slash (`/`)
 - must not contain any number of repeated forward slashes (`/`)
 - must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`
-- must have balanced curly braces (`{}`), i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`
+- must have balanced curly braces (`{}`) when used, i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`
 
 ### Fully Qualified Names
 
 The topic and service name rules allow for some convenience syntax, which in some cases requires additional context to expand to the fully qualified name and then to the DDS equivalent name.
-For example, names may be relative (e.g. `foo` versus the absolute `/foo`), they may contain the private namespace substitution character, or arbitrary substitutions which are within the bracket syntax.
+For example, as outlined in further detail in the sections that follow, names may be relative (e.g. `foo` versus the absolute `/foo`), they may contain the private namespace substitution character (`~`), or arbitrary substitutions which are within the curly braces (`{}`) syntax.
 With context, each of these features can be expanded to some simple string to form the fully qualified name.
 Fully qualified names have these additional restrictions:
 
@@ -82,12 +83,12 @@ Fully qualified names have these additional restrictions:
 Note that expanded substitutions must result in a valid name.
 For example, a name `foo_{bar}` is not valid if `bar` is expanded as `_baz`, because that would result in a fully qualified name of `foo__baz` which contains repeated underscores (`_`).
 
-### Uniform Resource Locators (URL's)
+### Uniform Resource Locators (URLs)
 
 Additionally, topic and service names can be represented in the [Uniform Resource Locator (URL)](https://en.wikipedia.org/wiki/Uniform_Resource_Locator) format to further disambiguate the resource name.
 A topic name may be preceded by a `rostopic://` scheme prefix, and a service name may be preceded by a `rosservice://` scheme prefix.
 For example, the absolute name `/foo` could also be represented as a topic with `rostopic:///foo` or as a service with `rosservice:///foo`.
-Note the triple forward slash (`/`), which is a similar style as the `file://` scheme.
+Note the triple forward slash (`/`), which is a similar style to the `file://` scheme.
 A relative name `foo/bar` could would be represented (as a topic) with `rostopic://foo/bar`.
 
 ### ROS 2 Name Examples
@@ -119,10 +120,10 @@ Topic and service names:
 
 - must not end with a forward slash (`/`)
 
-The last token is the topic or service basename, and any preceding tokens comprise the namespace of the topic or service.
+The last token is the topic or service basename, and any preceding tokens make up the namespace of the topic or service.
 
-For example, the topic name `/foo/bar/baz` comprises of a topic or service with the basename `baz` in the `/foo/bar` namespace.
-In another example, the name `/foo` splits into one token, such that it comprises of a topic or service with the basename `foo` in the `/` namespace (the root namespace).
+For example, the topic name `/foo/bar/baz` is composed of a topic or service with the basename `baz` in the `/foo/bar` namespace.
+In another example, the name `/foo` splits into one token, such that it is composed of a topic or service with the basename `foo` in the `/` namespace (the root namespace).
 
 Topic and service names:
 
@@ -138,46 +139,37 @@ However, if the topic or service name is `/foo/bar`, then the absolute name will
 
 ### Name Tokens
 
-"Name token" are the strings between the namespace delimiters, e.g. the tokens of the topic or service `/foo/bar/baz` are `foo`, `bar`, and `baz`.
+"Name tokens" are the strings between the namespace delimiters, e.g. the tokens of the topic or service `/foo/bar/baz` are `foo`, `bar`, and `baz`.
 
 Topic and service name tokens:
 
 - must not be empty, e.g. the name `//bar` is not allowed
 
-  - rational: it removes the chance for accidental `//` from concatenation
+  - rationale: it removes the chance for accidental `//` from concatenation and thereore the need to collapse `//` to `/`
 
-- may use alphanumeric characters (`[0-9|a-z|A-Z]`), an underscore (`_`), or balanced curly braces (`{}`)
+- may use alphanumeric characters (`[0-9|a-z|A-Z]`), single underscores (`_`), and/or balanced curly braces (`{}`)
 
 - must not start with numeric characters (`[0-9]`)
 
 - must not end with a single underscore (`_`)
 
-  - rational: if tokens are allowed to end with and start with `_` then `foo_/bar` is indistinguishable from `foo/_bar` if the `/` is replaced with `__` (as proposed in the "Substitution of the Namespace Delimiter" section), i.e. both result in `foo___bar`
+  - rationale: if tokens are allowed to end with and start with `_` then `foo_/bar` is indistinguishable from `foo/_bar` if the `/` is replaced with `__` (as proposed in the "Substitution of the Namespace Delimiter" section), i.e. both result in `foo___bar`
 
 - must not be a single underscore (`_`)
 
-  - rational: this is a special case of "must not end with an underscore"
+  - rationale: this is a special case of "must not end with an underscore"
 
 - must not have two or more underscores (`__`) repeated anywhere
 
-  - rational: this provides the implementation with a safe to use delimiter
+  - rationale: this provides the implementation with a safe to use delimiter
 
 - may be a single tilde character (`~`)
-
-### Substitutions
-
-The bracket syntax (`{substitution_name}`) maybe used in non-fully qualified names to substitute useful contextual information into the name.
-The set of substitution keys (names) are not set in this document, but some reasonable examples might be: `{node}` expands to the current node's name or `{ns}` expands to the current node's namespace.
-
-Substitutions are expanded before the private namespace substitution character is expanded.
-Therefore a substitution may contain the private namespace substitution character, i.e. `~`.
-For example, given the name `{private}foo`, that there is a substitution called `{private}` which expands to `~/_`, the current node name is `my_node`, and the current node's namespace is `my_ns`, then the fully qualified name would be `/my_ns/my_node/_foo`.
 
 ### Private Namespace Substitution Character
 
 The special single character token `~` will be replaced with a namespace snippet that is a concatenation of the namespace for the node and the node name.
-For example, for a node `node1` in a namespace `/foo` would result in `~` being replaced with `/foo/node1`.
-For another example, a node `node1` in a namespace `foo/bar` would result in `~` being replaced with `foo/bar/node1`.
+For example, a node `node1` in a namespace `/foo` would result in `~` being replaced with `/foo/node1`.
+As another example, a node `node1` in a namespace `foo/bar` would result in `~` being replaced with `foo/bar/node1`.
 It must be used at the beginning of a non-fully qualified name, if at all.
 Here is a table with some example expansions:
 
@@ -190,25 +182,36 @@ Here is a table with some example expansions:
 
 Note the private namespace substitution character makes the name absolute, and therefore the namespace is not added a second time.
 
+### Substitutions
+
+The bracket syntax (`{substitution_name}`) may be used in non-fully qualified names to substitute useful contextual information into the name.
+The set of substitution keys (names) are not set in this document, but some reasonable examples might be: `{node}` expands to the current node's name or `{ns}` expands to the current node's namespace.
+
+Substitutions are expanded before the private namespace substitution character is expanded.
+Therefore a substitution may contain the private namespace substitution character, i.e. `~`.
+For example, given the name `{private}foo`, if there is a substitution called `{private}` that expands to `~/_`, the current node name is `my_node`, and the current node's namespace is `my_ns`, then the fully qualified name would be `/my_ns/my_node/_foo`.
+
 ### Hidden Topic or Service Names
 
-Any topic or service name that contains any tokens (either namespaces or a topic or service name) which starts with an underscore (`_`) will be considered hidden and tools may not show them unless explicitly asked.
+Any topic or service name that contains any tokens (either namespaces or a topic or service name) that start with an underscore (`_`) will be considered hidden and tools may not show them unless explicitly asked.
 
 ## Mapping of ROS 2 Topic and Service Names to DDS Topics
 
-The ROS topic and service name constraints allow more characters than the DDS topic names because ROS allows for the forward slash (`/`) the tilde (`~`), and the balanced curly braces (`{}`).
-Since ROS 2 topic and service names are expanded to fully qualified names (which are absolute) before being used, there is always a leading forward slash (`/`) and any balanced bracket (`{}`) substitutions and tildes (`~`) will have been expanded.
-Additionally any URL related syntax is removed, e.g. the `rostopic://` prefix.
+The ROS topic and service name constraints allow more types of characters than the DDS topic names because ROS additionally allows the forward slash (`/`), the tilde (`~`), and the balanced curly braces (`{}`).
+These must be substituted or otherwise removed during the process of mapping to DDS topic names.
+Since ROS 2 topic and service names are expanded to fully qualified names, any balanced bracket (`{}`) substitutions and tildes (`~`) will have been expanded.
+Additionally any URL related syntax, e.g. the `rostopic://` prefix, will be removed once parsed.
 Therefore only forward slashes have to be substituted when converting to DDS topic names.
 
 ### Substitution of the Namespace Delimiter
 
 The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
+Note that as fully qualified ROS 2 topic and service names are absolute, there is always a leading forward slash (`/`).
 
 ### ROS Specific Name Prefix
 
-In order to differentiate ROS topics easily, all DDS topic names created by ROS shall be prefixed with `rX`, where `X` is a single character which indicates what subsystem of ROS to which the topic belongs.
-For example, a plain topic called `/foo` would translate to a DDS topic named `rt__foo`, which is a result of concatenating the prefix `rt` for being a ROS topic, plus `__` for the leading `/`, plus the topic name `foo`.
+In order to differentiate ROS topics easily, all DDS topic names created by ROS shall be prefixed with `rX`, where `X` is a single character that indicates to which subsystem of ROS the topic belongs.
+For example, a plain topic called `/foo` would translate to a DDS topic named `rt__foo`, which is the result of concatenating the prefix `rt` for being a ROS topic, with `__` for the leading `/`, and the topic name `foo`.
 
 For systems where Services are implemented with topics (like with OpenSplice), a different subsystem character can be used: `rq` for the request topic and `rr` for the response topic.
 On systems where the implementation is handled for us by DDS (like with Connext), we use `rs` as the common prefix.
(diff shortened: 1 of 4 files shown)


wjwwood:
I added a section for the capital letter alternative and fixed style issues. I'm going to wait on the section about differences from ROS 1 until I get feedback on the more questionable bits like the `~` expansion differences.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,313 @@
...
+For systems where Services are implemented with topics (like with OpenSplice), a different subsystem character can be used: `rq_` for the request topic and `rp_` for the response topic.
+On systems where the implementation is handled for us by DDS (like with Connext), we use `rS_` as the common prefix.
+
+Here is a non-exhaustive list of prefixes:
+
+| ROS Subsystem        | Prefix |
+|----------------------|--------|
+| ROS Topics           | rt_    |
+| ROS Service Request  | rq_    |
+| ROS Service Response | rs_    |
```
dirk-thomas: The paragraph above refers to this as `rp_` (which also collides with parameters from this table). I would propose `rr_` for the response / result. That leaves `q` for the query, `s` for the native service, and `p` for parameters`.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,313 @@
...
+| ROS Service          | rS_    |
+| ROS Parameter        | rp_    |
+| ROS Action           | ra_    |
+
+### Substitution of the Namespace Delimiter
+
+The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
+The leading forward slash (`/`) is not replaced with double underscores (`__`), and is replaced instead by the ROS specific name prefix.
```
dirk-thomas: Imo keeping the parsing simpler and also replacing the leading `/` with two underscores outweighs the "benefit" of shortening the name by a single character.

dirk-thomas: I just saw it covered later as an alternative: Prefix with Double Underscores

articles/115_topic_and_service_name_mapping.md
wjwwood: Sounds good to me: 2e8628c

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,313 @@
...
+This alternative differs only in that it uses two underscores in the prefix, i.e. `rt__` rather than `rt_`.
+
+Trade-offs:
+
+- More consistent with replacement of other forward slashes (`/`)
+- Uses one more character
+
+Rational:
+
+Slight preference given to the shorter alternative.
```
dirk-thomas: I certainly wouldn't agree with the rational. I think the increased parser complexity in potentially N implementations is pretty severe compared to the benefit of 1 extra character.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,313 @@
...
+
+- No prefix
+
+- The prefix is restructured to be a suffix, i.e. `rX_<topic>` -> `<topic>_rX_`
+
+  - this would be unique to user defined names because they cannot have a trailing underscore (`_`)
+
+Trade-offs:
+
+- more difficult to distinguish ROS created DDS topics from normal or built-in DDS topics when listing them using DDS tools because they are not sorted by a prefix
```
dirk-thomas: I think the opposite should be stated here too: the suffix makes topics, service, parameters, etc. in the same namespace much easier to see since they are close to each other, rather then with a prefix completely separated.

articles/115_topic_and_service_name_mapping.md
wjwwood: Well then we'll have to vote on it, because I disagree. We don't have any slack in the limitation on the name length, so any extra work we can do, which doesn't also restrict the user, to maximize the available characters is worth it in my opinion. This isn't a complex rule and in all likelihood would be a single line in any language where it is implemented.

dirk-thomas: If length is so crucial you might not want to use a three letter prefix in the first place. A single character could encode all the subsystem. For readability two would be reasonable, e.g. just `rX`. There is no need for the underscore to separate.

articles/115_topic_and_service_name_mapping.md
wjwwood: I don't think that's the opposite. This line is talking about distinguishing between DDS topics and ROS created DDS topics.

I think you're thinking about yet another alternative which is `r_<topic>_X_`, which keeps ROS stuff together, but sorts first by namespace and then by type. You can add another Alternative section if you like or push for that.

My opinion is that anything we choose will be preferring one sorting over another, e.g. namespace vs type vs ROS related or not, and no matter which we pick, someone at some point will want to do the other. So I place more importance on the other aspects of the alternatives like number of characters used, ability to prevent collisions by design, and obviousness to unfamiliar observers.

articles/115_topic_and_service_name_mapping.md
wjwwood: I agree, but in that case I think at least one underscore helps because it is less likely to share a common prefix with a non-ROS DDS topic, e.g. the sorting of `ramy_action`, `radical_non_ros_topic`, `raweird_action` versus `ra_my_action`, `ra_weird_topic`, `radical_non_ros_topic`. Also, `startswith('rt')` is more likely to overmatch more than `startswith('rt_')` in my opinion. However, in my opinion an additional `_` doesn't help with readability or reduce the likelihood of prefix congestion considerably (it does help a little). Mostly a double underscore would help with consistency in my opinion.

Admittedly this is all pretty subjective, but I think there is a significant enough distinction to be made between `rX` and `rX_` to justify one underscore and not enough of an advantage for `rX__` over `rX_` to justify a second. But we can vote on it as a group if we can't come to a consensus.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Name Tokens
+
+Name token's are the strings between the namespace delimiters, e.g. the tokens of the topic or service `/foo/bar/baz` are `foo`, `bar`, and `baz`.
+
+Topic and service name tokens:
+
+- must not be empty, e.g. the name `//bar` is not allowed
+
+  - rational: it removes the chance for accidental `//` from concatenation; also if the implementation chooses to use a trailing underscore (`_`) then it prevents `foo_/_bar` and `foo//bar` from being ambiguous
```
mikaelarguedas: given that having a trailing '_' in token names is prohibited this ambiguity shouldn't exist anymore. I'm not against keeping the rule though, but maybe rephrase/remove the rational

wjwwood: Yeah, I was thinking about how Connext does Services, where if you give it `foo` it will create to topics called `foo_Request` and `foo_Response` (or something like that). So if we give it `foo_s_` (as was proposed in an alternative) it would do `foo_s__Request`. But either way this is a weak part of the rational for sure. I'm agnostic to removing it, maybe I could get another +1 for removing it from someone else?

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,313 @@
...
+
+| ROS Name          | Subsystem | DDS Name              |
+|-------------------|-----------|-----------------------|
+| `/foo`            | Topic     | `rt_foo`              |
+| `/foo/bar`        | Topic     | `rt_foo__bar`         |
+| `/_foo/bar`       | Topic     | `rt__foo__bar`        |
+| `/foo/_bar`       | Topic     | `rt_foo___bar`        |
+| `/_foo/_bar`      | Topic     | `rt__foo___bar`       |
+| `/_foo/_bar/_baz` | Topic     | `rt__foo___bar___baz` |
+| `/foo/_/bar`      | Topic     | `rt_foo_____bar`      |
```
mikaelarguedas: related to previous remark. Shouldn't this be an invalid name given that trailing underscores are not allowed?

wjwwood: Well, it's not clear this is a trailing or starting `_`. This is a special case I was wondering about. I'd need to add a specific rule to say a single `_` is not allowed as a namespace or topic name.

wjwwood: Oh I guess `/foo/_/_bar` is a problem (maybe that's what you meant).

I'll update the document to not allow this at all.

wjwwood: 0d9be83

mikaelarguedas: Yeah this case would definitely be ambiguous. But in general, I was thinking that this example goes against the statement line 65 `must not have an underscore (`_`) proceeded by a forward slash (`/`), i.e. `_/``

wjwwood: Yeah, I think that line was enough, but I went ahead and clarified it more with:

```
- must not be a single underscore (`_`)

  - rational: this is a special case of "must not end with an underscore"
```

In the "name tokens" section.

articles/115_topic_and_service_name_mapping.md
tfoote: I think that a separator is important. And since we already have a separator reusing it makes sense. It will greatly simplify the parsing process and visually it will also be consistent. 

As much as I completely agree that the hard limit on the topic length in DDS is a big issue and we should work to minimize the overhead we use. I think that optimizing down to the character at this point is overkill. I would suggest that we reserve for ourselves several characters. Maybe 6 for the prefix. We will plan to use only 2 in our implementations. And assuming we are happy in the future we can decrease our reserved characters to the user. If we really wanted to preserve characters we'd want to use a single separator with an escape character. The requirement that there's not trailing `_` and leading numbers could make most separators only single width. But the introspection and readability gets much worse.

wjwwood: > It will greatly simplify the parsing process and visually it will also be consistent.

I agree it would be more visually consistent, but I take issue with both your and @dirk-thomas's characterization of how much more complicated one `_` rather than two `__` would be... I just don't think it's an issue at all. If you want the to split the header (`rX`) from the name, simply split it at the first `_` and prepend a namespace separator to the name. If you're creating the topic name, simply remove the leading separator, prepend an `_`, and then prepend the header (`rX`).

That being said, if no one agrees with me, then I'll update the document. Please everyone vote up/down on this proposed change.

> I would suggest that we reserve for ourselves several characters. Maybe 6 for the prefix. We will plan to use only 2 in our implementations. And assuming we are happy in the future we can decrease our reserved characters to the user.

I'm not sure I agree because I can't imagine why we'd need more characters (but it's hard to prove a negative). Especially given that we can start using the second character as an op-code if we need more options and then we have around 30 unique values (a-z + 0-9 - the ones we've already reserved). That would be a hit to readability and obviousness, but it's also something I don't think we'll have to do.

Also, when considering reserving more characters, I think that we should assume that we'll never give the characters back, because I think it would be a painful process to do so. So much so that it would never be worth rolling out. It would basically be a wire level breaking change to change the prefix. I'm not arguing that we shouldn't reserve space because of this, but it's like a mail-in rebate to me: I just assume the price does not include the rebate and that I'm not going to get it back, but then if I do it's icing on top.

> If we really wanted to preserve characters we'd want to use a single separator with an escape character.

What does this mean?

> The requirement that there's not trailing _ and leading numbers could make most separators only single width.

Can you give an example? I don't see how we could use those rules to implement a separator (I assume you mean the namespace delimiter).

tfoote: > I'm not sure I agree because I can't imagine why we'd need more characters (but it's hard to prove a negative). Especially given that we can start using the second character as an op-code if we need more options and then we have around 30 unique values (a-z + 0-9 - the ones we've already reserved). That would be a hit to readability and obviousness, but it's also something I don't think we'll have to do.
> 
> Also, when considering reserving more characters, I think that we should assume that we'll never give the characters back, because I think it would be a painful process to do so. So much so that it would never be worth rolling out. It would basically be a wire level breaking change to change the prefix. I'm not arguing that we shouldn't reserve space because of this, but it's like a mail-in rebate to me: I just assume the price does not include the rebate and that I'm not going to get it back, but then if I do it's icing on top.

Indeed we shouldn't assume they'll come back , and we don't need to use them in the prefix. If we just enforce that we leave empty space at the end of the strings it's a forward compatible change. All old topics will still work. The value in this that to users if we allow them x elements in their names, it's a non breaking change to allow them x+3 elements in their names. It's a potentially crippling change if we change it to allow x-3 which is almost impossible to audit if there are dynamically generated topic names. 

I agree that there are no strong use cases right now for going beyond the 32 characters. But there are some things we could consider in the future such as embedded ROS version semantics into the prefix such that you could differentiate versions of ROS running over DDS in parallel etc. Leaving ourselves some extra bits might be worth it. 

> If we really wanted to preserve characters we'd want to use a single separator with an escape character.
> What does this mean?

For example:
If '0' is prefixed by '_' then '0' else '/'
- a_0b -> a0b
- a0_b -> a/_b
- a_00b -> a0/b
- a__0b -> a_0b
- a_____b -> a_____b
- a0000000 -> b invald empty namespaces
- a_0_0_0_0b -> a0000b

Or looking the other way.
- trailing 0: a0/b -> a_00b
- leading underscore: a/_b -> a0_b
- zero after underscore: a_0b -> a__0b 
- zero before underscore: a0_b -> a_0_b

Overhead 1 character per namespace separator + 1 character per instance of 0 in the string. 

This is a lower overhead than 2 characters per namespace separator assuming that there are more namespace separators than zeros. But i'm not advocating for this but saying that if we need more space.

We could pick any other character such as a capital letter which is theoretically invalid in the ROS space and decrease the number of collisions. We've been talking about allowing them since we haven't enforced it before. 

> The requirement that there's not trailing _ and leading numbers could make most separators only single width.
> Can you give an example? I don't see how we could use those rules to implement a separator (I assume you mean the namespace delimiter).

Zero can never start a token and underscore can never end a token thus _0 can never be at the boundary. It limits where it could be but on expanding it out, it's not limited enough to avoid the need for an escape character. I think using the capitals would be a better optimization since they should appear less. 

But again I think we should use the `__` for clarity.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+
+Topic and service names:
+
+- may be split into tokens using a forward slash (`/`) as a delimiter
+- must not end with a forward slash (`/`)
+
+The last token is the topic or service name, and any preceding tokens comprise the namespace of the topic or service.
+
+For example, the topic name `/foo/bar/baz` comprises of a topic or service named `baz` in the `/foo/bar` namespace.
+In another example, the name `/foo` does not use splitting into tokens, so it comprises of a topic or service named `foo` in the `/` namespace (the root namespace).
```
gbiggs: The wording here could be confusing. '/foo' can be split into tokens, we just find that there are no additional tokens (parent names) between the root and the topic name, so it is in the root namespace.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+An absolute name begins with a forward slash (`/`) and does not respect namespaces, i.e. it can be considered "global".
+A relative name does not begin with a forward slash (`/`) and does respect the namespace of the node which created them.
+
+Relative names are appended to the namespace of the node which creates them.
+For example, if the node is put into a namespace `/ping/pong` and the topic or service name is `foo/bar`, then the absolute name will become `/ping/pong/foo/bar`.
+However, if the topic or service name is `/foo/bar`, then the absolute name will stay just `/foo/bar`, ignoring the node's namespace.
+
+### Name Tokens
+
+Name token's are the strings between the namespace delimiters, e.g. the tokens of the topic or service `/foo/bar/baz` are `foo`, `bar`, and `baz`.
```
gbiggs: "Name token's" -> "Name tokens"

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+| `123abc`   | `123`  | `__foo`  | `foo__bar` | `foo bar`  |
+| `foo__`    | ` `    | `foo_`   | `foo//bar` | `foo/`     |
+| `foo_/bar` | `~foo` | `foo~`   | `foo~/bar` | `foo/~bar` |
+| `/_/bar`   | `_`    | `_/_bar` |            |            |
+
+### Namespaces
+
+Topic and service names:
+
+- may be split into tokens using a forward slash (`/`) as a delimiter
```
gbiggs: Add a reference to the "Name Tokens" section for more detail on tokenisation?

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+
+  - rational: it removes the chance for accidental `//` from concatenation; also if the implementation chooses to use a trailing underscore (`_`) then it prevents `foo_/_bar` and `foo//bar` from being ambiguous
+
+- may use alphanumeric characters (`[0-9|a-z|A-Z]`), or an underscore (`_`)
+
+- must not start with numeric characters (`[0-9]`)
+
+- must not end with a single underscore (`_`)
+
+  - rational: if tokens are allowed to end with and start with `_` then `foo_/bar` is indistinguishable from `foo/_bar` if the `/` is replaced with `__`, i.e. both result in `foo___bar`
```
gbiggs: A reference to the section about namespace delimiter substitution is needed to make this rationale complete (i.e. why is the replacing happening at all?).

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+For convenience here is a summary of all rules for topic and service names in ROS 2:
+
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), tildes (`~`), or forward slashes (`/`)
+- must not start with a numeric character (`[0-9]`)
+- must not contain any number of repeated underscores (`_`)
+- must not end with an underscore (`_`)
+- must not have an underscore (`_`) proceeded by a forward slash (`/`), i.e. `_/`
+- must not end with a forward slash (`/`)
+- must not contain any number of repeated forward slashes (`/`)
+- must not have tilde (`~`) adjacent to anything other than forward slashes (`/`)
```
gbiggs: Perhaps add a note that this makes the tilde an entire token all on its own, and nothing else is allowed?

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+### Hidden Topic or Service Names
+
+Any topic or service name that contains any tokens (either namespaces or a topic or service name) which starts with an underscore (`_`) will be considered hidden and tools may not show them unless explicitly asked.
+
+## Mapping of ROS 2 Topic and Service Names to DDS Topics
+
+The ROS topic and service name constraints allow more characters than the DDS topic names because ROS allows for the forward slash (`/`) and the tilde (`~`).
+Since ROS 2 topic and service names are expanded to absolute names before being used, there is always a leading forward slash (`/`) and any tildes (`~`) will have been expanded.
+Additionally any URL related syntax is removed, e.g. the `rostopic://` prefix.
+Therefore the only forward slash has to be substituted when converting to DDS topic names.
```
gbiggs: "Therefore the only forward slash has" -> "Therefore only forward slashes have"

articles/115_topic_and_service_name_mapping.md
gbiggs: I agree with Tully that it may be worth having a few reserved characters even if they are not used now. It is easier to reduce the required prefix length than increase it.

I do not think that having `rt_` instead of `rt__` is too much of a chore. Although the hidden topic `/_foo` becomes `rt__foo`, which is awful, I'm assuming that users will not see it, and that the first step in processing a topic name will be lopping off the `rt_` prefix leaving `_foo` anyway.

If the first step of parsing a DDS topic name to a tokenised ROS topic is to lop off and process the prefix, then having one underscore or two will not complicate the processing implementation.

The only "against" I can think of for a single underscore in the prefix is that there is a higher chance that it will collide with a DDS topic name, but that risk seems fairly low to me without any data on frequently-used DDS topic names.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+
+In this section an outline of the proposed constrains for ROS 2 topic and service names will be enumerated along with rationales where appropriate.
+
+For convenience here is a summary of all rules for topic and service names in ROS 2:
+
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), tildes (`~`), or forward slashes (`/`)
+- must not start with a numeric character (`[0-9]`)
+- must not contain any number of repeated underscores (`_`)
+- must not end with an underscore (`_`)
+- must not have an underscore (`_`) proceeded by a forward slash (`/`), i.e. `_/`
```
jack-oquin: proceeded -> followed

tkruse:
Possibly it could be useful to have a section dedicated to potential mappings between ROS1 and ROS2 names. Or to other 3rd-party middlewares.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,317 @@
...
+Any resulting empty namespaces which result from expanding the `~` will be collapsed.
+For example, when `~` expands to `/foo/node1` and the name `ping/~/pong` is given, it will expand to `ping/foo/node1/pong` and not to `ping//foo/node1/pong`.
+Here is a table with some example expansions:
+
+| **Input Name** | Node: `my_node` NS: none | Node: `my_node` NS: `/foo`     |
+|----------------|--------------------------|--------------------------------|
+| `ping`         | *`/ping`*                | *`/foo/ping`*                  |
+| `/ping`        | *`/ping`*                | *`/ping`*                      |
+| `~`            | *`/my_node`*             | *`/foo/foo/my_node`*           |
+| `~/ping`       | *`/my_node/ping`*        | *`/foo/foo/my_node/ping`*      |
```
jack-oquin: Why is there a repeated `/foo/foo` in the previous two examples?

wjwwood: I assume you mean the last column of the previous two lines?

The formulae is `<name space>` + `<stuff before the ~>` + `<expanded ~>` + `<stuff after the ~>`. In the case of line 158 (example 2), it's `/foo` + `<nothing>` + `/foo/my_node` + `<nothing>` which results in `/foo/foo/my_node`. For line 159 (example 2), it's `/foo` + `<nothing>` + `/foo/my_node` + `/ping` which results in `/foo/foo/my_node/ping`.

Basically the `~` expands to "namespace" + "node name" and is expanded in place. In these cases the namespace is represented twice, once as the normal namespace and once in the expanded `~`. Does that make more sense?

In the following lines this doesn't happen because the input is absolute already and therefore not affected by the namespace implicitly.

gbiggs: I must admit that some of the examples in the table made my head hurt, too. But, when I thought about it, I decided that as painful as they are, there are probably minor use cases (such as auto-generated topic names for infrastructure and tooling) where expanding to "namespace + node name" within the node within the namespace, leading to odd-looking repitition, could be used.

wjwwood: I'm not necessarily in favor of this level of flexibility, but I suggested it because the current restrictions on how to use `~` in ROS 1 seemed arbitrarily constraining. Having the character is nice though because it makes remapping slightly easier.

jack-oquin: I see. That seems to violate the design principle of "least surprise".

The only useful ROS1 semantic I know of for the `~` prefix is at the beginning of the name. I can't think of any valid reason to expand the node's namespace and name in the middle of some string of tokens. 

Unless there is a valid use case, I recommend only expanding `~` at the beginning of a name (as the first token). For compatibility, it should probably _not_ be preceded by a `/`, but I would accept that too, as long as it's optional.

wjwwood: Yeah, I was imagining something like a topic called `/public_cameras/~/image_raw` and then two node instances per robot, where the node names are `left_camera` and `right_camera`, and each robot is in a namespace like `robot1` and `robot2`. You'd get something like this:
- `/public_cameras/robot1/left_camera/image_raw`
- `/public_cameras/robot1/right_camera/image_raw`
- `/public_cameras/robot2/left_camera/image_raw`
- `/public_cameras/robot2/right_camera/image_raw`

Now, I don't know why you'd want to do that, and you could do it in ROS 1, but not in a way that's easily remapped. In this scenario you can do a remap like `/public_cameras/~/image_raw:=/private_cameras/~/image_raw`. However, without this more flexible `~`, you'd have to write Python code like this to achieve it: `pub = Publisher('/public_camera/' + rospy.get_namespace() + '/' + rospy.get_name() + '/image_raw', ...)`. That's fine, except the remapping becomes non-trivial, something like `/public_camera/robot1/left_camera/image_raw:=/private_camera/robot1/left_camera/image_raw` where the `robot1/left_camera` part would be subject to change based on the configuration.

---

So that's what I was imagining, but I'm also ok with keeping it simple. What do others think?

articles/115_topic_and_service_name_mapping.md
wjwwood: At this point in the document the term token hasn't been introduced. I intended this list to be declarative. The purpose of these two rules is explained more in the following sections. I'd prefer not to complicate this list with context specific terminology. Maybe there's a way to reword this to be clearer without referencing other concepts like a "token"?

articles/115_topic_and_service_name_mapping.md
wjwwood: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: I removed it anyways: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: d78250c93ae3b4b60f482c86009d924a0b00724c

articles/115_topic_and_service_name_mapping.md
wjwwood: I switched the prefix to be `rX` and removed the stipulation that the first namespace delimiter is dropped, so now `/foo` becomes `rt__foo` (`rt` + `__` + `foo`): 70fb7043577673c762822d3a378d5c58e60a5c97

wjwwood:
Thanks for the feedback everyone. I addressed a lot of the inline comments and changed the prefixing to be `/foo` -> `rt__foo` rather than `/foo` -> `rt_foo`.

---

> Possibly it could be useful to have a section dedicated to potential mappings between ROS1 and ROS2 names. Or to other 3rd-party middlewares.

I plan to make a section that summarizes the differences between ROS 1 and ROS 2 after resolving the discussion around the `~`, since it will likely change a lot depending on that. I'm not sure what other middleware I'd additionally compare it to. Do you have a suggestion? ZeroMQ? YARP?

articles/115_topic_and_service_name_mapping.md
wjwwood: Instead of reserving characters, can we just say anything between `rX` and the first `__` is reserved? That way parsers will be written to handle either no new characters or arbitrarily more? In general I'm still not convinced enough that we'll need them to justify preallocating characters to the cause.

wjwwood:
> Do you have a suggestion? ZeroMQ? YARP?

Well on second thought ZeroMQ isn't a great thing to compare to, since it's not an "anonymous" or "by name" pub/sub. You have to specify the transport, something like `tcp://<ip>:port`.

I can't find YARP's guidelines on this (maybe it's not explicitly documented), but based on their tutorials I'd guess it's really similar to ROS's. Does anyone with more experience know what YARP's port name rules are?

I also looked for LCM's, but other than seeing them drop the topic name into a regex with the equivalent of `sscanf('^%s$', channel_name)` (https://github.com/lcm-proj/lcm/blob/master/lcm/lcm.c#L323), I haven't seen any other validation and I can't find anything in their documentation. So if anyone has a lead on that one, I'll happily add it too.

I looked into the port naming conventions of the orocos family of projects (orocos, rock, and orca), but also came up short. If anyone can link to or recite their name rules I'll add them too.

At this point I'm thinking there's something wrong with me, maybe it's the wine 🍷. If anyone wants to contribute name guidelines from other middlewares or robot kits I'll gladly add them.

articles/115_topic_and_service_name_mapping.md
gbiggs: The problem of saying "anything between `rX` and the first `__`" is that you run the risk of extending the prefix and suddenly invalidating some borderline topic names. If the reserved length is known in advance, then everyone knows where the limit lies.

articles/115_topic_and_service_name_mapping.md
gbiggs: "May contain tilde (`~`) characters only if each individual tilde is surrounded by forward slashes"?

articles/115_topic_and_service_name_mapping.md
jack-oquin: I agree with Will that we should not lightly decide to reduce the expressive power of the names. The example given does not seem very compelling (to me), but I can imagine similar patterns that make sense. 

However, I dislike the potential for confusion, because these rules are not intuitive and work very differently from common ROS 1 usages, especially `~/foo`. I suppose all those instances in existing packages would have to be converted to `/~/foo` to duplicate the expected ROS 1 semantics.

The main problem with the part of line 158 where `~` resolves to `/foo/foo/my_node` is that the namespace gets substituted twice: once before the tilde and again as part of the tilde expansion, not usually what one wants.  That's going to trip up lots of people.

I don't know of any rigorous description of the ROS 1 tilde semantics, but I believe it always resolved to an absolute name containing the namespace and node name. No leading `/` was required.

wjwwood: Ok, that's a good argument for keeping it simple. What do others think?

codebot: I just re-read the [bash documentation for the tilde character](http://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html#Tilde-Expansion). It looks like bash doesn't do anything with tildes that aren't the first character, and tilde-expansions always result in global path (i.e., with a leading slash). There is also some subtlety about whether or not the tilde is followed by a slash. But for dummies like me :confused:, once I start seeing strings with slashes and tildes, I get in "UNIX filesystem" mental mode and anything that is "different" would make me fall down :crying_cat_face:. Maybe we could have another magic character that would expand differently, or just continue following bash conventions and have more magic substrings, like ${HOME} or ${NODE} or ${MORESTUFF} ? :chicken:

tfoote: I think that bash is the canonical comparison for `~` usage. I didn't know about all those compound options. Limiting it to the front of the string seems reasonable. The availability of some shorthand substitution elements might make sense too, but I'd suggest we defer that for later.

wjwwood: I agree that being similar to the bash shell syntax is a good idea.

One thing I'll point out though is that ROS 1 allows `~foo` and none of the shells on my computer allow that. That's not an argument for or against the more complex substitutions, but you can considered it a separate proposal to not allow `~foo` in ROS 2 and require any instances of `~foo` to be changed to `~/foo`.

I also think the substitutions are potentially a nice thing to have. Two concerns about that for me is: complexity and overlap with `~`. As for complexity, I worry about it making the rules for topic names too complicated and slightly about implementing it with all the possible corner cases in C or C++. As for overlap, to me it would then become questionable as to why we have `~` if you could express the same thing as `/{ns}/{node}`. Obviously it's more convenient, but if we allow `~` why not others? I don't think it makes `~` unjustifiable, but I think it does pose the question of why we have two ways to do something.

As for when to decide on the more complex substitutions, I think we should not defer the decision. We should either decide to do it now because it's useful or not, but deferring will just increase the cost of implementing it. I think we have all the experience (from using ROS 1) and information we need to make the decision and I don't think anything will change in the next X amount of time that will help us make a better decision. I'm personally leaning towards no. If anyone else has an opinion on that issue, let's discuss it now.

codebot: We can always revisit this decision (or any decision) in the future. I agree we need to move forward. I would suggest we just follow the bash usage pattern. It's interesting to note that in bash there are often many different ways to do the same thing, and that's not necessarily a bad thing. Both `~/foo` and `${HOME}/foo` evaluate to the same result, and have for [many decades](http://unix.stackexchange.com/questions/34196/why-was-chosen-to-represent-the-home-directory). So I think providing multiple overlapping expansion options isn't necessarily a bad thing. For now, I think it's safe to just say that we'll provide the same tilde behavior as commonly used in ROS1 (namely, the leading tilde), and since curly braces aren't currently allowed in ROS names anyway, we can add that feature later when/if the utility of doing so for various applications becomes more clear.

I suppose we could come up with a whitelist of allowed tokens inside curly braces (e.g., `{ns}`, `{node}`) for advanced use cases right now. Are there any other expansion tokens that one would want? Should we prefix them with a dollar sign, as in canonical bash, like `${ns}`  ? :chicken:

jack-oquin: Actually, bash resolves `~foo` to the home directory of user `foo`. You probably don't have that name defined on your computer. It is definitely _not_ the same as `~/foo`, which looks for `foo` in the current user's home.

However, I have no objection to making them both equivalent as ROS names, simply to ease ROS 1 migration.

jack-oquin: I am inclined to agree with Will, not to add any more-complex substitutions.

wjwwood: > Actually, bash resolves ~foo to the home directory of user foo. You probably don't have that name defined on your computer. It is definitely not the same as ~/foo, which looks for foo in the current user's home.

If the analogy to the shell syntax is important then we should consider a point of confusion. I think it would be worth the migration cost to address this inconsistency. In pretty much every case it could be done automatically. For a ROS 1 shim it could be done automatically (always replace `~` with `~/` and remove redundant `/`'s), and if used in ROS 2 it could produce a warning but convert it implicitly.

wjwwood: > We can always revisit this decision (or any decision) in the future. I agree we need to move forward.

That's true, but it's also true that when it comes to software engineering, later == more expensive. I don't see any advantage to staying the decision.

> I suppose we could come up with a whitelist of allowed tokens inside curly braces (e.g., `{ns}`, `{node}`) for advanced use cases right now. Are there any other expansion tokens that one would want? Should we prefix them with a dollar sign, as in canonical bash, like `${ns}` ? 🐔

I think coming up with a substitution "pattern" is sufficient to decide now. If we have that then adding substitution keywords becomes relatively straightforward. Between `$`, `${}`, and `{}`, I like `{}` myself because it's more convenient to type but not as difficult to parse (unlike `$ns` or `$node` which has no definite bound). Unlike bash, we don't need to discriminate the substitution syntax from other things like function definitions, so I think we can live with a simple syntax. I'm still worried about every client having to deal with things like unmatched `{` and odd typos like `/foo{node/bar}/baz`.

My intention is not to obstruct moving forward, I just want to make sure we do our best to nail down anything that could influence the parsing of the names to avoid costly iterations in the future.

articles/115_topic_and_service_name_mapping.md
wjwwood: I had that before, but my goal was to have each rule be correct without other rules (otherwise the rules are somewhat contradicting each other). For example, the valid name `~/foo` when applied to "May contain tilde (`~`) characters only if each individual tilde is surrounded by forward slashes" is invalid because the `~` is not surrounded by `/`.

Either way, it sounds like we'll be changing this so I'll wait till that discussion is done.

articles/115_topic_and_service_name_mapping.md
codebot: Makes sense; I like your idea of nailing down the substitution syntax and dealing with the exact keyword mapping as needs arise. Perhaps because UNIX shells have been around for so long, there are usually many ways to say the same thing, like `$ns` and `${ns}`, for example. I worry about about omitting the dollar sign, since [brace expansion](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html) is a thing in bash. I'd vote for either `${ns}` or just `$ns` but hopefully other folks can also cast votes. I think the version with braces `${ns}` is considered best-practice in bash since it's more obvious where the variable name begins and ends, and like you said, it's more obvious when there is a syntax error, but I might be wrong on that. :chicken:

wjwwood: It's important to remember this is _not_ bash, and since we don't have brace expansion or function (scope) blocks, then we don't necessarily have to have a `$`. There are plenty of other systems where `{}` is used for substitution, like Python string templating: https://docs.python.org/2/library/string.html#format-string-syntax There are yet more ways of expressing substitution, like `%{}` for "hash" syntaxing in both Python and Ruby, or `#{}` for string interpolation in Ruby. Long story short, anything we pick will be at odds with something someone else is familiar with, so it's sort of arbitrary. I agree that it's best to see what a large group of people think.

mikaelarguedas: I agree with Morgan's point, I'd be in favor of ${ns}, while it may be more
cumbersome to write it will avoid confusion on the user side and be easier
to debug.

brunohexsel: I like the idea of having a subsitution token such as `${ns}` instead of "~". However I worry about the following case:

``` bash
# Somewhere this is defined:
export ns="some_value"

# The user now tries to manually instantiate a node passing the topic as parameter:
my_node my_parameter_with_namespace:=${ns}/some_topic
# bash would interpret it as:
my_node my_parameter_with_namespace:=some_value/some_topic
# but this could be avoided with:
my_node 'my_parameter_with_namespace:=${ns}/some_topic'
# which is a little cumbersome and not everyone knows about it.
```

I would go with what @wjwwood suggested with "{ns}" or "%{}" that would not be a source of very obscure and hard to debug errors.

gbiggs: I vote for treating `~` the same way as Bash, except for `~foo`. While it would be nifty to say "the namespace of node foo", it would be ambiguous if there are many foos in different name spaces.

I also vote for `{}` with no other options. As William said, this is not Bash so we do not need to be exactly like them. I do have one concern, though: It might need escaping in Python strings when .format() is used, which would mean that every instance of `{...}` becomes `{{...}}`.

As potentially usable substitutions, `{ns}` and `{node}` sound like they will be frequently useful.

wjwwood: Sorry for letting the ball drop on this. Let me try to recap:
- Only allow `~` at the beginning of names
- `~` must be followed by `/`
  - uses of `~foo` (which is allowed in ROS 1) can always be implicitly upgraded to `~/foo`, but in ROS 2 we can additionally choose to disallow that or warn about it
- use `{}` as a generic substitution escape sequence
  - actual substitutions, e.g. `{ns}` or `{node}`, to be determined later

I reiterate the proposal that we use `{}` as the substitution syntax because of the argument that `${}` would collide with expressing the substitutions within shell scripts which use the `${}` already. Did @brunohexsel's argument about bash collisions sway any of the `${}` proponents (@codebot or @mikaelarguedas)?

Anyone else want to weigh in with an opinion?

mikaelarguedas: Fine by me to use `{}` rather than `${}` to avoid collision in shell scripts :+1:

codebot: Thanks everyone for the excellent points; I hereby rescind my preference for `${foo}` and henceforth prefer `{foo}`. Thanks @brunohexsel for the concise and helpful example. :+1:

wjwwood:
I believe I've addressed all of the comments and I just added a section about the length limit and a comparison to ROS 1's rules.

dhood:
I've opened #94 with some suggested changes.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,421 @@
...
+- must start with a forward slash (`/`), i.e. they must be absolute
+- must not contain tilde (`~`) or curly braces (`{}`)
+
+Note that expanded substitutions must result in a valid name.
+For example, a name `foo_{bar}` is not valid if `bar` is expanded as `_baz`, because that would result in a fully qualified name of `foo__baz` which contains repeated underscores (`_`).
+
+### Uniform Resource Locators (URL's)
+
+Additionally, topic and service names can be represented in the [Uniform Resource Locator (URL)](https://en.wikipedia.org/wiki/Uniform_Resource_Locator) format to further disambiguate the resource name.
+A topic name may be preceded by a `rostopic://` scheme prefix, and a service name may be preceded by a `rosservice://` scheme prefix.
```
dhood, line 101: What would the fully-qualified name look like for Connext vs Opensplice?

wjwwood: The same, this is before it is translated into DDS topic names.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,421 @@
...
+- may be a single tilde character (`~`)
+
+### Substitutions
+
+The bracket syntax (`{substitution_name}`) maybe used in non-fully qualified names to substitute useful contextual information into the name.
+The set of substitution keys (names) are not set in this document, but some reasonable examples might be: `{node}` expands to the current node's name or `{ns}` expands to the current node's namespace.
+
+Substitutions are expanded before the private namespace substitution character is expanded.
+Therefore a substitution may contain the private namespace substitution character, i.e. `~`.
+For example, given the name `{private}foo`, that there is a substitution called `{private}` which expands to `~/_`, the current node name is `my_node`, and the current node's namespace is `my_ns`, then the fully qualified name would be `/my_ns/my_node/_foo`.
```
dhood: Are nested curly braces permitted?

articles/115_topic_and_service_name_mapping.md
wjwwood: Also this section is about the URL syntax, not the fully-qualified name :stuck_out_tongue:.

dhood: yeah, I meant DDS topic name when I said fully-qualified. I found myself wondering what would happen for the two, given the difference in `rs` vs `rq`/`rr` prefixes. You are right that extra info wouldn't necessarily go in this section though :smiley:

articles/115_topic_and_service_name_mapping.md
wjwwood: That's a good question. I'd default to no, just to keep the expansion simple.

wjwwood: I'll update the document, if anyone thinks they should be nested they can comment here.

articles/115_topic_and_service_name_mapping.md
wjwwood: Yeah, they will be different, but they would have been anyways, i.e. they won't inter-operate now. The resulting DDS topic names will differ though. Connext would be `rs` and Opensplice would be `rq` and `rr`.

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,424 @@
...
+| `/foo/bar`        | Topic     | `rt__foo__bar`         |
+| `/_foo/bar`       | Topic     | `rt___foo__bar`        |
+| `/foo/_bar`       | Topic     | `rt__foo___bar`        |
+| `/_foo/_bar`      | Topic     | `rt___foo___bar`       |
+| `/_foo/_bar/_baz` | Topic     | `rt___foo___bar___baz` |
+
+### ROS Topic and Service Name Length Limit
+
+Because DDS topic names must be limited to 255 characters, the length of a ROS topic or service name is also limited in length.
+Calculating this length is governed by an unfortunately non-trivial algorithm.
```
dhood: if we think the current expression seems too contrived, we can also refer to the name length limit as C+N+P <= 255 where C is the number of characters in the fully-qualified name and N is the number of name tokens. not essential.

wjwwood: It would still need to be `C+(N*2)+P <= 255` right?

dhood: not if you let C refer to any character, as opposed to any non-/ character

wjwwood: Ah, that's probably simpler, do you want to reword it or would you prefer I do it?

dhood: I'll do it

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,423 @@
...
+
+## ROS 2 Topic and Service Name Constraints
+
+In this section an outline of the proposed constrains for ROS 2 topic and service names will be enumerated along with rationales where appropriate.
+
+For convenience here is a summary of all rules for topic and service names in ROS 2:
+
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), or forward slashes (`/`)
+- may use balanced curly braces (`{}`) for substitutions
+- may start with a tilde (`~`), the private namespace substitution character
```
dirk-thomas, line 64: What is the difference between a relative topic (without a leading slash) and a topic starting with a tilde? Do we need the special semantic of the tilde?

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,423 @@
...
+- must not start with a numeric character (`[0-9]`)
+- must not contain any number of repeated underscores (`_`)
+- must not end with an underscore (`_`)
+- must not have an underscore (`_`) followed by a forward slash (`/`), i.e. `_/`
+- must not end with a forward slash (`/`)
+- must not contain any number of repeated forward slashes (`/`)
+- must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`
+- must have balanced curly braces (`{}`) when used, i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`
```
dirk-thomas, line 72: Should we instead define what the content inside the curly braces can be? Alphanumeric + underscore? Any other constraints?

articles/115_topic_and_service_name_mapping.md
```diff
@@ -0,0 +1,423 @@
...
+- must not contain tilde (`~`) or curly braces (`{}`)
+
+Note that expanded substitutions must result in a valid name.
+For example, a name `foo_{bar}` is not valid if `bar` is expanded as `_baz`, because that would result in a fully qualified name of `foo__baz` which contains repeated underscores (`_`).
+
+### Uniform Resource Locators (URLs)
+
+Additionally, topic and service names can be represented in the [Uniform Resource Locator (URL)](https://en.wikipedia.org/wiki/Uniform_Resource_Locator) format to further disambiguate the resource name.
+A topic name may be preceded by a `rostopic://` scheme prefix, and a service name may be preceded by a `rosservice://` scheme prefix.
+For example, the absolute name `/foo` could also be represented as a topic with `rostopic:///foo` or as a service with `rosservice:///foo`.
```
dirk-thomas, line 102: Are these to resource names colliding or can they be used at the same time? So does the protocol have a separate namespace or do they share the same namespace? This should be clearly stated (I assume it is the former?).

articles/115_topic_and_service_name_mapping.md
wjwwood: Put another way, if the tilde is used then the tilde may only be at the front of the name. It doesn't have anything to do with absolute versus relative names in my opinion.

dirk-thomas:
With this proposal all topics are being prefixed before they are being passed to DDS. I think it would be valuable to allow users to e.g. subscribe to "native" (non ROS created) DDS topics using the ROS api. E.g. if someone wrote a DDS participant (without using ROS) which publishes on the DDS topic "image" there should be a way to subscribe to that topic using ROS (maybe requiring special flags).

articles/115_topic_and_service_name_mapping.md
wjwwood: I believe they can be used at the same time, as the DDS topic names will differ. This is related to the on going discussion about "host-oriented" versus "content-oriented" addressing of services. What does "protocol" mean in this case? Which "namespace" are you referring to, the ROS namespace or the DDS topic namespaces (the ones separated by `__`) or maybe the prefix used in the DDS topic name?

articles/115_topic_and_service_name_mapping.md
dirk-thomas: Maybe you can clarify the difference based on an example. How are the following two topics resolved to fully qualified names in a node with the namespace `/ns`:
- `foo`
- `~/foo`

articles/115_topic_and_service_name_mapping.md
wjwwood: That's a good point, my instinct is to say that it has to follow the same rules as the rest of the topic name, just to simplify validation. For example, if we allow `!` or `__` in the substitution content, then the content of substitutions need to be removed from the topic name before doing checks.

wjwwood:
@dirk-thomas that use case is covered in the document.

articles/115_topic_and_service_name_mapping.md
dirk-thomas: But a trailing underscore wouldn't do any harm. So I think it needs to be specific about what is allowed and what isn't.

articles/115_topic_and_service_name_mapping.md
wjwwood: There are examples just as you've described further into the document. This bullet list is meant to be a quick reference of all the rules.

dirk-thomas:
Are you referring to this line? https://github.com/ros2/design/pull/90/files#diff-11cdcb7792e2ef4c9f19a62673484497R367

Can you please reference the line you have in mind. I couldn't find it myself. Thanks.

wjwwood:
https://github.com/ros2/design/pull/90/files#diff-11cdcb7792e2ef4c9f19a62673484497R378

dirk-thomas:
If this rational is valid for the proposed approach should this line be moved to a different section? It looks to me like it was only relevant for the alternative approach where it is mentioned.

wjwwood:
We can mention it in one of the sections other than the alternatives. I saw it as a prescription on how to address a specific application of the rules, so I didn't mention it directly. Which section do you think it belongs in, the "ROS Specific Name Prefix" section maybe?

articles/115_topic_and_service_name_mapping.md
dirk-thomas: I guess I don't see a significant benefit in have two separate concepts (the node name as well as a namespace) which together form the prefix of the nested topics.

The complexity of the table in "Private Namespace Substitution Character" could be reduced by only having a single concept. The namespace of a node is by default equal to the name of the node. And when you specify a custom namespace you are free to choose something completely different or use something which contains the node name. With that you only have to distinguish between absolute topics and relative topics (no need for ~) and the node provides the namespace to convert a relative topic into an absolute one.

dirk-thomas:
Maybe mention the use case to be able to subscribe to any DDS topics name at the end of section "ROS to DDS Name Conversion Examples"? Something like: 

> In order to be able to interoperate with "native" topic names the API should provide a way to skip the ROS specific prefixing.

articles/115_topic_and_service_name_mapping.md
wjwwood: I think the point is emulate the behavior of ROS 1:

http://wiki.ros.org/Names#Resolving

You can imagine that you have two nodes, one publishes on `foo` and the other subscribes on `foo`, if you put them both in the namespace `bar` then they should still communicate. If the node name is always included in the namespace then they wouldn't talk because one would publish to `/bar/node1/foo` and the other would subscribe to `/bar/node2/foo`.

The `~` is specifically for when you want it to be node private, but in that case it should probably still include the namespace of the node in addition to the node name.

With your proposal I do not think it is possible to have a relative name resolve to the same name but absolute, i.e. `foo` -> `/foo`, because the namespace is never empty.

wjwwood: Sorry my previous comment wasn't very concise. I understand your proposal as:

> Remove `~` and default the namespace to `<node name>`

If I interpret that correctly, then I think it no longer allows for simple relative names like `foo` to become simple absolute names like `/foo`. Is that right? Maybe that use case doesn't matter?
It also makes it harder to naturally map to ROS 1 where `~` is used.

dirk-thomas: My proposal is to unify the two concepts. As far as I can see the expressive power should be equivalent to what is possible with the two separate concepts.

You can any of the following:
- not set a custom ns: the name of the node is being used (e.g. `node_name`)
- you can set a custom ns: that name space is being used, that ns can be:
  - empty
  - any name (not related to the node name):  e.g. `foo` or `bar/baz` 
  - something nested the node name: e.g. `node_name/foo` (to make this easier to specify without retyping the node name the namespace of a node could use a substitution like `{nodeName}/bar`

articles/115_topic_and_service_name_mapping.md
wjwwood: Ok, I update the rules w.r.t. substitution content: 00e06d0

Let me know what you think.

wjwwood:
I added a section about the native topic names: d787ac2

articles/115_topic_and_service_name_mapping.md
wjwwood: I wasn't considering a difference between not set and empty for the namespace. I think what you've proposed is fine if the "empty" namespace is used, but I don't think it significantly simplifies the rules because now you have to know that an unset namespace implicitly implies node name. With the `~`, you may not know right off what it does, but it is more explicit that something is happening and you can look it up. To me it is less surprising when `foo` becomes `/foo` as opposed to `/node_name/foo`.

I'm the last person to choose an alternative solely for ROS 1 compatibility, but in this case I think the benefit of getting rid of `~` doesn't outweigh the problems of converting to and from ROS 1 topic names easily. Perhaps that is simpler than I am imagining, can you come up with a set of rules that could be followed to go from one to the other?

articles/115_topic_and_service_name_mapping.md
dirk-thomas: Thanks for updating it.

Since you mentioned the problem with a double underscore inside, should we allow that or not? Not allowing it would make the logic handling it easier. Therefore I would suggest the latter one.

dirk-thomas:
Great, the new section is perfect.

articles/115_topic_and_service_name_mapping.md
wjwwood: Yeah, I thought about that, but since you have to do the substitutions before checking for correctness (as mentioned in the document) I didn't think it would be a burden to be slightly less restrictive for the substitution name.

I think you have to do the substitution before validating the topic name because if the substituted value has a leading underscore, or has a trailing underscore and is followed by an underscore, then the expanded topic name could become invalid where the original one was not. I think you also cannot just replace substitutions with empty string either because an example like this `/{name}_foo` would not be valid if the substitution is just ignored to do pre-expansion validation.

The only case where not permitting `__` in the substitution text would be useful is a pre-expansion check which naively asserts that no double underscore (`__`) is contained in the name string. I'm just not sure how useful that check is, since you need to do it again after expanding.

Having said all that, I'm fine with not allowing `__` in the substitution name because I don't think it is something we need to support.

dirk-thomas: Thanks for the clarification. Then it is fine with me to not specify any further restrictions since there seems to be no need for it. So the current state should be good as-is.

tkruse:
Regarding names I would suggest formalizing the rules, e.g. using EBNF. Here is a working prototype:
https://gist.github.com/tkruse/e07d9c5af6d3caad945f4715b130d407 (had to cheat for double underscores, multiple substitutions in same segment missing, trailing underscore in path segment not failing yet (I got bored) )

I assume the benefits are obvious to you.

wjwwood:
An EBNF would be useful, but I won't have time to pick up where you left off right now. Does anyone else have the interest and time to finish it?

wjwwood:
I'm going to go ahead and merge this since it is in a pretty decent place. I've ticketed the task to finish the EBNF: https://github.com/ros2/design/issues/96

Also, there is a chance that this will need to be changed once we address the topic of Service name addressing, as mentioned in the original post of this pr, but we can handle that when it comes up.

wjwwood:
For anyone who was following this pr, but not the repository, I've proposed some changes to how substitutions are parsed: https://github.com/ros2/design/pull/127

PR #91, opened 2016-06-10
require Python 3.5

dirk-thomas:
Only merge after the switch to Xenial.


jack-oquin:
What is the rationale for this change?

wjwwood:
We're moving our target platform to be Xenial and it has Python 3.5. There are features in 3.5 that we want to use, so we're making it the minimum.

dirk-thomas:
This still needs review.

wjwwood:
+1

PR #92, opened 2016-06-15
update dependency list

dirk-thomas:
It looks like upstream has just released a new major version. I don't know yet what is necessary to make it work again.


dirk-thomas:
The current patch works for me locally. I don't know what the difference is on Travis. I will unassign myself for now and let someone else take a look.

dirk-thomas:
Finally got it to work using a newer npm version.

Please review.

wjwwood:
+1

PR #95, opened 2016-09-16
Various edits

dhood:
While reading through the design docs I've been making some edits.

Most are just minor text edits; I'm trying to keep the ones that might cause more of a discussion in separate commits.

### DIFF
--- a/articles/020_ros_with_dds.md
+++ b/articles/020_ros_with_dds.md
@@ -29,7 +29,7 @@ Terminology:
 ## Why Consider DDS
 
 When exploring options for the next generation communication system of ROS, the initial options were to either improve the ROS 1.x transport or build a new middleware using component libraries such as [ZeroMQ](http://zeromq.org/), Protocol Buffers, and zeroconf (Bonjour/Avahi).
-However, in addition to those options, which both involved us building a middleware from parts or scratch, other end-to-end middlewares were considered.
+However, in addition to those options, which all involved us building a middleware from parts or scratch, other end-to-end middlewares were considered.
 During our research, one middleware that stood out was DDS.
 
 ### An End-to-End Middleware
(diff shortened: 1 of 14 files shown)


articles/020_ros_with_dds.md
```diff
@@ -29,7 +29,7 @@ Terminology:
 ## Why Consider DDS
 
 When exploring options for the next generation communication system of ROS, the initial options were to either improve the ROS 1.x transport or build a new middleware using component libraries such as [ZeroMQ](http://zeromq.org/), Protocol Buffers, and zeroconf (Bonjour/Avahi).
-However, in addition to those options, which both involved us building a middleware from parts or scratch, other end-to-end middlewares were considered.
+However, in addition to those options, which all involved us building a middleware from parts or scratch, other end-to-end middlewares were considered.
```
wjwwood: Maybe that should be "both of which involved..."

articles/020_ros_with_dds.md
```diff
@@ -77,21 +77,21 @@ This type of evolution in the standard body for DDS is an encouraging thing to o
...
-- financial systems
-- space systems
-- flight systems
-- train switchboard systems
+- battleships,
+- large utility installations like dams,
+- financial systems,
+- space systems,
+- flight systems,
+- train switchboard systems,
```
wjwwood: Not sure this is necessary since they are already delimited by bullets.

articles/020_ros_with_dds.md
dhood: ah yep, this one I misinterpreted.

articles/050_ros_rpc_design.md
```diff
@@ -92,16 +92,16 @@ For logging/introspection purposes the RPC Server instance might publish all inc
 
 ### ROS Preemptible RPC API
 
-The ROS preemptible RPC API will extend the Asynchronous API to enable preemption of an RPC in progress using a UID.
+The ROS preemptible RPC API will extend the Asynchronous API to enable preemption of an RPC in progress using a user identifier (UID).
```
wjwwood: This stands for "unique identifier".

articles/050_ros_rpc_design.md
```diff
@@ -112,9 +112,9 @@ This will just be a thin layer on top of the Asynchronous API requiring no addit
 
 There are some issues with the above proposed approach, which are outlined below.
 
-### Visibility of UID's
+### Visibility of User Identifiers
```
wjwwood: Unique Identifiers.

contribute.md
```diff
@@ -70,9 +70,9 @@ If you need to reproduce the site locally and/or make extensive changes, you can
 
 ### Forking this Repository
 
-First check to see if you already have a fork of this repository:
+First check to see if you already have a fork of this repository at:
 
-[https://github.com/<your username>/design][]
+https://github.com/{your github username}/design
```
wjwwood: The `{}` typically denotes "pick one of", e.g. `{up, down}`, whereas `<thing>` means mandatory replace. This in the style of USAGE or man pages. So, I think that it should remain inside `<>`. Also, if you don't make a link out of it with the `[name](url)` then it will not link. I think instead it should just be in a pre using the ````.

contribute.md
```diff
@@ -81,9 +81,9 @@ If you do not, then browse to this [repository](https://github.com/ros2/design)
 Github will now go off and create a fork off this repository into your Github account.
 Then you can clone your fork of the repository by running this command:
 
-    git clone https://github.com/<your username>/design.git
+    git clone https://github.com/{your github username}/design.git
```
wjwwood: Same.

contribute.md
wjwwood: I think it's fine to leave it outside of the `[]()` syntax (which I'm assuming was intended here, but resulted in a typo).

dhood: ``s were the key to making the link display properly, thanks

contribute.md
```diff
@@ -81,7 +81,7 @@ If you do not, then browse to this [repository](https://github.com/ros2/design)
 Github will now go off and create a fork off this repository into your Github account.
 Then you can clone your fork of the repository by running this command:
 
-    git clone https://github.com/{your github username}/design.git
+    git clone https://github.com/<your github username>/design.git
```
wjwwood, line 84: Does this one display correctly then?

dhood: yeah because it's a codeblock

wjwwood (approved):

dhood:
I'll probably have some more changes, so will keep pushing to this PR

wjwwood:
Sounds good.

articles/050_ros_rpc_design.md
jack-oquin: Should it be `UUID`? Or, are they not universal?

wjwwood: That's a good question. I left it as `UID` because it could be `GUID` if the request is uniquely identified by the requester plus a sequence number, but otherwise you might need a full `UUID` if the id must be unique across all service requests.

wjwwood:
@dhood you think this will be merged before ROSCon? or is it not that important.

dhood:
yeah I think it will, main thing left to do is that I was going to re-organise the index a bit.

PR #97, opened 2016-10-30
Fix #96 Add EBNF grammar for ROS2 names in self-testing python script

### LINKED ISSUE #96: finish EBNF for topic and service names
See: https://github.com/ros2/design/pull/90#issuecomment-247329344

tkruse:
Prototype for #96. Since last time I extended a bit. Because of the fully-qualified vs. non-fully-qualified, the grammar has grown a bit.

The grammar still does not prevent double underscores, the python script around it can handle that much easier I think.

The script itself does not yet handle maximum name length constraints.

### DIFF
--- a/articles/ebnf_names.py
+++ b/articles/ebnf_names.py
@@ -1,0 +1,207 @@
+#! /usr/bin/env python
+
+# PoC Ros2Names parsing using EBNF and Grako (install from pypi)
+# Tested with grako 3.14.0
+#
+
+
+## TODO:
+# maximum length + example
+
+## ambiguous texts
+# "balanced" curly braces may imply nesting?
+# url scheme only 'rostopic' and 'rosservice' allowed?
+
+
+import re
+import grako
+
+
+POSITIVES_QUALIFIED = [
+    "/x", "/foo", "/abc123", "/_foo", "/Foo", "/BAR",
+    "/foo/bar", "/x/y/z",
+    "/_foo/bar", "/_f_o_o/bar",
+    "rosservice:///foo", "rostopic:///foo/bar"
+    ]
+
+POSITIVES_UNQUALIFIED = [
+    "x", "foo", "abc123", "_foo", "Foo", "BAR",
+    "x/y/z",
+    "foo/bar",
+     "{foo}", "{foo}_bar", "foo/{ping}/bar",
+     "foo{x1}bar{x2}baz", "{x1}bar{x2}baz{top}",
+     "foo{x1}/bar/{x2}baz", "{x1}/bar/{x2}",
+     "~", "~/foo", "~/foo/x/_123",
+     "_foo/bar", "_f_o_o/bar",
+     "rosservice://foo", "rostopic://foo/bar", "rostopic://~/_bar/foo{x}",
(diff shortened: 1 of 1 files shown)


tkruse:
Some things to clarify (consider clarifying in the article):
- is '{foo}' only allowed? -> current script assumes yes
- is /foo/123/bar allowed? -> current script assumes yes
- "balanced" curly braces may imply nesting? -> current script assumes no
- ~ only in starting position? -> current script assumes yes, only in starting position
- url scheme only 'rostopic' and 'rosservice' allowed? -> current script allows any alphabetic name

wjwwood:
Thanks for putting the current state into a pr, that's easier to track.

> is `{foo}` only allowed? -> current script assumes yes

I'd say yes, so long as the contents of `foo` are valid on their own, i.e. not empty.

> is /foo/123/bar allowed? -> current script assumes yes

I would say no, the section for name tokes says:

https://github.com/ros2/design/blob/gh-pages/articles/115_topic_and_service_name_mapping.md#name-tokens

> Topic and service name tokens:
> 
> ...
> - must not start with numeric characters ([0-9])

Since `123` in your example is a token, I think it must start with an alphabetical character.

> "balanced" curly braces may imply nesting? -> current script assumes no

I suppose that it could, but the document does not appear to be clear on that point right now. I vaguely remember discussing it, but I don't see it in the issue history at the moment, so I'll open an issue to address that concern: https://github.com/ros2/design/issues/98

> ~ only in starting position? -> current script assumes yes, only in starting position

Yes, that is described in the document I as:

https://github.com/ros2/design/blob/gh-pages/articles/115_topic_and_service_name_mapping.md#ros-2-topic-and-service-name-constraints

> For convenience here is a summary of all rules for topic and service names in ROS 2:
> ...
> - may contain alphanumeric characters ([0-9|a-z|A-Z]), underscores (_), or forward slashes (/)
> - ...
> - may start with a tilde (~), the private namespace substitution character

It is also mentioned later in the document in more explicit terms.

> url scheme only 'rostopic' and 'rosservice' allowed? -> current script allows any alphabetic name

I think that is find for the script, for now. I imagine we'll want to limit what goes there, but I also don't think `rostopic` and `rosservice` will be exhaustive in the end. We can leave it open ended for now (like with the substitution names).

wjwwood:
I just wanted to clarify my response to:

> is /foo/123/bar allowed? -> current script assumes yes

The document says, more specifically:

https://github.com/ros2/design/blob/gh-pages/articles/115_topic_and_service_name_mapping.md#name-tokens

> Topic and service name tokens:
> - ...
> - may use alphanumeric characters ([0-9|a-z|A-Z]), single underscores (_), and/or balanced curly braces ({})
> - must not start with numeric characters ([0-9])

So, it cannot start with a numeric (so `123` is not a valid token name), but it can start with more than just an alpha, i.e. `_` underscore or `{` curly brace.

Thanks @dhood for pointing this out.

tkruse:
Ah, somehow I missed the "token" bit. That makes the EBNF simpler to write. Updated the PR, added more positive and negative examples to tests.

PR #99, opened 2016-11-09
Define external life cycle management interface

gbiggs:
The previous version of the document only gave an ambiguous description of the external interface used to manage a node's life cycle. This pull request describes the interface in detail, including topic names, and message and service definitions.


dirk-thomas:
Some high level feedback for now:
- Do we need a specific service for every requested transition? Having a single service which takes an argument would make the API more flexible (since it would work for any state machine defined).
- Currently ROS 2 does neither have the concept of a namespace nor latching. While we hope to support something similar in the future it might be problematic to refer to those terms in the design doc for now.

Regarding the Travis error: every sentence in the markdown files should start on a new line. Hopefully if you update the patch the ugly error should go away.

dirk-thomas:
Regarding the Travis error: every sentence in the markdown files should start on a new line. Hopefully if you update the patch the ugly error should go away.

gbiggs:
Both good points. Thanks, Dirk. I've made changes for both.

Karsten1987:
I'd like to propose to rename `create` and `destroy` into `init` and `deinit` respectively. I am having trouble with the word create, since we don't have a factory which creates nodes and we encourage users to inherit from a generic base class. That being said, a create function is more like a constructor and a destroy is a destructor which never really gets called externally. 

Concretely speaking, this would result in an non-op constructor, getting called when a user instantiates a new node. The respective init function, then, which is responsible for enabling all pub/sub etc, can get triggered externally.

gbiggs:
I don't feel that renaming them to `init` and `deinit` is correct. These two functions are more intended to be provided by a container system that has a factory to create and destroy nodes on demand, rather than by the nodes themselves. This is why they are not included in the node's external interface specification. If a managed node is created by a user in their own `main()`, then there will be no `create` and `destroy` calls.

If I clarify the document to make this clear, would that be satisfactory?

jack-oquin:
> If I clarify the document to make this clear, would that be satisfactory?

+1; that would be a useful improvement.

jack-oquin:
Overall, this document looks good to me. 

Maybe I just missed it, but where and how does the management interface return the current state?

gbiggs:
The current state is broadcast on a topic, and is also accessible via a service.

gbiggs:
I clarified that `create` and `destroy` are functions of a container system, but I think the document could do with more detail on what a container is and what interface it should provide. However, that needs more discussion about what, if anything, we want to do with containers.

articles/node_lifecycle.md
```diff
@@ -196,51 +199,176 @@ Transitions to `ErrorProcessing` may be caused by error return codes in callback
...
+They shall not be disabled by the node shifting to the `Inactive` state, for example.
 
-A managed node will be exposed to the ROS ecosystem by the following interface, as seen by tools that perform the managing.
-This interface should not be subject to the restrictions on communications imposed by the lifecycle states.
+### Interface namespace
 
-It is expected that a common pattern will be to have a container class which loads a managed node implementation from a library and through a plugin architecture automatically exposes the required management interface via methods and the container is not subject to the lifecycle management.
-However, it is fully valid to consider any implementation which provides this interface and follows the lifecycle policies a managed node.
-Conversely, any object that provides these services but does not behave in the way defined in the life cycle state machine is malformed.
+The interface shall be provided in a namespace named "infra/lifecycle" underneath the node's namespace.
```
Karsten1987: As we don't have namespaces in ROS2.0 for the moment, I believe this paragraph can not be merged as is. Currently, the topics and services are advertised with `*__get_state`,  `*__change_state` and `*__transition_event`

articles/node_lifecycle.md
```diff
@@ -196,51 +199,176 @@ Transitions to `ErrorProcessing` may be caused by error return codes in callback
...
+    uint8 UNCONFIGURED=1
+    uint8 INACTIVE=2
+    uint8 ACTIVE=3
+    uint8 FINALIZED=4
+    uint8 CONFIGURING=10
+    uint8 CLEANING_UP=11
+    uint8 SHUTTING_DOWN=12
+    uint8 ACTIVATING=13
+    uint8 DEACTIVATING=14
+    uint8 ERROR_PROCESSING=15
```
Karsten1987: we can link to the actual msg: https://github.com/ros2/rcl_interfaces/blob/master/lifecycle_msgs/msg/State.msg

articles/node_lifecycle.md
```diff
@@ -196,51 +199,176 @@ Transitions to `ErrorProcessing` may be caused by error return codes in callback
...
+
+### Life cycle state changes topic
+
+When the node's life cycle changes, it shall broadcast the following message on the `infra/lifecycle/state_change` topic.
```
Karsten1987: `<node_name>__transition_event'

articles/node_lifecycle.md
```diff
@@ -196,51 +199,176 @@ Transitions to `ErrorProcessing` may be caused by error return codes in callback
...
+    next_state = ERROR_PROCESSING
+    trigger = "error in activating state"
+
+    previous_state = ERROR_PROCESSING
+    next_state = UNCONFIGURED
+    trigger = "error processing returned OK"
+
+### Current life cycle state service
+
+The node's current life cycle state shall be available via the `infra/lifecycle/get_state` service.
```
Karsten1987: `<node_name>__get_state`

articles/node_lifecycle.md
```diff
@@ -196,51 +199,176 @@ Transitions to `ErrorProcessing` may be caused by error return codes in callback
...
+    SHUTTING_DOWN    | shutting_down
+    ACTIVATING       | activating
+    DEACTIVATING     | deactivating
+    ERROR_PROCESSING | error_processing
+
+The `UNKNOWN`/`unknown` value shall be assumed by clients of the service to indicate that the node is in an unknown state and thus unusable.
+
+### Transition request service
+
+The service `infra/lifecycle/change_state` service shall be provided by the life cycle interface.
```
Karsten1987: `<node_name>__change_state`

articles/node_lifecycle.md
dirk-thomas: Which is still a kind of namespace - just not with a `/` but with `__`.

dirk-thomas:
@gbiggs Can you please address the pending comments.

articles/node_lifecycle.md
gbiggs: Now that namespace support is being implemented, is this paragraph acceptable?

articles/node_lifecycle.md
gbiggs: Should the msg file follow the specification, or should the specification reference the msg? One needs to be considered the master definition, and the other needs to be kept in sync with it.

Although I prefer keeping the specification document as the master definition, when considering maintenance, having the spec link to the msg file would be easier.

articles/node_lifecycle.md
gbiggs: See above about namespaces.

articles/node_lifecycle.md
gbiggs: See above about namespaces.

articles/node_lifecycle.md
gbiggs: See above about namespaces.

articles/node_lifecycle.md
wjwwood: It should probably make use of the leading `_` in topic names to make it "hidden":

http://design.ros2.org/articles/topic_and_service_names.html#hidden-topic-or-service-names

Otherwise, it uses a namespace as currently written, so I'm not sure what else you might be referring to.

articles/node_lifecycle.md
wjwwood: I don't mind either way. Linking to the spec might have the problem of tracking changes over time. Then again it may be just as hard to represent that in this document as well. So I don't see that one way is strictly better than the other. At least with the interface file being the sole authority, you can avoid some duplication of information if you choose.

articles/node_lifecycle.md
gbiggs: I was referring to @dirk-thomas 's comment that it doesn't use an actual namespace (separated with a `/`) but uses `__` to mimic a namespace. The proposal from @Karsten1987 is to change the topics used by the lifecycle interface to not use namespaces.

articles/node_lifecycle.md
gbiggs: Having thought about it more, I prefer keeping the spec as the master definition. In theory there could be other implementations. I will add a link, though.

articles/node_lifecycle.md
wjwwood: That's what the implementation is doing right now. I think you should write the spec to how it "should be" and we can catch the implementation up when we have finished implementing namespaces (in progress now).

gbiggs: That's what I was getting at.

And on going to add the underscore to make it a hidden interface, I found myself unsure where to add it. On `infra` or on `lifecycle`? Is `infra` going to be a widely-used thing for ROS infrastructure topics? (I hope it is.)

wjwwood: I don't know how I feel about `infra` over `infrastructure` or `impl` or `ros`, or even just letting individual things create their own "impl" namespaces. For example, maybe the common namespace for these things should be just `~/_lifecycle/`. But in general, namespacing the plumbing is something I think we should be doing.

gbiggs: I'm in favour of having all infrastructure stuff under a single namespace because it makes things tidier, but I can see the argument of "who defines/controls what is 'infrastructure' enough to go in there?".

wjwwood: I can see that angle too. It seems to be a bit a bikeshed thing atm. I'm fine with leaving it as-is because we can always come back and change it later once we get more of these "infrastructure" topics and services in place. Perhaps then a better pattern with emerge.

gbiggs: OK, I make the `infra` namespace hidden. We can wait a while and see if anything else starts using this same namespace. If not, we can remove it.

PR #103, opened 2016-12-06
add article containing stories

dirk-thomas:
I have these "stories" in the back in my head and bring them up frequently in conversations to motivate what we are working on. This article just collects (some of) them to be able to make the visible to a bigger audience.

### DIFF
--- a/articles/040_stories.md
+++ b/articles/040_stories.md
@@ -1,0 +1,82 @@
+---
+layout: default
+title: Stories driving the development
+permalink: articles/stories.html
+abstract:
+  This article captures some stories which drive the direction of features in ROS.
+published: true
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+categories: Overview
+---
+
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+The article enumerates a few stories which sketch what will be possible with ROS in the future.
+The list is by no means exhaustive.
+
+## Push ROS into the hardware nodes
+
+A couple of design decisions in ROS 1 make it difficult to integrate hardware devices (e.g. a sensor or actuator) natively into the ROS graph.
+The master is a central entity in the ROS graph and needs to be started before any of the nodes.
+Also the communication between the nodes and the master is done using XML-RPC which poses a significant dependency when being implemented on small resource constraint systems / micro controllers due to its recursive unbounded nature.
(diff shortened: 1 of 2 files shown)


articles/040_stories.md
```diff
@@ -0,0 +1,81 @@
...
+Original Author: {{ page.author }}
+
+The article enumerates a few stories which sketch what will be possible with ROS in the future.
+The list is by no means exhaustive.
+
+## Push ROS into the hardware nodes
+
+A couple of design decisions in ROS 1 make it difficult to integrate hardware devices (e.g. a sensor or actuator) natively into the ROS graph.
+The master is a central entity in the ROS graph and needs to be started before any of the nodes.
+Also the communication between the nodes and the master is done using XML-RPC which poses a significant dependency when being implemented on small resource constraint systems / micro controllers.
```
wjwwood: Might be worth working in the "why" there, which AFAIK is because XML is recursive (unbounded) and generally memory intensive.

articles/040_stories.md
```diff
@@ -0,0 +1,81 @@
...
+
+The advantage of this approach is twofold:
+
+- The integration effort to build a new system is being reduced since no custom drivers are required anymore.
+- By implementing a specific ROS-based interface the devices of specific class can easily be substituted without the need to spend time on integrating the software / hardware from e.g. a different vendor.
+
+## Delay decision on process layout to deploy time
+
+In ROS 1 nodes commonly use the `Node` API and implement their own `main` function.
+Only a few packages leverage the `Nodelet` API and compile their components into shared libraries instead.
```
wjwwood, line 44: Maybe it's obvious, but you could point out that whether or not you use the Nodelet API is a programming time decision, and that the Nodelet API was different enough from the Node API to make porting a pain.

wjwwood (approved):

wjwwood:
Also, perhaps this should be a section of the "Why ROS 2" article rather than its own?

articles/040_stories.md
dirk-thomas: Good point. I added the hint about the recursive unbounded nature.

articles/040_stories.md
dirk-thomas: I added an additional sentence to clarify that:

> The developer has to choose between these two different APIs and converting from using one to the other requires some non trivial effort.

dirk-thomas:
I see the "Why ROS 2" focused more on the question why to start with a new code base rather then extend existing ROS 1 code. I would prefer to keep these stories separate since they could be achieve in an extended version of ROS 1 too.

PR #104, opened 2016-12-13
update changes between ROS 1 and ROS 2 article

dirk-thomas:
Ready for review.

Please address cosmetic changes directly.

If you want to propose certain additions or changes please include a proposed text block which is copy-n-pastable.

### DIFF
--- a/articles/changes.md
+++ b/articles/changes.md
@@ -24,7 +24,9 @@ Original Author: {{ page.author }}
 Each change is described as briefly as possible but giving enough context and rationale for a reader familiar with ROS 1.
 If further external information is available (e.g. other articles) it should be linked to.
 
-Some of the described features are not yet available and are marked with ⏳.
+<div class="alert alert-info" markdown="1">
+  Some of the described features are not yet available and are marked with ⏳.
+</div>
 
 ## Platforms and dependencies
 
--- a/articles/changes.md
+++ b/articles/changes.md
@@ -53,7 +53,7 @@ ROS 2 requires at least Python version 3.5.
 ROS 1 uses a custom serialization format, a custom transport protocol as well as a custom central discovery mechanism.
 ROS 2 has an [abstract middleware interface](http://design.ros2.org/articles/ros_middleware_interface.html), through which serialization, transport, and discovery is being provided.
 Currently all implementations of this interface are based on the DDS standard.
-This enables ROS 2 to provide various [Quality of Service parameters](http://design.ros2.org/articles/qos.html) which improve communication over different networks.
+This enables ROS 2 to provide various [Quality of Service policies](http://design.ros2.org/articles/qos.html) which improve communication over different networks.
 
 ## Build system
 
--- a/articles/changes.md
+++ b/articles/changes.md
@@ -21,8 +21,10 @@ Original Author: {{ page.author }}
 
 ## Preface
 
-Each change is described as briefly as possible but giving enough context and rational for a reader familiar with ROS 1.
+Each change is described as briefly as possible but giving enough context and rationale for a reader familiar with ROS 1.
 If further external information is available (e.g. other articles) it should be linked to.
+
+Some of the described features are not yet available and are marked with ⏳.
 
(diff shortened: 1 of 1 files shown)


articles/changes.md
```diff
@@ -21,36 +21,39 @@ Original Author: {{ page.author }}
...
+
 ## Platforms and dependencies
 
 ### Platforms
 
 ROS 1 is only being CI tested on Ubuntu.
 It is actively supported by the community on other Linux flavors as well as OS X.
 
-ROS 2 will be CI tested and supported on Ubuntu, OS X as well as Windows (see [ci.ros2.org](http://ci.ros2.org/)).
+ROS 2 is being CI tested and supported on Ubuntu Xenial, OS X as well as Windows 10 (see [ci.ros2.org](http://ci.ros2.org/)).
```
mikaelarguedas: OSX 10.11.6, aka El Capitan

articles/changes.md
```diff
@@ -111,17 +116,23 @@ If the package follows common practice it might even be possible to detect some
...
+
+The generated Python code for messages and service is currently using the same module and class names in ROS 1 and ROS 2.
+Therefore they can not be imported both in a single application.
+This decision might be revisited if required.
+
 ### Optional default values in message definitions
 
 In ROS 2 primitive values in messages can now have default values, set when the message is constructed.
-Default values for non-primitive fields, i.e. nesting messages, is not yet possible.
+Default values for non-primitive fields, i.e. array of strings, nested messages, are not yet possible (⏳).
```
mikaelarguedas, line 137: not possible for standard strings either ATM

mikaelarguedas (commented):
Looks good to me. One or two minor comments though

articles/changes.md
dirk-thomas: I think you are wrong about this. A string field can have a default value: https://github.com/ros2/rosidl/blob/bf854d971f2afa95e194a795c0da4d57d9d8a683/rosidl_parser/rosidl_parser/__init__.py#L564

mikaelarguedas: oh my bad, taking that comment back then

articles/changes.md
dirk-thomas: Added: 4d30e7f3b770958fa8b1b7bbd5a8bceda2272006

articles/changes.md
```diff
@@ -133,7 +144,7 @@ This is useful for performance and for use cases like real-time.
 In ROS 1 the duration and time types are defined in the client libraries.
 The member names of the data structures are different in C++ (sec, nsec) and Python (secs, nsecs).
 
-In ROS 2 these types will be defined as messages and therefore be consistent across languages.
+In ROS 2 these types are defined as messages and therefore be consistent across languages.
```
mikaelarguedas: and therefore **are** consistent

dirk-thomas: 694e1481e5f707b3e874668868e8f4458b7ff0ca Thanks

articles/changes.md
```diff
@@ -21,36 +21,39 @@ Original Author: {{ page.author }}
 
 ## Preface
 
-Each change is described as briefly as possible but giving enough context and rational for a reader familiar with ROS 1.
+Each change is described as briefly as possible but giving enough context and rationale for a reader familiar with ROS 1.
 If further external information is available (e.g. other articles) it should be linked to.
 
+Some of the described features are not yet available and are marked with ⏳.
```
dhood: I would make this statement more noticeable (bold at the least) because towards the end of the article if you miss what the symbol represents, you will be mislead about the status of some features e.g. global parameter server

dirk-thomas: Wrapped the sentence in `alert-info`.

dhood (approved):
looks good. I would add a "last updated" note or similar so that people don't have to wonder what "currently" refers to

dirk-thomas:
@dhood Since none of the design article has that information I wouldn't attach it to this one. We should just make sure that all these documents stay "current".

PR #105, opened 2016-12-13
add information about mapping between ROS and DDS concepts

dirk-thomas:
Ready for review.

### DIFF
--- a/articles/060_ros_middleware_interface.md
+++ b/articles/060_ros_middleware_interface.md
@@ -369,3 +369,15 @@ Their package name starts with the prefix `rmw_` followed by the name of the use
 A user land executable using ROS nodes, publishers and subscribers must link against one specific middleware implementation library.
 The used middleware implementation library will only use the corresponding type support from each message package and ignore any additionally available type supports.
 Only middleware implementations can be selected for which the corresponding type support generators have been available when building the message packages.
+
+### Mapping between DDS and ROS concepts
+
+Every ROS node is one DDS participant.
+If multiple ROS nodes are being run in a single process they are still mapped to separate DDS participants.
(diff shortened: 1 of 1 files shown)


articles/060_ros_middleware_interface.md
```diff
@@ -369,3 +369,15 @@ Their package name starts with the prefix `rmw_` followed by the name of the use
...
 The used middleware implementation library will only use the corresponding type support from each message package and ignore any additionally available type supports.
 Only middleware implementations can be selected for which the corresponding type support generators have been available when building the message packages.
+
+### Mapping between DDS and ROS concepts
+
+Every ROS node is one DDS participant.
+If multiple ROS nodes are being run in a single process they are still mapped to separate DDS participants.
+If the containing process exposes its own ROS interface (e.g. to load nodes into the process at runtime) it is acting as a ROS node itself and is therefore also mapped to a separate DDS participant.
+
+The ROS publishers and subscribers are mapped to DDS publishers and subscribers.
```
dhood, line 379: We seem to say subscriptions instead of subscribers. But I am not 100% clear on the reasoning so it might not apply here.

wjwwood: The terms are used interchangeably. I pedantically defend my position that subscription is the correct term for what we have in ROS 2, but both ROS 1 and DDS call them subscribers. We've discussed changing them back to subscribers in ROS 2 for consistency, but no one has had the time to do it.

dirk-thomas: The high level concept is called publisher / subscriber. Therefore I think describing that these entities match is good to help people understand what in ROS equates to what in DDS. Even if the exact class name is different atm I don't think that should be reflected here.

Karsten1987: +1 for subscriptions

PR #106, opened 2016-12-14
services now use volatile QoS

dhood:
connects to https://github.com/ros2/rmw/pull/79

### DIFF
--- a/articles/qos.md
+++ b/articles/qos.md
@@ -70,7 +70,7 @@ Note: for each of the main bullets there is also the option of "system default",
 
 ROS 2.0 will provide QoS profiles based on the following use cases:
 
-- Default QoS settings for publishers and subscriptions ([rmw_qos_profile_default](https://github.com/ros2/rmw/blob/97008fdc646e3199c0a2f99b3307ee2ee807ede9/rmw/include/rmw/qos_profiles.h#L41-L47)).
+- Default QoS settings for publishers and subscriptions
 
   In order to make the transition from ROS1 to ROS2, exercising a similar network behavior is desirable.
   By default, publishers and subscriptions are reliable in ROS2.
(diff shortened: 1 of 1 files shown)


articles/qos.md
```diff
@@ -70,24 +70,26 @@ Note: for each of the main bullets there is also the option of "system default",
...
+- Sensor data
 
   For sensor data, in most cases it's more important to receive readings in a timely fashion, rather than ensuring that all of them arrive.
   That is, developers want the latest samples as soon as they are captured, at the expense of maybe losing some.
+  For that reason the sensor data profile uses best effort reliability and a smaller queue depth.
+
+- Parameters
+
+  Parameters are based on services, and as such have a similar profile.
+  The difference is that parameters use a much larger queue depth so that requests do not get lost when, for example, the parameter client is unable to reach the parameter service.
```
mikaelarguedas: I think we use `server` in that context for clarity because service has a double meaning (parameters are based on service clients and service servers)

dhood: I know what you mean, but things might be different for parameter services? https://github.com/ros2/tutorials/blob/1f64b078cf08a7933cfb6f94b80498146f426592/rclcpp_tutorials/src/parameters/parameter_events.cpp#L44

mikaelarguedas: I think it is the same for services https://github.com/ros2/tutorials/blob/1f64b078cf08a7933cfb6f94b80498146f426592/rclcpp_tutorials/src/services/add_two_ints_server.cpp#L40

dhood: ah I didn't realise! https://github.com/ros2/design/pull/106/commits/68e970b49b723d7bb6de1c16f82ad61f7c318f27

articles/qos.md
```diff
@@ -70,24 +70,26 @@ Note: for each of the main bullets there is also the option of "system default",
...
+- Sensor data
 
   For sensor data, in most cases it's more important to receive readings in a timely fashion, rather than ensuring that all of them arrive.
   That is, developers want the latest samples as soon as they are captured, at the expense of maybe losing some.
+  For that reason the sensor data profile uses best effort reliability and a smaller queue depth.
+
+- Parameters
+
+  Parameters are based on services, and as such have a similar profile.
+  The difference is that parameters use a much larger queue depth so that requests do not get lost when, for example, the parameter client is unable to reach the parameter service server.
```
mikaelarguedas, line 92: It looks like parameters keep all samples in the history cache: https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L59

mikaelarguedas:
What was the motivation for removing the links to the code where the profiles are defined ?

articles/qos.md
dhood: that's the profile for "parameter events" which watches all parameter-change events across the graph; parameters themselves are keep_last

mikaelarguedas: oh my bad! I didn't even see the "parameter" profile (that for some reason is declared between two pub/sub profiles)...
All good then, sorry for the noise

mikaelarguedas (approved):
Except question about removing code references, looks good to me

dhood:
> What was the motivation for removing the links to the code where the profiles are defined ?

just so that we don't have to maintain the links when the lines are moved

mikaelarguedas:
Yeah I assumed it was to not have to maintain them. Arguably keeping links would force us to keep the documentation up to date is we change the default qos profiles. I'm ok with both solutions, I was just wondering

PR #109, opened 2017-02-03
dds partitions for namespacing

Karsten1987:


Karsten1987:
up for review

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+At the same time, DDS Partitions don't support a partition hierarchy.
+If a publisher operates on two partition entries, e.g. `foo` and `bar` with a base name of `baz`, the resulting topic will be twofold:
+
+```
+- /foo/baz
+- /bar/baz
+```
+
+That implies, that DDS Partitions by default are not depicting any hierarchy.
+However, aliasing is easily douable with it.
```
tfoote: spelling: doable

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
 Additionally any URL related syntax, e.g. the `rostopic://` prefix, will be removed once parsed.
 Therefore only forward slashes have to be substituted when converting to DDS topic names.
 
-### Substitution of the Namespace Delimiter
+### DDS Partitions as namespace hierarchy
 
-The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
-Note that as fully qualified ROS 2 topic and service names are absolute, there is always a leading forward slash (`/`).
+Having looked closer to the various QoS settings inside the DDS protocol, paritions have been considered as the prefered method of implementing namespaces.
+DDS partitions allow up to 64 items in a string array, limiting the accumulated string size to 255 characters (TODO: Find spec ref).
```
tfoote: Dangling TODO

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+If a publisher operates on two partition entries, e.g. `foo` and `bar` with a base name of `baz`, the resulting topic will be twofold:
+
+```
+- /foo/baz
+- /bar/baz
+```
+
+That implies, that DDS Partitions by default are not depicting any hierarchy.
+However, aliasing is easily douable with it.
+That is, having an original topic such as `cmd_vel` in a namespace `robot1`, we can easily duplicate that topic as `robot2/cmd_vel`.
```
tfoote: How does this easy way of aliasing extend to aliasing including the basename? Are we expecting to expose this sort of aliasing to the user through the API?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+#### Remapping with DDS Partitions
+
+We have to differentiate remapping from aliasing at this point.
+As described before, aliasing implies a duplication a topic.
+Remapping, in contrast, alters an existing topic and thus remaps one topic into another.
+Given the nature of DDS Partitions, aliasing is natively supported by adding a second entry in the partition field.
+Remapping means, we have to change one existing partition field.
+However, as the completely hierarchy is illustrated within the first index of the partitions, the act of remapping is a simple string modification.
+If we want to remap a camera image from `camera1` to `camera2`, we have to modify the string in the respective field.
+Publishers or Subscribers are uneffected by this change and don't have to be shutdown and restarted.
```
tfoote: Is this true if we remap a basename?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+```
+warehouse/robot1/camera_left/image_raw
+```
+
+#### Remapping with DDS Partitions
+
+We have to differentiate remapping from aliasing at this point.
+As described before, aliasing implies a duplication a topic.
+Remapping, in contrast, alters an existing topic and thus remaps one topic into another.
+Given the nature of DDS Partitions, aliasing is natively supported by adding a second entry in the partition field.
```
dhood: this is only true if the basename is unchanged. e.g. aliasing `foo/bar` to `foo/baz` is not covered by partitions natively, we would need to create an additional publisher on topic `baz`, partition `foo`, right?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+
+- may contain substitutions which are delimited with balanced curly braces (`{}`)
+
+  - This is a more generic extension of the idea behind the tilde (`~`).
+
+- have length limits
+
+  - This is driven by the topic name length limit of DDS.
+
+- may be indicated as "hidden" by using a leading underscore (`_`) in one of the namespaces
```
tfoote: If we support leading underscores do we not need to restrict trailing underscores to avoid ambiguity with the `__` delimiter conversion? Might it not make sense to use a backwards compliant notation such as a trailing underscore to hide things? This seems like it is a change that will break a lot of things when turning on the bridge.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
 Additionally any URL related syntax, e.g. the `rostopic://` prefix, will be removed once parsed.
 Therefore only forward slashes have to be substituted when converting to DDS topic names.
 
-### Substitution of the Namespace Delimiter
+### DDS Partitions as namespace hierarchy
 
-The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
-Note that as fully qualified ROS 2 topic and service names are absolute, there is always a leading forward slash (`/`).
+Having looked closer to the various QoS settings inside the DDS protocol, paritions have been considered as the prefered method of implementing namespaces.
```
tfoote: This is temporally correct at the moment, but should be restated to be more timeless.

dhood: People will not have reference to the previous proposal of substituting `/` with `__` at this point in the document ("preferred over what?").  I suggest changing this to something that simply states that partitions will be used.

articles/140_topic_and_service_name_mapping.md
dhood: it is not clear at this point what the string array items represent, and perhaps not relevant yet that there is an array. Perhaps replace this sentence with something that states that DDS partitions represent an equivalent concept: datawriters/readers will only connect if both the topic name _and_ the partition match. The translation is: DDS topic name = ROS basename; DDS partition = ROS namespace.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+DDS partitions allow up to 64 items in a string array, limiting the accumulated string size to 255 characters (TODO: Find spec ref).
+The big benefit of handling the namespaces with partitions is that it generally stays ROS agnostic.
+This means if future implementations of the RMW interface (e.g. ZeroMQ etc.) want to introduce another, different namespace system, they are non-restricted.
```
dhood: I'm not sure that I understand this point, could you elaborate please? Is it that you are saying that the `/`->`__` substitution proposal meant that topics were "mangled" in RCL before they get to RMW, and using partitions alleviates the need for such mangling at the RCL level?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+
+#### DDS Partitions Characteristics
+
+As already said, DDS Partitions are implemented as a QoS setting.
+Hereby, the partitions field is implemented as an array of strings, where each string is an individual partitions.
+This means further that every partition index is stricly unique and independent from all other items.
+At the same time, DDS Partitions don't support a partition hierarchy.
+If a publisher operates on two partition entries, e.g. `foo` and `bar` with a base name of `baz`, the resulting topic will be twofold:
```
dhood: I think it is necessary to refer to DDS topics as conceptually being made up of a partition plus a topic, as opposed to referring to topic names as one concatenated string that encompasses the partition.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+That is, having an original topic such as `cmd_vel` in a namespace `robot1`, we can easily duplicate that topic as `robot2/cmd_vel`.
+
+#### Hierarchy with DDS Partitions
+
+Given that DDS Partitions are not hierarchical, we have to setup the hierarchy on top.
+For this, we use the forward slash `/` explicitely in each partition field.
+If we want to replicate a hierarchy of `warehouse`, `robot1`, `camera_left`, and finally the basename `image_raw`, the first entry of the partitions would look like:
+
+```
+warehouse/robot1/camera_left/image_raw
```
dhood: If I understand correctly, the partition would not include the basename. the basename would be the DDS 'topic'

articles/140_topic_and_service_name_mapping.md
dhood: I don't think we can make this claim yet because of the nuance that partitions are subscriber-based, not data-based. From https://community.rti.com/examples/partitions :
> If we start the subscriber late, it will receive all samples available, even those that were published while under incompatible partitions.

Because of that we may (hypothetically) decide to restart publishers to avoid this side effect. I don't think that the discussion on this is in the scope of this document, but I don't think we can state that publishers won't have to be restarted yet.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+### DDS Partitions as namespace hierarchy
 
-The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
-Note that as fully qualified ROS 2 topic and service names are absolute, there is always a leading forward slash (`/`).
+Having looked closer to the various QoS settings inside the DDS protocol, paritions have been considered as the prefered method of implementing namespaces.
+DDS partitions allow up to 64 items in a string array, limiting the accumulated string size to 255 characters (TODO: Find spec ref).
+The big benefit of handling the namespaces with partitions is that it generally stays ROS agnostic.
+This means if future implementations of the RMW interface (e.g. ZeroMQ etc.) want to introduce another, different namespace system, they are non-restricted.
+
+#### DDS Partitions Characteristics
```
dhood: I think this is where it would be good to mention character restrictions on DDS partitions (no regex special characters from what I read). I would also suggest moving the length limit to this section.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+
+  - This is done to avoid inconsistency with how `~foo` works in filesystem paths versus when used in a ROS name.
+
+- may contain substitutions which are delimited with balanced curly braces (`{}`)
+
+  - This is a more generic extension of the idea behind the tilde (`~`).
+
+- have length limits
+
+  - This is driven by the topic name length limit of DDS.
```
dhood: and the partition length limit

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +219,83 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+  - This is used to hide common but infrequently introspected topics and services.
+
+## Concerns/Alternatives
+
+This section lists concerns about the proposed design and alternatives that were considered.
 
 ### ROS Specific Name Prefix
 
+Initially, the idea of implementing ROS topics and namespaces based on specific name prefixes was chosen.
+This is not any longer the favorite solution as DDS partitions have proven to be more flexible and accurate since they are part of the protocol.
```
dhood: I don't know that this section (on using ROS name prefixes) is no longer part of the proposal. I think instead this section should be updated to propose that there all _partitions_ for ROS topics have a ROS-specific prefix (but please check with other team members on this as I am not certain on this)

articles/140_topic_and_service_name_mapping.md
```diff
@@ -265,7 +335,7 @@ Here are some examples of translations between ROS topic names and the correspon
 | `/_foo/_bar`      | Topic     | `rt___foo___bar`       |
 | `/_foo/_bar/_baz` | Topic     | `rt___foo___bar___baz` |
 
-### ROS Topic and Service Name Length Limit
+#### ROS Topic and Service Name Length Limit
```
dhood: This section still applies and shouldn't be under the "alternatives" section. It has to be re-worked to account for both the limit of partition length and topic name length

articles/140_topic_and_service_name_mapping.md
```diff
@@ -279,31 +349,10 @@ Services are governed by the same algorithm, but in some implementations may req
...
-In order to support a mapping to the more restrictive DDS topic name rules, these rules are in some ways more constraining than the rules for ROS 1.
-Other changes have been proposed for convenience or to remove a point of confusion that existed in ROS 1.
-In ROS 2, topic and service names differ from ROS 1 in that they:
-
-- must separate the tilde (`~`) from the rest of the name with a forward slash (`/`)
-
-  - This is done to avoid inconsistency with how `~foo` works in filesystem paths versus when used in a ROS name.
-
-- may contain substitutions which are delimited with balanced curly braces (`{}`)
+#### Substitution of the Namespace Delimiter
```
dhood: could you move this subsection higher up in the "alternatives" section so that the "ROS to DDS Name Conversion Examples" section has context

dhood (commented):
This is looking good but I don't think it quite captures all that we've spoken about offline yet. I've added some suggestions that will give more context to outside readers. I am happy to help with the modifications directly if others agree with the suggestions

articles/140_topic_and_service_name_mapping.md
Karsten1987: well, so I haven't found any official statement limiting the length of the partition.

articles/140_topic_and_service_name_mapping.md
Karsten1987: i left it within the alternative section, as there is no ros prefix IMO.

articles/140_topic_and_service_name_mapping.md
Karsten1987: we could still reserve the leading double underscore as "ros2 reserved" in analogy what the C++ standard says to types with leading underscore.

articles/140_topic_and_service_name_mapping.md
Karsten1987: IMO, there is no ROS name prefix anymore. But yes, up for discussion.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +201,94 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
-### Substitution of the Namespace Delimiter
+### DDS Partitions as namespace hierarchy
 
-The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
-Note that as fully qualified ROS 2 topic and service names are absolute, there is always a leading forward slash (`/`).
+We chose to implement the ROS namespace hierarchy with the help of DDS Partitions.
+DDS Paritions are part of the Subscription/Publisher QoS settings and have been considered as the prefered method of implementing namespaces.
+A subscriber and publisher only match when not only their basename topic matches, they also have to be specified on the same partiton.
+The big benefit of handling the namespaces with partitions is that it is an official QoS within the DDS specification.
+Furthermore, we can separate a ROS topic hierarchy from a DDS implementation, by forwarding the ROS topic in form of `/ns1/ns2/ns3/base_name` to the DDS implementation, which then takes take about setting up a hierarchy.
```
dirk-thomas: "takes take"? Not sure what it should say.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +201,94 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+All this stays generally ROS agnostic and open to alternative implementations of the RMW interface (e.g. ZeroMQ etc.), which may want to introduce another, different namespace system.
```
dirk-thomas: I am not sure if I would call this "generally ROS agnostic". Also "introduce another, different namespace system" sounds weird to me.

I would try to formulate it like this: the mapping from ROS names (using a forward slash to separate the namespace and the basename) to DDS partitions and topic names is an implementation details of the RMW implementation. While we use the same mapping for all DDS-based implementations other RMW implementation (e.g. Zero MQ) can decide on a different mapping depending on what the middleware provides.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +201,94 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+If a publisher operates on two partition entries, e.g. `foo` and `bar` with a base name of `baz`, the resulting topic will be twofold:
+
+```
+- /foo/baz
+- /bar/baz
+```
+
+That implies, that DDS Partitions by default are not depicting any hierarchy.
+On the other hand, it allows an easy aliasing or duplication of the specified topic, given that the basename topic does not change.
+A dynamic duplication of the original topic such as `cmd_vel` in a namespace `robot1` to `robot2/cmd_vel` can thus be easily achieved by expanding the DDS partition array.
```
dirk-thomas: This aliasing approach will only work if only the namespace is different. If the topic name needs to be different this will likely require multiple separate DDS publishers. Therefore I am not sure that we should describe this here. This is an implementation detail and I am not convinced that we should commit to this implementation option (since we will likely have to implement another option anyway).

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +201,94 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+warehouse/robot1/camera_left
+```
+
+#### Remapping with DDS Partitions
+
+We have to differentiate remapping from aliasing at this point.
+As described before, aliasing implies a duplication of a topic.
+Remapping, in contrast, alters an existing topic and thus remaps one topic into another.
+Given the nature of DDS Partitions, aliasing is natively supported by adding a second entry in the partition field.
+Remapping means, we have to change one existing partition field.
```
dirk-thomas: As mentioned before: that is not the case if the topic name should be changed. So I think the previous two lines are not accurate in that case.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,13 +201,94 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
...
+However, as the complete hierarchy is illustrated within the first index of the partitions, the act of remapping resolves into a string modification.
+If we want to remap a camera image from namespace `camera1` to namespace `camera2`, we have to modify the string in the respective field.
+Publishers or Subscribers are uneffected by this change and don't have to be shutdown and restarted.
```
dirk-thomas: I am not sure that we should make this guarantee. Depending on the rmw impl. this might be impossible to do without closing the existing publisher and creating a new one.

articles/140_topic_and_service_name_mapping.md
dirk-thomas: I agree with @Karsten1987's conclusion.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -246,13 +309,18 @@ Here is a non-exhaustive list of prefixes:
...
+#### Communicating with Non-ROS Topics
 
 Since all ROS topics are prefixed when being converted to DDS topic names, it makes it impossible to subscribe to existing DDS topics which do not follow the same naming pattern.
 For example, if an existing DDS program is publishing on the `image` topic (and is using the DDS equivalent to the ROS message type) then a ROS program could not subscribe to it because of the name mangling.
 Therefore to allow ROS programs to interoperate with "native" DDS topic names the API should provide a way to skip the ROS specific prefixing.
 
-### ROS to DDS Name Conversion Examples
+#### Substitution of the Namespace Delimiter
+
+The namespace delimiter in ROS 2 topic and service names, a forward slash (`/`), will be replaced with double underscores (`__`).
```
dirk-thomas: At the top of the document you removed the constraints about parts not being allowed to start / end / contain underscores. If the forward slashes here are still replaced with double underscores that constrain can't go away. Otherwise the resulting string would be ambiguous. Can the partition not contain forward slashed according to the standard? Or does it allow them?

articles/140_topic_and_service_name_mapping.md
dhood: wasn't the point of the ROS prefix to help tools that wanted to ignore non-ROS DDS topics? if so, how have things changed by using partitions?

articles/140_topic_and_service_name_mapping.md
Karsten1987: I updated the section to make clear that this replacement only has to be done when not using partitions.

articles/140_topic_and_service_name_mapping.md
wjwwood: > wasn't the point of the ROS prefix to help tools that wanted to ignore non-ROS DDS topics?

It was the rationale for it. Also, so that it is easy to answer the question "which of the DDS topics are also ROS topics?"

I don't know why @Karsten1987 arrived at that statement. I can only guess he didn't see a need for it or confused it's purpose as some mechanism specific for the substitution method proposed before.

I still think it has merit. If it is removed it should at least be covered by the alternatives (which currently cover why we *did* keep the prefix):

> This alternative would:
>
> - not prefix topics
> - optionally prefix other kinds of "implementation detail topics"
>
> Trade-offs:
>
> - it would be easier to have ROS subscribe to a DDS created topic
>
>   - e.g. DDS publisher on topic image could be subscribed to in ROS using just image
>   - however, the types would need to match as well
>   - in the current proposal the ROS topic image would become `rt__image`, so DDS topics would need to follow the ROS topic conversion scheme to interoperate with ROS components
>   - it would be hard to distinguish ROS created DDS topics and normal DDS topics
>
> - services would still need to be differentiated
>
>   - e.g. service `/foo` would need to make two topics, something like `foo_Request` and `foo_Reply`
>
> Rationale:
>
> Slight preference was given to easily categorizing ROS created topics with DDS created topics over easily connecting to existing DDS topics. Connecting to DDS topics could be achieved by having an option when subscribing or publishing to "alias" to an implementation topic name, e.g. something like `sub->alias_to_explicit_topic('dds_topic')`. Also, the workaround for separating ROS created topics from other DDS topics was considered to be more complicated than the suggested solution of allowing users to specify specific DDS topic names for their publishers and subscriptions.
>
> -- https://github.com/ros2/design/blob/e31ca2bacfff154f1cf034effe3d07b271fbcd94/articles/140_topic_and_service_name_mapping.md#limited-prefixes-alternative

This section hasn't been changed in this pr (yet): https://github.com/ros2/design/pull/109/files#diff-51ea0dd0254ef288483a09be8b28379dR403

> if so, how have things changed by using partitions?

As far as I can tell the partitions don't have anything to do with the decision to implicitly namespace ROS topics so they stand apart from other DDS topics. I think they could be made independently. The only thing I can think of is that one of the arguments against the prefix (and why it was so short) was that it cut into the 255 topic length. The partitions alleviate that slightly since the combine namespace + topic name limit is now longer.

articles/140_topic_and_service_name_mapping.md
wjwwood: > i left it within the alternative section, as there is no ros prefix IMO.

I agree with @dhood, there is a topic length restriction and it's based on the partition length limit and the topic name limit in DDS. This doesn't have anything to do with "ros prefix" as I understand it.

articles/140_topic_and_service_name_mapping.md
wjwwood: There are actually two alternatives sections which talk about this issue. The second one:

> This alternative is the same as the "Suffix Alternative" except:
>
> - Topics would not have a suffix or prefix at all
>
> Trade-offs:
>
> - same trade-offs as the "Suffix Alternative"
>
> - but also easier to have ROS subscribe to a DDS created topic
>
>   - e.g. DDS publisher on topic image could be subscribed to in ROS using just image
>   - the types would need to match
>   - in the current proposal the ROS topic image would become rt__image, so DDS topics would need to follow our naming scheme to interoperate with ROS components
>
> Rationale:
>
> While this alternative provided the benefits of both the suffix and limited prefix alternatives, the rationale for the limited prefix alternative still applies here.
>
> -- https://github.com/ros2/design/blob/e31ca2bacfff154f1cf034effe3d07b271fbcd94/articles/140_topic_and_service_name_mapping.md#limited-suffix-alternative

wjwwood: Except for suffixes rather than prefixes.

articles/140_topic_and_service_name_mapping.md
Karsten1987: well, given the reasoning which comes after it has something to do with the ros prefix.

articles/140_topic_and_service_name_mapping.md
Karsten1987: I may have misunderstood something here. Why I removed that section was because of not prefixing topics with `rt` and `rx` etc. 
But I can leave it in if you wish.

articles/140_topic_and_service_name_mapping.md
wjwwood: The limit was effected by the fact that there was a prefix, but even if we drop the prefix, there is a length limit. It doesn't go away or stay based on the ros prefix.

articles/140_topic_and_service_name_mapping.md
wjwwood: Using `rt__` was the "how", not the "why". We could still choose to have a prefix, or not, but we would just achieve it differently with the partitions. For example, you might implicitly join `/ros` to the beginning of all partitions, rather than implicitly starting them with `/`. That's just a suggestion. But you should either capture this section as an alternative (if we decide to not have the implicit prefix) or you should keep it in the proposal and update it so it would work with partitions.

articles/140_topic_and_service_name_mapping.md
Karsten1987: the length limit is described at the very beginning of the article, where the complete ros topic cannot exceed 255 characters. Am I confusing things here?

wjwwood: Talking with you offline, there is still a limit of 255, which is the partition name + the topic name in DDS. So this section gets simpler, but it still deserves to be put into the proposal (i.e. not just in the alternatives section). It is also affected by the "ros prefix", because that also eats into the topic length limit.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +221,72 @@ Here is a non-exhaustive list of prefixes:
...
-While all planned prefixes consist of two characters, i.e. `rX`, anything proceeding the first namespace separator, i.e. `__`, can be considered part of the prefix.
+While all planned prefixes consist of two characters, i.e. `rX`, anything proceeding the first namespace separator, i.e. `/`, can be considered part of the prefix.
 The standard reserves the right to use up to 8 characters for the prefix in case additional prefix space is needed in the future.
 
-### Communicating with Non-ROS Topics
+#### Communicating with Non-ROS Topics
 
 Since all ROS topics are prefixed when being converted to DDS topic names, it makes it impossible to subscribe to existing DDS topics which do not follow the same naming pattern.
 For example, if an existing DDS program is publishing on the `image` topic (and is using the DDS equivalent to the ROS message type) then a ROS program could not subscribe to it because of the name mangling.
 Therefore to allow ROS programs to interoperate with "native" DDS topic names the API should provide a way to skip the ROS specific prefixing.
```
dirk-thomas: Do we agree that this is a use case we want to support? If that is the case I would propose to clarify here that when any API is being used in this "native mode" (e.g. command line tools, all the API to create publishers, subscribers, etc.) it should still be possible to optionally ignore vendor specific topics as well as distinguish the type topics (is it a pub/sub topic or used for req/res).

articles/140_topic_and_service_name_mapping.md
```diff
@@ -50,7 +50,7 @@ TOPICNAME - A topic name is an identifier for a topic, and is defined as any ser
 
 *Note:* that the DDS specification has a known typo, where it says `-` are allowed, but the RTI documentation correctly lists `_` as allowed.
 
-Additionally, DDS has a hard limit on topic names of 256 characters, so an additional goal is to minimize the number of extra characters used when mapping from ROS to DDS names.
+Additionally, DDS has a hard limit on topic names of 255 characters, so an additional goal is to minimize the number of extra characters used when mapping from ROS to DDS names.
```
wjwwood: Can you add a citation for this? Especially if I got it wrong before, it would be to good to back it up with a source.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -76,14 +73,15 @@ The content of substitutions:
 
 - must not be empty
 
+- must be not longer than 256 characters
```
wjwwood: Why? The content of the substitution does not go into the final topic name. Was this added due to some discussion already?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -76,14 +73,15 @@ The content of substitutions:
...
+
 - may contain alphanumeric characters (`[0-9|a-z|A-Z]`) and underscores (`_`)
 
 - must not start with a numeric character (`[0-9]`)
 
 - are unlike topic and service names in that they:
 
   - may start with or end with underscores (`_`)
-  - may contain repeated underscores, e.g. `__`
```
wjwwood: Why was this removed? (again we may have already discussed this, but if we did I don't remember)

articles/140_topic_and_service_name_mapping.md
```diff
@@ -96,7 +94,6 @@ Fully qualified names have these additional restrictions:
 - must not contain tilde (`~`) or curly braces (`{}`)
 
 Note that expanded substitutions must result in a valid name.
-For example, a name `foo_{bar}` is not valid if `bar` is expanded as `_baz`, because that would result in a fully qualified name of `foo__baz` which contains repeated underscores (`_`).
```
wjwwood: This should be reworded to have another example, rather than to be removed completely. The purpose of this line was to point out how a substitution could result in an invalid name, not to reiterate the `__` rule (which no longer applies).

articles/140_topic_and_service_name_mapping.md
```diff
@@ -116,11 +113,9 @@ For example, these are valid names:
...
 But these are not valid names:
 
-| `123abc`    | `123`  | `__foo`  | `foo__bar` | `foo bar`   |
-| `foo__`     | ` `    | `foo_`   | `foo//bar` | `foo/`      |
-| `foo_/bar`  | `~foo` | `foo~`   | `foo~/bar` | `foo/~bar`  |
-| `/_/bar`    | `_`    | `_/_bar` | `/~`       | `foo/~/bar` |
-| `foo/_/bar` |        |          |            |             |
+| `123abc`    | `123`  | `foo bar`| ` `        | `foo//bar` | `foo/` |
+| `foo_/bar`  | `~foo` | `foo~`   | `foo~/bar` | `foo/~bar` | `/~`   |
+| `foo/~/bar` |
```
wjwwood: Please check that this renders correctly, as the other tables finish out the last line with empty cells (see the one just above this).

articles/140_topic_and_service_name_mapping.md
```diff
@@ -219,15 +202,10 @@ Since ROS 2 topic and service names are expanded to fully qualified names, any b
 Additionally any URL related syntax, e.g. the `rostopic://` prefix, will be removed once parsed.
 Therefore only forward slashes have to be substituted when converting to DDS topic names.
```
wjwwood: This is no longer the case in the current proposal (the forward slashes are kept in the partition name, AFAIU). Unless you're thinking about the forward slash between the namespace (kept in the DDS partition) and the basename (kept in the DDS topic name) which is redundant. If so you should update this sentence to clarify that.

articles/140_topic_and_service_name_mapping.md
wjwwood: It's probably also good to point out that DDS puts a length limit on both topic names and partition names.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +221,72 @@ Here is a non-exhaustive list of prefixes:
 | ROS Parameter        | rp     |
 | ROS Action           | ra     |
 
-While all planned prefixes consist of two characters, i.e. `rX`, anything proceeding the first namespace separator, i.e. `__`, can be considered part of the prefix.
+While all planned prefixes consist of two characters, i.e. `rX`, anything proceeding the first namespace separator, i.e. `/`, can be considered part of the prefix.
```
wjwwood, line 255: This needs to be reworded, "anything proceeding the first namespace separator" no longer makes sense because all the partitions appear to start with `/` and contain nothing before it, even the prefix described here.

articles/140_topic_and_service_name_mapping.md
wjwwood: > Do we agree that this is a use case we want to support?

I think it's worth supporting, but just supporting matching of the topic name isn't really sufficient. To make this option useful, you need to also address the fact that the topic types will probably not match either. So in addition to telling our API not to use the ROS specific prefix in the partition or topic name, you'll have to have some custom type support which can take the DDS image type and convert it to some ROS message type on the fly. At that point, I begin to question why you're using a ROS subscriber or publisher at all. Perhaps it would be better to just use DDS directly to get this info and republish it (or some derived information) to ROS when you're done using it. I support this option (to leave off the ROS prefix) so there is a path to this use case working, but I'm not completely convinced it is practical.

> If that is the case I would propose to clarify here that when any API is being used in this "native mode" (e.g. command line tools, all the API to create publishers, subscribers, etc.) it should still be possible to optionally ignore vendor specific topics as well as distinguish the type topics (is it a pub/sub topic or used for req/res).

I don't fully understand what you're asking for here, can you reword this? Or maybe suggest an edit so it is clearer what you mean?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +221,72 @@ Here is a non-exhaustive list of prefixes:
...
 
-### Communicating with Non-ROS Topics
+#### Communicating with Non-ROS Topics
 
 Since all ROS topics are prefixed when being converted to DDS topic names, it makes it impossible to subscribe to existing DDS topics which do not follow the same naming pattern.
 For example, if an existing DDS program is publishing on the `image` topic (and is using the DDS equivalent to the ROS message type) then a ROS program could not subscribe to it because of the name mangling.
 Therefore to allow ROS programs to interoperate with "native" DDS topic names the API should provide a way to skip the ROS specific prefixing.
 
-### ROS to DDS Name Conversion Examples
+### DDS Partitions as namespace hierarchy
```
wjwwood: Throughout this you use "namespace hierarchy", this might be a of a nitpick but I see this as a redundant term. From my experience, multi-element namespaces are hierarchical by design. So to me it would be clearer to just say "namespace" instead. Also, the notion of splitting namespaces into parts with the `/` is already covered in sections above, so the hierarchy within the namespace is an established concept already.

For this title I'd reword it to something like "ROS Namespaces as DDS Partitions" (think of it as, "we already have ROS namespaces which are described above, now we'll explain how we'll implement them with DDS partitions"). Also, I think this section should go before the "ROS Specific Name Prefix" subsection and the "Communicating with Non-ROS Topics" subsection, basically it should be the first subsection of the "Mapping of ROS 2 Topic and Service Names to DDS Topics" section. Because I think it is probably confusing to talk about the ROS prefix and other special cases before explaining that we will use DDS partitions and what the partitions are. (so also move the "DDS Partitions Characteristics" subsection up there as well)

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +221,72 @@ Here is a non-exhaustive list of prefixes:
...
 
-Here are some examples of translations between ROS topic names and the corresponding DDS topic names:
+We chose to implement the ROS namespace hierarchy with the help of DDS Partitions.
+DDS Paritions are part of the Subscription/Publisher QoS settings and have been considered as the prefered method of implementing namespaces.
+A subscriber and publisher not only match when their basename topic matches, they also have to be specified on the same partition.
+The big benefit of handling the namespaces with partitions is that it is an official QoS within the DDS specification.
+Furthermore, we can separate a ROS topic hierarchy from a DDS implementation, by forwarding the ROS topic in form of `/ns1/ns2/ns3/base_name` to the DDS implementation, which then takes care about setting up a hierarchy.
+We can therefore make the translation from a ROS Topic (namespaces and basename separated by forward slashes) to a DDS Topic (topic plus paritions) an implementation detail of each RMW implementation.
+This further upholds the flexibility of each alternative to DDS (e.g. ZeroMQ) to realize a different mapping depending on what the middleware provides.
```
wjwwood: This paragraph is a bit confusing to me, it might just be the wording. Would you mind if I open a pr against this with a different wording?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +221,72 @@ Here is a non-exhaustive list of prefixes:
...
-Note that this algorithm must be applied on a fully qualified name, i.e. after expanding all substitutions and the private namespace substitution character (`~`), after removing any URL related syntax (e.g. without the `rostopic://` prefix).
+### ROS Topic and Service Name Length Limit
 
-Services are governed by the same algorithm, but in some implementations may require additional characters to be subtracted from the limit depending on how the request and response topics are created by the middleware.
-In the specific case of RTI Connext's Request-Reply implementation, they append the `Request` and `Reply` strings to the topic names.
-Therefore, it would be safest to assume the Service name limit to be less 8 more characters.
+The length of the DDS topic must not exceed 255 characters.
+The actual length of a partition field may be limited to 255 characters, however this varies drastically depending on the vendor.
+RTI Connext does not allow a creation of a publisher/subscription with a partition length of 248 characters, whereas FastRTPS does not have any limitation in length.
```
wjwwood: Why is it 248? Can you add a citation to this as well?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -305,6 +314,24 @@ In ROS 2, topic and service names differ from ROS 1 in that they:
 
 This section lists concerns about the proposed design and alternatives that were considered.
 
+#### Substitution of the Namespace Delimiter
```
wjwwood: This subsection should be `###` as the one before it is `##`. (the subsections of this subsection are also affected)

wjwwood: Also, this needs a section preceding it to describe the context in which it was rejected. Just moving it here is insufficient in my opinion...

wjwwood (changes requested):
If you don't mind, I'll open a pull request against this pull request to address some of the comments I made.

Karsten1987:
please go ahead and push directly on my branch for any changes to wording or language.

wjwwood (approved):
This lgtm me, I think it's ready for final (re)review from people who already commented on things and comments from community.

Thanks for working on this @Karsten1987.

articles/140_topic_and_service_name_mapping.md
wjwwood: Never mind on this one, makes sense to me now.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -211,23 +194,51 @@ For example, given the name `{private}foo`, if there is a substitution called `{
...
+You can read more about partitions in RTI's documentation:
 
-### ROS Specific Name Prefix
+- [PARTITION_QosPolicy](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/PARTITION_QosPolicy.htm)
 
-In order to differentiate ROS topics easily, all DDS topic names created by ROS shall be prefixed with `rX`, where `X` is a single character that indicates to which subsystem of ROS the topic belongs.
-For example, a plain topic called `/foo` would translate to a DDS topic named `rt__foo`, which is the result of concatenating the prefix `rt` for being a ROS topic, with `__` for the leading `/`, and the topic name `foo`.
+The array of strings have no hierarchy and the order does not matter.
+Each entry in the partition array is directly combined with the DDS topic and they are not sequentially combined.
+If a publisher operates on two partition entries, e.g. `foo` and `bar` with a base name of `baz`, this would result in the moral equivalent of two topics: `/foo/baz` and `/bar/baz`
```
gbiggs: Moral equivalent?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -211,23 +194,51 @@ For example, given the name `{private}foo`, if there is a substitution called `{
...
+
+#### Hierarchy with DDS Partitions
+
+Since DDS partitions have no hierarchy or order to the array of strings, this proposal would use only one of the strings in the array to hold the entire ROS name's namespace.
+Therefore we preserve the forward slashes (`/`) that exist in the ROS name's namespace already.
+
+### ROS Specific Namespace Prefix
+
+In order to differentiate ROS topics easily, all DDS topics created by ROS shall be automatically prefixed with a namespace like `/rX`, where `X` is a single character that indicates to which subsystem of ROS the topic belongs.
+For example, a plain topic called `/foo` would translate to a DDS partition `/rt/` and a DDS topic `foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name `foo`.
```
gbiggs: Is there value in having the leading and trailing slashes on the partition? Particularly as the use of partitions does not appear to get us unlimited name lengths in some DDS implementations (e.g. RTI).
If we know that a partition and topic will always be joined by a slash, then the trailing slash does not need to be stored.
If ROS partitions always begin with a known prefix then the leading slash could also be dropped, although there is a possibility of (likely very rare) cases where an existing system uses the same partition.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +254,76 @@ Here is a non-exhaustive list of prefixes:
...
+| `/robot1/camera_left/image_raw` | `image_raw` | [`/rt/robot1/camera_left/`] |
+
+### Changing ROS Names During Runtime
+
+If the ROS name needs to be changed during runtime (after the publisher or subscriber has already been created and is in use), then some details about DDS partitions might be useful to consider.
+
+Since this proposal splits the ROS name into namespace and base name, we'll consider how that might affect DDS partitions and DDS topic names respectively.
+When changing the namespace of a ROS name, i.e. everything but the base name, then this could be done by changing just the existing partition string.
+As partitions are part of QoS settings, it technically doesn't require a full restart of the publisher and subscription when changing the partitions field, however we cannot change the DDS topic name on the fly.
+With this in mind, we may want to always require a full restart of the publisher or subscription instance to avoid inconsistency in how these changes behave from the ROS user's perspective.
```
gbiggs, line 275: Since this is something that should be handled behind the API where the user can't see it, I think it is worth aiming for partition changes to be handled without a restart of the publisher or subscriber. Restarting for topic name changes should be done where the user can't see it, and the API call documentation should make it clear that changing any part of the topic name is not a fixed-time operation, not real-time safe, etc.

gbiggs (commented):
Only three small comments, so in general it looks great.

articles/140_topic_and_service_name_mapping.md
wjwwood: A phrase I've shamelessly adopted from @gerkey. It just means "a la" or "in the spirit of". I'll just reword it to avoid any turn of phrase. 401d5bb

articles/140_topic_and_service_name_mapping.md
wjwwood: I left both slashes in the example as that is the easiest way to implement it, just split the string in the right place and put the first half in the partition and the second in the topic name. But you're right that both the leading and trailing forward slashes could be left off without losing any information.

Leaving off the trailing slash is a no brainer. I don't see why we couldn't do that.

Leaving off the leading slash leaves also seems doable and it helps to address the issue of "native" topic names which I describe else where in the article.

I'd like to hear input from others before committing to it, but both seem like obvious improvements to me.

articles/140_topic_and_service_name_mapping.md
wjwwood: Yeah, I left this note, or implementation hint, in because @Karsten1987 went to the trouble to test it out and I didn't want to lose the information. But I agree that the API should make it clear that you should just assume that the publisher or subscriber may need to be recreated, i.e. we shouldn't try to make any guarantee about the possible shortcut that can be taken when the basename is unchanged.

articles/140_topic_and_service_name_mapping.md
wjwwood: @ros2/team If some of you could comment on this idea soon I'd appreciate it. I think it should be addressed before merging this, and I'd like to avoid holding this up any longer.

dirk-thomas: I agree that the DDS partition should neither have a leading or trailing slash.

wjwwood: Ok, I'll draft a change to reflect that today.

wjwwood: bfc7ba2

wjwwood:
@Karsten1987 @gbiggs or @dirk-thomas can one of you review my last change bfc7ba2?

Once that's been proofread, then I think we should merge this. It's been available for comment for quite some time now.

Karsten1987:
lgtm

dhood:
There was still an open question on this line https://github.com/ros2/design/pull/109/files#diff-51ea0dd0254ef288483a09be8b28379dR76 that I think is work addressing

I opened https://github.com/ros2/design/pull/114 to remove some references to the double-underscore restriction that I believe are no longer appropriate

dirk-thomas:
lgtm.

articles/140_topic_and_service_name_mapping.md
Karsten1987: That is only partially correct. The size of the RTPS topic name is limited to 256 characters and thus the ROS base name is limited to 256.
I'll update that.

articles/140_topic_and_service_name_mapping.md
Karsten1987: how can it result in an invalid name if we don't do the underscore substitution anymore?

wjwwood: There are other things that aren't allowed, for example `/foo/{sub}/bar` would be invalid if `{sub}` were replaced with an empty string resulting int `/foo//bar`. Or if `{sub}` were replaced with `invalid!characters` which would result in `/foo/invalid!characters/bar`.

articles/140_topic_and_service_name_mapping.md
wjwwood: To be clear this bullet is talking about "The content of substitutions", which is the stuff inside the `{}` (which is to say not what they get replaced with), i.e. if you have `{foo}` which expands to `bar`, we're talking about the rules for what `foo` can contain.

Based on that, I don't think there's an issue with `/foo/{somethingreallylong...[> 256 characters]...junkjunkjunk}/bar` so long as that substitution is replaced with something that doesn't make the resulting topic invalid. I don't know why you'd do this, but I also don't see why it is a restriction we have to put in place.

articles/140_topic_and_service_name_mapping.md
wjwwood: This is no longer an issue either.

articles/140_topic_and_service_name_mapping.md
wjwwood: This change still doesn't look right to me:

https://github.com/ros2/design/commit/bb4b2f4#diff-51ea0dd0254ef288483a09be8b28379dR77

wjwwood (approved):

wjwwood:
lgtm, merge whenever you're ready! 🎉

articles/140_topic_and_service_name_mapping.md
Karsten1987: Do you want me to remove that sentence or rather explain that we truncate the leading and trailing forward slashes ?

wjwwood: I think it's fine as-is, because a DDS observer would see `['rt/my_topic/is/'] + cool` (`['partition1', ...] + topic_name`), and so "anything proceeding the first namespace separator" would just be `rt`. However, if you want to explain why this is ok, i.e. because of the truncating there is no leading `/`, then that's fine too.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -243,41 +252,71 @@ Here is a non-exhaustive list of prefixes:
...
 
-### ROS Topic and Service Name Length Limit
+For example with an option in the API:
 
-Because DDS topic names must be limited to 255 characters, the length of a ROS topic or service name is also limited in length.
-In the case of a topic, the length is governed by the following algorithm:
+| ROS Name              | ROS Prefix | DDS Topic   | DDS Partition               |
+|-----------------------|------------|-------------|-----------------------------|
+| `rostopic://image`    | with       | `image`     | [`rt`]                      |
+| `rostopic://image`    | without    | `image`     | [``]                        |
```
dirk-thomas, line 304: Should this be [] instead of ['']?

wjwwood: It is currently ```[``]```, which is the same as your suggestion I think.

dirk-thomas: Oh, I see. These are actual backticks which are only used for formatting. Not actually quotes. Thanks for clarifying.

ros-discourse:

https://discourse.ros.org/t/rfc-using-dds-partitions-to-implement-ros-2-namespaces/1390/1

PR #110, opened 2017-02-03
update string value constraints

mikaelarguedas:
based on [the recent discussion](https://github.com/ros2/rosidl/pull/204#discussion_r99045799) here is a proposal for requiring users to escape quotes when defining strings in IDL

### DIFF
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -209,6 +209,14 @@ Depending on the type the following values are valid:
 - `string`:
 
   - a string value which can optionally be quoted with either single quotes (`'`) or double quotes (`"`)
+  - a double-quoted (`"`) string (respectively single-quoted (`'`)) should have any inner double quotes (respectively single quotes) escaped. Example:
+
+    - `string my_string "I heard \"Hello\""` is valid
+    - `string my_string "I heard "Hello""` is **not** valid
+    - `string my_string "I heard 'Hello'"` is valid
+    - `string my_string 'I heard \'Hello\''` is valid
+    - `string my_string 'I heard 'Hello''` is **not** valid
+    - `string my_string 'I heard "Hello"'` is valid
 
 ### Service file format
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This looks good to me. It would be good to have a corresponding PR to update the parser to match the modified rules / behavior.

mikaelarguedas:
Sounds good, putting this `in progress` until corresponding parser PR is ready

mikaelarguedas:
Corresponding parser PR: https://github.com/ros2/rosidl/issues/205

serge-nikulin:
@mikaelarguedas, do you have a design document for this one? If not, can you create one, please?

Thanks!

CC: @dejanpan

mikaelarguedas:
@serge-nikulin  I do not have a design doc for this ATM.

Can you confirm the scope of the design document you are looking for ? (I think you mean option 1 below) 

1. What is allowed in strings defined in ROS IDL files? 
1. What is allowed in strings in the language specific code ?

wjwwood:
@mikaelarguedas I'll sync with you about this when we finish the release.

mikaelarguedas:
This design doc change (and [corresponding code change](https://github.com/ros2/rosidl/pull/205)) focused on providing rules for escaping quotes in externally quoted (starting and ending with `'` or `"`) strings defined in ROSIDL files (.msg/.srv).

The rules are:
- If a string is double quoted, any double quote inside the string must be escaped
- If a string is single quoted, any single quote inside the string must be escaped
- If a string is double quoted and contains single quotes, they inner quotes don't need to be escaped
- If a string is single quoted and contains double quotes, the inner quotes don't need to be escaped

Rationale:
- Consistency in string declaration is crucial.
- Escaping should be language agnostic and clear at string declaration time.

Consequence:
- All language specific generators should do the extra work to make sure the resulting string is represented correctly in the specific language

Future work:
- Decide and implement how escaping should work for non-quoted strings.
  - the discussion lied toward "no escaping should be necessary, any escaped quote should result in an escaped backslash in the resulting string".
  - this has not been ratified and added to the design document (neither has been implemented)

@serge-nikulin FYI

PR #111, opened 2017-02-03
add rules for array default values

mikaelarguedas:
this adds rules defining default values for arrays in IDL.
It also specifies that string arrays elements must be quoted

related to https://github.com/ros2/rosidl/issues/206

### DIFF
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -111,9 +111,11 @@ A field of type `array` can optionally specify a default value.
 - default values for an array must start with an opening square bracket (`[`) and end with a closing square bracket (`]`)
 - each value within the array must be separated with a comma (`,`)
 
-Additional rule for string arrays:
+Additional rule for `string` arrays:
+- string arrays must contain only `string`s respecting the following rules:
+  - a string value which can optionally be quoted with either single quotes (`'`) or double quotes (`"`)
 
-- every element of a string array must be quoted using either single quote (`'`) or double quote (`"`)
+  - a double-quoted (`"`) string (respectively single-quoted (`'`)) should have any inner double quotes (respectively single quotes) escaped
 
 <div class="alert alert-warning" markdown="1">
   <b>TODO:</b> default values are currently not supported for <i>complex</i> fields
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -103,6 +103,8 @@ If no default value is specified a common default value is used:
 - for `bool` it is `false`
 - for *numeric* types it is a `0` value
 - for `string` it is an *empty* string
+- for *static size arrays* it is an array of N elements with its fields zero-initialized
+- for *bounded size arrays* and *dynamic size arrays* it is an empty array `[]`
 
 #### Array default values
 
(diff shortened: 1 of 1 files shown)


articles/110_interface_definition.md
```diff
@@ -104,8 +104,19 @@ If no default value is specified a common default value is used:
 - for *numeric* types it is a `0` value
 - for `string` it is an *empty* string
 
+#### Array default values
+
+A field of type `array` can optionally specify a default value.
+
+- default values for an array must start with an opening square bracket (`[`) and end with a closing square bracket (`]`)
+- each value within the array must be separated with a comma (`,`)
```
wjwwood, line 114: May elements end with a `,` even if no other element comes after it?

The difference between:

```
['a', 'b', 'c']
```

and:

```
['a', 'b', 'c',]
```

wjwwood: Also, it's obvious to me, but it might be worth mentioning that the values in the array must all be of the same type, i.e. the array is homogeneous.

wjwwood (dismissed):
lgtm, other than some small comments.

mikaelarguedas:
Thanks @wjwwood for reviewing
Putting this back in ready until comments addressed and corresponding PR on the parser is ready to be reviewed

mikaelarguedas:
@ros2/team looking for feedback on this before wrapping up the implementation. Particularly on @wjwwood's question https://github.com/ros2/design/pull/111#discussion_r99690955: How do we want to treat unnecessary commas?

Option 1: starting or ending a sequence with a comma is illegal. Having two consecutive commas in a sequence (without an element in between) is illegal.
Option 2: leading or trailing commas in a sequence will be ignored.  Having two consecutive commas in a sequence (without an element in between) is illegal.
Option 3: leading or trailing commas in a sequence will be ignored.  Two consecutive commas in a sequence will be ignored as well.

tfoote:
Trailing commas are valuable if you can wrap and might expect to extend a list element by element per line. However I don't think we support line wrapping at the moment as as such, I would advocate for being strict with option 1. Double comma's I would definitely reject. That would be a null/None type in the middle of the sequence, unless we want to insert the default value of that datatype there. But that seems dangerous and would get into discussions whether a trailing or leading comma is implicitly a None appended to the beginning or end of the list.

nuclearsandwich:
Trailing commas are really valuable for reducing line-by-line diff noise. I don't think there's a similar argument to be made in favor of leading or internal spurious commas.

The absence of trailing commas is one of the more unfortunate aspects of json. If there is ever an intent to support hard-wrapped lines within array items, I think it is worth allowing the trailing comma from the outset.

clalancette:
Agreed with the other commenters here.  I'd allow trailing commas, but reject double commas.

wjwwood:
+1 for no leading comma and no repeated commas (even with whitespace between, i.e. `, ,` is invalid).

+0 for trailing comma as it is not useful unless we have multiline support, and we could add support for trailing comma when we add support for that too.

dirk-thomas:
Currently we parse the default value on our own. In the future when we allow nested initialization etc. we might want to use an existing format (e.g. yaml). At that point we have to follow the specification of that file format. Therefore I think we should consider those in this decision rather than "defining our own".

mikaelarguedas:
> Currently we parse the default value on our own. In the future when we allow nested initialization etc. we might want to use an existing format (e.g. yaml). At that point we have to follow the specification of that file format. 

YAML seems to be matching the feedback we got on that thread so far:
- Option 4 / YAML: 
  - No leading comma and no repeated commas (even with whitespace between, i.e. `, ,` is invalid)
  - Whitespaces around separators are ignored (striped)
  - Single trailing comma is allowed and is ignored (striped)

If we're planning on focusing on the ROS IDL file format next cycle I'm happy to delay this a bit more. But if its uncertain I'd prefer getting https://github.com/ros2/rosidl/pull/197 (slightly modified) merged earlier rather than later and iterate on it when we get a final decision on the file format.

mikaelarguedas:
@wjwwood @dirk-thomas Can you please have another look at this when you get a chance?
This has been updated based on discussion and the implementation in https://github.com/ros2/rosidl/pull/197 has been updated accordingly

dirk-thomas (approved):
LGTM - matches the implementation.

PR #112, opened 2017-02-03
Draft of document for remapping

sloretz:
PR to get feedback. 

I'm trying to define static remapping, but I'm finding I need to define remapping first. If I've understood correctly, this should be the same as what was discussed yesterday with a couple changes. 

I removed "Literal name token" replacement because appears to be everything else, just done prior to names being expanded to FQN. I tried to expand "Namespace Replacement" to be aware of the position of a namespace.


articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+ - Replace any substring with another string
+ 
+- Exact replacement (This is ROS 1)
+
+ - Replace a name with another name exactly
+
+ROS 2 may support remapping rules being applied either before or after the names are expanded to FQN. ROS 1 only supports the latter. 
+
+#### Pre-FQN expansion remapping
+Applying rules prior to FQN expansion enables this use case. Say the source code for a node use two names `"cat"` and `"/ns/cat"`. The node is in namespace `"ns"`. With just Post-FQN expantion it is impossible to remap these names individually. With Pre-FQN expansion remapping, a rule can be made to remap just one of them.
```
Karsten1987: Nice point here. However, maybe stress a bit more that "/ns/cat" is the global identifier and the resulting topics are identical.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### Requirements for remapping capabilities
+##### 1. Namespace Replacement
+**A user can remap the first namespace token without replacing all namespace tokens with the same name.**
+
+*Example*
+Node uses names `"/big/cat"`, `"/big/red/dog"`, `"/red/big/cat"`. It must be possible to make a rule to that replaces "big" with "small" such that the final names are `"/small/cat"`, `"/small/red/dog"`, `"/red/big/cat"`
```
Karsten1987: does this also apply to the basename?

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
+### Remapping Names
+#### ROS 1
+In ROS 1 [Names](http://wiki.ros.org/Names) can be remapped to other names by passing [remapping arguments](http://wiki.ros.org/Remapping%20Arguments) via the command line. `roscpp` additionally allows remaps to be passed via Init as strings in STL containers. A ROS1 remap consists of two names: one that should be replaced with another.
+
+In ROS1 remapping works with FQN. When a name is to be checked for remappings, it is first expanded to a FQN. Both parts of the remapping rule are expanded to a FQN. Then a name is changed to the remap's second name only if it exactly matches the first name.
```
wjwwood: I'm not completely certain this is how it works, though it might be correct. In fact it doesn't matter if you expand them first or not, I think. For example, if you create a topic called `foo`, and then you pass the remapping `foo:=bar` and you're in a namespace `/baz/`, then if you replace before expanding to a FQN you'd replace `foo` with `bar` and then expand that to `/baz/bar` (correct answer) but if you expand the replacements first you get the same result, i.e. `foo` becomes `/baz/foo` and the remapping becomes `/baz/foo:=/baz/bar` and then you replace it to get the correct answer again. So I think it's potentially unimportant when the remapping happens, at least in this case, but it would be better if you could cite this by linking to ROS 1 docs or even to the code which does this in `roscpp`.

Also, a nitpick, you should use the full name the first time you use an acronym, i.e. `Fully Qualified Name (FQN)` first and then use just `FQN` after that.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### ROS 2
+In ROS2 remapping will be more powerful. There are more types of remappings that will be supported
```
wjwwood: End with a `:` or some other punctuation.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+ 
+- Basename replacement
+
+ - Replace all of the basename with another basename
+ 
+- Arbitrary substring replacement
+
+ - Replace any substring with another string
+ 
+- Exact replacement (This is ROS 1)
```
wjwwood: I'm not sure about this one, see above comments about how this works in ROS 1.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+
+ - Replace all or part of a namespace with another namespace
+ 
+- Basename replacement
+
+ - Replace all of the basename with another basename
+ 
+- Arbitrary substring replacement
+
+ - Replace any substring with another string
```
wjwwood: It's not clear to me that we should support this in ROS 2. I mentioned it in the meeting for completeness. I'm not sure if it makes sense to support this in ROS 2, as I think that it might lead to really messy systems that rely on arcane remapping rules, i.e. I think the structure enforced by namespaces and basenames adds value over a free-for-all string replacement feature.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+
+*Examples*
+Node uses names `"/big/red/dog"`, `"/very_big/red/dog"`. It must be possible to make a rule to that replaces "big/red" with "small" such that the final names are `"/small/dog"`, `"/very_small/dog"`
+
+Node uses names `"/big/dog"`, `"/dog/big"`. It must be possible to make a rule to that replaces "dog" with "cat" such that the final names are `"/big/cat"`, `"/cat/big"`
+
+Node uses names `"/big/red/dog"`. It must be possible to make a rule to that replaces "/" with "_" such that the final name is `"/_big_red_dog"`
+
+#### Static and Dynamic Remapping
+Static remapping is the ability to provide a node with remap rules prior to launching it. A node stores this remapping rules for the duration of it's life. Any names it uses get have these rules applied to it before it is used.
```
Karsten1987: What are the consequences for the original topics written in code? It make sense to state that they technically never make it into a DDS world.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+ 
+- Arbitrary substring replacement
+
+ - Replace any substring with another string
+ 
+- Exact replacement (This is ROS 1)
+
+ - Replace a name with another name exactly
+
+ROS 2 may support remapping rules being applied either before or after the names are expanded to FQN. ROS 1 only supports the latter.
```
wjwwood: Again, I would have thought ROS 1 does only pre substitution.

Karsten1987 (approved):
I like it!

articles/remapping.md
wjwwood: Nitpick, just use ``` `cat` ``` rather than ``` `"cat"` ```.

wjwwood: Adding on to the ROS 1 issue, I think you can do this in ROS 1 already.

articles/remapping.md
wjwwood: > Any names it uses get have these rules

Not sure what you're trying to say here, looks like a typo.

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+Node uses names `"/big/red/dog"`, `"/very_big/red/dog"`. It must be possible to make a rule to that replaces "big/red" with "small" such that the final names are `"/small/dog"`, `"/very_small/dog"`
+
+Node uses names `"/big/dog"`, `"/dog/big"`. It must be possible to make a rule to that replaces "dog" with "cat" such that the final names are `"/big/cat"`, `"/cat/big"`
+
+Node uses names `"/big/red/dog"`. It must be possible to make a rule to that replaces "/" with "_" such that the final name is `"/_big_red_dog"`
+
+#### Static and Dynamic Remapping
+Static remapping is the ability to provide a node with remap rules prior to launching it. A node stores this remapping rules for the duration of it's life. Any names it uses get have these rules applied to it before it is used. 
+
+Dynamic remapping is the ability to remap a name after it has already been used. For a neat user experience, Dynamic remapping requires rules to be changed. A user should be able to remap the name they see rather than the original name that was used. This suggests remapping rules have an order in which they should be applied (chronological?).
```
wjwwood: Maybe? This was something we were debating amongst ourselves during the meeting. I personally believe you need to at least store the "static" remappings for when you create new entities that use "Names" (as you mention above), but it's less clear to me about the dynamic remappings. We need to first decide what kinds of remappings may be used in dynamic remapping and if they apply to all existing things, all new things, or are scoped more narrowly than static remappings (i.e. applies to a single entity versus a whole node).

wjwwood (commented):
Looks like a good start, though I think you may have to look into the description of how ROS 1 works in this regard.

Also, a high level nitpick is that we usually do "one sentence per line" when writing markdown. This is a compromise between avoiding line wrapping manually (which is tedious and makes diffs hard to read) and having extremely long lines in the source (one sentence per line versus one paragraph per line).

Thanks for starting the conversation here.

articles/remapping.md
sloretz: I'll check roscpp's implementation and link it. I got the impression that's how it worked from the [third paragraph of this page](http://wiki.ros.org/Remapping%20Arguments). I think it makes a difference if: remapping `foo:=bar`, namespace: `ns`, names used in code `foo` `/ns/foo`. If done pre-FQN only the first name will get remapped; if done post-FQN both will get remapped. 

I'll update to make sure FQN is defined.

articles/remapping.md
sloretz: I included it to match the meeting notes, but I also think it shouldn't be supported.

I think with just namespace and basename remapping it's possible to check if a rule is valid at startup. With arbitrary substring replacement it would be possible to make a rule that appears legal at startup, but fails later when a name gets used. Example: replace "x" with "_", name used 20 minutes into program's run "/xxx".

articles/remapping.md
sloretz: [names::init()](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#a377ff8fede7b95398fd5d1c5cd49246b) resolves both sides of the remap before storing to `g_remappings`. [names::remap()](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#ab2eebaf734abfbdccb4122f8e24f547f) resolves the name to be remapped prior to looking up the remapping in `g_remappings`.

articles/remapping.md
dhood: for readability I suggest changing "enables this use case" to "enables the following use case" (otherwise I thought it was referring to the section heading).

articles/remapping.md
dhood: I would mention here that examples are provided below

articles/remapping.md
```diff
@@ -0,0 +1,77 @@
...
+
+##### 4. Arbitrary Substring Replacement
+**A user can replace any substring with another string**
+
+*Examples*
+Node uses names `"/big/red/dog"`, `"/very_big/red/dog"`. It must be possible to make a rule to that replaces "big/red" with "small" such that the final names are `"/small/dog"`, `"/very_small/dog"`
+
+Node uses names `"/big/dog"`, `"/dog/big"`. It must be possible to make a rule to that replaces "dog" with "cat" such that the final names are `"/big/cat"`, `"/cat/big"`
+
+Node uses names `"/big/red/dog"`. It must be possible to make a rule to that replaces "/" with "_" such that the final name is `"/_big_red_dog"`
```
dhood: couldn't there be a case where you would want `/big/red/dog` to map to `_big_red_dog` as a non-global name?

articles/remapping.md
dhood: if nothing else I think the conclusion was that dyanmic remapping rules should be chainable, so that systems can have names dynamically remapped again and again without restarting (especially if they are running for years at a time)

articles/remapping.md
wjwwood: Yup, you're right. I never noticed that it processed them this way. Probably because the effect is the same in pretty much every case I can think of. Thanks for tracking down the citations, I'd still recommend putting the citations in the document, so if someone else comes along with the same question as me it is clear to them too.

articles/remapping.md
wjwwood: It still might be clearer to call this "exact replacement after expansion" or in the following bullet say "Replace a FQN with another FQN by exact match" or something along those lines. Just to make it clear.

articles/remapping.md
wjwwood: You're right here as well. I guess the rationale for "pre-FQN" replace is weaker now. In the next paragraph, however, you point out that it might help with more narrowly specifying what to remap.

articles/remapping.md
wjwwood: > Adding on to the ROS 1 issue, I think you can do this in ROS 1 already.

Actually, you are probably right that ROS 1 cannot do this as-is, but it would be good to verify this.

articles/remapping.md
sloretz: I intended the examples in this first section to only change the namespace, and not the basename.

articles/remapping.md
sloretz: I think that case might be exact name replacement. I think in ROS 1 `/big/red/dog:=_big_red_dog` would become `/ns/_big_red_dog`. I'll add multiple topics to this example to clarify how this remapping might operate.

sloretz: > I think that case might be exact name replacement.

Maybe this example is actually that case. The string after doing that replacement would be `_big_red_dog`, so maybe it makes more sense that the final name would be `/ns/_big_red_dog`.

dhood: while you could do `/big/red/dog:=_big_red_dog` with exact name replacement, to accomplish the generality of the `/`->`_` mapping that is referred to in this bullet point, you would have to do that for every name that matches the pattern (e.g. also `/small/red/cat:=_small_red_cat`)

dhood: having said that, I am not advocating for inclusion or exclusion of this requirement, just pointing out that the example would be more general if the final name is not global

sloretz: Good point. I'll update it.

articles/remapping.md
sloretz: > it would be good to verify this.

I verified the behavior in  kinetic. 

```
#include <ros/ros.h>
#include <std_msgs/String.h>

int main(int argc, char **argv)
{
    ros::init(argc, argv, "remap_test");
    ros::NodeHandle nh;
    ros::Publisher pub1 = nh.advertise<std_msgs::String>("cat", 10);
    ros::Publisher pub2 = nh.advertise<std_msgs::String>("/ns/cat", 10);
    ros::spin();
    return 0;
}

```

`rosrun remap_test remap_test_node __ns:=ns cat:=cat2`

and 

`rosrun remap_test remap_test_node __ns:=ns /ns/cat:=cat2`

Both result in

```
s$ rostopic list
/ns/cat2
/rosout
/rosout_agg
```

sloretz:
I used the feedback from the first draft to update the document. To keep the design discussion going, I also added a paragraph proposing to make `rcl` responsible for remapping in the `Static Remapping` section at the bottom of the document.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+A node stores all remapping rules for the duration of it's life.
+All names have remapping rules applied before they are used.
+This means the DDS interface, and maybe all of `rmw`, is ignorant of remapping.
+
+In ROS 1 remap rules are passed into the node at startup.
+`roscpp` received remap rules as arguments to `ros::init()`.
+`rospy` loaded remap rules from `sys.argv` during `rospy.init_node()`.
+
+In ROS 2 this code can be shared between client implementations by doing it in `rcl`.
+[rcl_init()](http://docs.ros2.org/beta1/api/rcl/rcl_8h.html#a6abc5188c50b3a4a5fc41b357b49d6e5) already requires command line arguments to be passed to it.
```
dirk-thomas: `rcl_init` is only called once per process. So it can only do the separation of the command line args. They later have to be passed to `rcl_node_init` since each node should get its specific remapping arguments.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+It could parse the rules into an intermediate representation and store it internally.
```
dirk-thomas: I would suggest to not mention that the information is stored internally. First that is an implementation choice and I think it is to early to make that call (especially with no rational to why that is the case). Second returning the representation and passing it explicitly to the next functions make the API much more explicit. Each part can be reused separately as well as tested separately.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+Methods like [rcl_publisher_init()](http://docs.ros2.org/beta1/api/rcl/publisher_8h.html#a0f09bd85795259f6c96f38c875d19d13) and [rcl_subscription_init()](http://docs.ros2.org/beta1/api/rcl/subscription_8h.html#a2b8a58f9ae9fef8adba35b5a45db10a0) would apply remapping rules automatically.
```
dirk-thomas: I would suggest that these function call functions on the containing node to perform the name mapping. That encapsulates all the knowledge about remapping and the logic to do it in a single location rather then spreading it across publisher, subscribers, requesters, responders, etc. But again that is already an implementation detail which doesn't need to be covered in the design doc.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+Making name remapping an automatic part of `rcl` suggests that features using names, such as parameters, would be implemented in `rcl` and not `rclcpp` and `rclpy` as indicated on [slide 31](http://roscon.ros.org/2016/presentations/ROSCon%202016%20-%20ROS%202%20Update.pdf).
+
+#### Dynamic Remapping
+
+Dynamic remapping is the ability to remap a name after it has already been used.
+For a neat user experience, Dynamic remapping requires rules to be chainable.
```
dirk-thomas: I think this needs more explanation. While we have talked about it a lot and know why it would be good to provide more information to readers why that is "neat" / "sane".

dirk-thomas (commented):
Regarding the overall structure I would suggest to not start with ROS 1 and try to describe the difference in the beginning. Instead I would mentioned how namespaces and topic names are embedded in each nodes code and the ability to remap these enables reusing existing code without the need to modify it. Then describe that each FQN should be remappable to any other FQN (mentioning the different ways to remap). Later a comparison to ROS 1 is great to describe what the differences are and why we chose to make changes to the way ROS 1 does it.

The rational for that is that reader should be able to follow the design process without relying on ROS 1 knowledge.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
+### Remapping Names
+#### ROS 1
+In ROS 1 [Names](http://wiki.ros.org/Names) can be remapped to other names by passing [remapping arguments](http://wiki.ros.org/Remapping%20Arguments) via the command line.
+`roscpp` additionally allows remaps to be passed via `init()` as strings in STL containers.
+A ROS1 remap consists of two names: one that should be replaced with another.
+
+ROS 1 remapping works on **Fully Qualified Names** (FQN).
+`roscpp` [names::init()](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#a377ff8fede7b95398fd5d1c5cd49246b) expands both sides of a remap to FQNs before storing it.
+[names::remap()](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#ab2eebaf734abfbdccb4122f8e24f547f) expands a name to a FQN before checking for applicable remapping rules.
+The name is remapped to the right side only if it exactly matches the left side of a rule.
```
tfoote: This needs to be fleshed out. The ROS1 remappings are not this simple. They match both the fully qualified values as well as the specific name used in the code without the namespace. 

ROS1 relies on matching the token used in the code.  The token used inside the code is important as something publishing "foo/bar" will not match `foo` or `bar` like above would. 

The above full match can also be manipulated by using nodehandles to move the resolution elements around.

clalancette:
I agree with @dirk-thomas ' comments, though I might go a bit further.  For someone coming into this fresh (like me), a set of use cases/rationales/requirements up-front would go a long way to help understand what we are trying to accomplish.  Then we can go on to explain the ways in which remapping will allow us to accomplish those use cases.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### ROS 2
+In ROS2 remapping will be more powerful.
```
tfoote: This is subjective. Maybe something more like:

In ROS2 we expect to target more use cases.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+See the requirements section below for examples.
```
tfoote: Why not put the requirements at the top here with the reference, and have the implementation afterwards?

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Pre-FQN expansion remapping
+Say the source code for a node use two names `cat` and `/ns/cat`.
+If the node runs in namespace `ns`, the FQN of both topics will be identical: `/ns/cat`.
+No rule could remap these names individually once they have been expanded.
+A rule applied Pre-FQN expansion could remap just one of them.
+
+#### Requirements for remapping capabilities
+##### 1. Namespace Replacement
+These are rules that only apply to the namespace part of the name.
```
tfoote: Is it really different that we have different remapping types? Since this is a list of requirements it shouldn't presume how to implement this.

The namespace/basename distinction seems to collapse if we support filesystem like regexes. But the requirement to support both use cases is still valid.

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+
+Node uses names `/big/red/dog`, `/small/cat`.
+Node is run in namespace `ns`.
+It must be possible to make a rule that replaces "/" with "_" such that the final names are `/ns/_big_red_dog`, `/ns/_small_cat`
+
+#### Static Remapping
+Static remapping is the ability to provide a node with remap rules prior to launching it.
+A node stores all remapping rules for the duration of it's life.
+All names have remapping rules applied before they are used.
+This means the DDS interface, and maybe all of `rmw`, is ignorant of remapping.
```
tfoote: How should precedence work for conflicting/overlapping rules like this?

`./node foo:=bar foo:=baz`

articles/remapping.md
```diff
@@ -0,0 +1,119 @@
...
+[rcl_init()](http://docs.ros2.org/beta1/api/rcl/rcl_8h.html#a6abc5188c50b3a4a5fc41b357b49d6e5) already requires command line arguments to be passed to it.
+It could parse the rules into an intermediate representation and store it internally.
+Methods like [rcl_publisher_init()](http://docs.ros2.org/beta1/api/rcl/publisher_8h.html#a0f09bd85795259f6c96f38c875d19d13) and [rcl_subscription_init()](http://docs.ros2.org/beta1/api/rcl/subscription_8h.html#a2b8a58f9ae9fef8adba35b5a45db10a0) would apply remapping rules automatically.
+Making name remapping an automatic part of `rcl` suggests that features using names, such as parameters, would be implemented in `rcl` and not `rclcpp` and `rclpy` as indicated on [slide 31](http://roscon.ros.org/2016/presentations/ROSCon%202016%20-%20ROS%202%20Update.pdf).
+
+#### Dynamic Remapping
+
+Dynamic remapping is the ability to remap a name after it has already been used.
+For a neat user experience, Dynamic remapping requires rules to be chainable.
+When a user dynamically remaps a name, it should apply to the name they see rather than the original name that was used.
```
tfoote: Also the persistance of remapping is important. For example does the node keep a history of all remappings and apply historic remappings to new graph entities on startup?

articles/remapping.md
sloretz: As in `rcl_node_init()` should have an argument for remapping rules similar to `ros::init()`?

articles/remapping.md
sloretz: Makes sense. I'll limit the amount of content mentioning implementation details and focus on the behavior as a user would see it.

articles/remapping.md
sloretz: It is pretty sparse. I'll expand it to explain why dynamic remapping is useful.

articles/remapping.md
sloretz: Requirements at the top sounds like a good idea. Do you think the content about implementation has a place in this design doc?

articles/remapping.md
sloretz: What do you mean by filesystem regexes? Got an example?

articles/remapping.md
wjwwood: In short, yes. In ROS 1, initialization (normally once per process) and node creation are conflated into `ros::init()`. In ROS 2, initialization is separate from node creation, and so we have to separate the responsibilities of `ros::init()` accordingly. For this case, processing command line arguments only really makes sense to do once per process, so maybe that should be inherent in the `rcl_init` function. So given the command line arguments as input, you should get back out modified command line arguments and some remappings. Then the remappings should be given somehow to each node at creation time. One way to do this is by passing them in as options to `rcl_node_init`

That all being said, I want to second the idea that we avoid any specifics like this, and instead speak in language neutral terms. For example, I'd say something like:

> Arguments passed on the command line can be converted into remapping rules, which can be shared with nodes and other ROS entities that the remapping rules might affect.

This implies what is possible without going to the point of calling out how it's done in C versus C++ or Python and without deciding if they should be implicitly shared or explicitly passed and to which functions. The former is more important information for the reader to gain a good conceptual understanding anyways. The discussion of implementation details may just add confusion at this point.

articles/remapping.md
sloretz: How about applying rules  in the order they're specified? In that case, a name `foo` would become `bar` after the first rule, and the second rule would change nothing because there are no `foo` names to act on.

I think starting with that logic makes dynamic remapping easier later on, because dynamic rules should chain to let a user remap the names they see at runtime.

articles/remapping.md
sloretz: Do you have an example where ROS1 uses the token in code and not the FQN? I would have expected a different outcome [here](https://github.com/ros2/design/pull/112#discussion_r100630697).

sloretz:
@dirk-thomas @tfoote @clalancette @dhood @wjwwood  I used your feedback to refactor the document. Would you mind reviewing it again?

I removed details that could be considered implementation specific, but left a couple thoughts that I think are important but didn't fit into a use case.

I also added a proposal for a command line syntax at the bottom to keep the discussion going.

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+ROS 1 remapping works on **Fully Qualified Names** (FQN).
+Both sides of a rule are [expanded to FQN](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#a377ff8fede7b95398fd5d1c5cd49246b).
+Before a name is remapped it is also [expanded to FQN](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#ab2eebaf734abfbdccb4122f8e24f547f).
+The name is remapped to the right side only if it exactly matches the left side of a rule.
+
+
+### Thoughts on ROS 2 implementation
+`rmw` is likely ignorant of remapping.
+The APIs offered by client implementations should remap names automatically to prevent creation of names that are impossible to remap.
```
dirk-thomas: Maybe avoid the term `remap` here and replace it with `transform` to indicate that it is a different mechanism?

dirk-thomas (dismissed):

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
+## Remapping Names
+Topics, parameters, and services are identified by [Names](http://wiki.ros.org/Names).
+Names are hard coded in ROS nodes.
+Remapping is what allows names to be changed at runtime.
+Without remapping nodes would have to be modified and rebuilt for every application.
```
wjwwood: Maybe `Without remapping, nodes would...` or `Without remapping a node would...`?

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+- `/foo/bar`
+- `foo/bar/baz`
+- `foo/bar`
+- `bar`
+
+Each part of a name between a slash is a **token**.
+The examples above use the tokens `foo`, `bar` and `baz`.
+FQN are conceptually divided into two pieces: **namespace** and **basename**.
+The basename is the last token in a name.
+The namespace is all tokens and slashes prior to the basename in a FQN.
```
wjwwood: Technically the namespace is still "all tokens and slashes prior to the basename" even if it is not a FQN. For example, for `foo/bar` the namespace is `foo/`, it's just a relative namespace.

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+Without remapping nodes would have to be modified and rebuilt for every application.
+**Remapping rules** are the instructions given to a node to change the names it uses.
+
+### Structure of a Name
+Names are hierarchical strings delineated by `/`.
+If they begin with `/` they are said to be **global** or **Fully Qualified Names** (FQN).
+Otherwise they are said to be **relative**.
+Relative names are resolved to FQN prior to being used.
+
+**Example names**:
```
wjwwood: Missing a `:`?

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Example:**
+
+- Node is in namespace `/fiz/buz`
+- Node uses relative name `foo/bar`
+- The FQN is `/fiz/buz/foo/bar`
+
+
+### ROS 2 Use Cases
+Remapping in ROS 2 will likely support the following use cases:
```
wjwwood: I think it would be better to enumerate them as a list here and then say they are each described in a section below. Or just drop the `:` and say they are described below. Either way a `:` followed by subsections looks strange to me.

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+
+#### A user can run multiple instances of a node using the same executable in different data paths
```
wjwwood: I'd suggest trying to avoid long section names, instead I'd recommend doing some like "Running Identical Nodes in Different Namespaces" and then move the current section name into a sentence in the first paragraph.

wjwwood: I'd suggest doing something similar in each of the sections below.

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+Remapping rules have two parts:
+
+1. a part used to determine if the rule applies to a name
+2. a replacement for the names that are matched
+
+
+### Remapping Names in ROS 1
+Remapping is a feature that also exists in ROS 1.
+In ROS 1 remapping works by passing in [arguments](http://wiki.ros.org/Remapping%20Arguments) to each node.
+Client implementations also have APIs in code to pass remapping rules when the node is initialized.
```
wjwwood: I'd suggest using the term "client libraries" rather than "client implementations".

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+- `%`
+ - During matching it is a wildcard that matches a single token
+ - It may appear at most once on the matching side of a rule
+ - During replacement it is replaced with the matched token
+
+- `%%`
+ - During matching it is a wildcard that matches zero or more tokens with at least one slash
+ - It may appear at most once on the matching side of a rule
+ - During replacement it is replaced with the matched slashes and tokens
```
wjwwood: Could you add some rationale to these two items? For example, why did you land on `%` instead of something like `*`? What's the use case for them? I agree that they're logically useful, but it would best to pose their purpose in terms of what was difficult in ROS 1 that these allow for. It's not a deal breaker if we can't identify anything, we've added features before just because they seem like a good idea, but it's always better to have a motivation in place.

articles/remapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+- `:=` divides the two parts of a remapping rule: matching and replacement
+
+- `~` expands to `/namespace/nodename`
+
+- `nodename|` prefixed to a rule makes it apply only to a node with that unique name
+ - If not specified, the rule applies to all nodes in the process
```
wjwwood: Again, a rationale for the syntax selection would be nice to see here. I don't have a better suggestion off-hand, but it would be nice to see this justified (if there is one). Also will using `|` make it hard to type this in a shell?

I think this is a fine way to narrow down individual rules to individual nodes. However, I was thinking about this a lot this weekend and I was unable to convince myself that we "must" provide this feature. The only time you would need to provide different remapping arguments to a single executable with multiple nodes via the command line would be when someone has written their own main function and instantiated their own nodes. This would be a nice feature to provide these people since they would have to implement the argument parsing on their own otherwise, but I'm not convinced it will be a commonplace thing.

I think it's far more likely that two other modes of operation will be used frequently, and neither would require passing remapping for multiple nodes over the command line.

The first mode is a single node in a single executable. I expect this to be commonplace, but the ROS 1 style command line arguments should work here.

The second mode will be that a "container" is run for multiple nodes (same thing as Nodelet manager in ROS 1) and then the nodes it runs will be loaded via a configuration file or after it has started using ROS services. The process would be: start a container, add some stuff to it, and specifying the remappings for each as you add them.

So I think we should go through the exercise of coming up with a syntax, but it doesn't have to be especially fancy in my opinion because I believe it should be rarely used. That being said, I think the proposed syntax should work out fine.

wjwwood (changes requested):
Most of my comments are formatting related, which is good! Thanks for trying to hammer down the static remapping ideas. This is already better described than it is in ROS 1!

A couple more formatting comments, this article is missing a preamble (see other articles for what I'm talking about or the http://design.ros2.org/contribute.html page has a template) . Also this file should have a number in front (See the other articles). If you're unsure about what number to use, just pick one of the ones near the topic and service name mapping article.

articles/remapping.md
sloretz: Oops, I landed on `%` because I thought `*` would be expanded by the shell and `%` is the wildcard in SQL, but now I see that `*` only gets expanded it it is surrounded by whitespace (at least in bash) and the separator `|` does make it hard to type in a shell. I can change them to `*` and `:`

I could add how these enable the use cases above, would that be the same as adding rationale?

wjwwood: I don't mind `%` over `*`, I was just curious why you selected that one. In fact, `*` always gets expanded in `zsh` (which is what I use) unless you escape them or put `noglob` before the command, so I don't see it as clearly better. I was just curious 😄.

> I could add how these enable the use cases above, would that be the same as adding rationale?

A rationale as to why it is necessary for aforementioned use cases would be good, yeah.

wjwwood: I don't have a strong preference about `:` versus something else, but I do think that `|` would be inconvenient if we expect people to use this on the terminal ever.

tfoote: I agree that the `|` will be problematic on the command line. I'd suggest that we try to reuse existing syntax. I'd suggest that we consider using the POSIX fnmatch wildcard substitution rules which is already used in the DDS API would be both low cognative load to learn, easy to implement, and no new implementation or dependency. Note that if we want to extend it I believe that all the application specific things like the node specific requirement can simply be done as a preprocessing step before we pass it to fnmatch which implements the wildcard replacements in C.

tfoote (changes requested):

articles/remapping.md
wjwwood: In this case @sloretz was wanting to use `|` as a separator of some sort. What's the `fnmatch` equivalent? Maybe `/`?

I agree that using `*` would be consistent with us using `fnmatch` compatible patterns in other parts of the system.

sloretz: I like that the fnmatch syntax is standardized. How does it avoid new implementations and dependencies? It sounds like it requires an implementation that captures the text matched by wild cards or inclusion of a regular expression library with capturing support.

The right side syntax needs to reference what was matched on the left. The syntax in this PR uses a wildcard on the right to do so. Example remapping actionlib topics:

```
/some_action_interface/%%:=/some_other_name/%%
```

fnmatch doesn't have capturing or back references. One solution is to assume every wild card on the left is capturing, and borrow the back reference syntax from POSIX BRE.

```
/some_action_interface/*:=/some_other_name/\1
```

sloretz:
I've made some edits that I believe fix all the comments above. Major changes have been made to the formatting of requirements; they all have short names now. The section proposing the syntax has new content for each requirement showing how the syntax meets it.

It also displays nicely in Jekyll.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,341 @@
...
+
+#### Supporting: Change a Namespace
+There are two cases: changing part of a namespace, and changing the entire namespace.
+The first case requires a wildcard to match the rest of a namespace.
+The second requires a wildcard to match the basename at the end.
+
+*Example: Partial namespace replacement*
+
+- Node uses names `/foo`, `/foo/bar/baz`
+- Node given rule `/foo**:=/fizz\1`
```
dirk-thomas: I find this rule difficult to read and interpret. I would expect `/foo/**` to be more "natural".

The defintion says that `**` matches at least match one slash. But from just reading the rule I would expect it also to match things like `/foobar` (which I know it wouldn't).

articles/141_static_remapping.md
```diff
@@ -0,0 +1,341 @@
...
+- Resulting names `/bar/foo/baz`, `/foo/bar/fee/biz`
+
+#### Supporting: Change a Basename
+Changing a basename requires a wildcard which matches the entire namespace.
+The wildcard `**` is useful because it matches every possible namespace.
+
+*Example:*
+
+- Node uses names `/foo`, `/buz/foo`, `/biz/buz/foo`
+- Node given rule `**foo:=\1bar`
```
dirk-thomas: Same here. For readability I would expect the rule to look like this instead: `**/foo` Especially since the definition doesn't mention anything about a trailing slash.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,341 @@
...
+## Structure of a Remapping Rule
+**Remapping rules** are the instructions describing how a node should change the names it uses.
+Remapping rules have two parts.
+The first part is used to determine if the rule applies to a name.
+The second part is the replacement for a matched name.
+The act of replacing one name with another is **remapping**.
+
+
+## ROS 2 Remapping Requirements
+These requirements are being considered for remapping in ROS 2:
```
wjwwood: These are really worded more like "use cases" rather than "requirements" in my opinion. There's a semi-formal language for requirements (shall, should, may, won't, etc have very explicit meanings, see http://www.ietf.org/rfc/rfc2119.txt), if you want to phrase them as such. For example, if I wanted to make these into requirements I might rewrite a few of them like this:

- It shall be possible to remap a single node within a process, even in multi node processes
- It shall be possible to change the Namespace of a topic (when? always? anytime?)

I think more details would need to be sussed out to make them into requirements. I actually don't think that's the right thing to do here. I like the way you've worded these as use cases already. So my suggestion (if you decide take it) is to just refer to them as use cases rather than requirements.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,342 @@
...
+Requirements not supported
+- Change a Token
+- Pre-FQN Remapping
+
+Special Operators
+- `*` wild card that matches a single token
+- `**` wild card that matches at least one slash and zero or more tokens
+- `:=` divides the two parts of a remapping rule: matching and replacement
+- `~` expands to `/namespace/nodename` and may only be at the very beginning of a match or replacement
+- `__ns` matches the default namespace if used by itself on the match side of a rule
```
wjwwood: Do you think this is sufficient? I'm thinking of a case where the remappings are calculated in order, and you have one normal remap, then a remap that starts with `__ns:=`, then another. In that case the left hand side (the "match") would be affected in the last remapping by the middle one. I don't that happens with other remappings. That's what makes it "special" in my mind.

I don't know how it works in ROS 1 or if it even matters in that case, but originally I was thinking that changing the node's name or it's namespace (`__name:=` and `__ns:=` respectively) might need to be handled before all other remappings. That might evolve if we allow either or both of those to be changed during runtime. Perhaps it's worth distinguishing between changing the namespace with `__ns:=` and a normal namespace remap whose "match" part is just the namespace of the node. The former would potentially be retroactive, whereas the latter might be applied in order like the other remap rules.

Not sure if I have a concrete change suggestion, but I'd like to know what you think about this point.

articles/141_static_remapping.md
wjwwood: I agree that `/foo/**` looks better. I think that if you wanted to make the existing `/foo**` make sense, you'd have to have an input topic like `/foo_bar/baz` which would become `/fizz_bar/baz`. However, I think to do what @dirk-thomas suggested, i.e. `/foo/**`, and have it match `/foo/bar` (for example) then the above rule about `**` would need to be modified to so that it could match zero to many slashes rather than at least one.

At that point, however, it's hard to make a rule to match like this:

- `/foo`: does not match
- `/foo/bar`: matches, `/1` is `/bar` for `/foo**` or `bar` for `/foo/**`
- `/foo/bar/baz`: matches, `/1` is `/bar/baz` for `/foo**` or `bar/baz` for `/foo/**`

The current setup `/foo**` would work in most of these cases (except `/1` is `/bar` for `/foo**`), but if you changed it so `**` doesn't have to match at least one `/`, then it would also match `/foo`. I'm not sure it matters which, but you'll have to choose between a more flexible `**` and a stricter `**` which might give you more control.

articles/141_static_remapping.md
wjwwood: This is also a good point. `**` not only matches all namespaces, but it can (as far as I understand) also match the entire namespace plus a bit of the base name. For example, if you add an input topic `/ping/pong/baz_foo` does it remain unchanged or change to `/ping/pong/baz_bar`?

I'm assuming the `**` is greedy when it comes to trailing `/` and it will change. In which case just changing the example to `**/foo` would fix it.

wjwwood (commented):
This is looking really good! Thanks for working on it @sloretz.

I had some questions and comments which I'd like to discuss, but overall I think this is going really well so far.

I also pushed a few small changes, mostly syntax and grammar.

articles/141_static_remapping.md
sloretz: @dirk-thomas I see what you mean about it feeling more natural. Without the slash it looks like it would match `/foobar`, but I didn't intend that. 

@wjwwood If `**` matched zero or more slashes and whole tokens only, the distinction of `\1` being `/bar` vs `bar` could be made unimportant by the replacement side of a rule automatically adding or removing slashes. For example if `\1` is `/bar` and the replacement rule is `/\1` the result would be `/bar` (slash auto-deleted), and if `\1` is `bar` and the replacement rule is `/foo\1` the result would be `/foo/bar` (slash auto-added).

articles/141_static_remapping.md
sloretz: @wjwwood I think `/ping/pong/baz_foo` should remain unchanged in this example. If `**` matched partial tokens, it would be difficult to remap short basenames. `**a` would match both `/a` and `/camera` for example. 

If `**` instead matches zero or more slashes and partial tokens, remapping short basenames is easy `**/a`. On the replacement side of the rule replacing partial tokens means it matters if `\1` has a slash or not, which as you pointed out is more difficult to control with `**` matching zero or more slashes.

articles/141_static_remapping.md
sloretz: I agree that `__ns` has to be handled prior to all remappings. Remap rules happen Post-FQN expansion, but FQN expansion requires knowing what the default namespace is.

The description of `__ns` here is confusing. It is not really a remap rule. It just looks a little bit like one because it is prefixable with `nodename:` and has a right hand side `:=/some/namespace`. I can move `__ns` to its own section away from the remapping syntax.

dirk-thomas:
Would it be helpful to actually rename a node (so that a node with the name `camera_driver` can be renamed to `left_camera_driver`)? That could affect e.g. log messages which then use the new name which allows to distinguish them if multiple instances of the same node have been started.

sloretz:
@dirk-thomas @wjwwood would you mind reviewing it again? I've used your comments to update the document. I also added a few paragraphs about the remapping syntax to nail down the behavior in corner cases.

wjwwood:
@sloretz sure thing, I'll do that now.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,435 @@
...
+Nodes are said to be in a **namespace**.
+This namespace becomes a prefix to all relative names used by the node.
+Relative names are expanded to FQN by prepending the node's namespace to them.
+
+
+**Example:**
+
+- Node is in namespace `/fiz/buz`
+- Node uses relative name `foo/bar`
+- The FQN is `/fiz/buz/foo/bar`
```
wjwwood: This is a good summary of the names (which lets this document stand on it's own, and that's good imo), but it might be good to link to the other design doc that talks about them as well, since it has a slight stricter definition of what can and cannot be in a name:

http://design.ros2.org/articles/topic_and_service_names.html#ros-2-topic-and-service-name-constraints

Later in this document you talk about `~` and how it can be used. Again, I think it's good if there is enough information there for this doc to stand on its own, but we should try to avoid duplication of information. It might be worth putting the core of the rules in its own doc and have the "maps to DDS" part of the other doc and the static remapping in this doc point to it.

The other option is to merge this doc with the other one, so that rules about names, static remapping, and mapping to dds are all in the same document. The downside to that is that this "super" document might be too long or overwhelming.

wjwwood: I don't know if there is any action that needs to be taken before we merge this document, but it would be nice to hear what you think would be best to have in the end.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,435 @@
...
+
+## Remapping Names in ROS 1
+Remapping is a feature that also exists in ROS 1.
+In ROS 1 remapping works by passing in [arguments](http://wiki.ros.org/Remapping%20Arguments) to each node.
+Client libraries also have APIs in code to pass remapping rules when the node is initialized.
+A remap rule consists of two names: one that should be replaced with another.
+
+ROS 1 remapping works on **Fully Qualified Names** (FQN).
+Both sides of a rule are [expanded to FQN](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#a377ff8fede7b95398fd5d1c5cd49246b).
+Before a name is remapped it is also [expanded to FQN](http://docs.ros.org/api/roscpp/html/namespaceros_1_1names.html#ab2eebaf734abfbdccb4122f8e24f547f).
```
wjwwood: I'd recommend picking a version of ROS to link to, because `docs.ros.org/api` I think is a soft link to the default ROS distribution, and therefore can change over time. Then the anchor hash may stop working. You could instead use something starting with `http://docs.ros.org/kinetic/api`, which should never change unless the code in that file changes for Kinetic. That is still possible, but less likely than the `/api` link changing, which is pretty certain to happen at some point.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,435 @@
...
+- `nodename:~/foo:=foo`
+- `**/foo:=\1/bar`
+
+#### Match Part of a Rule
+
+The match part of a rule uses these operators:
+
+- `*` matches a single token
+- `**` matches zero or more tokens delimited by slashes
+- `~` is replaced with the node's name anywhere it is used
```
wjwwood: This is true, but implicitly that becomes `/ns/node` because the intermediate step is the relative `node`, which gets the namespace prefixed.

I explain it differently in the other document:

> The special single character token `~` will be replaced with a namespace snippet that is a concatenation of the namespace for the node and the node name. For example, a node `node1` in a namespace `/foo` would result in `~` being replaced with `/foo/node1`. As another example, a node `node1` in a namespace `foo/bar` would result in `~` being replaced with `foo/bar/node1`. It must be used at the beginning of a non-fully qualified name, if at all.

-- http://design.ros2.org/articles/topic_and_service_names.html#private-namespace-substitution-character

My description allows for the "default namespace" to be relative, which on reflection doesn't really make sense. So maybe we need to come to some conclusion about this behavior and update the other doc or this one or both, whatever may be.

Related to this is generic substitutions, i.e. things inside of `{}`, see: http://design.ros2.org/articles/topic_and_service_names.html#substitutions

I don't think these are mentioned in this document at all. I always thought of `~` as something that could be replaced with something like `{private}` or `{ns}/{node_name}`. Anyways, how does remapping interact with these substitutions? Does it assume they are replaced before any work is done? If so, then maybe this document can just say that `~`, like substitutions within `{}`, in both the "match" and the "replacement" are expanded before remappings are considered, and leave it at that? Then there's no need to explain them any further.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,435 @@
...
+- `nodename:` prefixed to the match makes it apply only to a node with that name
+
+The operators `*` and `**` are similar to the globbing behavior in bash.
+`**` behaves similar to its use in bash>=4.0 with the globstar option set.
+
+The operators `~`, `*`, and `**` match whole tokens only.
+`*bar` looks like it would match `foobar`, but that would mean matching a partial token.
+To avoid confusion the operators are required to be separated from tokens by a `/`.
+For example `*/bar` is allowed, but `*bar` is invalid.
+This is incompatible with the ROS 1 syntax where `~bar` is used, but now `~/bar` must be used instead.
```
wjwwood: This is also supported by the rules in the other document:

> - must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`

-- http://design.ros2.org/articles/topic_and_service_names.html#ros-2-topic-and-service-name-constraints

wjwwood (approved):
lgtm

I had one comment that should help prevent links breaking which I think should be addressed. Other than that I have some more questions, but mostly they are to do with deduplicating rules and text in the other document, but shouldn't necessarily prevent this from moving forward.

This is really coming along, thanks @sloretz for the updates.

articles/141_static_remapping.md
```diff
@@ -0,0 +1,435 @@
...
+
+**Example names:**
+
+- `/foo`
+- `/foo/bar`
+- `foo/bar/baz`
+- `foo/bar`
+- `bar`
+
+Each part of a name between a slash is a **token**.
```
dirk-thomas: Wording:

> between a slash

Should this be?

> between slashes

(since it can't be "between" a single slash but only between two slashes.)

articles/141_static_remapping.md
```diff
@@ -0,0 +1,435 @@
...
+- Remap One Node in a Process
+- Change a Namespace
+- Change a Basename
+- Change a Token
+- Pre-FQN Remapping
+- Exact FQN Replacement
+- Exact Relative Name Replacement
+- Remap via Command Line
+- Change the Default Namespace
+- Change the Node Name
```
dirk-thomas, line 76: Do we need to define a specific order how these remappings need to be applied?

articles/141_static_remapping.md
sloretz: I would rather have this document link to an authoritative definition of a name than duplicate it.

~~My first thought is can the definition of a name stand alone in a third document that both articles link to?~~ **Edit oops, you had that thought too. I should read more closely**

However, I'm leaning towards this remapping article linking to the DDS/Name mapping article so that the definition of a name is not separated from the rationale for restrictions that come from mapping names to DDS topics.

sloretz: On merging the docs, I think the person who needs to know how names map to DDS topics probably doesn't need to know how remapping works and vice versa. I vote to keep them separate.

wjwwood: > On merging the docs, I think the person who needs to know how names map to DDS topics probably doesn't need to know how remapping works and vice versa. I vote to keep them separate.

Makes sense, separate is fine by me too.

> However, I'm leaning towards this remapping article linking to the DDS/Name mapping article so that the definition of a name is not separated from the rationale for restrictions that come from mapping names to DDS topics.

Whatever works best for you. We can deduplicate and link to the other document for now (which sounds like the easiest thing to do), with the intention to separate those rules out at a later date (maybe move them into a REP or into the core documentation directly). Let me know if you'd like help with that, or if restructuring the other doc would make it easier for you to cross reference it more precisely.

articles/141_static_remapping.md
sloretz: >  If so, then maybe this document can just say that ~, like substitutions within {}, in both the "match" and the "replacement" are expanded before remappings are considered, and leave it at that?

I think that would be good enough.

> So maybe we need to come to some conclusion about [the behavior of ~] and update the other doc or this one or both, whatever may be.

I changed `~` from the ROS 1 definition to a node name substitution because it could automatically become `/namespace/nodename` during FQN expansion. I really like the python-like `{}` substitutions. What do you think about removing `~` entirely in favor of `{node}`? Then the equivalent of ROS 1 private names would be `{node}/some_basename`.

articles/141_static_remapping.md
sloretz: If I understand correctly I don't think so. `Change the Default Namespace` and `Change the Node Name` are special in that they have to happen before any other remapping, and that's described in the syntax section. I can add something saying the rest of the types of remapping rules have no priority over one another, or are applied in the order in which a user gives the rules regardless of the type of rule.

dirk-thomas: I just found it in the syntax section after your comment. I didn't find it before. So it can stay as-is.

articles/141_static_remapping.md
sloretz: b8c3c69a034a6182429a361b67cd718880ea9d39

I chose to link to the DDS mapping document and condense the "Structure of a Name" section to summarize the terms used in this article. I think this gets rid of the duplicated stuff except for `~`.

articles/141_static_remapping.md
wjwwood: > What do you think about removing ~ entirely in favor of {node}?

(we already talked about this in person since @sloretz asked here, but to summarize...)

Yeah, I believe that was discussed at some point, but the consensus (IIRC) was to leave it since it follows the filesystem metaphor and because it exists in ROS 1. The compromise was to explicitly disallow the `~foo` style syntax that ROS 1 allowed, but that differed from how most systems treat it in a filesystem path. There might be some details here: https://github.com/ros2/design/pull/90#r78666048 and perhaps other locations in that pr.

wjwwood: Also this in the alternatives section:

http://design.ros2.org/articles/topic_and_service_names.html#more-versatile-private-namespace-substitution-character

sloretz: f9429d65a595cc3d77e623a6d74f4a17195754ba

I removed the duplicate/conflicting definition of `~` and stated that substitutions get replaced first when remapping.

articles/141_static_remapping.md
sloretz: I added a section about the order of remapping rules anyways. It was an opportunity to add that rules stack. It's pointless for static remapping, but important for dynamic remapping the visible name.
8cff468696f1881ada90c30a75a88e26e7eee0ee

tfoote (approved):

ros-discourse:

https://discourse.ros.org/t/looking-for-input-on-proposal-for-remapping-names-in-ros-2/1589/1

PR #113, opened 2017-02-23
reword and reorganize

wjwwood:
@Karsten1987 I Moved one or two the the sections around, so the diff isn't very easy to read. I'll go in and do line comments on my changes to point out what I did.

This is targeted at https://github.com/ros2/design/pull/109

### DIFF
--- a/articles/140_topic_and_service_name_mapping.md
+++ b/articles/140_topic_and_service_name_mapping.md
@@ -231,14 +231,14 @@ Therefore we cannot use the array of strings to represent our implied namespace 
 
 #### Hierarchy with DDS Partitions
 
-Since DDS partitions have no hierarchy or order to them array of strings, this proposal would use only one of the strings in the array to hold the entire ROS name's namespace.
-Therefore we preserve the forward slashes (`\`) that exist in the ROS name's namespace already.
+Since DDS partitions have no hierarchy or order to the array of strings, this proposal would use only one of the strings in the array to hold the entire ROS name's namespace.
+Therefore we preserve the forward slashes (`/`) that exist in the ROS name's namespace already.
 
 ### ROS Specific Namespace Prefix
 
 In order to differentiate ROS topics easily, all DDS topics created by ROS shall be automatically prefixed with a namespace like `/rX`, where `X` is a single character that indicates to which subsystem of ROS the topic belongs.
-For example, a plain topic called `/foo` would translate to a DDS partition `/rt/` and a DDS topic `foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name name `foo`.
-As another example, a topic called `/left/image_raw` would translate to a DDS partition `/rt/left/` and a DDS topic `image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name name `image_raw`.
+For example, a plain topic called `/foo` would translate to a DDS partition `/rt/` and a DDS topic `foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name `foo`.
+As another example, a topic called `/left/image_raw` would translate to a DDS partition `/rt/left/` and a DDS topic `image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name `image_raw`.
 
 For systems where Services are implemented with topics (like with OpenSplice), a different subsystem character can be used: `rq` for the request topic and `rr` for the response topic.
 On systems where the implementation is handled for us by DDS (like with Connext), we use `rs` as the common prefix.
--- a/articles/140_topic_and_service_name_mapping.md
+++ b/articles/140_topic_and_service_name_mapping.md
@@ -131,10 +131,10 @@ Topic and service names:
 
 - must not end with a forward slash (`/`)
 
-The last token is the topic or service basename, and any preceding tokens make up the namespace of the topic or service.
+The last token is the topic or service base name, and any preceding tokens make up the namespace of the topic or service.
 
-For example, the topic name `/foo/bar/baz` is composed of a topic or service with the basename `baz` in the `/foo/bar` namespace.
-In another example, the name `/foo` splits into one token, such that it is composed of a topic or service with the basename `foo` in the `/` namespace (the root namespace).
+For example, the topic name `/foo/bar/baz` is composed of a topic or service with the base name `baz` in the `/foo/bar` namespace.
+In another example, the name `/foo` splits into one token, such that it is composed of a topic or service with the base name `foo` in the `/` namespace (the root namespace).
 
 Topic and service names:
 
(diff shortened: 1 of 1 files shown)


articles/140_topic_and_service_name_mapping.md
```diff
@@ -131,10 +131,10 @@ Topic and service names:
 
 - must not end with a forward slash (`/`)
 
-The last token is the topic or service basename, and any preceding tokens make up the namespace of the topic or service.
+The last token is the topic or service base name, and any preceding tokens make up the namespace of the topic or service.
```
wjwwood, line 134: We were using `basename` and `base name` throughout, and I decided to use `base name` because the spell checker was happier with it and the snippet from ROS 1 uses that form.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -194,18 +194,51 @@ For example, given the name `{private}foo`, if there is a substitution called `{
...
+
+The array of strings have no hierarchy and the order does not matter.
+Each entry in the partition array is directly combined with the DDS topic and they are not sequentially combined.
+If a publisher operates on two partition entries, e.g. `foo` and `bar` with a base name of `baz`, this would result in the moral equivalent of two topics: `/foo/baz` and `/bar/baz`
+Therefore we cannot use the array of strings to represent our implied namespace hierarchy and we have to continue to use the forward slash to accomplish this.
+
+#### Hierarchy with DDS Partitions
+
+Since DDS partitions have no hierarchy or order to them array of strings, this proposal would use only one of the strings in the array to hold the entire ROS name's namespace.
+Therefore we preserve the forward slashes (`\`) that exist in the ROS name's namespace already.
```
wjwwood: I maintained as much of these three sections as possible, but I also reworded it a lot. Let me know if you have any questions about it.

The diff isn't great because I also moved it from below so the order of the sections makes more sense.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -194,18 +194,51 @@ For example, given the name `{private}foo`, if there is a substitution called `{
...
+
+### ROS Specific Namespace Prefix
+
+In order to differentiate ROS topics easily, all DDS topics created by ROS shall be automatically prefixed with a namespace like `/rX`, where `X` is a single character that indicates to which subsystem of ROS the topic belongs.
+For example, a plain topic called `/foo` would translate to a DDS partition `/rt/` and a DDS topic `foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name name `foo`.
+As another example, a topic called `/left/image_raw` would translate to a DDS partition `/rt/left/` and a DDS topic `image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name name `image_raw`.
```
wjwwood: I updated the example and added a second one.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -224,69 +257,73 @@ Here is a non-exhaustive list of prefixes:
 While all planned prefixes consist of two characters, i.e. `rX`, anything proceeding the first namespace separator, i.e. `/`, can be considered part of the prefix.
 The standard reserves the right to use up to 8 characters for the prefix in case additional prefix space is needed in the future.
 
-#### Communicating with Non-ROS Topics
+### Examples of ROS Names to DDS Concepts
```
wjwwood, line 260: I moved the examples section up and modified the examples to have a table with DDS topic and partition in them.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -224,69 +257,73 @@ Here is a non-exhaustive list of prefixes:
...
+| `/robot1/camera_left/image_raw` | `image_raw` | [`/rt/robot1/camera_left/`] |
 
-We chose to implement the ROS namespace hierarchy with the help of DDS Partitions.
-DDS Paritions are part of the Subscription/Publisher QoS settings and have been considered as the prefered method of implementing namespaces.
-A subscriber and publisher not only match when their basename topic matches, they also have to be specified on the same partition.
-The big benefit of handling the namespaces with partitions is that it is an official QoS within the DDS specification.
-Furthermore, we can separate a ROS topic hierarchy from a DDS implementation, by forwarding the ROS topic in form of `/ns1/ns2/ns3/base_name` to the DDS implementation, which then takes care about setting up a hierarchy.
-We can therefore make the translation from a ROS Topic (namespaces and basename separated by forward slashes) to a DDS Topic (topic plus paritions) an implementation detail of each RMW implementation.
-This further upholds the flexibility of each alternative to DDS (e.g. ZeroMQ) to realize a different mapping depending on what the middleware provides.
+### Changing ROS Names During Runtime
```
wjwwood, line 270: I reworded this section about remapping so it doesn't refer to remapping, and instead just talks about changing the topic name during runtime (same implications, just avoids the concepts of static or dynamic remapping).

articles/140_topic_and_service_name_mapping.md
```diff
@@ -224,69 +257,73 @@ Here is a non-exhaustive list of prefixes:
...
-```
+The length of the DDS topic must not exceed 255 characters.
+The actual length of a partition field may be limited to 255 characters, however this varies drastically depending on the vendor.
+RTI Connext does not allow a creation of a publisher/subscription with a partition length of 248 characters, whereas FastRTPS does not have any limitation in length.
+Please bare in mind, that the length of the partition gets further diminished due to the introduction of a ros specific prefix.
+The actual length of a ROS Topic, including the namespace hierarchy and the base name of the topic, may thus be varying in length as well.
+Yet, the base name token must not exceed the length of 255 characters as this is getting mapped direclty as the DDS topic.
 
-That implies, that DDS Partitions by default are not depicting any hierarchy.
+### Communicating with Non-ROS Topics
```
wjwwood, line 288: I moved this section here, added notes about how the leading `/` of absolute topics might affect this and then added an examples and ideas section as well.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -314,24 +351,6 @@ In ROS 2, topic and service names differ from ROS 1 in that they:
...
-
-| ROS Name          | Subsystem | DDS Name               |
-|-------------------|-----------|------------------------|
-| `/foo`            | Topic     | `rt__foo`              |
-| `/foo/bar`        | Topic     | `rt__foo__bar`         |
-| `/_foo/bar`       | Topic     | `rt___foo__bar`        |
-| `/foo/_bar`       | Topic     | `rt__foo___bar`        |
-| `/_foo/_bar`      | Topic     | `rt___foo___bar`       |
-| `/_foo/_bar/_baz` | Topic     | `rt___foo___bar___baz` |
-
```
wjwwood, line 334: I replaced these with a section summarizing them below. I moved it down so it is closer to the sections that were alternatives to this (now alternative) proposal.

wjwwood (commented):
I think this is ready, merge at your leisure.

articles/140_topic_and_service_name_mapping.md
Karsten1987: them array of strings? Is that a typo?
Also forward slash ('\') : That's a backslash, no?

articles/140_topic_and_service_name_mapping.md
Karsten1987: base name name?

wjwwood:
Thanks @Karsten1987 I fixed those typos.

PR #115, opened 2017-03-03
create build tool article

dirk-thomas:
This is a draft for an article describing the goal of a "universal" build tool. It describes the current state as well as the rational, goals, and use cases. The proposal section is intentionally left empty for now. It can be filled with the decision we made after some discussion.

It would be great to get feedback on the rational / goals / use cases first so that we are all on the same page before moving forward to deciding which path to take.

(Please feel free to fix spelling / improve wording directly.)

### DIFF
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -1,0 +1,212 @@
+---
+layout: default
+title: A universal build tool
+permalink: articles/build_tool.html
+abstract:
+  This article describes the rational for a universal build tool.
+published: true
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+categories: Overview
+---
+
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Preface
+
+ROS is developed with a federated model.
+It consists of many separately developed and maintained packages.
+
+The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
+For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
+From an efficiency point of view that manual process does not scale well.
+
+This article describes the steps from the current build tools to a single universal build tool which automates the process and works for various different use cases.
+
+## Goal
+
+The goal of the build tool is to build a set of packages with a single invocation automating the process.
+It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
+It would be great if the tool could also be used to build other packages (e.g. Gazebo including its dependencies).
+
+### Out of Scope
+
+The build tool does not cover the steps necessary to fetch the sources of the to-be-build packages.
+This can be done with existing tools (e.g. [rosinstall_generator](http://wiki.ros.org/rosinstall_generator), [vcstool](https://github.com/dirk-thomas/vcstool)).
+
+The build tool also does not provide a mechanism to install any dependencies required to build the packages.
+In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
+
+The build tool also does not create binaries packages (e.g. a Debian package).
+In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generated the required metadata and than platform dependent tools like `dpkg-buildpackage` to build binary packages.
+
+## Build Tool vs. Build System
+
+A build tool operates on a set of packages.
+It determines the dependency graph and invokes the package specific build tool of each package in topological order.
+The build tool itself should know as little as possible about the build system used for a specific package.
+Just enough in order to know how to setup the environment for it, invoke the build, and setup the environment to use the built package.
+
+The build system on the other hand operates on a single package.
+In the case a package uses e.g. `CMake` the responsibility of the tool comes down to invoke the common steps `cmake`, `make`, `make install` for this package.
+As another example for a package using `Autotools` the steps could look like `configure`, `make`, `make install`.
+
+### Environment Setup
+
+A very important part beside the actual build of a package is the environment setup.
+For a CMake package e.g. a dependency discovered using the CMake function `find_package` needs to be either in a common location or the location must be provided through the environment variable `CMAKE_PREFIX_PATH`.
+On the other hand after a package has been build and installed the environment might need to be extended to use the package.
+E.g. when a package installs a shared library in a non-default location the environment variable `LD_LIBRARY_PATH` (or `PATH` on Windows) needs to be extended to include the containing folder.
+
+The functionality to setup these environment variables can be provided by either the build tool or the build system.
+In the later case the build tool only needs to know how the build system exposes the environment setup in order to reuse it.
+
+Considering the use case that a user might want to invoke the build system of each package manually it is beneficial if the build system already provides as much of the environment setup as possible.
+That avoids that the user has to manually take care of the environment setup when not using a build tool.
+
+## Existing Build Systems
+
+In the following the build systems being used in the ROS ecosystem are briefly described.
+
+### catkin
+
+[catkin](http://wiki.ros.org/catkin) is based on CMake and provides a set of convenience functions to make writing CMake package easier.
+It automates the generation of CMake config files as well as pkg-config files.
+It additionally provides functions to register different kinds of tests.
+
+### ament_cmake
+
+[ament_cmake](https://github.com/ament/ament_cmake) is an evolution of `catkin` and is also based on CMake.
+The main difference between `ament_cmake` and `catkin` is described in [another article](http://design.ros2.org/articles/ament.html).
+In the context of the build tool the biggest difference is that `ament_cmake` generates package-specific files to setup the environment to use the package after it has been built and installed.
+
+### Python setuptools
+
+`setuptools` is the common tool to package Python packages.
+A Python package use a `setup.py` file to describe the dependencies as well as how and what to build and install.
+In ROS 2 a package can be a "vanilla" Python package whereas in ROS 1 any Python functionality is triggered from a CMake file.
+
+## Existing Build Tools
+
+Several different build tools are already being used in the ROS ecosystem.
+Their method of operating is being described in the following subsections together with their advantages as well as disadvantages.
+
+### catkin_make
+
+`catkin_make` is provided by the ROS package `catkin` which contains the build system for ROS 1.
+It has been designed as the successor of `rosbuild` for ROS Fuerte.
+
+The tool invoked CMake only a single time and uses CMake's `include` function to process all packages in a single context.
+While this is an efficient approach since all targets across all packages can be parallelized it comes with a significant disadvantages.
+Due to the single context all function names, targets and tests share a single namespace across all packages and on a larger scale easily leads to collisions.
+The single context is also subject to side effect between the packages and sometimes requires to add additional target dependencies across package boundaries.
+
+### catkin_make_isolated
+
+`catkin_make_isolated` is provided by the ROS package `catkin` which contains the build system for ROS 1.
+It was developed after `catkin_make` to address the problems involved with building multiple packages in a single CMake context.
+
+The tool only supports CMake-based packages and builds each package in topological order using the command sequence common for CMake packages: `cmake`, `make`, `make install`.
+While each package can parallelize the build of its targets the packages are processed sequentially even if they are not (recursive) dependencies of each other.
+
+### ament_tools
+
+`ament_tools` is provided by a standalone Python 3 package used to build ROS 2 packages.
+It was developed to bootstrap the ROS 2 project, is therefore only targeting Python 3, and works on Linux, MacOS and Windows.
+In addition to CMake packages it also supports building Python packages and can infer meta information without requiring an explicit package manifest.
+The tool performs an "isolated" build like `catkin_make_isolated` (one CMake invocation per package) but also parallelizes the build of packages which have no (recursive) dependencies on each other.
+
+### catkin_tools
+
+[catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build ROS 1 packages.
+It was developed after `catkin_make` / `catkin_make_isolated` to build packages in parallel as well as provide significant usability improvements.
+The tool supports building CMake packages and builds them in isolation as well as support parallelizing the process across packages.
+
+### Naming
+
+The existing build tools are all named by the build system they are supporting.
+In general it should be possible for a build tool to support multiple different build systems.
+Therefore a name for a build tool being derive from a single build system might mislead the users that the tool only works for that specific build system.
+To avoid confusion of the user it would be beneficial if the build tool would have a different name to avoid implying an undesired correlation.
+
+## Rational
+
+Currently each of the existing tools performs similar tasks and duplicates a significant amount of the logic.
+Certain features are only available in some of the tools while other tools lack those.
+
+The reason to work on a single universal build tool comes down to reduce the effort necessary for development and maintenance.
+Additionally this makes new features developed for one ROS version / build system available to the other supported ROS versions / build systems.
+
+## Requirements
+
+The unified build tool should provide a superset of the functionality provided by the existing tools.
+In the following a few use cases are described as well as desired software criteria.
+
+### Use Cases
+
+The following uses cases should be satisfied by the unified build tool.
+
+#### Build ROS 1 workspaces
+
+The tool needs to be able to build ROS 1 workspaces which can already be built using `catkin_make_isolated` / `catkin_tools`.
+
+#### Build ROS 2 workspaces
+
+The tool needs to be able to build ROS 2 workspaces which can already be built using `ament_tools`.
+
+#### Build Gazebo including dependencies
+
+After cloning the repositories containing Gazebo and all its dependencies (excluding system packages) the tool needs to be able to build the set of packages.
+Meta information not inferable from the sources can be provided externally without adding or modifying any files in the workspace.
+After the build a single file can be sourced / invoked to setup the environment to use Gazebo (e.g. `GAZEBO_MODEL_PATH`).
+
+## Software Criteria
+
+The tool aims to support a variety of build systems, use cases, and platforms.
+The above mentioned ones are mainly driven by the needs in the ROS ecosystem but the tool should also be usable outside the ROS ecosystem (e.g. for Gazebo).
+Therefore it should be designed in a way which enables extending its functionality.
+
+Assuming that the tool will be implemented in Python (since that is the case for existing tools) the entry point mechanism provides a convenient way to make the software extensible.
+Extensions don't even have to be integrated into the Python package containing the core logic of the build tool but can easily be provided by additional Python packages.
+This approach will not only foster a modular design and promote clear interfaces but enable external contributions without requiring them to be integrated in a single monolithic package.
+
+Several well known software principles apply:
+
+- Separation of concerns
+- Single Responsibility principle
+- Principle of Least Knowledge
+- Don’t repeat yourself
+- Keep it stupid simple
+- "Not paying for what you don't use"
+
+### Extension Points
+
+The following items are possible extension points to provide custom functionality:
+
+- contribute `verbs` to the command line tool (e.g. `build`, `test`)
+- contribute command line options for specific features (e.g. `build`, `test`)
+- discovery of packages (e.g. recursively crawling a workspace)
+- identification of packages and their meta information (e.g. from a `package.xml` file)
+- process a package (e.g. build a CMake package, test a Python package)
+- execution model (e.g. sequential processing, parallel processing)
+- output handling (e.g. console output, logfiles, status messages, notifications)
+- setup the environment (e.g. `sh`, `bash`, `bat`)
+- completion (e.g. `bash`, `Powershell`)
+
+## Alternative Approaches
+
+Their are two different approaches possible to reach the goal of a universal build tool.
+
+- Incrementally evolve one of the existing tools to satisfy the described goals.
+- Implement a new build tool "from scratch" (while existing building blocks can still be "copied" from the existing solutions).
+
+## Proposal
+
+*to be decided*
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -21,62 +21,49 @@ Original Author: {{ page.author }}
 
 ## Preface
 
-ROS is developed with a federated model.
-It consists of many separately developed and maintained packages.
+In the ROS ecosystem the software is separated into numerous packages.
+It is very common that a developer is working on multiple packages at the same time.
+This is in contrast to workflows where a developer only works on a single software package at a time, and all dependencies are provided once but not being iterated on.
 
 The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
 For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
-From an efficiency point of view that manual process does not scale well.
+Such a workflow is impracticable at scale without a tool that automates that process.
+
+A build tool performs the task of building a set of packages with a single invocation.
+For ROS 1 multiple different tools provide support for this, namely `catkin_make`, `catkin_make_isolated`, and `catkin_tools`.
+For ROS 2 the build tool providing this functionality is called `ament_tools`.
+
+This article describes the steps to unify these build tools as well as extend the field of application.
 
 ## Goal
 
-The goal of the build tool is to build a set of packages with a single invocation automating the process.
+The goal of a unified build tool is to build a set of packages with a single invocation.
 It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
 It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
-This will allow the build tool to be utilized for non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
+This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its dependencies).
 
 In the ROS ecosystems several tools already exist which support this use case (see below).
 Each of the existing tools performs similar tasks and duplicates a significant amount of the logic.
 As a consequence of being developed separately certain features are only available in some of the tools while other tools lack those.
 
 The reason to work on a single universal build tool comes down to reducing the effort necessary for development and maintenance.
-Additionally this makes new features developed for one ROS version / build system available to the other supported ROS versions / build systems.
+Additionally this makes new features developed once available for all the use cases.
 
-### Out of Scope
+### Build Tool vs. Build System
 
-The build tool does not cover the steps necessary to fetch the sources of the to-be-built packages.
-There are already tools to help with this.
-For example, the list of repositories that need to be fetched is provided either by a hand crafted `.rosinstall` or `.repos` file or by using [rosinstall_generator](http://wiki.ros.org/rosinstall_generator) to generate one.
-The list of repositories can then be fetched with one of several tools, like [rosinstall](http://wiki.ros.org/rosinstall) or [wstool](http://wiki.ros.org/wstool) in the case of a `.rosinstall` file, or [vcstool](https://github.com/dirk-thomas/vcstool) in the case of a `.repos` file.
-
-The build tool also does not provide a mechanism to install any dependencies required to build the packages.
-In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
-
-The build tool also does not create binary packages (e.g. a Debian package).
-In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generate the required metadata and then platform dependent tools like `dpkg-buildpackage` build binary packages.
-
-## Build Tool vs. Build System
+Since this article focuses on the build tool the distinction to a build system needs to be clarified.
 
 A build tool operates on a set of packages.
-It determines the dependency graph and invokes the specific build tool for each package in topological order.
+It determines the dependency graph and invokes the specific build system for each package in topological order.
 The build tool itself should know as little as possible about the build system used for a specific package.
 Just enough in order to know how to setup the environment for it, invoke the build, and setup the environment to use the built package.
+The existing ROS build tools are: `catkin_make`, `catkin_make_isolated`, `catkin_tools`, and `ament_tools`.
 
 The build system on the other hand operates on a single package.
-In the case a package uses e.g. `CMake` the responsibility of the tool comes down to invoke the common steps `cmake`, `make`, `make install` for this package.
-As another example for a package using `Autotools` the steps could look like `configure`, `make`, `make install`.
+Examples are `Make`, `CMake`, `Python setuptools`, or `Autotools` (which isn't used in ROS atm).
+A CMake package is e.g. build by invoking these steps: `cmake`, `make`, `make install`.
 
-### Examples of Build Tools vs. Build Systems
-
-|                                             | Build System | Build Tool |
-|---------------------------------------------|:------------:|:----------:|
-| CMake project                               |       x      |            |
-| catkin (CMake that calls catkin macros)     |       x      |            |
-| ament_cmake (CMake that calls ament macros) |       x      |            |
-| catkin_make                                 |              |      x     |
-| catkin_make_isolated                        |              |      x     |
-| catkin_tools                                |              |      x     |
-| ament_tools                                 |              |      x     |
+`catkin` as well as `ament_cmake` are based on CMake and offer some convenience functions described below.
 
 ### Environment Setup
 
@@ -92,9 +79,46 @@ In the latter case the build tool only needs to know how the build system expose
 Considering the use case that a user might want to invoke the build system of each package manually it is beneficial if the build system already provides as much of the environment setup as possible.
 That avoids forcing the user to manually take care of the environment setup when not using a build tool.
 
+### Out of Scope
+
+To clarify the scope of this article a few related topics are explicitly enumerated even though they are not being considered.
+
+#### Build System
+
+Any build system related functionality (which is not directly relevant for the build tool) is not considered in this article.
+
+##### Mixing Different Build Systems
+
+The unified build tool will support different build systems in order to satisfy the described goals.
+If packages using different build system inter-operate with each other correctly depends also to a large degree on the build system.
+While the build tool should ensure that it doesn't prevent that use case this article will not cover the use case of mixing multiple build systems in a single workspace (e.g. ROS 1 packages using `catkin` with ROS 2 packages using `ament_cmake`).
+
+#### Fetch Source Code
+
+The build tool does not cover the steps necessary to fetch the sources of the to-be-built packages.
+There are already tools to help with this.
+For example, the list of repositories that need to be fetched is provided either by a hand crafted `.rosinstall` or `.repos` file or by using [rosinstall_generator](http://wiki.ros.org/rosinstall_generator) to generate one.
+The list of repositories can then be fetched with one of several tools, like [rosinstall](http://wiki.ros.org/rosinstall) or [wstool](http://wiki.ros.org/wstool) in the case of a `.rosinstall` file, or [vcstool](https://github.com/dirk-thomas/vcstool) in the case of a `.repos` file.
+
+#### Install Dependencies
+
+The build tool also does not provide a mechanism to install any dependencies required to build the packages.
+In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
+
+#### Create Binary Packages
+
+The build tool also does not create binary packages (e.g. a Debian package).
+In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generate the required metadata and then platform dependent tools like `dpkg-buildpackage` build binary packages.
+
 ## Existing Build Systems
 
 In the following the build systems being used in the ROS ecosystem are briefly described.
+
+### CMake
+
+[CMake](https://cmake.org/) is a cross-platform build system generator.
+Projects specify their build process with platform-independent `CMakeLists.txt` files.
+Users build a project by using CMake to generate a build system for a native tool on their platform, e.g. `Makefiles` or `Visual Studio projects`.
 
 ### catkin
 
@@ -102,11 +126,16 @@ In the following the build systems being used in the ROS ecosystem are briefly d
 It automates the generation of CMake config files as well as pkg-config files.
 It additionally provides functions to register different kinds of tests.
 
+A package using `catkin` specifies its meta data in a manifest file named `package.xml`.
+The format of the manifest file is specified in the [ROS REP 140](http://www.ros.org/reps/rep-0140.html).
+
 ### ament_cmake
 
 [ament_cmake](https://github.com/ament/ament_cmake) is an evolution of `catkin` and is also based on CMake.
 The main difference between `ament_cmake` and `catkin` is described in [another article](http://design.ros2.org/articles/ament.html).
 In the context of the build tool the biggest difference is that `ament_cmake` generates package-specific files to setup the environment to use the package after it has been built and installed.
+
+A package using `ament_cmake` uses the same manifest file as `catkin` (except that it only allows the newer format version 2).
 
 ### Python setuptools
 
@@ -183,11 +212,18 @@ After cloning the repositories containing Gazebo and all its dependencies (exclu
 Meta information not inferable from the sources can be provided externally without adding or modifying any files in the workspace.
 After the build a single file can be sourced / invoked to setup the environment to use Gazebo (e.g. `GAZEBO_MODEL_PATH`).
 
-#### Mixing different build systems
+### Development Environment Setup
 
-The build tool will support using different build systems within a single workspace.
-If these packages inter-operate with each other correctly depends also to a large degree on the build system.
-The build tool should ensure that it doesn't prevent that use case.
+Invoking a build system for a package implies also setting up environment variables before the process, e.g. the `CMAKE_PREFIX_PATH`.
+It should be possible for developers to manually invoke the build system for one package.
+The environment variable might be partially different from the environment variables necessary to use a package after it has been built.
+To make that convenient the tool should provide an easy to use mechanism to setup the development environment necessary to invoke the build system.
+
+### Beyond Building
+
+Building packages is only one task the build tool can perform on a set of packages.
+Additional tasks like e.g. running tests should also be covered by the build tool.
+The build tool must provide these abstract tasks and then map them to the steps necessary for each supported build system.
 
 ### Software Criteria
 
@@ -221,6 +257,10 @@ The following items are possible extension points to provide custom functionalit
 - output handling (e.g. console output, logfiles, status messages, notifications)
 - setup the environment (e.g. `sh`, `bash`, `bat`)
 - completion (e.g. `bash`, `Powershell`)
+
+Assuming that the tool will be implemented in Python (since that is the case for existing tools) the entry point mechanism provides a convenient way to make the software extensible.
+Extensions don't have to be integrated into the Python package containing the core logic of the build tool but can easily be provided by additional Python packages.
+This approach will not only foster a modular design and promote clear interfaces but enable external contributions without requiring them to be integrated in a single monolithic package.
 
 ## Possible Approaches
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -76,7 +76,7 @@ As another example for a package using `Autotools` the steps could look like `co
 ### Environment Setup
 
 A very important part beside the actual build of a package is the environment setup.
-For example, in order for a CMake project to discover a dependency using the CMake function `find_package`, the CMake module (e.g. `FindFoo.cmake`) or the CMake config file (e.g. `FooConfig.cmake`) for that dependency must either be in a path that CMake searches implicitly (e.g. `/usr/share/Foo/cmake`) or the location must be provided through the environment variable `CMAKE_PREFIX_PATH`.
+For example, in order for a CMake project to discover a dependency using the CMake function `find_package`, the CMake module (e.g. `FindFoo.cmake`) or the CMake config file (e.g. `FooConfig.cmake`) for that dependency must either be in a prefix that CMake searches implicitly (e.g. `/usr`) or the location must be provided through the environment variable `CMAKE_PREFIX_PATH` / `CMAKE_MODULE_PATH`.
 
 In addition to building a package on top of another package (using `find_package` in the case of CMake), you may need to adjust the environment in order to run an executable from a package.
 For example, when a package installs a shared library in a non-default location then the environment variable `LD_LIBRARY_PATH` (or `PATH` on Windows) needs to be extended to include the containing folder before trying to run executables that load that library at runtime.
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -34,7 +34,8 @@ This article describes the steps from the current build tools used in the ROS ec
 
 The goal of the build tool is to build a set of packages with a single invocation automating the process.
 It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
-It would be great if the tool could also be used to build other packages (e.g. Gazebo including its dependencies).
+It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
+This will allow the build tool to be utilized for non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
 
 ### Out of Scope
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -189,7 +189,7 @@ While each package can parallelize the build of its targets the packages are pro
 It was developed after `catkin_make` / `catkin_make_isolated` to build packages in parallel as well as provide significant usability improvements.
 The tool supports building CMake packages and builds them in isolation as well as supports parallelizing the process across packages.
 
-`catkin_make_isolated` supports building the following packages:
+`catkin_tools` supports building the following packages:
 
 - ROS 1 `catkin` package with a `package.xml` file.
 - Plain CMake packages with a `package.xml` file.
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -3,7 +3,7 @@ layout: default
 title: A universal build tool
 permalink: articles/build_tool.html
 abstract:
-  This article describes the rational for a universal build tool.
+  This article describes the rationale for a universal build tool.
 published: true
 author: '[Dirk Thomas](https://github.com/dirk-thomas)'
 categories: Overview
@@ -38,14 +38,14 @@ It would be great if the tool could also be used to build other packages (e.g. G
 
 ### Out of Scope
 
-The build tool does not cover the steps necessary to fetch the sources of the to-be-build packages.
+The build tool does not cover the steps necessary to fetch the sources of the to-be-built packages.
 This can be done with existing tools (e.g. [rosinstall_generator](http://wiki.ros.org/rosinstall_generator), [vcstool](https://github.com/dirk-thomas/vcstool)).
 
 The build tool also does not provide a mechanism to install any dependencies required to build the packages.
 In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
 
-The build tool also does not create binaries packages (e.g. a Debian package).
-In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generated the required metadata and than platform dependent tools like `dpkg-buildpackage` to build binary packages.
+The build tool also does not create binary packages (e.g. a Debian package).
+In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generate the required metadata and then platform dependent tools like `dpkg-buildpackage` build binary packages.
 
 ## Build Tool vs. Build System
 
@@ -62,11 +62,11 @@ As another example for a package using `Autotools` the steps could look like `co
 
 A very important part beside the actual build of a package is the environment setup.
 For a CMake package e.g. a dependency discovered using the CMake function `find_package` needs to be either in a common location or the location must be provided through the environment variable `CMAKE_PREFIX_PATH`.
-On the other hand after a package has been build and installed the environment might need to be extended to use the package.
+On the other hand after a package has been built and installed the environment might need to be extended to use the package.
 E.g. when a package installs a shared library in a non-default location the environment variable `LD_LIBRARY_PATH` (or `PATH` on Windows) needs to be extended to include the containing folder.
 
 The functionality to setup these environment variables can be provided by either the build tool or the build system.
-In the later case the build tool only needs to know how the build system exposes the environment setup in order to reuse it.
+In the latter case the build tool only needs to know how the build system exposes the environment setup in order to reuse it.
 
 Considering the use case that a user might want to invoke the build system of each package manually it is beneficial if the build system already provides as much of the environment setup as possible.
 That avoids that the user has to manually take care of the environment setup when not using a build tool.
@@ -77,7 +77,7 @@ In the following the build systems being used in the ROS ecosystem are briefly d
 
 ### catkin
 
-[catkin](http://wiki.ros.org/catkin) is based on CMake and provides a set of convenience functions to make writing CMake package easier.
+[catkin](http://wiki.ros.org/catkin) is based on CMake and provides a set of convenience functions to make writing CMake packages easier.
 It automates the generation of CMake config files as well as pkg-config files.
 It additionally provides functions to register different kinds of tests.
 
@@ -90,7 +90,7 @@ In the context of the build tool the biggest difference is that `ament_cmake` ge
 ### Python setuptools
 
 `setuptools` is the common tool to package Python packages.
-A Python package use a `setup.py` file to describe the dependencies as well as how and what to build and install.
+A Python package uses a `setup.py` file to describe the dependencies as well as how and what to build and install.
 In ROS 2 a package can be a "vanilla" Python package whereas in ROS 1 any Python functionality is triggered from a CMake file.
 
 ## Existing Build Tools
@@ -103,10 +103,10 @@ Their method of operating is being described in the following subsections togeth
 `catkin_make` is provided by the ROS package `catkin` which contains the build system for ROS 1.
 It has been designed as the successor of `rosbuild` for ROS Fuerte.
 
-The tool invoked CMake only a single time and uses CMake's `include` function to process all packages in a single context.
-While this is an efficient approach since all targets across all packages can be parallelized it comes with a significant disadvantages.
-Due to the single context all function names, targets and tests share a single namespace across all packages and on a larger scale easily leads to collisions.
-The single context is also subject to side effect between the packages and sometimes requires to add additional target dependencies across package boundaries.
+The tool invokes CMake only a single time and uses CMake's `include` function to process all packages in a single context.
+While this is an efficient approach since all targets across all packages can be parallelized it comes with significant disadvantages.
+Due to the single context all function names, targets and tests share a single namespace across all packages and on a larger scale this easily leads to collisions.
+The single context is also subject to side effects between the packages and sometimes requires to add additional target dependencies across package boundaries.
 
 ### catkin_make_isolated
 
@@ -127,21 +127,21 @@ The tool performs an "isolated" build like `catkin_make_isolated` (one CMake inv
 
 [catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build ROS 1 packages.
 It was developed after `catkin_make` / `catkin_make_isolated` to build packages in parallel as well as provide significant usability improvements.
-The tool supports building CMake packages and builds them in isolation as well as support parallelizing the process across packages.
+The tool supports building CMake packages and builds them in isolation as well as supports parallelizing the process across packages.
 
 ### Naming
 
 The existing build tools are all named by the build system they are supporting.
 In general it should be possible for a build tool to support multiple different build systems.
-Therefore a name for a build tool being derive from a single build system might mislead the users that the tool only works for that specific build system.
+Therefore a name for a build tool being derived from a single build system might mislead the users that the tool only works for that specific build system.
 To avoid confusion of the user it would be beneficial if the build tool would have a different name to avoid implying an undesired correlation.
 
-## Rational
+## Rationale
 
 Currently each of the existing tools performs similar tasks and duplicates a significant amount of the logic.
 Certain features are only available in some of the tools while other tools lack those.
 
-The reason to work on a single universal build tool comes down to reduce the effort necessary for development and maintenance.
+The reason to work on a single universal build tool comes down to reducing the effort necessary for development and maintenance.
 Additionally this makes new features developed for one ROS version / build system available to the other supported ROS versions / build systems.
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


tkruse:
* It would be nice to announce this in a mailing list. I am not sure which one, https://groups.google.com/forum/#!forum/ros-sig-buildsystem or discourse
* It would be nice to link to past discussions / articles about buildsystems in ROS (I can dig up links if that is desired)
* Since this is not a REP, it would be nice to explain the decision-making process / planned timeline
* Overall I missed some mentioning of ROS2 message generation / lower-middleware flexibility and why this could not be done easily with catkin. I assume this is an important force in the decision making, maybe that's wrong.

dirk-thomas:
> It would be nice to link to past discussions / articles about buildsystems in ROS (I can dig up links if that is desired)

That would be great to add. Since I am not sure which discussions you are referring to can you please either comment with the plain links here and I am happy to add them "somewhere" to the article or make a PR against this branch to make specific suggestions where / how to reference them. Thanks!

> Since this is not a REP, it would be nice to explain the decision-making process / planned timeline

The first step is to get feedback on the "facts" and "goal" parts of the current draft. After that we have to describe possible paths forward in more details and gathering the pros and cons of each. Then we have to come up with a proposal. Only with these parts being "finished" I think a time estimate can be made.

> It would be nice to announce this in a mailing list. I am not sure which one, https://groups.google.com/forum/#!forum/ros-sig-buildsystem or discourse

I certainly plan to announce this on the mailing list to receive feedback from a wider audience. For now I will wait a little bit until the article contains more information. I think it needs at least some of the missing information described in the previous paragraph to be helpful.

> Overall I missed some mentioning of ROS2 message generation / lower-middleware flexibility and why this could not be done easily with catkin. I assume this is an important force in the decision making, maybe that's wrong.

I am not sure if I understand your comment. The article is about the build tool. It is not related to any message generation or middleware at all. As mentioned in the article this tool should even be usable for non-ROS use cases like Gazebo. Can you please clarify your feedback on this.

tkruse:
Regarding ament, I mixed up something, ignore my comment.

sbarthelemy:
Hi,

> catkin allows users to build multiple packages within a single CMake context (using catkin_make).
> While this is an efficient approach since all targets across all packages can be parallelized it comes
> with a significant disadvantages.
> Due to the single context all function names, targets and tests share a single namespace across all
> packages and on a larger scale easily leads to collisions.
> [...] ament does not provide that feature.

the "single CMake context" seems especially valuable to me during development of two projects/packages (eg. a lib and a closely related service): CMake will nicely parallelize incremental rebuilds, and also partial rebuilds:

```
while not fixed:
  vi this_buggy_library.cpp
  make this_service_failing_test && bin/this_service_failing_test
```

It would be great to keep support for this use case.

The point is not to build  all the packages in a single CMake context (which indeed does not scale, especially with a federated development model), but to enable *some of them* to be in the same CMake context.

dirk-thomas:
@sbarthelemy This article is only about the build tool which does not make that decision. The build system `ament_cmake` is covered in a separate article which describes the rational why support for a single CMake context is being dropped: http://design.ros2.org/articles/ament.html#building-within-a-single-cmake-context I just created #116 to elaborate on the rational a bit more.

Basically it comes down to the following: if you build multiple packages within a single CMake context their decoupling is basically lost. The packages need to know about the internal structure of their dependencies: avoid duplicate symbols, avoid side effects, declare cross-package target level dependencies. At that point you might want to put those tightly coupled packages into a single package.

Also when considering the bigger picture the user trying to build a set of packages has no way of knowing if he can use this approach / if it is supported by the set of packages or not.

sbarthelemy:
@dirk-thomas 

> Basically it comes down to the following: if you build multiple packages within a single CMake context their decoupling is basically lost.

I agree on this.

> At that point you might want to put those tightly coupled packages into a single package.

I agree on this too.

Still, my experience is that  there is pressure to split packages: separate interface from implementation, have finer package-level dependency descriptions (users of a package don't want to compile what they won't use, to install what they won't need),...
In the end of the day, packages tend to be smaller than the "comfortable development size".
The shared CMake context was a useful workaround.

> Also when considering the bigger picture the user trying to build a set of packages has no way of knowing if he can use this approach / if it is supported by the set of packages or not.

Agreed.

Maybe a better solution would be to let a single build produce several packages. A bit like debian packages: a single source package build creates several binary packages (if I recall correctly).

Thus a big single source package might produce several tightly (coupled) binary packages.

dirk-thomas:
> Maybe a better solution would be to let a single build produce several packages. A bit like debian packages: a single source package build creates several binary packages (if I recall correctly).

> Thus a big single source package might produce several tightly (coupled) binary packages.

But where would the necessary information for how to split the package come from? Also each subpart would require the following meta information for packaging: name, version, maintainer, changelog, dependencies etc. (which is exactly what a package.xml file specifies).

For Debian this process is performed by a human by hand which is quite some work. I doubt any ROS maintainer would like to spend that much time to release a single package.

tkruse:
This kind of discussion is not ideal on github. Rosbuild had the concept of stacks and packages. A stack was something like all packages living in a single VCS root. Those could in theory be build cleanly (enough) in one cmake context, but released individually. There should be an old thread about this.

sbarthelemy:
@tkruse 
> This kind of discussion is not ideal on github.

indeed, and the matter is not really related to this pull request. Is there a better place?

@dirk-thomas 
> But where would the necessary information for how to split the package come from?

I'll try to find the time to make up a real life example.

dirk-thomas:
> > This kind of discussion is not ideal on github.

> indeed, and the matter is not really related to this pull request. Is there a better place?

I don't think there is a discourse category for it yet. Hence I would recommend posting to the buildsystem mailing list: https://groups.google.com/forum/#!forum/ros-sig-buildsystem

sloretz:
The naming section sounds like `ament_tools` would stop being developed in favor of a new build tool with a name that didn't sound similar to `ament_cmake`. I think it would be fine for the tool to share a name with a build system. To a lot of developers the "tool" is just the commands they've memorized to build their code. Knowing the build tool is a different from the build system doesn't change their workflow.

Adding another build tool might increase the maintenance workload. The existing build tools `catkin_make`, `catkin_make_isolated`, and `catkin_tools` are all commonly used in ROS 1. While there is a goal to not disturb ROS 1, if the tool works with ROS 1 packages I think deprecating those tools in favor of the new one would go a long ways towards consolidating the work to the new tool.

Extensions adding extra flags to verbs might be too small to be worth supporting. Flags tend to be small changes to behavior. Supporting extending these means adding hooks throughout the logic without really knowing where hooks will be useful to future extension writers, or encouraging python monkey patching. If someone really wants a small change in behavior then it not being extensible may increase the chance they modify the tool and contribute the change upstream.

articles/101_build_tool.md
```diff
@@ -0,0 +1,212 @@
...
+## Preface
+
+ROS is developed with a federated model.
+It consists of many separately developed and maintained packages.
+
+The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
+For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
+From an efficiency point of view that manual process does not scale well.
+
+This article describes the steps from the current build tools to a single universal build tool which automates the process and works for various different use cases.
```
dhood: "current build tools" -> "current build tools used in the ROS ecosystem" (otherwise the scope is not clear)

articles/101_build_tool.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Build ROS 1 workspaces
+
+The tool needs to be able to build ROS 1 workspaces which can already be built using `catkin_make_isolated` / `catkin_tools`.
+
+#### Build ROS 2 workspaces
+
+The tool needs to be able to build ROS 2 workspaces which can already be built using `ament_tools`.
+
+#### Build Gazebo including dependencies
```
dhood, line 239: From the 'goal' section above I understood that this was just a 'nice-to-have'. It would be good to either mention again that this is a nice-to-have, or explain why it is that this is a requirement

dhood (commented):
this reads well to me

articles/101_build_tool.md
dirk-thomas: Good point: done in ae668e5b5659b765319161d1e3f3a4b2f8ed0cc7

articles/101_build_tool.md
dirk-thomas: In the goal section it was more meant as "since we already have all this functionality it should also work for this case". Would you have a suggestion to clarify that so that it doesn't look like a nice-to-have?

dirk-thomas:
@sloretz

> Adding another build tool might increase the maintenance workload. The existing build tools catkin_make, catkin_make_isolated, and catkin_tools are all commonly used in ROS 1. While there is a goal to not disturb ROS 1, if the tool works with ROS 1 packages I think deprecating those tools in favor of the new one would go a long ways towards consolidating the work to the new tool.

That is certainly the goal. The "new unified" build tool should work for both ROS 1 as well as ROS 2 (and hopefully other use cases) and we would recommend it to be used for both ROS versions. For backward compatibility the current ROS 1 tools will still be around but we wouldn't plan to spend any time on them (except urgent bug fixes of course).

> The naming section sounds like ament_tools would stop being developed in favor of a new build tool with a name that didn't sound similar to ament_cmake. I think it would be fine for the tool to share a name with a build system. To a lot of developers the "tool" is just the commands they've memorized to build their code. Knowing the build tool is a different from the build system doesn't change their workflow.

Since the build tool is supposed to be used with multiple different build system which have different names I think the association with one of them by name is to be avoided (since users can a) often not distinguish what is build tool and what is build system and b) expect a build tool called `ament_tools` not to work for `catkin`).

> Extensions adding extra flags to verbs might be too small to be worth supporting. Flags tend to be small changes to behavior. Supporting extending these means adding hooks throughout the logic without really knowing where hooks will be useful to future extension writers, or encouraging python monkey patching. If someone really wants a small change in behavior then it not being extensible may increase the chance they modify the tool and contribute the change upstream.

After having developed quite a few build systems and build tools in the past I have to disagree with you on this. A verb like `build` has to handle **a lot** of different things. Just take a look at the command line arguments for it in `catkin_tools` (https://github.com/catkin/catkin_tools/blob/master/catkin_tools/verbs/catkin_build/cli.py#L94-L193). The number of different aspects it has to handle is enormous. And I am sure people can easily come of with more ideas which it should handle additionally. In my opinion this is the exact opposite of the goal of the single responsibility principle. (Contributing to this code base sets a pretty high bar. I would expect a more modular code base to attract more external contributions.)

articles/101_build_tool.md
```diff
@@ -0,0 +1,212 @@
...
+For a CMake package e.g. a dependency discovered using the CMake function `find_package` needs to be either in a common location or the location must be provided through the environment variable `CMAKE_PREFIX_PATH`.
+On the other hand after a package has been built and installed the environment might need to be extended to use the package.
+E.g. when a package installs a shared library in a non-default location the environment variable `LD_LIBRARY_PATH` (or `PATH` on Windows) needs to be extended to include the containing folder.
+
+The functionality to setup these environment variables can be provided by either the build tool or the build system.
+In the latter case the build tool only needs to know how the build system exposes the environment setup in order to reuse it.
+
+Considering the use case that a user might want to invoke the build system of each package manually it is beneficial if the build system already provides as much of the environment setup as possible.
+That avoids that the user has to manually take care of the environment setup when not using a build tool.
+
```
clalancette: This might be better worded as:

This avoids forcing the user to manually take care of the environment setup when not using a build tool.

articles/101_build_tool.md
```diff
@@ -0,0 +1,212 @@
...
+### catkin_make
+
+`catkin_make` is provided by the ROS package `catkin` which contains the build system for ROS 1.
+It has been designed as the successor of `rosbuild` for ROS Fuerte.
+
+The tool invokes CMake only a single time and uses CMake's `include` function to process all packages in a single context.
+While this is an efficient approach since all targets across all packages can be parallelized it comes with significant disadvantages.
+Due to the single context all function names, targets and tests share a single namespace across all packages and on a larger scale this easily leads to collisions.
+The single context is also subject to side effects between the packages and sometimes requires to add additional target dependencies across package boundaries.
+
```
clalancette: Better worded as:

The single context is also subject to side effects between the packages and sometimes requires adding additional target dependencies across package boundaries.

clalancette:
This looks pretty good to me.  There will definitely have to be more clarification in the future as things evolve, but I think this is a good start.

dirk-thomas:
> There will definitely have to be more clarification in the future as things evolve

Can you please clarify which aspects you have in mind?

clalancette:
In particular, the part at the end for the Proposal section, which is TBD.

articles/101_build_tool.md
dhood: please see https://github.com/ros2/design/pull/118

mikaelarguedas:
All my minor comments and questions have been addressed by previous feedback, hence my +1 on the first comment.

Few points that could be clarified (maybe directly in the proposal):
- What do we mean by multi-platform ? The platforms targeted by ament natively will be supported by the core of the tool? and then users can use extension points to add support for other platforms?
- In the goal section: does "work with ROS1 packages as well as ROS2 packages" implies that no modification of their CMake will be necessary, right?
- While it's possible to do cross-compilation with some of the existing tools it doesn't appear in this document at all, should this be taken into account in this document?
- Will the proposal specify a way to [provide meta information externally](https://github.com/ros2/design/pull/115/commits/634dd8fe23cfcc9b1d7281e9efabdd6a73c9215e#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR37)?

dirk-thomas:
> What do we mean by multi-platform ? The platforms targeted by ament natively will be supported by the core of the tool? and then users can use extension points to add support for other platforms?

The "unified" build should certainly cover all platforms currently supported by the existing tools it aims to replace. If that is part of the "core" or realized in plugins is an implementation detail. The article doesn't prescribe that.

> In the goal section: does "work with ROS1 packages as well as ROS2 packages" implies that no modification of their CMake will be necessary, right?

No, the build tool will not provide any such magic. It aims to support building ROS 1 packages like catkin_* does as well as ROS 2 package like ament_tools. Any mixture is the responsibility of the build system - not the build tool.

The idea as mentioned on the roadmap draft is to create a ROS 2 package named "catkin" which provides the same CMake API as catkin does in ROS 1 and map the calls to ament_cmake. That should allow to process existing ROS 1 packages for ROS 2 on a CMake level (!). If anything within such a package depends on ROS 1 names etc. that is a different story which can't be addressed by this CMake API shim.

> While it's possible to do cross-compilation with some of the existing tools it doesn't appear in this document at all, should this be taken into account in this document?

I don't consider cross-compilation to be relevant for the build tool. The build sytem is responsible to support this. Maybe it needs to be invoked with certain parameters. So the build tools responsibilty is to allow invoking the build system of each package with arbitrary arguments. But I don't think this is specific to cross-compilation. The same requirement already arises from native builds. E.g. passing an individual build type to a specific package in the workspace.

> Will the proposal specify a way to [provide meta information externally](https://github.com/ros2/design/pull/115/commits/634dd8fe23cfcc9b1d7281e9efabdd6a73c9215e#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR37)?

I don't plan to specify the details about how the meta information is being provided. It could be yaml, xml, whatever else. The requirement is that it needs to be possible from outside of the source space. It could even be that different plugins of the build tool provide different mechanisms (locally for building a Gazebo workspace I have three different ways to provide these information).

mikaelarguedas:
> So the build tools responsibilty is to allow invoking the build system of each package with arbitrary arguments.

Agreed, I think this should be a requirement for the new tool. Maybe some tools (catkin_tools?) allow that already so I guess it's implied in the "should provide a superset of the functionality provided by the existing tools " but it wasn't obvious to me when I read it. Hence the question

wjwwood:
> No, the build tool will not provide any such magic. It aims to support building ROS 1 packages like catkin_* does as well as ROS 2 package like ament_tools. Any mixture is the responsibility of the build system - not the build tool.

When you say "Any mixture" do you mean the shim you mentioned or do you mean a workspace, with two packages, one `catkin` and the other `ament_cmake`? In the latter case I think the build tool (or it's plugins) will need to deal with this issue. Either that or the proposal will need to include what we intend to do in the build systems to make them behave with one another.

> The idea as mentioned on the roadmap draft is to create a ROS 2 package named "catkin" which provides the same CMake API as catkin does in ROS 1 and map the calls to ament_cmake.

Where is this being discussed? It sounds like an interesting idea, but I don't remember talking about it. Why is it necessary to have this shim? Also, how would the shim handle the required differences in the contents of the `package.xml` (think different dependencies beyond the build system)? Can't you just have `catkin` packages use `catkin` directly and let them coexist with `ament_cmake` packages? (this is perhaps off-topic, I can repost the question where ever this is being discussed once you point me to it)

> I don't consider cross-compilation to be relevant for the build tool. The build sytem is responsible to support this.

I'm not convinced by this argument, because `catkin_tools` is mostly divested from the "build system" part of `catkin` and yet it had to be modified to not break `DESTDIR` and in other ways to support cross-compilation. I agree most of the responsibility lies mostly with cmake or python's setuptools or whatever underlying build system it is to support cross-compilation, but I also don't think the tool will be unaffected.

You could be making a separate argument that the "core" build tool doesn't need to handle this, but rather the catkin or ament specific plugin, but I think this just kicks the can down the road. In my opinion the proposal needs to address details about these plugins and how they'll work together. I think it's good to separate them conceptually, but both need to be represented in this document I think.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+The goal of the build tool is to build a set of packages with a single invocation automating the process.
+It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
+It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
+This will allow the build tool to be utilized for non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
+
+### Out of Scope
+
+The build tool does not cover the steps necessary to fetch the sources of the to-be-built packages.
+This can be done with existing tools (e.g. [rosinstall_generator](http://wiki.ros.org/rosinstall_generator), [vcstool](https://github.com/dirk-thomas/vcstool)).
```
wjwwood: If you're not going to list the tools exhaustively, then I'd recommend the pair `rosinstall_generator` and `wstool` since that's what the ROS 1 from source tutorials use. In ROS 2 we use `vcstool`, but we don't use it with `rosinstall_generator` right now. So this comes off as mismatched in my opinion. Perhaps a compromise would read like this:

> There are already tools to help with this.
> For example, the list of repositories that need to be fetched is provided either by a hand crafted `.rosinstall` or `.repos` file or by using `rosinstall_generator` to generate one.
> The list of repositories can then be fetched with one of several tools, like `rosinstall` or `wstool` in the case of a `.rosinstall` file, or `vcstool` in the case of a `.repos` file.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+The build tool also does not provide a mechanism to install any dependencies required to build the packages.
+In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
+
+The build tool also does not create binary packages (e.g. a Debian package).
+In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generate the required metadata and then platform dependent tools like `dpkg-buildpackage` build binary packages.
+
+## Build Tool vs. Build System
+
+A build tool operates on a set of packages.
+It determines the dependency graph and invokes the package specific build tool of each package in topological order.
```
wjwwood: Suggestion:

> It determines the dependency graph and invokes the ~~package~~ specific build tool ~~of~~ for each package in topological order.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+The build tool itself should know as little as possible about the build system used for a specific package.
+Just enough in order to know how to setup the environment for it, invoke the build, and setup the environment to use the built package.
+
+The build system on the other hand operates on a single package.
+In the case a package uses e.g. `CMake` the responsibility of the tool comes down to invoke the common steps `cmake`, `make`, `make install` for this package.
+As another example for a package using `Autotools` the steps could look like `configure`, `make`, `make install`.
```
wjwwood: I think it might be useful to create a table of things and how we would categorize them according to these descriptions. For example, you could list something like `CMake which calls catkin macros`, `Plain CMake`, `CMake that calls ament_cmake macros`, ```Python package that only has a `setup.py` ```, etc as a build system and then list `catkin_make`, `catkin_make_isolated`, `catkin build (catkin_tools)`, `ament build (ament_tools)`, etc as build tools. I think an explicit, if not exhaustive list, would help cement this idea for people first encountering this conceptual distinction.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Environment Setup
+
+A very important part beside the actual build of a package is the environment setup.
+For a CMake package e.g. a dependency discovered using the CMake function `find_package` needs to be either in a common location or the location must be provided through the environment variable `CMAKE_PREFIX_PATH`.
```
wjwwood: Suggested edit:

> For example, in order for a CMake project to discover a dependency using the CMake function `find_package`, the CMake module (e.g. `FindFoo.cmake`) or the CMake config file (e.g. `FooConfig.cmake`) for that dependency must either be in a path that CMake searches implicitly (e.g. `/usr/share/Foo/cmake`) or the location must be provided through the environment variable `CMAKE_PREFIX_PATH`.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+On the other hand after a package has been built and installed the environment might need to be extended to use the package.
```
wjwwood: I think this could be clearer. I know you mean "On the other hand" where the other hand is running something rather than building on top of it, but that's isn't clear enough I think. Perhaps:

> In addition to building a package on top of another package (using `find_package` in the case of CMake), you may need to adjust the environment in order to run an executable from a package.

Really just need to clarify what "use" means in "to use the package". So you can also just modify what you have to be clearer. This ambiguity isn't really addressed by the example that comes next in my opinion either. It could be cleared up by explaining the shared libraries may need to be on a particular PATH so they can be loaded at runtime of an executable.

articles/101_build_tool.md
wjwwood: I realize this overlaps with what's listed in sections below, but I still think it would be helpful to have a concise list here as well.

dirk-thomas:
> > No, the build tool will not provide any such magic. It aims to support building ROS 1 packages like catkin_* does as well as ROS 2 package like ament_tools. Any mixture is the responsibility of the build system - not the build tool.

> When you say "Any mixture" do you mean the shim you mentioned or do you mean a workspace, with two packages, one `catkin` and the other `ament_cmake`? In the latter case I think the build tool (or it's plugins) will need to deal with this issue. Either that or the proposal will need to include what we intend to do in the build systems to make them behave with one another.

From my point of view the build tool identifies a package and decides how to process it. Afterwards it knows how to setup the environment to "make it available" to downstream packages.

A package using the CMake API shim is still an ament package since it uses ament_cmake under the hood (it registers itself at the ament resource index, it creates setup files like other ament packages, etc.).

The primary goal for the unified build tool is to support building a ROS 1 workspace and (unrelated from that !) a ROS 2 workspace. Mixing ROS 1 and ROS 2 packages in the same workspace might be technically possible since the build tool knows how to process them. But there is much more to it then just invoking the build: packages names might collide or be different, etc.

> > The idea as mentioned on the roadmap draft is to create a ROS 2 package named "catkin" which provides the same CMake API as catkin does in ROS 1 and map the calls to ament_cmake.

> Where is this being discussed? It sounds like an interesting idea, but I don't remember talking about it. Why is it necessary to have this shim? Also, how would the shim handle the required differences in the contents of the `package.xml` (think different dependencies beyond the build system)? Can't you just have `catkin` packages use `catkin` directly and let them coexist with `ament_cmake` packages? (this is perhaps off-topic, I can repost the question where ever this is being discussed once you point me to it)

Currently the CMake code of every ROS 1 package we try to use with ROS 2 is being completely changed from catkin to ament_cmake. Imo most of that is unnecessary. We might need a couple of changes (e.g. names as well as a CMake call at the end where `ament_package` usually is invoked) but it would be beneficial to be able to keep all the rest as is. We have talked about it in verb discussions only. It is also mentioned on the (still internal) roadmap draft.

To fully work in a ROS 2 workspace it is not sufficient to just be able to build the catkin package with a tool which knows how to do that. A package using `catkin` will e.g. by definition not register itself at the ament resource index.

> > I don't consider cross-compilation to be relevant for the build tool. The build sytem is responsible to support this.

> I'm not convinced by this argument, because `catkin_tools` is mostly divested from the "build system" part of `catkin` and yet it had to be modified to not break `DESTDIR` and in other ways to support cross-compilation. I agree most of the responsibility lies mostly with cmake or python's setuptools or whatever underlying build system it is to support cross-compilation, but I also don't think the tool will be unaffected.

Completely agreeing with you. The build tool will need to make sure not to be in the way of what is possible with the build system.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+
+Several different build tools are already being used in the ROS ecosystem.
+Their method of operating is being described in the following subsections together with their advantages as well as disadvantages.
+
+### catkin_make
+
+`catkin_make` is provided by the ROS package `catkin` which contains the build system for ROS 1.
+It has been designed as the successor of `rosbuild` for ROS Fuerte.
+
+The tool invokes CMake only a single time and uses CMake's `include` function to process all packages in a single context.
```
wjwwood: Technically, doesn't it use the `add_subdirectory` function? It doesn't really matter I think, but as long as you mention it, then we might as well try to be precise.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+The following items are possible extension points to provide custom functionality:
+
+- contribute `verbs` to the command line tool (e.g. `build`, `test`)
+- contribute command line options for specific features (e.g. `build`, `test`)
+- discovery of packages (e.g. recursively crawling a workspace)
+- identification of packages and their meta information (e.g. from a `package.xml` file)
+- process a package (e.g. build a CMake package, test a Python package)
+- execution model (e.g. sequential processing, parallel processing)
+- output handling (e.g. console output, logfiles, status messages, notifications)
+- setup the environment (e.g. `sh`, `bash`, `bat`)
```
wjwwood, line 288: Is this not covered already by "process a package" and "completion"? What type of arbitrary extension to these setup files are you imagining that lies outside of environment contributed by specific packages or tab completion for the tool itself?

A more basic question is, do this mean setup of a workspace or setup for the tool?

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+- completion (e.g. `bash`, `Powershell`)
+
+## Alternative Approaches
```
wjwwood: You might want to use the term "Possible" rather than "Alternative", since to me at least alternative means possible but not selected.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+
+There are two different approaches possible to reach the goal of a universal build tool.
+
+- Incrementally evolve one of the existing tools to satisfy the described goals.
+- Implement a new build tool "from scratch" (while existing building blocks can still be "copied" from the existing solutions).
```
wjwwood: This starts to overlap with an actual proposal, but I feel there is a lot more than just "two" approaches represented here. Even within these two branches, there are lots of details that could go one way or the other. I think it would be helpful to discussion to head in this direction in general. Trying to sort out the individual details.

articles/101_build_tool.md
```diff
@@ -0,0 +1,213 @@
...
+[catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build ROS 1 packages.
+It was developed after `catkin_make` / `catkin_make_isolated` to build packages in parallel as well as provide significant usability improvements.
+The tool supports building CMake packages and builds them in isolation as well as supports parallelizing the process across packages.
+
+### Naming
+
+The existing build tools are all named by the build system they are supporting.
+In general it should be possible for a build tool to support multiple different build systems.
+Therefore a name for a build tool being derived from a single build system might mislead the users that the tool only works for that specific build system.
+To avoid confusion of the user it would be beneficial if the build tool would have a different name to avoid implying an undesired correlation.
```
wjwwood: Would you classify this as nice-to-have or a requirement?

wjwwood:
> To fully work in a ROS 2 workspace it is not sufficient to just be able to build the catkin package with a tool which knows how to do that. A package using catkin will e.g. by definition not register itself at the ament resource index.

Why does it need to register in the ament resource index? How is a catkin package which doesn't add itself to the ament resource index any different from a plain cmake package?

Are you proposing to make having a file in the ament resource index a required artifact of all packages built by the tool? I don't actually mind that, but if that's the case the decision is only whether to add that to catkin itself or make the "catkin plugin" do it on behalf of the catkin package it builds (like we do with plain cmake and python packages right now in ROS 2).

articles/101_build_tool.md
wjwwood: Oh and if you'd like me to or this or any other suggestion I'm happy to do so.

dirk-thomas:
> > To fully work in a ROS 2 workspace it is not sufficient to just be able to build the catkin package with a tool which knows how to do that. A package using catkin will e.g. by definition not register itself at the ament resource index.

> Why does it need to register in the ament resource index? How is a catkin package which doesn't add itself to the ament resource index any different from a plain cmake package?

It doesn't need to. A plain CMake package does neither. But then certain feature in ROS 2 might just not work. I only want to point out that simply building a package doesn't make it work 100%.

> Are you proposing to make having a file in the ament resource index a required artifact of all packages built by the tool? I don't actually mind that, but if that's the case the decision is only whether to add that to catkin itself or make the "catkin plugin" do it on behalf of the catkin package it builds (like we do with plain cmake and python packages right now in ROS 2).

I am not proposing that. The build tool doesn't care imo if the build system does something like that or not.

articles/101_build_tool.md
dirk-thomas: Please go ahead and create a PR for any changes you suggest. It makes it much easier to iterate on it and integrate the change. Thanks!

articles/101_build_tool.md
dirk-thomas: I would not phrase it as "nice-to-have" since I interpret it to be almost as weak as "not going to happen" :wink: It is clearly not a requirement since any name would "work". So I would call it a "goal" - so stating that this is what we want to do.

articles/101_build_tool.md
dirk-thomas: The set of supported shells should be extensible by external plugins. So beside having a `setup.bash` there could be variations for `fish`, `tcsh`, `zsh`, `powershell`, etc.

articles/101_build_tool.md
dirk-thomas: I think the two bullet points state the major directions possible. For sure there is a lot of variation in the details but I don't see a third major direction?

articles/101_build_tool.md
dirk-thomas: All the line comments with suggestions which I didn't explicitly commented on sounds like good improvements to me.

wjwwood:
> I am not proposing that. The build tool doesn't care imo if the build system does something like that or not.

Ok, but then how do you handle plain CMake packages? If you build one in a ros 1 workspace (with an existing tool) it won't have that file, but if you do with ros 2 it will. Which should the unified tool do? Should it be dependent on the workspace you're building? Doesn't the tool then need to know which is which at that point?

More generally, I'm trying to make the point that the plugins for different build systems need to either have knowledge of each other so they interact correctly (at least right now) or there has to be some rules for pre- and post-conditions for each package built regardless of the build system. I think the latter is better, and I also think that the rules can be lean and can be fulfilled by the build system plugin for the build tool to avoid changes in upstream if that's more desireable.

This is again getting into design details, but I at least feel that some of these things would need to be hashed out to some level before I could venture to say a particular proposal is a good idea or not.

articles/101_build_tool.md
wjwwood: > as weak as "not going to happen"

Well, I don't see it that way 😃. A nice-to-have would be in favor of a "from scratch" tool or forking an existing tool to rename it, which might lead us to favor those options too. However, if it were a requirement it would categorically exclude a proposal to iterate on an existing tool without changing the name. I'm happy to have this as a nice-to-have as I've described here, but to not consider iterating on a tool without changing its name moving forward goes too far for me to agree right now.

I would need to see more discussion from more people about this for my mind to be changed.

> It is clearly not a requirement since any name would "work".

But according to this section, a name that is based on an existing build system would be bad, so not any name would work by that reasoning.

> so stating that this is what we want to do.

In that case, unfortunately 😞, I'm not sure I can agree with this. I'm certainly not completely opposed to it, but I wouldn't go so far as to say I'm confident that's the right direction to take. There are some possible negative impacts of having a new name. I think @sloretz pointed them out well enough that I don't have to reiterate them.

----

I'm still not clear on the distinction you're making. If a goal is a weaker version of a requirement, then how is that different than a nice-to-have? Maybe you could clarify: what's the difference (in your opinion) in a requirement and a goal and a nice-to-have?

Perhaps there's more precise terminology we can draw on, to avoid the nuances.

articles/101_build_tool.md
wjwwood: But doesn't that need to be supported (at least in part) by the build system plugins? `sh` works for all the `*sh` on unix, but I think `bat` could not be used like this in powershell for example. I'm fine to leave this in, but I'm concerned it is not practical to decouple this completely from other parts.

dirk-thomas:
> > I am not proposing that. The build tool doesn't care imo if the build system does something like that or not.

> Ok, but then how do you handle plain CMake packages? If you build one in a ros 1 workspace (with an existing tool) it won't have that file, but if you do with ros 2 it will. Which should the unified tool do? Should it be dependent on the workspace you're building? Doesn't the tool then need to know which is which at that point?

I would say neither of it. The build tool chooses on process to apply to a package. If we define only a single "plain CMake" process. The same will be applied in both cases. If we want to distinguish them we need to classify the package differently - I don't think the build tool should do any "workspace-type detection". In my tools I use locally if they can't distinguish this case based on the package metadata they will require me to resolve the ambiguity by specifying manually which of the two "processes" I would like to apply.

> More generally, I'm trying to make the point that the plugins for different build systems need to either have knowledge of each other so they interact correctly (at least right now) or there has to be some rules for pre- and post-conditions for each package built regardless of the build system. I think the latter is better, and I also think that the rules can be lean and can be fulfilled by the build system plugin for the build tool to avoid changes in upstream if that's more desireable.

I think this gets way too detailed. In general I don't think it will be possible to use packages "with no changes". But we should minimize the required changes to whatever is actually necessary - hopefully those changes can be introduced in way compatible with the other build tools.

> This is again getting into design details, but I at least feel that some of these things would need to be hashed out to some level before I could venture to say a particular proposal is a good idea or not.

For now the article doesn't even contain a proposal. In my first comment I was "only" asking for feedback on the ration / goals / use cases and requirements.

articles/101_build_tool.md
wjwwood: > I think the two bullet points state the major directions possible.

It does, but I guess I'm saying, that I don't see that it helps us move forward. It's a coarse distinction to the point of being arbitrary. I think we need to subdivide the two directions into more detailed, mutually exclusive directions and then categorize orthogonal choices as "add-ons" to any of the proposal. I think that could be a helpful step between these goals and an actual proposal.

wjwwood:
> > > I am not proposing that. The build tool doesn't care imo if the build system does something like that or not.

> > Ok, but then how do you handle plain CMake packages? If you build one in a ros 1 workspace (with an existing tool) it won't have that file, but if you do with ros 2 it will. Which should the unified tool do? Should it be dependent on the workspace you're building? Doesn't the tool then need to know which is which at that point?

> I would say neither of it. The build tool chooses on process to apply to a package. If we define only a single "plain CMake" process. The same will be applied in both cases. If we want to distinguish them we need to classify the package differently - I don't think the build tool should do any "workspace-type detection". In my tools I use locally if they can't distinguish this case based on the package metadata they will require me to resolve the ambiguity by specifying manually which of the two "processes" I would like to apply.

That's not neither, that's the way it is in ROS 1 😛.

> > More generally, I'm trying to make the point that the plugins for different build systems need to either have knowledge of each other so they interact correctly (at least right now) or there has to be some rules for pre- and post-conditions for each package built regardless of the build system. I think the latter is better, and I also think that the rules can be lean and can be fulfilled by the build system plugin for the build tool to avoid changes in upstream if that's more desireable.

> I think this gets way too detailed. In general I don't think it will be possible to use packages "with no changes". But we should minimize the required changes to whatever is actually necessary - hopefully those changes can be introduced in way compatible with the other build tools.

If you mean for the goals and use cases, I agree, but if you mean for the proposal then I disagree, because these are important things to consider because they will likely require changes in either `catkin`, `ament_cmake`, or both. And I think that any changes that might be required in existing systems due to a proposal will be a considerable factor in the decision making process.

Of course as you pointed out, I'm getting ahead of your current question, and so we can table this for now, but I do think it needs to be discussed at this level of detail before we've decided on a proposal.

articles/101_build_tool.md
wjwwood: https://github.com/ros2/design/pull/120

articles/101_build_tool.md
```diff
@@ -0,0 +1,228 @@
...
+There are already tools to help with this.
+For example, the list of repositories that need to be fetched is provided either by a hand crafted `.rosinstall` or `.repos` file or by using [rosinstall_generator](http://wiki.ros.org/rosinstall_generator) to generate one.
+The list of repositories can then be fetched with one of several tools, like [rosinstall](http://wiki.ros.org/rosinstall) or [wstool](http://wiki.ros.org/wstool) in the case of a `.rosinstall` file, or [vcstool](https://github.com/dirk-thomas/vcstool) in the case of a `.repos` file.
+
+The build tool also does not provide a mechanism to install any dependencies required to build the packages.
+In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
+
+The build tool also does not create binary packages (e.g. a Debian package).
+In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generate the required metadata and then platform dependent tools like `dpkg-buildpackage` build binary packages.
+
```
nuclearsandwich: Source fetching, dependency resolution, and binary package generation are out of scope for the build tool itself.

Is the intended use to chain the build tool with the existing infrastructure:
```
$ fetch-sources && resolve-deps && build-tool && generate-packages
```
where fetch-sources, resolve-deps, and generate-packages are standins for vcstool, rosdep, and bloom with applicable arguments

Or is the intent to have the build tool be aware of those utilities and invoke them as necessary in order to have a single invocation for the entire process?

dirk-thomas: They will be used as they are right now - through separate invocation.

nuclearsandwich: :cool: Is that worth adding to the document or is it an unnecessary detail?

dirk-thomas: Since that is exactly what is already the case with the current build tools I don't think I needs to be mentioned.

nuclearsandwich:
> > I think it would be fine for the tool to share a name with a build system. To a lot of developers the "tool" is just the commands they've memorized to build their code. Knowing the build tool is a different from the build system doesn't change their workflow.
> 
> Since the build tool is supposed to be used with multiple different build system which have different names I think the association with one of them by name is to be avoided (since users can a) often not distinguish what is build tool and what is build system and b) expect a build tool called ament_tools not to work for catkin).

Coming into the ROS ecosystem with no prior familiarity, it took a few days of doodling charts before I really had an idea of how the different build system and tool components fit together. There are quite a lot of them, especially if you start looking at both ROS 1 and ROS 2 but having distinct names creates boundaries that make the tool responsibility organization more approachable for newcomers.

dirk-thomas:
To summarize the important points of the draft to verify that we have a consensus on:

* We want to develop a unified build tool which replaces the existing tools (catkin_make, catkin_make_isolated, catkin_tools, ament_tools)
* The goal is to name it differently than the build systems (so not have `catkin` or `ament` in the name)
* We want to support use cases beyond ROS (e.g. but not limited to building Gazebo including its dependencies)
* The unified build tool should support a superset of features of the current build tools
* The list of proposed extension points sounds reasonable (maybe propose additional ones, or question some of the proposed ones?)
* The stated software criteria make sense and should be folllowed / aimed for

articles/101_build_tool.md
clalancette: Also, are there any other tools out there which provide some of this functionality, particularly the ones about being extensible?  Some other ecosystem of existing tooling could help us get over the initial startup cost of something new.

clalancette:
> To summarize the important points of the draft to verify that we have a consensus on:
> 
>     We want to develop a unified build tool which replaces the existing tools (catkin_make, catkin_make_isolated, catkin_tools, ament_tools)

I would say we "want a unified build tool which replaces the existing tools".  Whether we have to develop it from scratch isn't 100% clear to me, though you probably have a better idea of the ecosystems out there.

>     The goal is to name it differently than the build systems (so not have catkin or ament in the name)

Agreed.

>     We want to support use cases beyond ROS (e.g. but not limited to building Gazebo including its dependencies)

Agreed.

>     The unified build tool should support a superset of features of the current build tools

Agreed.

>     The list of proposed extension points sounds reasonable (maybe propose additional ones, or question some of the proposed ones?)

They sound reasonable to me.

>     The stated software criteria make sense and should be folllowed / aimed for

Yes, although they are pretty general (i.e. basically all software should aim for them).

mikaelarguedas:
> We want to develop a unified build tool which replaces the existing tools (catkin_make, catkin_make_isolated, catkin_tools, ament_tools)

:+1: 

> The goal is to name it differently than the build systems (so not have catkin or ament in the name)

:+1:  for avoiding user confusion

> We want to support use cases beyond ROS (e.g. but not limited to building Gazebo including its dependencies)

:+1: 

> The unified build tool should support a superset of features of the current build tools

:+1: 

> The list of proposed extension points sounds reasonable (maybe propose additional ones, or question some of the proposed ones?)

:+1: I assume here that the "process a package" part includes being able to add support for new build systems

>The stated software criteria make sense and should be folllowed / aimed for

:+1:

wjwwood:
> To summarize the important points of the draft to verify that we have a consensus on:

> - We want to develop a unified build tool which replaces the existing tools (catkin_make, catkin_make_isolated, catkin_tools, ament_tools)

+1, I agree if the source of the replacement may be an evolution of an existing tool (which I think is implied but not explicitly stated here)

> - The goal is to name it differently than the build systems (so not have catkin or ament in the name)

-0, I'm not convinced this is worth it due to the drawbacks, but could be convinced otherwise by feedback from other stakeholders (ROS community members, candidates for new users like the Gazebo team, contributors to catkin, catkin_tools, ament_tools, etc).

> - We want to support use cases beyond ROS (e.g. but not limited to building Gazebo including its dependencies)

+1

> - The unified build tool should support a superset of features of the current build tools

+1

> - The list of proposed extension points sounds reasonable (maybe propose additional ones, or question some of the proposed ones?)

+1, I made some comments on them inline in the document, but none of the questions are deal breakers.

> - The stated software criteria make sense and should be folllowed / aimed for

+1, I agree that the criteria a good individually, but I would rather make a judgement on the application of them on a particular design decision. Put another way, it's far more valuable (in my opinion) to ask if using a particular criteria to justify (or argue against) doing X instead of Y makes sense, than to ask if the software criteria is good in general.

On the last point I agree with @clalancette:

> Yes, although they are pretty general (i.e. basically all software should aim for them).

dhood:
The goals seem reasonable to me and I am in favour of changing the name to
make it different from the build systems

On 14 Mar. 2017 14:21, "Dirk Thomas" <notifications@github.com> wrote:

To summarize the important points of the draft to verify that we have a
consensus on:

   - We want to develop a unified build tool which replaces the existing
   tools (catkin_make, catkin_make_isolated, catkin_tools, ament_tools)
   - The goal is to name it differently than the build systems (so not have
   catkin or ament in the name)
   - We want to support use cases beyond ROS (e.g. but not limited to
   building Gazebo including its dependencies)
   - The unified build tool should support a superset of features of the
   current build tools
   - The list of proposed extension points sounds reasonable (maybe propose
   additional ones, or question some of the proposed ones?)
   - The stated software criteria make sense and should be folllowed /
   aimed for

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/ros2/design/pull/115#issuecomment-286564693>, or mute
the thread

iluetkeb:
I just noticed this, and read through the discussion so far with some amazement.

I mean, I didn't understand the case for ament, and I understand the case for replacing it again even less.

'nuff said.

tfoote:
@dirk-thomas +1 to all the summarized points

gbiggs:
What are the reasons (technical or not) that ament cannot be developed to satisfy the requirements of this article? Rather than making yet another build tool that, as one of its features, can build ament packages, I think that continuing development of ament to meet these requirements would be better. Its already a modular, extensible design, so it seems likely that it would be possible to extend it to allow building catkin packages and generic software.

tkruse:
> The primary goal for the unified build tool is to support building a ROS 1 workspace and (unrelated from that !) a ROS 2 workspace. 

If that is a primary goal, why not state it in the goals section like this? I only understood it from the comment. And would this functionality be trivially achievable by having a tiny meta-buildtool that does:

1. heuristically detect if in a catkin or ament workspace.
2. invoke catkin_make or ament respectively

> The goal is to name it differently than the build systems

And why make it a 'goal' to name it differently? That sounds more like a decision than a goal. A goal would be to introduce it with minimal confusion. And personally I think catkin2 or ament2 would cause much less confusion than yet another name.

A motivation is still entirely missing, as in what is the problem with having both a catkin and an ament build tool, given they server different buildsystems and workspaces anyway?

> It is also mentioned on the (still internal) roadmap draft.

Keeping drafts internal does not exactly help the public discussion. As long as it is marked as draft, why not make it public?

> A package using catkin will e.g. by definition not register itself at the ament resource index.

Briefly comparing the anatomy of the catkin / ament cmake files would help others understand the problem. Same for briefly describing the ament resource index. Since those are apparently major forces in the decision making, they belong in that document.

dirk-thomas:
@tkruse 

> If that is a primary goal, why not state it in the goals section like this? I only understood it from the comment. And would this functionality be trivially achievable by having a tiny meta-buildtool that does:

The goal is to not maintain N separate tools anymore which more or less implement the same things. See this paragraph: https://github.com/ros2/design/pull/115/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR157

> And why make it a 'goal' to name it differently? That sounds more like a decision than a goal.

Please see this paragraph which describes the rational: https://github.com/ros2/design/pull/115/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR150

> And personally I think catkin2 or ament2 would cause much less confusion than yet another name.

Same as the first comment: we don't want to maintain N tools anymore.

> Keeping drafts internal does not exactly help the public discussion. As long as it is marked as draft, why not make it public?

This is a public repository - so nobody is keeping anything "internal".

> Briefly comparing the anatomy of the catkin / ament cmake files would help others understand the problem. Same for briefly describing the ament resource index. Since those are apparently major forces in the decision making, they belong in that document.

Please refer to existing documentation about those which we don't want to reiterate here: e.g. http://design.ros2.org/articles/ament.html and https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/doc/resource_index.md

dirk-thomas:
@iluetkeb 

> I mean, I didn't understand the case for ament, and I understand the case for replacing it again even less.

Please see the existing article about ament(_cmake) which describe how it is different from catkin and why it has been created instead of catkin being changed: http://design.ros2.org/articles/ament.html

dirk-thomas:
@gbiggs 

> What are the reasons (technical or not) that ament cannot be developed to satisfy the requirements of this article? Rather than making yet another build tool that, as one of its features, can build ament packages, I think that continuing development of ament to meet these requirements would be better. Its already a modular, extensible design, so it seems likely that it would be possible to extend it to allow building catkin packages and generic software.

The goal is to not maintain N separate tools anymore which more or less implement the same things. See this paragraph: https://github.com/ros2/design/pull/115/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR157

Please also see [this paragraph](https://github.com/ros2/design/pull/115/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR150) which describes why we think coupling the name of the build tool to the name of one build system (even though it should support others) is not a good idea since it misleads users.

Also don't expect that a "new unified build tool" implies that every line has to be implemented "from scratch". In my opinion it will look more like:
* define the necessary extension points (with the awareness of the goals, requirements, and capabilities of the existing build tools)
* implement a plugin architecture which is flexible enough to suit the needs of the new tool
* port existing code blocks from the various existing build tools to the new architecture

tkruse:
So is this a valid pseudo-code for the universal build-tool?

    1. detect workspace folder to use (e.g. current folder, or parent folder with marker file, ...)
    2. detect buildsystem name to use for this workspace (e.g. 'catkin', 'ament', 'cmake', 'setuptools', 'foobuild')
    3. load plugin/wrapper for this buildsystem
    4. ask plugin/wrapper for package declarations (name, folder and dependencies of each package)
    5. create DAG of packages for building the workspace
    6. invoke buildsystem plugin command for each package in build order

So the universal build tool would incorporate the abstraction of a workspace containing a set of packages with dependencies, and the ability to create a DAG and invoke tasks in parallel. The buildsystem plugins/wrappers would contain all the buildsystem specific logic (e.g. specific manifest files, environment setting)?

wjwwood:
I think there are two concepts which we're not communicating well enough.

First, the build tool is not the same as the build system. For example, this idea of a universal build tool would be useful even if we dropped support for the `ament` build system (specifically `ament_cmake`) completely or if someone wanted to support autotools or yet another build system based on CMake or SCons -- it doesn't matter what. The real idea is to finish formalizing the process by which we build individual packages of different types and make sure that one tool supports all the types we care about for now. Though it might require adjustments in catkin or ament to make this easier, it doesn't replace any of them, only their build tools, which have lots of overlap already.

My personal favorite idea for doing that is to just improve `catkin_tools` so that it can easily be extended to build ament packages and pure Python packages (both of which we need for ROS 2). But there are lots of little details that have to be sorted out to do that. Another way to do that would be to build something from scratch that takes all the best things from `catkin_tools` and `ament_tools` and others.

In both cases we have an orthogonal decision to change the name so that it's less confusing, i.e. many people see `catkin_tools` and are completely unaware that it can currently build non-catkin CMake packages or that it could easily build autotools packages or pure Python packages or ament CMake packages with just some new plugins to support those types.

Second, we want to have non-homogeneous workspaces. That is to say a single workspace which contains packages of different types simultaneously. `catkin_tools` already does this with pure CMake packages and catkin packages. `ament_tools` can do this with ament CMake packages, pure CMake packages and pure Python packages. We do not have to support all combinations to start with, for example if you have an ament CMake package and a catkin CMake package in a single workspace we can just have an assert early on in the tool that prevents you from doing this. The more important thing is that the tool can be used on a workspace full of catkin packages and then the same tool can be used on a workspace full of ament packages on a separate occasion. We can later support mixing those (or any other combination) in the future. For this reason, the idea of a "routing" tool that just selects a different underlying tool depending on some context isn't particularly appealing to me, but it is technically an option.

wjwwood:
Sorry @tkruse, for some reason I didn't see your post until just now.

> So is this a valid pseudo-code for the universal build-tool?

It's close, the second point I think is unnecessary. Ideally there would be no difference in a workspace with just caktin or just cmake or just ament, but the tool can and should smooth over as many of the differences as it can. Instead, bullets 2, 3, and 4 would be reorganized as something like "load all build system plugins; discover all packages (of all kinds); use the matching plugin to collect the necessary metadata for each package (no matter the type)", at which point a DAG could be made and the iterative process of building each package could start.

> So the universal build tool would incorporate the abstraction of a workspace containing a set of packages with dependencies, and the ability to create a DAG and invoke tasks in parallel. The buildsystem plugins/wrappers would contain all the buildsystem specific logic (e.g. specific manifest files, environment setting)?

Yes, though there would need to be (in my opinion) some pre and post condition rules so that we can successfully chain one package to the next. For example, a pure Python package would need to make sure it is added to the `PYTHONPATH` before a dependent package should be built (regardless of that next package's type). Finding a way to generalize that is tricky, but we're really close with `ament_tools` and `catkin_tools`. This proposal is more about figuring out how to combine those efforts in the best way.

iluetkeb:
@wjwwood Thanks for the explanation, that clarifies it, and I understand (and agree with) the motivation now. I also really like the idea of just extending catkin_tools.

That said, the rationale for heterogeneous workspaces is still a bit fuzzy. I mean, I can understand wanting compatibility with catkin, it's the previous build-system, and support would ease the transition process. I do not see why anybody would want to support autotools or the like.

OF course, some people might say, the more build systems supported the better, but I completely disagree. I believe that one of the biggest advantages of the ROS build eco-system has been that it does not support everything, but pushes people towards a common system, which increases cohesion, and community integration.

Just think of passing flags: If everything is cmake-based, I can just give catkin tools the line
  --cmake-args -DCMAKE_CXXFLAGS="-O2 -ggdb" -G"Eclipse CDT4 - Unix Makefiles"
Try that on a build-system-independent tool... You would have to translate everything, and not every feature is supported everywhere, etc.

Other things are still integrated through the rosdep mechanism and binary packages, which I think is totally fine.

btw, what I **really** think we should be looking more at is container-based integration, where we don't build as much anymore, but just compose pre-built containers. That's an orthogonal problem, sure, but it has received very little attention so far.

iluetkeb:
@dirk-thomas The only argument on that page that I could identify as explaining why catkin has not been evolved is that it might cause regressions on ROS1. I do not agree with that argument, because I think there would have been ways to prevent that from happening.

However, that's all water under the bridge now. The decision has been made, you guys put a lot of effort into ament and improving upon catkin, and the result is nice, so it's here to stay. 

I really didn't mean to re-start the discussion, the only reason I mentioned is because I thought you guys wanted to do it *again*. I now see that I was mistaken about that, and that that's not what you're doing.

dirk-thomas:
Thank you @ all for the feedback so far. Based on that I have tried to clarify the goals at the very top of the article as well as mentioned some more use cases which should continue to be supported (e619b85a7c18c215d3c9bc9fe65b5405701b8081). I also described the highlevel steps necessary for the two possible approaches to give an idea what is involved to get there (a980b7ccfe74362ebaef85ae331ce05e49c07693).

tkruse:
It seems the decision on whether to use catkin_tools or ament_tools as a basis might have a personal/political aspect to it. Assuming we are professionals but also humans, might be worth mentioning that.

When I look at the commit-contributors of catkin_tools, I see:

    jbohren      211 commits
    wjwwood      204 commits 
    ...
    dirk-thomas    4 commits

When I look at ament_tools:

    dirk-thomas   127 commits
    wjwwood        65 commits
    ... others less than 3 commits each

So some of the players here might have strong emotional attachments to one of those tools, not the other. Is that something to discuss at all?

* https://github.com/ament/ament_tools/graphs/contributors
* https://github.com/catkin/catkin_tools/graphs/contributors

dirk-thomas:
> It seems the decision on whether to use catkin_tools or ament_tools as a basis might have a personal/political aspect to it. Assuming we are professionals but also humans, might be worth mentioning that.

Please see the two approaches current mentioned in the article. The first option describes "evolving catkin_tools" and the second option described "starting with more flexible architecture from scratch". Evolving from `ament_tools` is explicitly mentioned as not an reasonable option. So I am not sure what you are trying to imply with your statement and the statistics.

Imo the choice boils down to a software engineering decision. What is easier to achieve and / or what results in a better quality end result. The conclusion will likely be influenced by personal opinion. That is why the current draft explicitly doesn't decide to do either or but only describe two possible paths forward .

articles/101_build_tool.md
```diff
@@ -0,0 +1,261 @@
...
+## Possible Approaches
+
+In terms of flexibility neither of the existing build tools can already support the superset of features described in this article.
+There are two different paths possible to reach the goal of a universal build tool:
+
+### Evolve catkin_tools
+
+One approach is to incrementally evolve one of the existing tools to satisfy the described goals.
+Since `catkin_tools` is in many aspects the most complete build tool it should be one being evolved.
+While `ament_tools` has a few features `catkin_tools` currently lacks (e.g. plain CMake support without a manifest, Windows support) the feature richness of `catkin_tools` makes it a better starting point.
```
tkruse: 'Evolve ament' should be it's own section, even if it is rejected early on.

articles/101_build_tool.md
```diff
@@ -0,0 +1,261 @@
...
+
+The following items are highlighting some of the necessary efforts (not a complete list):
+
+- Refactor the software architecture of the existing code base to support the flexibility sketched by the extension points listed above.
```
tkruse, line 314: Which of the extension points are missing in catkin_tools?

tkruse:
Regarding the commit statistics, I'll let them speak for themselves without further comment.

Another aspect I have not seen mentioned so far is the problem of catkin_tools not being the default build tool for ROS1 yet. This was planned in 2014 (https://groups.google.com/d/msg/ros-sig-buildsystem/UdgxD7AZTzM/k2az2aY1_24J), but did not happen until now (despite the many contributions of OSRF visible in the commit statistics).

The impact of catkin_tools remains the same though. When package developers start using catkin_tools only to build packages in isolation, those packages will not necessarily work with catkin_make, as far as I am aware. The same would also happen for creating a universal build tool from scratch. So being able to build catkin packages/workspaces does not automatically imply being compatible with catkin_make.

Is that something that should be mentioned somewhere in the document? I.e. will the universal build tool finally lead to deprecation of catkin_make?

articles/101_build_tool.md
```diff
@@ -0,0 +1,261 @@
...
+- Move `catkin` specific concepts out of the core of the build tool into a catkin specific extension (e.g. manifest format, *devel space*).
+- Support for ROS 2 which includes:
+
+  - Support for Python 3 and Windows.
+  - Support for pure Python packages as well as packages without an in-source manifest file.
+  - Environment setup of `ament` packages
+
+- Rename the tool to use a name unrelated to one build system.
+- Investigate if a feature like continued support of the *devel space* is feasible since it doesn't apply to other build system and might be complicated to separate without sacrificing usability.
```
tkruse, line 323: How is this specific to evolving catkin_tools? What would the 'from scratch' solution do w.r.t. the devel space?

articles/101_build_tool.md
```diff
@@ -0,0 +1,261 @@
...
+### Evolve catkin_tools
+
+One approach is to incrementally evolve one of the existing tools to satisfy the described goals.
+Since `catkin_tools` is in many aspects the most complete build tool it should be one being evolved.
+While `ament_tools` has a few features `catkin_tools` currently lacks (e.g. plain CMake support without a manifest, Windows support) the feature richness of `catkin_tools` makes it a better starting point.
+
+The following items are highlighting some of the necessary efforts (not a complete list):
+
+- Refactor the software architecture of the existing code base to support the flexibility sketched by the extension points listed above.
+- Move `catkin` specific concepts out of the core of the build tool into a catkin specific extension (e.g. manifest format, *devel space*).
```
tkruse, line 315: Is that aspect important to the success of the tool? Extracting functionality can be done at any time after release 1, no?

articles/101_build_tool.md
```diff
@@ -0,0 +1,261 @@
...
+- process a package (e.g. build a CMake package, test a Python package)
+- execution model (e.g. sequential processing, parallel processing)
+- output handling (e.g. console output, logfiles, status messages, notifications)
+- setup the environment (e.g. `sh`, `bash`, `bat`)
+- completion (e.g. `bash`, `Powershell`)
+
+## Possible Approaches
+
+In terms of flexibility neither of the existing build tools can already support the superset of features described in this article.
+There are two different paths possible to reach the goal of a universal build tool:
```
tkruse: What I suggested in a comment was a third way. Create a tool sitting on top of both ament_tools and catkin_tools, detecting the workspace nature, then invoking either of those tools. Basically treating catkin_tools and ament_tools as libraries (they can be forked as libraries without the CLI part). Possibly even enforce that catkin_tools as library is only used using Python2, and ament only with Python3. Once this runs, release it, then iteratively refactor moving common functionality to the universal layer. Should this approach be mentioned?

articles/101_build_tool.md
dirk-thomas: Done in e99e8a966b1b12202717c1d4c15d72ea5ec8b685

articles/101_build_tool.md
dirk-thomas: `catkin_tools` only has two extension points atm:

* verbs: https://github.com/catkin/catkin_tools/blob/fb324ddafd8a9d450b24df46e510f3d08364ab26/setup.py#L143
* build types: https://github.com/catkin/catkin_tools/blob/fb324ddafd8a9d450b24df46e510f3d08364ab26/setup.py#L154

articles/101_build_tool.md
dirk-thomas: A "from scratch" tool will hopefully encapsulate all devel space specific knowledge in a catkin / ROS 1 specific plugin. The question it it want to offer the feature at all or not is the same.

articles/101_build_tool.md
dirk-thomas: One of the software goals says:

> Not paying for what you don't use

So if I am building a ROS 2 workspace with this tools which doesn't feature a devel space I don't want the code being involved building the workspace having to deal with that concept.

articles/101_build_tool.md
dirk-thomas: Since this approach doesn't meet the state goals (deduplicating the effort of maintaining multiple tools) I don't think it needs to be mentioned.

tkruse:
What about migrating all ament-packages to be build with catkin_tools (extended for python 3 and windows)? This seems to also satisfy a lot of goals. The document could briefly touch upon that option.

dirk-thomas:
> What about migrating all ament-packages to be build with catkin_tools (extended for python 3 and windows)? This seems to also satisfy a lot of goals. The document could briefly touch upon that option.

Any progress toward the goals is a good thing. But since there are arbitrary subsets possible I don't think it makes sense to enumerate them all.

dirk-thomas:
I have added some more feedback. The last commit (1db7d0344ca292f532b51fa2aa6881f1064f2af9) describes the re-prioritization and that the implementation of this goal is suspended for the foreseeable future.

If I could get a couple of final reviews on this I would like to go ahead and merge the article in the current state rather sooner then later.

articles/101_build_tool.md
```diff
@@ -166,19 +170,37 @@ It was developed after `catkin_make` to address the problems involved with build
...
 ### catkin_tools
 
 [catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build ROS 1 packages.
 It was developed after `catkin_make` / `catkin_make_isolated` to build packages in parallel as well as provide significant usability improvements.
 The tool supports building CMake packages and builds them in isolation as well as supports parallelizing the process across packages.
 
+`catkin_make_isolated` supports building the following packages:
+
+- ROS 1 `catkin` package with a `package.xml` file.
+- Plain CMake packages with a `package.xml` file.
```
wjwwood: This looks like a copy-paste error?

wjwwood (approved):
lgtm

It's a good starting point for discussion. Thanks for compiling it @dirk-thomas.

articles/101_build_tool.md
dirk-thomas: The headline is certainly a copy-n-paste error. Good catch :+1:

The supported build types are still correct, right?

wjwwood: Yeah, sorry I mean the labeling. The list looks ok.

dirk-thomas: Great, fixed: 70d11e484ec41769359263994684b8c74da468aa

nuclearsandwich (approved):

PR #116, opened 2017-03-07
extend rational why support for a single CMake context was dropped

dirk-thomas:
Elaborate why support for a single CMake context was dropped from ament based on this comment: https://github.com/ros2/design/pull/115#issuecomment-284733417

### DIFF
--- a/articles/100_ament.md
+++ b/articles/100_ament.md
@@ -227,9 +227,11 @@ With the different design of ament it becomes possible to implement a package si
 #### Building within a single CMake context
 
 catkin allows users to build multiple packages within a single CMake context (using `catkin_make`).
-While this significantly speeds up the process it is not scalable at large.
-This is due to the fact that all packages share the same CMake namespace.
-Therefore the package might have colliding target names or fail to build correctly due to side effects between the packages.
+While this significantly speeds up the process it has several drawbacks.
+Since all packages within a workspace share the same CMake context all targets are within the same namespace and must therefore be unique across all packages.
+The same applies to global variables, functions, macros, tests, etc.
+Furthermore a package might need to declare target level dependencies to targets in other packages to avoid CMake targets being parallelized when they ought to be sequential (but only when being built in the same CMake context).
+This requires internal knowledge about the build structure of other packages and subverts the goal of decoupling packages.
 ament does not provide that feature due to these drawbacks.
 
 ### Additional improvements in ament over catkin


wjwwood (approved):
The only thing I would add is that if you add target level dependencies across packages then you need to make that conditional, otherwise you code may not compile with an isolated build (since the target won't exist in the second package anymore). With that the non-isolated tool now only works with packages that correctly handle intra-package target dependencies and all-catkin workspaces (no plain cmake and no pure python or other things).

dirk-thomas:
The catkin variable `<pkgname>_EXPORTED_TARGETS` already takes care of that. Therefore I didn't mention it. I consider that an "implementation detail".

If you think strongly about it please add a sentence to clarify this.

wjwwood:
My point is that people have to know to use that or their packages will likely fail at some point with the non-isolated tool due to a race condition. So that puts pressure to support the non-isolated case on package maintainers who may not use that tool.

I don't feel strongly about it, and I agree it's probably too detailed to be mentioned in that article without a lot of supporting material (so people can understand it without a lot of context).

articles/100_ament.md
```diff
@@ -227,9 +227,11 @@ With the different design of ament it becomes possible to implement a package si
 #### Building within a single CMake context
 
 catkin allows users to build multiple packages within a single CMake context (using `catkin_make`).
-While this significantly speeds up the process it is not scalable at large.
-This is due to the fact that all packages share the same CMake namespace.
-Therefore the package might have colliding target names or fail to build correctly due to side effects between the packages.
+While this significantly speeds up the process it has several drawbacks.
+Since all packages within a workspace share the same CMake context all targets are within the same namespace and must therefore be unique across all packages.
+The same applies to global variables, functions, macros, tests, etc.
+Furthermore a package might need to declare target level dependencies to targets in other packages to avoid CMake targets to be parallelized which need to be sequential.
```
dirk-thomas: @wjwwood Maybe appending the following will clarify enough without adding to much additional information about how it works?

> which need to be sequential (but only when being built in the same CMake context).

wjwwood: Yeah, and if I can suggest a rewording, I think this might be clearer:

> ... declare target level dependencies to targets in other packages to avoid CMake targets being parallelized when they ought to be sequential (but only when being built in the same CMake context).

dirk-thomas: :+1: 2184a7538eba1347d1ad92889edff3fd5f68e18b

PR #117, opened 2017-03-07
Add article about WideStrings.

clalancette:
This incorporates a bunch of research and thinking I've done about the problem, as well as feedback from people internally.  Note a few things about this article:

1.  It takes the stance that we should have a wide string type separate from the string type.

2.  It takes the stance that we should use UTF-8.  We haven't 100% decided this, but it is the most logical choice; almost all other encodings are deprecated at this point.

3.  It takes the stance that we should map the wide string onto the DDS concept of a wchar.  As Mikael pointed out, the XTypes spec on top of DDS mandates that wchar == Char32.  While the spec does *not* mandate what size Char32 is, it would be pretty surprising if it would be anything but 32 bits.  The other option we have here is to map the wide string on the DDS concept of uint8, which would shrink the size on the wire.  As it stands right now, none of the tools I looked at (including the RTI Connext tools) have the ability to properly display a wchar/wstring, so there is currently no compelling reason to go with wchar.  Still TBD.

4.  The C++ example below is really simple.  There actually exists a library called "utf8" that has a utf8_string class that behaves a lot like std::string, and that is what the example below uses.  What other APIs do we think we would need to model?

connects to  ros2/rosidl#208

### DIFF
--- a/articles/WideStrings.md
+++ b/articles/WideStrings.md
@@ -48,87 +48,83 @@ For more information, the following links provide introductory material:
 
 The rest of this document is generally going to assume Unicode, unless stated otherwise.
 
+## Unicode Characters in Strings
+ROS 1 says string fields are to contain ASCII encoded data, but allows UTF-8.
+DDS-XTYPES mandates UTF-8 be used as the encoding of IDL type `string`.
+To be compatible with ROS 1 and DDS-XTYPES, in ROS 2 the content of a `string` is expected to be UTF-8.
+
 ## Introducing Wide Strings
-To support multi-byte characters, ROS 2 will introduce the concept of a wide string.
-There are a few basic questions to answer on how a wide string will be integrated into ROS 2:
+ROS 2 messages will have a new [primitive field type](/articles/interface_definition.html) `wstring`.
+This purpose is to allow ROS 2 nodes to comminicate with non-ROS DDS entities using an IDL containing a `wstring` field.
+The encoding of data in this type should be UTF-16 to match DDS-XTYPES 1.2.
+Since both UTF-8 and UTF-16 can encode the same code points, new ROS 2 messages should prefer `string` over `wstring`.
 
-* What is the size impact of the wide string?
-* What is the encoding of the wide string?
-* What does the API look like to a user of ROS 2?
+## Encodings are Recommendations
+The choice of UTF-8 or UTF-16 for `string` and `wstring` are recommendations.
+A `string` or `wstring` field is allowed to be populated and published with an unknown encoding.
+If subscriber receives a string containing an unknown ncoding then it is up to the subscriber how to handle it.
+It could decode it using an encoding that has been agreed to out of band.
+Other subscribers like `ros2 topic echo` may echo the bytes in hexadecimal.
 
-Each of these questions will be examined in more detail below.
+In practice using different encodings may not be possible.
+The IDL specification forbids `string` from containing `NULL` values.
+For compatibility a ROS message `string` field must not contain zero bytes, and a `wstring` field must not contain zero words.
+An encoding like UTF-32 cannot be used since it may have zero bytes or words in a code point.
+If UTF-32 encoded strings need to be transmitted then one should either convert to UTF-8 and use a `string` or use a `uint32` array.
 
-### What is the size impact of the wide string?
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> Can ROS 1 publish a string with NULL bytes?
+</div>
 
-There are 2 parts to this question.
-One is how the wide string is mapped onto the underlying DDS implementation, which defines how large messages will be on-the-wire (and on disk).
-The other part of the question deals with the runtime impact (in terms of speed and code size) of having to deal with wide strings.
+## Unicode Strings Across ROS 1 Bridge
+Since ROS 1 and 2 both allow `string` to be UTF-8, the ROS 1 bridge will pass values unmodified between them.
+If a message sent from ROS 2 to ROS 1 fails to serialize because a string contains values that are not legal UTF-8 then it won't be published by the bridge.
 
-**Mapping onto the DDS implementation**
+If a ROS 2 message has a field of type `wstring` then the bridge will attempt to convert it from UTF-16 to UTF-8.
+The resulting UTF-8 encoded string will be published as a `string` type.
+If the conversion fails then the bridge will not publish the message.
 
-There are a couple of different ways that a wide string could be mapped onto a DDS implementation.
-The first way it could be mapped is onto a uint8_t byte array.
-While this would work, it has the downside that the byte array would be "opaque", and any introspection tools into DDS wouldn't see anything but a stream of bytes.
-The second way it could be mapped is onto the "wchar/wstring" type defined by the DDS specification.
-The DDS specification itself doesn't define a size for wchar, but the X-Types specification (for DDS introspection) does say that a wchar should be mapped onto a Char32.
-Thus, all known encodings would fit into a wchar, and DDS introspection tools could look into wstrings.
-Thus, the current design of wide strings map onto wchar DDS types.
+## Size of a Wide String
 
-**Runtime impact of wide string**
+Both UTF-8 and UTF-16 are variable width encodings.
+To minimize the amount of memory used, the `string` and `wstring` types are to be stored in client libraries according to the smallest possible code point.
+This means `string` must be specified as a sequence of bytes, and `wstring` is to be specified as a sequence of words.
 
-Dealing with wide strings puts more strain on the software of a system, both in terms of speed and in terms of code size.
-This is felt most acutely when talking about small microcontrollers, where both flash size (code space) and performance (processor speed) are at a premium.
-Some of the common encodings (including UTF-8) are defined as being variable width, meaning that a character can take 1, 2, or 4 characters to represent.
-These encoding also typically have the ability to combine one character with the next in non-trivial ways.
-As one of the goals of ROS 2 is to support small, constrained systems, dealing with wide strings may be out of the question.
-Thus, the current ROS 2 design for wide strings makes them optional, defines them as a separate type from regular strings, and recommends against using them for common messages.
+Some DDS implementations use 32bit types to store wide strings values.
+This may be due to DDS-XTYPES 1.1 spec saying `wchar` is a 32bit value.
+ROS 2 will instead aim to be compatible with DDS-XTYPES 1.2 and use 16bit storage for wide characters.
+Generated code for ROS 2 messages will automatically handle the conversion when a message is serialized or deserialized.
 
-### What is the encoding of the wide string?
+### Bounded wide strings
 
-There are two main ways that ROS 2 can provide for the encoding of a wide string.
-Either ROS 2 can define exactly what the encoding will be for all strings, or it can allow the user to specify the encoding when handing it to ROS 2.
-There are pros and cons to each approach.
+Embedded systems accepting strings may use message definitions that restrict the maximum size of a string.
+These are referred to as bounded strings.
+Their purpose is to restrinct the amount of memory used, so the bounds must be specified as units of memory.
+If a `string` field is bounded then the size is given in bytes.
+Similarly the size of a bounded `wstring` is to be specified in words.
+It is the responsibility of whoever populates a bounded `string` or `wstring` to make sure it contains whole code points only.
+Partial code points are indistinuguishable from unknown encodings, so a bounded string whose last code point is incomplete will be published without error.
 
-**ROS 2 defines the encoding**
+## Runtime impact of wide string
 
-Pros:
+Dealing with wide strings puts more strain on the software of a system, both in terms of speed and code size.
+UTF-8 and UTF-16 are both variable width encodings, meaning a code point can take 1 (UTF-8 only), 2, or 4 bytes to represent.
+It may take multiple code points to represent a single user perceived character.
+One of the goals of ROS 2 is to support microcontrollers that are contrained by both code size and processor speed.
+Some wide string operations like splitting a string on a user perceived character may not be possible on these devices.
 
-* All strings known to be in the same encoding.
+However, whole string equality checking is the same whether using wide strings or not.
+Further splitting a UTF-8 string on an ASCII character is identical to splitting an ASCII character on an ASCII string.
+If a microcontroller must do other types of string manipluation, then a `string` type should still be used.
+Code in the subscriber callback should stop processing a message when it encounters a byte greater than 127 in a `string` field.
 
-Cons:
+## What does the API look like to a user of ROS 2?
 
-* If the user-level code uses a different encoding, the user is responsible for doing a conversion to the ROS 2 defined encoding.
-  This can involve some runtime cost.
+### Python 3
 
-**User defines the encoding**
-
-Pros:
-
-* No conversions necessary for user-to-ROS 2 API.
-
-Cons:
-
-* The user must always tell ROS 2 what encoding the data is in.
-* The encoding needs to be transmitted on the wire to the other side.
-* It is not clear how ROS 2 will transmit the encoding.
-  Fixed list?
-  Which encodings go into the list?
-
-While a user defined encoding looks somewhat attractive at first, the downsides to it make it difficult to determine exactly which encodings are in use.
-The downsides also mean that any program that wants to parse ROS 2 messages (or bag files) may potentially have to deal with many different encodings.
-Additionally, Unicode (and in particular, UTF-8) are very commonplace now, while other encodings are becoming more esoteric.
-Even earlier Unicode (UTF-16 and UTF-32) encodings are essentially deprecated in favor of UTF-8.
-For these reasons, the current design of the ROS 2 wide string is to use a UTF-8 encoding for all strings.
-
-### What does the API look like to a user of ROS 2?
-
-#### Python 3
-
-Python 3 splits byte arrays from strings pretty cleanly.
-Thus, in Python 3, a string is a sequence of characters, where each character can take 1 or more bytes.
+In Python 3 a string is a sequence of characters where each character can take 1 or more bytes.
 Byte arrays are sequences of bytes.
-Thus, the ROS 2 API for dealing with wide strings (such as the publish API) will take just a string type in, and encode it as utf-8.
-The ROS 2 API for dealing with regular strings will take a string type in, and encode it as ASCII.
+Thus the ROS 2 API for dealing with `wstring` will take a string type in internally and convert it to UTF-16.
 
 **Example**
 
@@ -168,26 +164,23 @@ if __name__ == '__main__':
     main()
 ```
 
-Notice that this code looks almost identical to the same demo for String.
-That's because in python, the characters in the string and the byte representation are cleanly separated.
-Thus, the API looks exactly the same as it would for the String type.
+This code looks almost identical to the same demo for String.
 
-#### C++
+### C++
 
-C++ doesn't have good built-in support for wide strings.
-Using wchar_t is generally not a good option, as wchar_t has different sizes on different platforms (2 bytes on Windows, 4 bytes on Linux).
-Thus, ROS 2 will define a new type to deal with wide strings.
-The APIs that deal with wide strings will expect this new type to be passed in.
+In c++ wsing `wchar_t` has different sizes on different platforms (2 bytes on Windows, 4 bytes on Linux).
+Instead ROS 2 will use `char16_t` for characters of wide strings, and `std::u16string` for wide strings themselves.
 
 **Example**
 
 ```
+#include <codecvt>
 #include <iostream>
 #include <memory>
 
 #include "rclcpp/rclcpp.hpp"
 
-#include "std_msgs/msg/w_string.hpp"
+#include "std_msgs/msg/wstring.hpp"
 
 int main(int argc, char * argv[])
 {
@@ -204,9 +197,10 @@ int main(int argc, char * argv[])
 
   auto msg = std::make_shared<std_msgs::msg::WString>();
   auto i = 1;
+  std::wstring_convert<std::codecvt<char16_t, char, std::mbstate_t>, char16_t> convert_to_u16;
 
   while (rclcpp::ok()) {
-    utf8_string hello("Hello World: " + std::to_string(i++));
+    std::u16string hello(u"Hello World: " + convert_to_u16.from_bytes(std::to_string(i++)));
     msg->data = hello;
     std::cout << "Publishing: '" << msg->data.cpp_str() << "'" << std::endl;
     chatter_pub->publish(msg);
@@ -217,29 +211,3 @@ int main(int argc, char * argv[])
   return 0;
 }
 ```
-
-Here, ROS 2 has implemented a new type called "utf8_string" (this is still to be implemented, and the name may change).
-The publisher takes in types of utf8_string, and generates the correct data on the wire.
-
-### Bounded wide strings
-
-A note on bounded wide strings.
-ROS 2 allows for "bounded" strings and wide strings.
-These are strings that can be no longer than their bounded number.
-In the case of strings, the width of a character == the width of a byte in most cases, so the number of bytes in a bounded string is generally the bound (plus 1 for the NULL-termination).
-Wide strings are a little bit different.
-When dealing with bounded wide strings, the bound is on the number of characters, not bytes.
-Thus the maximum number of bytes the bounded wide string can be may be, and most likely will be, larger than the number of characters.
-
-## Summary
-
-To summarize, the current ROS 2 design for wide strings:
-
-* Maps wide characters onto the DDS type wchar.
-  This in turn maps wide strings on the DDS type wstring.
-* Allows them to be optionally supported by an implementation.
-* Defines wide strings as a separate type from strings.
-* Recommends against using wide strings for common messages.
-* Defines the encoding for all wide string to be UTF-8.
-* For python the APIs that deal with wstrings will take a regular str type in.
-* For C++, ROS 2 will define a new type to deal with wstrings.
--- a/articles/WideStrings.md
+++ b/articles/WideStrings.md
@@ -2,7 +2,8 @@
 layout: default
 title: Unicode Support
 abstract:
-  This article desribes how ROS 2 will support multi-byte character sets.
+  This article describes how ROS 2 will support sending multi-byte character data using the [Unicode](https://en.wikipedia.org/wiki/Unicode) standard.
+  It also describes how such data will be sent over the ROS 1 bridge.
 author: '[Chris Lalancette](https://github.com/clalancette)'
 published: true
 ---
@@ -22,8 +23,6 @@ Original Author: {{ page.author }}
 
 Some users would like to send text data in languages that cannot be represented by ascii characters.
 Currently ROS 1 only supports ASCII data in the string field but [allows users to populate it with UTF-8](http://wiki.ros.org/msg).
-This article describes how ROS 2 will support sending multi-byte character data using the [Unicode](https://en.wikipedia.org/wiki/Unicode) standard.
-It also describes how such data will be sent over the ROS 1 bridge.
 
 Note that topic names cannot use multi-byte characters as they are disallowed by the DDS specification.
 See the [topic and service name](/articles/topic_and_service_names.html) design document for more information.
--- a/articles/WideStrings.md
+++ b/articles/WideStrings.md
@@ -1,6 +1,7 @@
 ---
 layout: default
 title: Unicode Support
+permalink: articles/wide_strings.html
 abstract:
   This article describes how ROS 2 will support sending multi-byte character data using the [Unicode](https://en.wikipedia.org/wiki/Unicode) standard.
   It also describes how such data will be sent over the ROS 1 bridge.
--- a/articles/WideStrings.md
+++ b/articles/WideStrings.md
@@ -149,7 +149,7 @@ Instead ROS 2 will use `char16_t` for characters of wide strings, and `std::u16s
 
 ```
 #include <codecvt>
-#include <iostream>
+#include <cstdio>
 #include <memory>
 
 #include "rclcpp/rclcpp.hpp"
@@ -176,7 +176,7 @@ int main(int argc, char * argv[])
   while (rclcpp::ok()) {
     std::u16string hello(u"Hello World: " + convert_to_u16.from_bytes(std::to_string(i++)));
     msg->data = hello;
-    std::cout << "Publishing: '" << msg->data.cpp_str() << "'" << std::endl;
+    std::printf("Publishing: '%s'\n", msg->data.cpp_str().c_str());
     chatter_pub->publish(msg);
     rclcpp::spin_some(node);
     loop_rate.sleep();
--- a/articles/WideStrings.md
+++ b/articles/WideStrings.md
@@ -126,8 +126,8 @@ It should be computationally inexpensive for it to stop processing a string when
 
 ### Python 3
 
-In Python 3 a string is a sequence of characters where each character can take 1 or more bytes.
-Thus the ROS 2 API for dealing with `wstring` will take a string type in internally and convert it to UTF-16.
+In Python the `str` type will be used for both strings and wide strings.
+Bytes of a known encoding should be converted to a `str` using [bytes.decode](https://docs.python.org/3/library/stdtypes.html#bytes.decode) before being assigned to a field.
 
 **Example**
 
--- a/articles/WideStrings.md
+++ b/articles/WideStrings.md
@@ -2,7 +2,8 @@
 layout: default
 title: WideStrings
 abstract:
-  Some ROS2 users would like to use multi-byte characters for topic data, suitable for supporting multi-byte languages.  This article lays out the problem, and explores potential solutions to this problem.
+  Some ROS2 users would like to use multi-byte characters for topic data, suitable for supporting multi-byte languages.
+  This article lays out the problem, and explores potential solutions to this problem.
 author: '[Chris Lalancette](https://github.com/clalancette)'
 published: true
 ---
@@ -20,31 +21,25 @@ Original Author: {{ page.author }}
 
 ## Background
 
-ROS2 currently supports the concept of strings, which are sequences of single-byte
-characters.  When working with ROS 2.0 topic data, some users would like to use multi-byte
-characters (e.g. UTF-8) for their topic data.  This article explores the
-problem and the possible solutions.  Note that this article specifically does
-not talk about binary data, as that is already handled by using uint8_t arrays.
-It also does not talk about using multi-byte characters for the topic *names*,
-as this is disallowed by the DDS specification.
+ROS2 currently supports the concept of strings, which are sequences of single-byte characters.
+When working with ROS 2.0 topic data, some users would like to use multi-byte characters (e.g. UTF-8) for their topic data.
+This article explores the problem and the possible solutions.
+Note that this article specifically does not talk about binary data, as that is already handled by using uint8_t arrays.
+It also does not talk about using multi-byte characters for the topic *names*, as this is disallowed by the DDS specification.
 
 ## Multi-byte character background
 
-Before delving into ROS 2.0 specifics, some time should be spent discussing multi-byte
-characters in general.  What is meant by multi-byte characters?  The original ASCII
-character set only provided for printable characters using binary sequences 0 to 127.
-Several extensions were made to this to use 128-255, but that approach only extended the
-character set to be able to cover a few more languages.  Several other encodings were
-created as the need arose, and it became clear by the 1990's that a worldwide standard
-was required.  Thus, Unicode was created to be a single specification that could cover
-all of the languages of the world.  The early versions of Unicode actually failed to
-live up to this expectation, and some of those early versions of Unicode are still
-baked into widely used software (Microsoft Windows, Java, etc).  Later on, the Unicode
-standard evolved to truly be able to cover the languages of the entire world, and this
-is what should generally be used for new designs.  As a side benefit, the newer versions of
-Unicode (UTF-8) are backwards compatible with ASCII.  There is a lot of history here, and
-things are much more complicated then the brief introduction above.  For more information,
-the following links provide introductory material:
+Before delving into ROS 2.0 specifics, some time should be spent discussing multi-byte characters in general.
+What is meant by multi-byte characters?
+The original ASCII character set only provided for printable characters using binary sequences 0 to 127.
+Several extensions were made to this to use 128-255, but that approach only extended the character set to be able to cover a few more languages.
+Several other encodings were created as the need arose, and it became clear by the 1990's that a worldwide standard was required.
+Thus, Unicode was created to be a single specification that could cover all of the languages of the world.
+The early versions of Unicode actually failed to live up to this expectation, and some of those early versions of Unicode are still baked into widely used software (Microsoft Windows, Java, etc).
+Later on, the Unicode standard evolved to truly be able to cover the languages of the entire world, and this is what should generally be used for new designs.
+As a side benefit, the newer versions of Unicode (UTF-8) are backwards compatible with ASCII.
+There is a lot of history here, and things are much more complicated then the brief introduction above.
+For more information, the following links provide introductory material:
 
 * [http://kunststube.net/encoding/](http://kunststube.net/encoding/)
 * [http://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms](http://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms)
@@ -54,8 +49,8 @@ the following links provide introductory material:
 The rest of this document is generally going to assume Unicode, unless stated otherwise.
 
 ## Introducing Wide Strings
-To support multi-byte characters, ROS 2.0 will introduce the concept of a wide string.  There are a
-few basic questions to answer on how a wide string will be integrated into ROS 2.0:
+To support multi-byte characters, ROS 2.0 will introduce the concept of a wide string.
+There are a few basic questions to answer on how a wide string will be integrated into ROS 2.0:
 
 * What is the size impact of the wide string?
 * What is the encoding of the wide string?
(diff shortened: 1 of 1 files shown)


articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+
+Original Author: {{ page.author }}
+
+## Background
+
+ROS2 currently supports the concept of strings, which are sequences of single-byte
+characters.  When working with ROS 2.0 topic data, some users would like to use multi-byte
+characters (e.g. UTF-8) for their topic data.  This article explores the
+problem and the possible solutions.  Note that this article specifically does
+not talk about binary data, as that is already handled by using uint8_t arrays.
```
dirk-thomas: I would avoid language specific things like `uint8_t` in this section of the document.

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+encodings go into the list?
+
+While a user defined encoding looks somewhat attractive at first, the downsides
+to it make it difficult to determine exactly which encodings are in use.  The
+downsides also mean that any program that wants to parse ROS 2 messages (or bag
+files) may potentially have to deal with many different encodings.  Additionally,
+Unicode (and in particular, UTF-8) are very commonplace now, while other encodings
+are becoming more esoteric.  Even earlier Unicode (UTF-16 and UTF-32) encodings are
+essentially deprecated in favor of UTF-8.  For these reasons, the current design of
+the ROS 2.0 wide string is to use a UTF-8 encoding for all strings.
```
dirk-thomas: I see two separate dimensions here. Both can be arbitrarily combined:

* The encoding used on the wire:

  * We use one specific (agreed upon) encoding on the wire.

  * The data on the wire is annotated with the used encoding.

* The user facing API - some options might only make sense in some languages:

  * The user passes a unicode string (e.g. in Python) which is agnostic to any encoding.

  * The user passes a string with a specific (agreed upon) encoding. The encoding does not necessarily have to match the wire - an automatic conversion could be applied.

  * The user passes a string together with the information identifying the encoding of that string.

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+        # TODO(wjwwood): need to spin_some or spin_once with timeout
+        sleep(1)
+
+
+if __name__ == '__main__':
+    main()
+```
+
+Notice that this code looks almost identical to the same demo for String.  That's because
+in python, the characters in the string and the byte representation are cleanly
```
dirk-thomas: Nitpick: `python` is name and should therefore be camelcase: `Python`

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+separated.  Thus, the API looks exactly the same as it would for the String type.
+
+#### C++
+
+C++ doesn't have good built-in support for wide strings.  Using wchar_t is generally
+not a good option, as wchar_t has different sizes on different platforms (2 bytes on
+Windows, 4 bytes on Linux).  Thus, ROS 2.0 will define a new type to deal with wide strings.
```
dirk-thomas: Imo this conclusion is drawn a bit early. If the user would interact with the API using UTF-8 encoded string I don't see a need for a new type.

dirk-thomas (commented):
Please see https://github.com/ros2/ros2/wiki/Developer-Guide#markdown for the style guide for Markdown files. (I am not sure why Travis marked this as "green" - locally the tests fail for me as I expected).

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+
+There are 2 parts to this question.  One is how the wide string is mapped onto the underlying
+DDS implementation, which defines how large messages will be on-the-wire (and on disk).  The
+other part of the question deals with the runtime impact (in terms of speed and code size) of
+having to deal with wide strings.
+
+**Mapping onto the DDS implementation**
+
+There are a couple of different ways that a wide string could be mapped onto a
+DDS implementation.  The first way it could be mapped is onto a uint8_t byte
```
mikaelarguedas: overlapping @dirk-thomas comment's above, I think in this section we should refer to the DDS types like you do below when referring to the X-types spec

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+the ROS 2.0 wide string is to use a UTF-8 encoding for all strings.
+
+### What does the API look like to a user of ROS 2.0?
+
+#### Python 3
+
+Python 3 splits byte arrays from strings pretty cleanly.  Thus, in Python 3, a
+string is a sequence of characters, where each character can take 1 or more bytes.
+Byte arrays are sequences of bytes.  Thus, the ROS 2.0 API for dealing with wide strings
+(such as the publish API) will take just a string type in, and encode it as utf-8.
```
mikaelarguedas: Nit: UTF-8 like in the rest of the document

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+    if args is None:
+        args = sys.argv
+
+    rclpy.init(args)
+
+    node = rclpy.create_node('talker')
+
+    chatter_pub = node.create_publisher(WString, 'chatter', qos_profile_default)
+
+    msg = String()
```
mikaelarguedas: Nit: WString

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+
+    i = 1
+    while True:
+        msg.data = 'Hello World: {0}'.format(i)
```
mikaelarguedas: We could also pass a string with non-ASCII characters for this example (e.g. `こんにちは`)

articles/WideStrings.md
mikaelarguedas: It would be interesting to see how Connext deals with it, they support a much wider variety of operating systems than us and use 4-byte chars everywhere from what I recall (need to be verified, last time I look into it was a long time ago). Maybe they just match the X-types spec and don't bother with the flexible wchar defined in the DDS one

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+#include "std_msgs/msg/w_string.hpp"
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+
+  auto node = rclcpp::node::Node::make_shared("talker");
+
+  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
+  custom_qos_profile.depth = 7;
```
mikaelarguedas: Nit: we can remove all qos specific stuff and use the default. That'll make the example shorter and more focused on the wstring aspect

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Bounded wide strings
+
+A note on bounded wide strings.  ROS 2.0 allows for "bounded" strings and wide strings.  These
+are strings that can be no longer than their bounded number.  In the case of strings, the
+width of a character == the width of a byte in most cases, so the number of bytes in a bounded
+string is generally the bound (plus 1 for the NULL-termination).  Wide strings are a little bit different.
+When dealing with bounded wide strings, the bound is on the number of characters, not bytes.  Thus
+the maximum number of bytes the bounded wide string can be may be, and most likely will be, larger
+than the number of characters.
```
mikaelarguedas: It would be interesting to explore how preallocation would work in this case. The main 2 use cases for bounded strings are memory constrained system (microcontrollers) and realtime. One conclusion could be that for both these use cases we don't recommend using this type. Otherwise we would need to explore memory allocation strategies (e.g. allocate 4byte per character and we're happy if it takes less)

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Summary
+
+To summarize, the current ROS 2.0 design for wide strings:
+
+* Maps wide characters onto the DDS type wchar.  This in turn maps wide strings on the DDS type wstring.
+* Allows them to be optionally supported by an implementation.
+* Defines wide strings as a separate type from strings.
```
mikaelarguedas: Is there any implication on whether we should or not implement a wchar type accordingly ?

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+* Recommends against using wide strings for common messages.
+* Defines the encoding for all wide string to be UTF-8.
+* For python the APIs that deal with wstrings will take a regular str type in.
```
mikaelarguedas: Nit: Python

clalancette:
About the markdown linter; it passes for me locally.  I'm using nodejs 4.2.6 (the built-in version from Ubuntu 16.04), and did an npm install of remark-cli as mentioned in the documentation.  I wonder what I'm doing differently; any thoughts?

articles/WideStrings.md
clalancette: Sure, I changed the language here a bit.

articles/WideStrings.md
clalancette: Fixed.

dirk-thomas:
@clalancette You need to install additional packages like `remark-lint-sentence-newline` (https://github.com/ros2/design/blob/0cf5be5594345cc8d808d8c398be3525d4212bff/.travis.yml#L10).

articles/WideStrings.md
tfoote: Although I agree that generally strings should be recommended against for use in realtime. It does seem that we would need a preallocation scheme for bounded strings.

articles/WideStrings.md
tfoote: It would be good to address these separately. It's quite possible our "wide string" datatype could implicitly do these conversions automatically.

clalancette:
@dirk-thomas What version of remark and the plugins do you have installed?  No amount of installing plugins and/or invoking remark will get me anything but green status.  From some preliminary looks (https://github.com/why-jay/remark-lint-sentence-newline/issues/25), it looks as though remark may have changed, and remark-lint-sentence-newline is no longer compatible, but I am just guessing, since I can't make it work.

articles/WideStrings.md
clalancette: I actually have come to the idea that adding this new type for C++ is the whole point of the WideString exercise.  The fact is that today, you can deal with UTF-8 data in ROS1 *or* ROS2 in 2 different ways: either by passing it through a byte array, or by passing it through the String type.  Neither place constraints on what bytes can be placed in them, and both of them will carry the data over to the destination, where it can be pulled back out.

Where WStrings come into play is putting some structure around this and guaranteeing that all parties agree on the encoding and agree on an API for dealing with them.  Since the story around UTF-8 in C++ is generally so weak, this would be an opportunity for ROS2 to make things easier for users, both at the API level, and when dealing with the data (as they'll always know what the encoding in).

So with all of that being said, we don't really *need* a WString type; it just makes dealing with certain things easier.  Given that, I'm not quite sure where we go with this.

dirk-thomas:
@clalancette My output for `npm list | grep remark`:

```
├─┬ remark-cli@1.0.0
│ ├─┬ remark@5.0.1
│ │ ├─┬ remark-parse@1.0.0
│ │ ├─┬ remark-stringify@1.0.0
├─┬ remark-lint@4.0.2
│ ├─┬ remark-message-control@2.0.2
└── remark-lint-sentence-newline@2.0.0
```

articles/WideStrings.md
clalancette: For preallocation, I agree that we'll probably have to use 4 bytes per character, and just hope it takes less.  This is yet another reason why we wouldn't want to use this type down on microcontrollers, as this would waste precious RAM.

articles/WideStrings.md
clalancette: Oh, true, good point.  I'll fix that.  Thanks.

clalancette: (and for the above comment)

As far as I can tell, Connext just calls a wchar 4 bytes.  That being said, their implementation of wchar/wstring is pretty immature, and their tools (for instance) don't know how to deal with wstrings at all.

articles/WideStrings.md
clalancette: Hm, yeah.  When I was testing stuff out I did do that.  Maybe I should switch to that to show more about what we are talking about.

k-okada: If I understand correctly, this article proposes new ROS `std_msgs/WString`. If this is correct. I have impression that this is not convenient for those who uses non-English speaker. We'll consider both ASCII string and Non-ASCII string as "string"

For example, we'll write 
https://github.com/jsk-ros-pkg/trans_system/blob/88be85431823a9a7cdd5e846dcb887bb16fa844d/jaxon_ros_bridge/scripts/robot_command_server.py#L33-L35
```
        self.sound_pub.publish(SoundRequest(sound=SoundRequest.SAY,
                                            command=SoundRequest.PLAY_ONCE,
                                            arg="Servo off"))
```
And if we want robot to speak in Japanese, we have to change `SoundRequest` to different message type, that's feels strange to me. We might want to write something like
```
        if language_type == "English":
           text = "Servo off"
        else:
           text = "サーボオフ"
        self.sound_pub.publish(SoundRequest(sound=SoundRequest.SAY,
                                            command=SoundRequest.PLAY_ONCE,
                                            arg=text))
```

I think you have already discussed this issue and there might be huge drawback on this strategy...

clalancette: @k-okada Yes, you understand correctly.  This proposal was to add a separate WString type.  Our main motivation for that was because proper UTF-8 decoding libraries can be expensive (both in memory and code size), and we didn't want to force all users, particularly microcontrollers, to have to decode UTF-8.  That being said, this is just a proposal, and we are happy to consider something different.

It might be helpful to get a better idea of what problems you have with the current string type (in ROS1 and/or ROS2).  What specific problems have you had, and what would you like to see the string type do differently?

k-okada: At this moment, ROS1 string type is ok for us (academic research level development), it works fine. Still there is something we want to fix (https://github.com/ros/genpy/pull/31), we can use them for now, after patching to application layer (https://github.com/ros/dynamic_reconfigure/pull/45)

On ROS2, we can not send multibyte language over ros string message (see https://github.com/ros2/rosidl/issues/208 for issue ticket, the data is totally corrupted), But this problem is only for a python client.

So maybe what I'm talking here is about primitive string data type, and all client should/try to send utf-8 encoded byte sequence as a string data.
On C/C++ it user's responsibility to pass utf-8 based string, and on Python, it uses string data and need to convert to byte sequence on rosidl layer (see https://github.com/ros2/rosidl/pull/209 for implementation).

- I just found MSVC uses non-utf8 string as a default, so may be we need something like https://msdn.microsoft.com/ja-jp/library/mt708823.aspx, if we want to ask users to write their code on utf-8.
- If this proposal means "if you wan to use non-ascii character, use `WString`", that's not a good idea. We (non-ascii people) usually regards utf-8 as a string. We'll often write a code like `msg = "Hello" if True else "こんにちは"`, so it is difficult to decide if variable `a` is `String` or `WString`
- I would not against `WString` itself, may be commercial level developer who design the message type very carefully would like this idea.

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+is what should generally be used for new designs.  As a side benefit, the newer versions of
+Unicode (UTF-8) are backwards compatible with ASCII.  There is a lot of history here, and
+things are much more complicated then the brief introduction above.  For more information,
+the following links provide introductory material:
+
+* [http://kunststube.net/encoding/](http://kunststube.net/encoding/)
+* [http://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms](http://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms)
+* [http://www.diveintopython3.net/strings.html](http://www.diveintopython3.net/strings.html)
+* [http://stackoverflow.com/questions/402283/stdwstring-vs-stdstring](http://stackoverflow.com/questions/402283/stdwstring-vs-stdstring)
+
```
sbarthelemy, line 38: I found that article quite helpful: http://utf8everywhere.org/

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Background
+
+ROS2 currently supports the concept of strings, which are sequences of single-byte
+characters.  When working with ROS 2.0 topic data, some users would like to use multi-byte
```
sbarthelemy: Maybe worth to mention that ROS2 has no encoding convention besides ASCII (7 bits).
Some of the issues you raise (agreeing on encoding) already exist when trying to use non-ASCII (but single-byte) character set.

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+Several extensions were made to this to use 128-255, but that approach only extended the
+character set to be able to cover a few more languages.  Several other encodings were
+created as the need arose, and it became clear by the 1990's that a worldwide standard
+was required.  Thus, Unicode was created to be a single specification that could cover
+all of the languages of the world.  The early versions of Unicode actually failed to
+live up to this expectation, and some of those early versions of Unicode are still
+baked into widely used software (Microsoft Windows, Java, etc).  Later on, the Unicode
+standard evolved to truly be able to cover the languages of the entire world, and this
+is what should generally be used for new designs.  As a side benefit, the newer versions of
+Unicode (UTF-8) are backwards compatible with ASCII.  There is a lot of history here, and
```
sbarthelemy: UTF-8 is not a newer version of unicode.

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
+---
+layout: default
+title: WideStrings
+abstract:
+  Some ROS2 users would like to use multi-byte characters for topic data, suitable for supporting multi-byte languages.  This article lays out the problem, and explores potential solutions to this problem.
```
sbarthelemy: do you mean "language" as in programing language or human language?

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+
+* [http://kunststube.net/encoding/](http://kunststube.net/encoding/)
+* [http://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms](http://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms)
+* [http://www.diveintopython3.net/strings.html](http://www.diveintopython3.net/strings.html)
+* [http://stackoverflow.com/questions/402283/stdwstring-vs-stdstring](http://stackoverflow.com/questions/402283/stdwstring-vs-stdstring)
+
+The rest of this document is generally going to assume Unicode, unless stated otherwise.
+
+## Introducing Wide Strings
+To support multi-byte characters, ROS 2.0 will introduce the concept of a wide string.  There are a
```
sbarthelemy: I understand "wide string" from the context of C++'s std::wstring, and it implies "fixed size" encoding.
Maybe it is worth clarifying.

articles/WideStrings.md
clalancette: Actually, that's bad wording.  I should fix that.

articles/WideStrings.md
clalancette: It depends on your definition of "newer"; the original Unicode (that came out in 1996 or whatever) didn't have UTF-8, which is why Win32 and Java both use UTF-16 (really UCS-2).  That being said, UTF-8 has been around a long time, so I'll change the wording here.

articles/WideStrings.md
clalancette: > On ROS2, we can not send multibyte language over ros string message (see ros2/rosidl#208 for issue ticket, the data is totally corrupted), But this problem is only for a python client.

Yeah, this just looks like a bug to me, not a fundamental limitation.  We should fix that.

> 
> So maybe what I'm talking here is about primitive string data type, and all client should/try to send utf-8 encoded byte sequence as a string data.

Yes, if we just keep the string type as-is, then we should be able to send UTF-8 data (modulo bugs).

> On C/C++ it user's responsibility to pass utf-8 based string, and on Python, it uses string data and need to convert to byte sequence on rosidl layer (see ros2/rosidl#209 for implementation).

Right.  One other question here would be whether it would be worthwhile to force the API user to use a UTF-8 type in C++, so that we could ensure it was always a valid UTF-8 string.  There are pros and cons to doing this, but it is one way to make sure it is always UTF-8.

> 
>     I just found MSVC uses non-utf8 string as a default, so may be we need something like https://msdn.microsoft.com/ja-jp/library/mt708823.aspx, if we want to ask users to write their code on utf-8.
>     If this proposal means "if you wan to use non-ascii character, use WString", that's not a good idea. We (non-ascii people) usually regards utf-8 as a string. We'll often write a code like msg = "Hello" if True else "こんにちは", so it is difficult to decide if variable a is String or WString

All right, that is good to know.

>     I would not against WString itself, may be commercial level developer who design the message type very carefully would like this idea.

I'm not sure I'd want to make that distinction.  If you don't think this is useful for non-ASCII data (which is the main purpose), then we should reconsider the whole proposal.

k-okada:
2017-03-22 23:53 GMT+09:00 Chris Lalancette <notifications@github.com>:

> One other question here would be whether it would be worthwhile to force
> the API user to use a UTF-8 type in C++, so that we could ensure it was
> always a valid UTF-8 string. There are pros and cons to doing this, but it
> is one way to make sure it is always UTF-8.


Do you mean put `#pragma execution_character_set("utf-8") ` to somewhere in
the header file? or check if the string is UTF-8 before serialization?(or
somewhere else)?, or produce compiler error when we assing non-utf8 to
string variable (not sure if we can do that)

--
◉ Kei Okada

clalancette:
> Do you mean put `#pragma execution_character_set("utf-8") ` to somewhere in the header file? or check if the string is UTF-8 before serialization?(or somewhere else)?, or produce compiler error when we assing non-utf8 to string variable (not sure if we can do that)

I'm thinking more along the lines of making the ->publish() call for the string type only take a UTF8String object (or whatever).  Similarly, the subscription callback would get a UTF8String object.  That way, when the UTF8String is instantiated, we can ensure that it is valid UTF-8, we know the encoding, etc.  Does that make sense?

mikaelarguedas:
I'll try to give more context here and take a step back:

> Maybe worth to mention that ROS2 has no encoding convention besides ASCII (7 bits).
Some of the issues you raise (agreeing on encoding) already exist when trying to use non-ASCII (but single-byte) character set.

That is true and should be mentioned here. This is the convention specified in the [interfaces design document](http://design.ros2.org/articles/interface_definition.html) in the todo: "string does not specify any encoding yet and the transport is agnostic to it, this means commonly it can only contain ASCII"

> If this proposal means "if you wan to use non-ascii character, use WString", that's not a good idea. We (non-ascii people) usually regards utf-8 as a string. We'll often write a code like msg = "Hello" if True else "こんにちは", so it is difficult to decide if variable a is String or WString
I would not against WString itself, may be commercial level developer who design the message type very carefully would like this idea.

The rationale behind this proposal is that the ROS2 message format allows more control on the size of the messages on the wire that is more suitable for embedded or realtime applications (allowing preallocation of memory, relax CPU by skipping any encoding/decoding step etc).
For example in ROS2 you can specify the maximum sixe (in bytes) of your strings:
`string<=10 my_10byte_string`
Being able to pass UTF-8 strings with multibyte characters of variable length will prevent this mechanism to work properly and result in either: trying to access unauthorized memory segments, allocating memory on the fly to fit the received message or crop the message to fit in preallocated memory. None of them being the desired behavior IMO.
I think that the general guideline offered by this proposal would be: "if you want to send ascii strings efficiently (bandwidth, CPU, memory allocation...) you should use `String`, if you want to send human readable strings with multi-(human)language support you should use `WString`". In that scenario `msg = "Hello" if True else "こんにちは"` would be a great candidate for a WString message type.

k-okada:
@clalancette 
> I'm thinking more along the lines of making the ->publish() call for the string type only take a UTF8String object (or whatever)

I think it's nice if we can do that. but may be I suppose people do not want to write code like
```
        msg.data = UTF8String("Hello World")
```
I think they want write code like
```
        msg.data = std::string("Hello World")
```
So what we can do is just to check encode type of string code within `->publish()` call, I suppose.


@mikaelarguedas 

I think basically, C++/C's basic string type deals string as byte sequence, so size/length/strlen of the multibyte string is not a length of the character, but the size of memory. So I think you can control the memory size very easily. 
```
k-okada@p40-yoga:/tmp$ cat hoge.cpp
#include <string>
#include <iostream>
#include <string.h>

int main () {
  std::string msg1 = "Hello";
  std::string msg2 = "こんにちは";
  std::cerr << msg1 << ", " << msg1.size() << std::endl;
  std::cerr << msg2 << ", " << msg2.size() << std::endl;
  std::cerr << msg1.c_str() << ", " << strlen(msg1.c_str()) << std::endl;
  std::cerr << msg2.c_str() << ", " << strlen(msg2.c_str()) << std::endl;
}
k-okada@p40-yoga:/tmp$ g++ -std=c++11-o hoga hoge.cpp ; ./hoga
Hello, 5
こんにちは, 15
Hello, 5
こんにちは, 15
```

If you want to use character length, it seems we can use `u32string` or `u16string`, but my opinion here is instead of using `std:u16string`, use `std:string` and let client library assume default encoding system as 'utf-8'. Of course, If you want more, then you can use WString.
```
#include <string>
#include <iostream>
#include <string.h>

int main () {
  std::u16string msg1 = u"Hello";
  std::u16string msg2 = u"こんにちは";
  std::cerr << msg1.length() << std::endl;
  std::cerr << msg2.length() << std::endl;
}
```
This will outputs `5` for both cases.

This is different from python case.
```
In [1]: hoge = u'こんにちは'

In [2]: len(hoge)
Out[2]: 5
```

> In that scenario msg = "Hello" if True else "こんにちは" would be a great candidate for a WString message type.

I'm afraid that people start creating nodes or messages based on String, for example ( https://github.com/ros-drivers/audio_common/blob/master/sound_play/msg/SoundRequest.msg#L29-L30) and if that happens, it is very hard to change the message type after wards. We do not want to create SoundRequesetUTF8 or sound_node_utf8.py, just for econding.

articles/WideStrings.md
```diff
@@ -0,0 +1,264 @@
...
+When dealing with bounded wide strings, the bound is on the number of characters, not bytes.  Thus
+the maximum number of bytes the bounded wide string can be may be, and most likely will be, larger
+than the number of characters.
+
+## Summary
+
+To summarize, the current ROS 2.0 design for wide strings:
+
+* Maps wide characters onto the DDS type wchar.  This in turn maps wide strings on the DDS type wstring.
+* Allows them to be optionally supported by an implementation.
```
jack-oquin: Are there other "optional" data types in ROS 2 messages? How are they handled?

These may be general questions, but they arise here: 

* What happens when a ROS 2 node attempts to advertise or subscribe to a message that contains a field its platform does not support? 
* Does it fail immediately?
* Does it "succeed", but ignore those fields?
* Is there some other result?

asmodehn:
Joining a bit late here, but I agree with @k-okada and I think English being on-byte is the exception here.

Few quick remarks : 

- ROS supporting multiple programming languages, the message field types should not mirror the types of any one language in particular.
- message field type can relate to the user semantics, with high level types, like `string` (one or multi bytes) `int` (8, 32, 64, ...), etc. OR relate to the operational semantics, and this is not the programming language semantics, it is the semantics of the data sent on the network, ie the (de)serialization, like a `uint8`, `int32`, `string`, `u16string` type indicating how it will be serialized.
- ros message field types should be types on their own, to avoid confusion of semantics between programming languages, even if they can provide a familiar interface. So a `ros::string`, like a `ros::u16string` could store and transmit their encoding, and extend `std::string` and `std::u16string`

asmodehn:
Another relevant piece of information for this discussion that I found is in http://www.omg.org/spec/DDS-XTypes/1.2/Beta1/PDF : 

> 7.2.2.2.1.2.4 String<Char8> type
The default encoding for String<Char8> shall be UTF-8. This encoded shall be used for the
externalized Data Representation (see section 7.4). Language bindings (see section 7.5) may use
the representation that is most natural in that particular language. If this is different from UTF-8
the language binding shall manage the transformation to/from UTF-8 external Data
Representation.

I am not following closely the DDS specs timeline but I translate this to ROS as `string` would mean utf-8 encoding. In [python2](https://docs.python.org/2.7/howto/unicode.html#python-2-x-s-unicode-support) this translate to `unicode` type and in [python3](https://docs.python.org/3/howto/unicode.html#python-s-unicode-support) to `str`. In C++-11 this is `std::string` because [a `char` has to be able to store an 8bit coding fo utf-8](https://en.wikipedia.org/wiki/C%2B%2B11#New_string_literals).

Said otherwise, the "ASCII" implicit meaning of the ROS1 `string` would disappear (UTF8 will provide ASCII backward compat for devs habits). If one wants to send raw ASCII bytes (without encoding), then `uint8[]` is the right type. This is similar to the [change done in python between 2.7 and 3.0](https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit) ( `str` became `bytes` ), so I would say it makes sense to do it this way as well for ROS, even if it might confuse some users at first.

This put in perspective with http://utf8everywhere.org/#conclusions seems to indicate that we should really review the meaning of the ROS `string` message field type to transmit encoding and be utf-8 encoded by default. It will help us to decide how to deal with wide strings when they are _actually_ needed...

articles/WideStrings.md
```diff
@@ -48,87 +48,83 @@ For more information, the following links provide introductory material:
...
 
-While a user defined encoding looks somewhat attractive at first, the downsides to it make it difficult to determine exactly which encodings are in use.
-The downsides also mean that any program that wants to parse ROS 2 messages (or bag files) may potentially have to deal with many different encodings.
-Additionally, Unicode (and in particular, UTF-8) are very commonplace now, while other encodings are becoming more esoteric.
-Even earlier Unicode (UTF-16 and UTF-32) encodings are essentially deprecated in favor of UTF-8.
-For these reasons, the current design of the ROS 2 wide string is to use a UTF-8 encoding for all strings.
+However, whole string equality checking is the same whether using wide strings or not.
+Further splitting a UTF-8 string on an ASCII character is identical to splitting an ASCII character on an ASCII string.
+If a microcontroller must do other types of string manipluation, then a `string` type should still be used.
+Code in the subscriber callback should stop processing a message when it encounters a byte greater than 127 in a `string` field.
```
mikaelarguedas: Can you clarify this recommendation?

articles/WideStrings.md
```diff
@@ -48,87 +48,83 @@ For more information, the following links provide introductory material:
...
-### What does the API look like to a user of ROS 2?
+## What does the API look like to a user of ROS 2?
 
-#### Python 3
+### Python 3
 
-Python 3 splits byte arrays from strings pretty cleanly.
-Thus, in Python 3, a string is a sequence of characters, where each character can take 1 or more bytes.
+In Python 3 a string is a sequence of characters where each character can take 1 or more bytes.
 Byte arrays are sequences of bytes.
```
mikaelarguedas: Is this line still necessary?

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+The resulting UTF-8 encoded string will be published as a `string` type.
+If the conversion fails then the bridge will not publish the message.
+
+## Size of a Wide String
+
+Both UTF-8 and UTF-16 are variable width encodings.
+To minimize the amount of memory used, the `string` and `wstring` types are to be stored in client libraries according to the smallest possible code point.
+This means `string` must be specified as a sequence of bytes, and `wstring` is to be specified as a sequence of words.
+
+Some DDS implementations use 32bit types to store wide strings values.
```
dirk-thomas: An important point is that we expect DDS impl. to use 16 bit in the future. E.g. RTI has already stated that future versions will do so.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+This may be due to DDS-XTYPES 1.1 spec saying `wchar` is a 32bit value.
```
dirk-thomas: Ref to section in DDS-XTYPES 1.1.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+ROS 2 will instead aim to be compatible with DDS-XTYPES 1.2 and use 16bit storage for wide characters.
```
dirk-thomas: Ref to section in DDS-XTYPES 1.2.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+Generated code for ROS 2 messages will automatically handle the conversion when a message is serialized or deserialized.
+
+### Bounded wide strings
+
+Embedded systems accepting strings may use message definitions that restrict the maximum size of a string.
```
dirk-thomas: I would remove the "embedded systems" part since this happens in the message definition and affect all platforms.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+These are referred to as bounded strings.
+Their purpose is to restrinct the amount of memory used, so the bounds must be specified as units of memory.
+If a `string` field is bounded then the size is given in bytes.
+Similarly the size of a bounded `wstring` is to be specified in words.
+It is the responsibility of whoever populates a bounded `string` or `wstring` to make sure it contains whole code points only.
+Partial code points are indistinuguishable from unknown encodings, so a bounded string whose last code point is incomplete will be published without error.
```
dirk-thomas: > will be published without error.

Is this what you actually wanted to say?

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Runtime impact of wide string
+
+Dealing with wide strings puts more strain on the software of a system, both in terms of speed and code size.
+UTF-8 and UTF-16 are both variable width encodings, meaning a code point can take 1 (UTF-8 only), 2, or 4 bytes to represent.
```
dirk-thomas: In the case of UTF-8 3 bytes is also an option.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+It may take multiple code points to represent a single user perceived character.
+One of the goals of ROS 2 is to support microcontrollers that are contrained by both code size and processor speed.
+Some wide string operations like splitting a string on a user perceived character may not be possible on these devices.
+
+However, whole string equality checking is the same whether using wide strings or not.
+Further splitting a UTF-8 string on an ASCII character is identical to splitting an ASCII character on an ASCII string.
+If a microcontroller must do other types of string manipluation, then a `string` type should still be used.
+Code in the subscriber callback should stop processing a message when it encounters a byte greater than 127 in a `string` field.
```
dirk-thomas: > should

I would rather say "could" and only if it can't handle the encoding.

articles/WideStrings.md
sloretz: It's trying to justify using the `string` type on microcontrollers that need to do string manipulation by saying it is cheap to filter non-ascii values when using UTF-8. I'll expand this paragraph to clear it up.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+
+## What does the API look like to a user of ROS 2?
+
+### Python 3
+
+In Python 3 a string is a sequence of characters where each character can take 1 or more bytes.
+Byte arrays are sequences of bytes.
+Thus the ROS 2 API for dealing with `wstring` will take a string type in internally and convert it to UTF-16.
```
dirk-thomas: I find this description a bit confusing in a Python section. The internal conversion is something the user doesn't "see". I would rather emphasize that both `string` and `wstring` are represented in Python as `str` and depending on the field tpe in the message they will be encoded differently when bein serialized.

articles/WideStrings.md
dirk-thomas: @k-okada Please wait until the proposal is "complete". There is a clear rational for providing both `string` as well as `wstring` in IDL. It will be recommended to use `string` with UTF-8 encoded data. `wstring` is being defined for interoperability only. Both types allow any unicode code point to be stored.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
+---
+layout: default
+title: WideStrings
+abstract:
+  Some ROS2 users would like to use multi-byte characters for topic data, suitable for supporting multi-byte languages.
```
dirk-thomas: The combination of title and abstract is rather confusing. The main point is that you don't need to use `wstring` to send unicode. So maybe changing the title to mention both `string` as well as `wstring` would make sense.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Multi-byte character background
+
+Before delving into ROS 2 specifics, some time should be spent discussing multi-byte characters in general.
+What is meant by multi-byte characters?
+The original ASCII character set only provided for printable characters using binary sequences 0 to 127.
+Several extensions were made to this to use 128-255, but that approach only extended the character set to be able to cover a few more languages.
+Several other encodings were created as the need arose, and it became clear by the 1990's that a worldwide standard was required.
+Thus, Unicode was created to be a single specification that could cover all of the languages of the world.
+The early versions of Unicode actually failed to live up to this expectation, and some of those early versions of Unicode are still baked into widely used software (Microsoft Windows, Java, etc).
```
dirk-thomas: This should have a reference to "back the claim" and be more specific about what the "problem" is.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Introducing Wide Strings
+ROS 2 messages will have a new [primitive field type](/articles/interface_definition.html) `wstring`.
+This purpose is to allow ROS 2 nodes to comminicate with non-ROS DDS entities using an IDL containing a `wstring` field.
+The encoding of data in this type should be UTF-16 to match DDS-XTYPES 1.2.
+Since both UTF-8 and UTF-16 can encode the same code points, new ROS 2 messages should prefer `string` over `wstring`.
+
+## Encodings are Recommendations
+The choice of UTF-8 or UTF-16 for `string` and `wstring` are recommendations.
+A `string` or `wstring` field is allowed to be populated and published with an unknown encoding.
```
dirk-thomas: The rest of the paragraph reads like "you can do something different but it is probably gone blow". So I am not sure why we should "allow" this. I would rather lean towards being strict and say the encoding shall be UTF-8 / UTF-16. And keep mentioning the technical restrictions from below what they can't contain.

articles/WideStrings.md
```diff
@@ -0,0 +1,213 @@
...
+An encoding like UTF-32 cannot be used since it may have zero bytes or words in a code point.
+If UTF-32 encoded strings need to be transmitted then one should either convert to UTF-8 and use a `string` or use a `uint32` array.
+
+<div class="alert alert-warning" markdown="1">
+  <b>TODO:</b> Can ROS 1 publish a string with NULL bytes?
+</div>
+
+## Unicode Strings Across ROS 1 Bridge
+Since ROS 1 and 2 both allow `string` to be UTF-8, the ROS 1 bridge will pass values unmodified between them.
+If a message sent from ROS 2 to ROS 1 fails to serialize because a string contains values that are not legal UTF-8 then it won't be published by the bridge.
```
dirk-thomas: This might go away if we decide on a mandatory encoding. Anyway I don't think we should set in stone that such messages will be dropped. There are fall back behaviors like ignoring or replacing not-encodable characters which could be considered.

articles/WideStrings.md
sloretz: In 354da47cb1cf7d9fcc1384e0ac4f6d84f2f1e5a2 encodings are required but not guaranteed to be enforced.

articles/WideStrings.md
sloretz: It seems like in some cases replacing or dropping invalid characters could be unsafe. It is possible the string message is a command with an invalid character that would be changed by that strategy. Worst case it might turn a string like  `DELETE ■*` into `DELETE *` where ■ is standing in for an invalid code point.

articles/WideStrings.md
sloretz: It was. In 7d2ca06ce8f8dbd420f6e2e2ce2bc233c82e87a9 I changed it to say they're not guaranteed to be published to match 354da47cb1cf7d9fcc1384e0ac4f6d84f2f1e5a2

articles/WideStrings.md
sloretz: e4cd96a90926ba81071364d9c1576683adebf565

articles/WideStrings.md
sloretz: Removed in 45a8a506c05b38c3f3877a878457c5e11e8ed101

articles/WideStrings.md
sloretz: ddf99014af761f0965d54086b1b49f8fb420eb4a

articles/WideStrings.md
sloretz: 9c4e7b5ab5316d27d62d4e083b035ec27201cd73

articles/WideStrings.md
sloretz: Tried to make this clearer in 9c4e7b5ab5316d27d62d4e083b035ec27201cd73

articles/WideStrings.md
sloretz: ae17066ad6bbe954095e516f4c35975a4cb84e7c

articles/WideStrings.md
sloretz: ae17066ad6bbe954095e516f4c35975a4cb84e7c

articles/WideStrings.md
sloretz: ae17066ad6bbe954095e516f4c35975a4cb84e7c

articles/WideStrings.md
sloretz: 5124c5a1738a9a7d9d407383a07e84db9105eef1

articles/WideStrings.md
sloretz: I made the title and abstract about unicode support rather than just wide strings in 4102e8e9bc44f4b7c73de2ce89dfbdbcb21f358b

articles/WideStrings.md
sloretz: I changed the wording to say dropping the entire message is the default behavior while allowing one to opt-in to another strategy 4a21e094e0ea3ab7bc8beeeca67395a781bbf810

articles/WideStrings.md
sloretz: I condensed this section into the `Background` section above in b23f68076d8aff2f0bc1f9adcd28d0f0e06ce84a . It no longer includes this claim, and instead points users to links where they can read more. I did this because I think the rest of the document can be understood without knowing about this point.

articles/WideStrings.md
sloretz: Included in b23f68076d8aff2f0bc1f9adcd28d0f0e06ce84a

sloretz:
I moved this PR to `in review`. I think all the feedback and discussion conclusions have been included.

articles/WideStrings.md
```diff
@@ -0,0 +1,186 @@
...
+    msg.data = 'Hello World' + bytes([0x00, 0x00, 0x26, 0x3A]).decode('utf-32-be')
+    print('Publishing: "{0}"'.format(msg.data))
+    chatter_pub.publish(msg)
+    node.destroy_node()
+    rclpy.shutdown()
+```
+
+### C++
+
+In c++ wsing `wchar_t` has different sizes on different platforms (2 bytes on Windows, 4 bytes on Linux).
```
sloretz: wsing -> wstring

Omegastick:
@clalancette @k-okada @asmodehn Perhaps we can have the best of both worlds by making "wide strings" the default and having a "narrow string" type for embedded systems.

While having both "strings" and "wide strings" technically allows for UTF-8 everywhere it is needed, by making the ASCII the default we do two things: we steepen the learning curve for newcomers (they will be surprised when they can't use the regular "string" type for some of their strings) and, inevitably, some applications will use the ASCII "string" type where they should really use the "wide string" type (because it is the default).

As a user who primarily writes uses ROS for non-English applications, trying to use a package only to find it doesn't play nicely with UTF-8 is a nightmare. It brings me back to days spent fiddling with `.encode()` and `.decode()` in Python 2.7. This should not be a problem in modern frameworks. If we have the "default" string type consider UTF-8 anything other than a first-class citizen, some people will use that when they should be using the "UTF-8 string" type.

On the other hand, developers of embedded systems sometimes can't handle UTF-8. This can be dealt with by giving them a "narrow string" type (the DDS "string"), that supports their use-case. But, by default, strings would support UTF-8.

articles/WideStrings.md
```diff
@@ -0,0 +1,186 @@
...
+The IDL specification forbids `string` from containing `NULL` values.
+To be compatible, a ROS message `string` field must not contain zero bytes, and a `wstring` field must not contain zero words.
+This restriction will be enforced.
+
+## Unicode Strings Across ROS 1 Bridge
+Since ROS 1 and 2 both allow `string` to be UTF-8, the ROS 1 bridge will pass values unmodified between them.
+If a message with a string field fails to serialize because the content is not legal UTF-8 then the default behavior will be to drop the entire message.
+Other strategies like replacing invalid bytes could unintentionally change the meaning, so they will be opt-in if available at all.
+
+If a ROS 2 message has a field of type `wstring` then the bridge will attempt to convert it from UTF-16 to UTF-8.
```
dirk-thomas, line 76: Is that currently implemented?

articles/WideStrings.md
```diff
@@ -0,0 +1,186 @@
...
+
+if __name__ == '__main__':
+    rclpy.init(sys.argv)
+
+    node = rclpy.create_node('talker')
+
+    chatter_pub = node.create_publisher(WString, 'chatter', qos_profile_default)
+
+    msg = WString()
+    msg.data = 'Hello World' + bytes([0x00, 0x00, 0x26, 0x3A]).decode('utf-32-be')
```
dirk-thomas: I would suggest to also have an example which contains the non-ASCII character literally.

articles/WideStrings.md
```diff
@@ -0,0 +1,186 @@
...
+  rclcpp::WallRate loop_rate(2);
+
+  auto msg = std::make_shared<std_msgs::msg::WString>();
+  auto i = 1;
+  std::wstring_convert<std::codecvt<char16_t, char, std::mbstate_t>, char16_t> convert_to_u16;
+
+  while (rclcpp::ok()) {
+    std::u16string hello(u"Hello World: " + convert_to_u16.from_bytes(std::to_string(i++)));
+    msg->data = hello;
+    std::cout << "Publishing: '" << msg->data.cpp_str() << "'" << std::endl;
```
dirk-thomas: Use `printf` according to our guidelines.

articles/WideStrings.md
```diff
@@ -0,0 +1,186 @@
...
+
+    msg = WString()
+    msg.data = 'Hello World' + bytes([0x00, 0x00, 0x26, 0x3A]).decode('utf-32-be')
+    print('Publishing: "{0}"'.format(msg.data))
+    chatter_pub.publish(msg)
+    node.destroy_node()
+    rclpy.shutdown()
+```
+
+### C++
```
dirk-thomas, line 145: There needs to be a note about the need for a BOM on Windows.

articles/WideStrings.md
sloretz: Probably not. Are you asking for a note to be added saying it's not implemented yet?

articles/WideStrings.md
sloretz: Added non-ASCII character in f82dfd9

articles/WideStrings.md
sloretz: `std::printf` in 5f3c97fd8326deb728fe3ebf6f573212bd8d2646

articles/WideStrings.md
dirk-thomas: Yes, e.g.

articles/WideStrings.md
sloretz: Where is the BOM needed? At the start of the source file? At the start of all string literal?

dirk-thomas: Same as ros2/rcl_interfaces#79.

articles/WideStrings.md
sloretz: Added note in 9818383

dirk-thomas: I don't think anything needs to happen for `strings`. They are just copied one-to-one already. So the note only applies to `wstrings` afaik.

articles/WideStrings.md
sloretz: Does 3996b8d describe it accurately?

articles/WideStrings.md
sloretz: Limited note to `wstring` fields in ecf689b

articles/WideStrings.md
dirk-thomas: :+1: (I added 5588980f06a0970a5e232bec65b933505355704e for clarification).

dirk-thomas (approved):

dirk-thomas (approved):

dirk-thomas (approved):
LGTM

PR #118, opened 2017-03-10
Clarify that support for non-ROS packages is a requirement

dhood:
The key point to make here is that we are not just supporting other packages because it would be nice to do so, but because we need it.


dirk-thomas:
Great improvement!

PR #119, opened 2017-03-12
text edits

tkruse:
Mostly suggestions to keep focus on change, and add more context. None of it is essential, I was writing while reading. 

I added package-managers for larger audiences.

I started structuring of essential and non-essential features.

Feel free to cherry pick what you like and close PR without merge without further discussion.

Some comments independent from my PR (but less chat in the main PR)
* The goal is not motivated yet. I hint at issues with the current situation, but without explaining those issues, it is difficult to make a decision solving them.
* I noticed rosbuild was not mentioned much so far, nor tools relying on the ROS_PACKAGE_PATH. Not sure if that is important for the discussion
* while packaging is out of the scope for the tool, I believe the package manifest provides packaging data, might be worth mentioning (Also because of catkin_simple)
* it might be helpful to compare the "anatomy" of catkin and ament files, to see the problems in migrating from one to the other
* Devel_spaces not mentioned yet
* I think catkin_isolated and catkin_auto can also build non-catkin cmake packages, no? The current text suggests otherwise

The realistic alternative approaches should be more detailed.
So far I see:
* keeping catkin packages same vs. requiring migration
* evolving catkin_tools vs. evolving ament_tools ('from scratch' is equivalent to one of those)

### DIFF
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -245,9 +245,6 @@ The tool aims to support a variety of build systems, use cases, and platforms.
 The above mentioned ones are mainly driven by the needs in the ROS ecosystem but the tool should also be usable outside the ROS ecosystem (e.g. for Gazebo).
 Therefore it should be designed in a way which enables extending its functionality.
 
-Assuming that the tool will be implemented in Python (since that is the case for existing tools) the entry point mechanism provides a convenient way to make the software extensible.
-Extensions don't even have to be integrated into the Python package containing the core logic of the build tool but can easily be provided by additional Python packages.
-This approach will not only foster a modular design and promote clear interfaces but enable external contributions without requiring them to be integrated in a single monolithic package.
 
 Several well known software principles apply:
 
@@ -271,6 +268,11 @@ The following items are possible extension points to provide custom functionalit
 - output handling (e.g. console output, logfiles, status messages, notifications)
 - setup the environment (e.g. `sh`, `bash`, `bat`)
 - completion (e.g. `bash`, `Powershell`)
+
+Assuming that the tool will be implemented in Python (since that is the case for existing tools) the entry point mechanism provides a convenient way to make the software extensible.
+Extensions don't even have to be integrated into the Python package containing the core logic of the build tool but can easily be provided by additional Python packages.
+This approach will not only foster a modular design and promote clear interfaces but enable external contributions without requiring them to be integrated in a single monolithic package.
+
 
 ## Possible Approaches
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -129,13 +129,13 @@ In the following the build systems being used in the ROS ecosystem are briefly d
 
 ### catkin
 
-[catkin](http://wiki.ros.org/catkin) is based on CMake and provides a set of convenience functions to make writing CMake packages easier.
+[catkin](http://wiki.ros.org/catkin) is an addition to CMake and provides a set of convenience functions to make writing CMake packages easier.
 It automates the generation of CMake config files as well as pkg-config files.
 It additionally provides functions to register different kinds of tests.
 
 ### ament_cmake
 
-[ament_cmake](https://github.com/ament/ament_cmake) is an evolution of `catkin` and is also based on CMake.
+[ament_cmake](https://github.com/ament/ament_cmake) is an evolution of `catkin` and is an addition to CMake.
 The main difference between `ament_cmake` and `catkin` is described in [another article](http://design.ros2.org/articles/ament.html).
 In the context of the build tool the biggest difference is that `ament_cmake` generates package-specific files to setup the environment to use the package after it has been built and installed.
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -21,13 +21,16 @@ Original Author: {{ page.author }}
 
 ## Preface
 
-ROS is developed with a federated model.
-It consists of many separately developed and maintained packages.
+In the ROS ecosystem teams typically solve problems by programming in many software packages at the same time.
+This is in contrast to workflows where teams work on a single software package at a time only, and all dependencies are installed and updated by package managers.
+Such a workflow is impracticable without a tool that automatically rebuilds many packages in topological order of dependency.
+The rosbuild tool enabled this workflow first and was a critical factor in the success of ROS.
 
-The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
-For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
-From an efficiency point of view that manual process does not scale well.
+Today, the ROS1 ecosystem has the deprecated rosbuild and the catkin build systems for this purpose.
+The ROS2 ecosystem has the ament buildsystem.
 
+
+This article describes the steps from the current build tools used in the ROS ecosystem to a single universal build tool.
 ## Goal
 
 The goal of the build tool is to build a set of packages with a single invocation automating the process.
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -34,9 +34,12 @@ This article describes the steps from the current build tools used in the ROS ec
 ## Goal
 
 The goal of the build tool is to build a set of packages with a single invocation automating the process.
-It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
-It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
-This will allow the build tool to be utilized for non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
+It should work with:
+* ROS 1 packages
+* ROS 2 packages
+* other software sources when sufficient meta information is externally provided.
+
+The last point allows to build non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
 
 In the ROS ecosystems several tools already exist which support this use case (see below).
 Each of the existing tools performs similar tasks and duplicates a significant amount of the logic.
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -162,7 +162,7 @@ Their method of operating is being described in the following subsections togeth
 
 ### rosbuild
 
-Rosbuild was the first build system for ROS, it was based on Makefiles.
+Rosbuild was the first combination of build system and build tool for ROS, it was based on Makefiles.
 Rosbuild did not provide any install target, and kept sources and build results in the same folder structure.
 Rosbuild was replaced because it did not support packaging, cross-compiling, and because build speed was an issue.
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -39,8 +39,8 @@ This article describes the steps to unify the build systems of ROS1 and ROS2, bu
 
 The goal of the build tool is to build a set of packages with a single invocation automating the process.
 It should work with:
-* ROS 1 packages
-* ROS 2 packages
+* ROS 1 packages (inside a ROS1 workspace)
+* ROS 2 packages (inside a ROS2 workspace)
 * other software sources when sufficient meta information is externally provided.
 
 The last point allows to build non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -85,9 +85,8 @@ Examples are dpkg, rpm, homebrew, portage, robotpkg.
 | catkin_tools                                |              |      x     |
 | ament_tools                                 |              |      x     |
 
-### Environment Setup
 
-A very important part beside the actual build of a package is the environment setup.
+The following describes the essential functional requirements for a build tool, and further functional and non-functional requirements that are to be decided.
 For example, in order for a CMake project to discover a dependency using the CMake function `find_package`, the CMake module (e.g. `FindFoo.cmake`) or the CMake config file (e.g. `FooConfig.cmake`) for that dependency must either be in a prefix that CMake searches implicitly (e.g. `/usr`) or the location must be provided through the environment variable `CMAKE_PREFIX_PATH` / `CMAKE_MODULE_PATH`.
 
 In addition to building a package on top of another package (using `find_package` in the case of CMake), you may need to adjust the environment in order to run an executable from a package.
@@ -96,8 +95,33 @@ For example, when a package installs a shared library in a non-default location 
 The functionality to setup these environment variables can be provided by either the build tool or the build system.
 In the latter case the build tool only needs to know how the build system exposes the environment setup in order to reuse it.
 
-Considering the use case that a user might want to invoke the build system of each package manually it is beneficial if the build system already provides as much of the environment setup as possible.
-That avoids forcing the user to manually take care of the environment setup when not using a build tool.
+### Development Environment Setup
+
+Invoking a buildsystem for a package implies also setting up environment variables before the process.
+Examples are the `CMAKE_PREFIX_PATH` and the `LD_LIBRARY_PATH`.
+For consistency of the build result, those variables should be restricted, allowing the build to only access declared dependencies.
+
+For various reasons, it is beneficial to also allow developers to easily manually invoke the buildsystem for one package.
+This requires the buildsystem to provide the environment setup by itself.
+
+### Subtask invocation
+
+Build systems define various tasks, such as testing, compiling, generating documentation, installing.
+A build tool must provide it's own abstract set of tasks to be mapped to the tasks of different build systems.
+
+### Isolated installation
+
+For the ROS ecosystem it is important that packages can be installed to any target location in a filesystem.
+
+After a package has been built and installed to a target location, the environment might need to be extended to use the package.
+
+### Convenience features
+
+* Build Parallelity
+* Invocation from various locations, roscd
+* Workspace chaining
+* rosbuild compatibility (ROS_PACKAGE_PATH)
+
 
 ## Existing Build Systems
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -178,9 +178,10 @@ While each package can parallelize the build of its targets the packages are pro
 
 ### catkin_tools
 
-[catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build ROS 1 packages.
+[catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build catkin packages.
 It was developed after `catkin_make` / `catkin_make_isolated` to build packages in parallel as well as provide significant usability improvements.
 The tool supports building CMake packages and builds them in isolation as well as supports parallelizing the process across packages.
+
 
 ### ament_tools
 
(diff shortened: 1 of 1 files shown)


tkruse:
More things I notice missing in goals, maybe because they seem too obvious:
* Cross-compilation support
* Cross plattform compatibility
* cmake support as a constraint

articles/101_build_tool.md
```diff
@@ -21,21 +21,27 @@ Original Author: {{ page.author }}
...
-It consists of many separately developed and maintained packages.
+In the ROS ecosystem teams typically solve problems by programming in many software packages at the same time.
+This is in contrast to workflows where teams work on a single software package at a time only, and all dependencies are installed and updated by package managers.
+Such a workflow is impracticable without a tool that automatically rebuilds many packages in topological order of dependency.
+The rosbuild tool enabled this workflow first and was a critical factor in the success of ROS.
 
-The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
-For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
-From an efficiency point of view that manual process does not scale well.
+Today, the ROS1 ecosystem has the deprecated rosbuild and the catkin_make build systems for this purpose.
```
dirk-thomas: This sentence seem incomplete. Should there be something between "the ROS1 ecosystem has the deprecated rosbuild and" and "the catkin_make build systems for this purpose"?

articles/101_build_tool.md
```diff
@@ -21,21 +21,27 @@ Original Author: {{ page.author }}
...
+The ROS2 ecosystem has the ament_tools build system.
 
-This article describes the steps from the current build tools used in the ROS ecosystem to a single universal build tool which automates the process and works for various different use cases.
+For historical reasons those buildsystems are incompatible with each other, which causes diverse issues in the present.
+
+This article describes the steps from the current build tools used in the ROS ecosystem to a single universal build tool.
```
dirk-thomas: I think this comes from a misunderstanding of the difference of build tool and build system. The fact that ROS 1 and ROS 2 use different build systems is described in another article (http://design.ros2.org/articles/ament.html). It provides the rational why the CMake API as well as the feature set is different.

The "univeral" build tool described in this article will no change any of that. The goal of the universal build tool is solely to reduce duplication of efforts across the various existing build tools and aim to even support use cases beyond pure ROS applications.

articles/101_build_tool.md
```diff
@@ -21,21 +21,27 @@ Original Author: {{ page.author }}
...
+In the ROS ecosystem teams typically solve problems by programming in many software packages at the same time.
+This is in contrast to workflows where teams work on a single software package at a time only, and all dependencies are installed and updated by package managers.
+Such a workflow is impracticable without a tool that automatically rebuilds many packages in topological order of dependency.
+The rosbuild tool enabled this workflow first and was a critical factor in the success of ROS.
 
-The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
-For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
-From an efficiency point of view that manual process does not scale well.
+Today, the ROS1 ecosystem has the deprecated rosbuild and the catkin_make build systems for this purpose.
+The ROS2 ecosystem has the ament_tools build system.
```
dirk-thomas: `ament_tools` is a *build tool* - not a build system. Distinguishing these two is crucial for this article to make clear what the scope is.

articles/101_build_tool.md
```diff
@@ -21,21 +21,27 @@ Original Author: {{ page.author }}
...
 The goal of the build tool is to build a set of packages with a single invocation automating the process.
-It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
-It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
-This will allow the build tool to be utilized for non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
+It should work with:
+* ROS 1 packages
+* ROS 2 packages
+* other software sources when sufficient meta information is externally provided.
+
+The last point allows to build non-ROS dependencies of ROS packages (e.g. Gazebo including its dependencies).
```
dirk-thomas: To clarify: this does **not** imply that you can or can not mix ROS 1 and ROS 2 packages within the same workspace. Supporting that is only related to the build system.

articles/101_build_tool.md
```diff
@@ -48,43 +54,74 @@ In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
...
 
-## Build Tool vs. Build System
+## Build Tool vs. Build System vs. Package manager
+
+A build system schedules and executes atomic tasks for creating runnable software from sources.
+Task examples are test, compile, install.
+Examples are Gnu Make, cmake, python setuptools.
+
+A build tool schedules and invokes the build systems for separate source trees (packages) in topological order of dependency.
+Example are rosbuild, catkin_make, ament.
```
dirk-thomas: I would suggest using the name `ament_tools` here to not confuse this with `ament_cmake`.

articles/101_build_tool.md
```diff
@@ -48,43 +54,74 @@ In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
...
+
+The following describes the essential functional requirements for a build tool, and further functional and non-functional requirements that are to be decided.
+
+### Topological invocation of build systems
 
-A build tool operates on a set of packages.
-It determines the dependency graph and invokes the package specific build tool of each package in topological order.
-The build tool itself should know as little as possible about the build system used for a specific package.
-Just enough in order to know how to setup the environment for it, invoke the build, and setup the environment to use the built package.
+A build tool defines a manifest format for packages to follow to define the invocation of the respective build tool, and the dependencies to other packages.
```
dirk-thomas: I disagree that a build tool must be tighly coupled to a manifest format. A build tool needs the information which can e.g. be provided by a manifest file like the `package.xml` files in ROS. But that is not the only possible source. It should be possible to obtain the necessary information from any arbitrary source, e.g. an out-of-source configuration file (in whatever format you like), or like `rosdep` pull the information from a URL.

The tight coupling of the current build tools to the ROS manifest is one big reason why they can't be used in application outside the ROS ecosystem (e.g. Gazebo).

articles/101_build_tool.md
```diff
@@ -48,43 +54,74 @@ In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
...
+### Isolated installation
+
+For the ROS ecosystem it is important that packages can be installed to any target location in a filesystem.
+
+After a package has been built and installed to a target location, the environment might need to be extended to use the package.
 E.g. when a package installs a shared library in a non-default location the environment variable `LD_LIBRARY_PATH` (or `PATH` on Windows) needs to be extended to include the containing folder.
 
-The functionality to setup these environment variables can be provided by either the build tool or the build system.
-In the latter case the build tool only needs to know how the build system exposes the environment setup in order to reuse it.
+The build tool make it easy to extend the environment to run code from installed packages, or to build other packages on top of installed ones.
```
dirk-thomas: I think the edit looses an important point of the previous state: either the build tool or the build system can provide significant parts of this environment setup. And it describes the benefit if most of that is performed by the build system rather than the build tool.

articles/101_build_tool.md
```diff
@@ -99,37 +136,45 @@ In ROS 2 a package can be a "vanilla" Python package whereas in ROS 1 any Python
 Several different build tools are already being used in the ROS ecosystem.
 Their method of operating is being described in the following subsections together with their advantages as well as disadvantages.
 
+### rosbuild
+
+Rosbuild was the first build system for ROS, it was based on Makefiles.
```
dirk-thomas: `rosbuild` is both a build system as well as a build tool.

dirk-thomas (commented):
> The goal is not motivated yet. I hint at issues with the current situation, but without explaining those issues, it is difficult to make a decision solving them.

The goal is to consolidate the effort from the various build tools, avoid duplicate work and create a single tool which covers all the current use cases (and hopefully more).

> I noticed rosbuild was not mentioned much so far, nor tools relying on the ROS_PACKAGE_PATH. Not sure if that is important for the discussion

This is a good example why the unified build tool needs to be very modular. The `ROS_PACKAGE_PATH` is a ROS specific concept and therefore not relevant in non-ROS use cases. Therefore the core part of the unified build tool should not know about this. Only a ROS specific plugin should introduce the concept of that environment variable.

> it might be helpful to compare the "anatomy" of catkin and ament files, to see the problems in migrating from one to the other

Again, this is related to the build system and imo out of scope for this build tool article. From the build tools point of view it treats both ROS build system as separate and doesn't care how they operate (and if they are similar or how they are different).

> Devel_spaces not mentioned yet

The devel space is a build system concept (only in catkin). The core of the unified build tool therefore doesn't care about it. The catkin specific extension might or might not expose that feature - that is up to the implementation to decide.

> keeping catkin packages same vs. requiring migration

The tool only decides how the build systems of multiple packages are being invoked in the right order. There is conceptionally nothing to migrate.

> evolving catkin_tools vs. evolving ament_tools ('from scratch' is equivalent to one of those)

Trying to refactoring an existing code base of significant size to a more modular architecture is a big effort. Designing a modular architecture from scratch with the experience and knowledge about the existing systems and then porting over existing features is very different. Therefore I don't agree with your statement. Which one to choose is a different question :wink:

> Cross-compilation support

This boils down to being in the realm of the build system again. Consider the build tool only a fancy "command line invocation improver". It doesn't do anything for cross-compilation which the build system can't do.

> Cross plattform compatibility, cmake support as a constraint

The universal build tool needs to cover all use cases the current build tools already cover, that includes but is not limited to:

* Build catkin packages with Python 2 on Linux and OS X
* Build ament packages with Python 3 on Linux, OS X and Windows
* Support processing packages in parallel 
* Support packages without a manifest file (e.g. eProsima's FastRTPS)
* Provide the usability of catkin_tools with `config` and `profile`
* *many many more*

wjwwood:
> The devel space is a build system concept (only in catkin). The core of the unified build tool therefore doesn't care about it. The catkin specific extension might or might not expose that feature - that is up to the implementation to decide.

It doesn't have to be in the requirements or goals, but a proposal for what we should do and how that fails to address the devel space concept in catkin would be incomplete in my opinion.

> > Cross-compilation support
> This boils down to being in the realm of the build system again. Consider the build tool only a fancy "command line invocation improver". It doesn't do anything for cross-compilation which the build system can't do.

It still may need to be addressed in some way by the tool. Especially if the tool has to connect two different build systems differently in the context of a cross-compile. You can hand wave it away as "should be in a plugin", but again that can be absent from the goals and requirements, but probably ought to be addressed in a proposal, even if just to say "it should not require anything extra out of the tool because ...".

articles/101_build_tool.md
tkruse: No, the sentence should be complete. It is equivalent to: The ros1 ecosystem has 2 things for this purpose. 1. The deprecated rosbuild build system. 2. The catkin cmake build system.

articles/101_build_tool.md
tkruse: Clarifying it in the document would be good. I had no idea.

articles/101_build_tool.md
tkruse: Is it a goal *not* to depend on any fixed manifest format? Then that would belong in the goals section (or somewhere else in the document, anyway)

articles/101_build_tool.md
tkruse: So it would be a "nice-to-have" feature to extract the responsibility for the environment from the build-tool? If so it should be kept, could be marked more explicitly as a possible goal.
I removed it because it looked like a technical remark, not something related to a goal.

tkruse:
Regarding all buildsystem-specific things: I believe they should be mentioned in the document exactly to make other people know why the goal is not simply to create a unified build-tool instead. The reason might be obvious to you, but others will not see this easily.

I am not sure if I should try editing according to the comments, or if this was too far off the mark anyway. Please advise. Close this PR without further comments if it served it's purpose in your eyes.

dirk-thomas:
> I am not sure if I should try editing according to the comments, or if this was too far off the mark anyway. Please advise

I will leave this up to you. If you would like to update the PR with the goal to get it merged as a whole that is fine by me. Otherwise I can go through the patch and cherry pick the changes which are good as is into the main PR.

tkruse:
I might find time on this weekend.

tkruse:
rebased, added commit, squashed one commit for restoring original text

articles/101_build_tool.md
tkruse: rephrased

articles/101_build_tool.md
tkruse: restored original text

articles/101_build_tool.md
tkruse: removed my change, original text changed changed upstream

articles/101_build_tool.md
tkruse: changed

tkruse:
Added more commits, mostly tidying up the requirements section

articles/101_build_tool.md
```diff
@@ -21,19 +21,29 @@ Original Author: {{ page.author }}
...
 
-ROS is developed with a federated model.
-It consists of many separately developed and maintained packages.
+In the ROS ecosystem teams typically solve problems by programming in many software packages at the same time.
+This is in contrast to workflows where teams work on a single software package at a time only, and all dependencies are installed and updated by package managers.
+Such a workflow is impracticable without a tool that automatically rebuilds many packages in topological order of dependency.
+The rosbuild tool enabled this workflow first and was a critical factor in the success of ROS.
 
-The "manual" approach to build a set of packages consists of building all packages in their topological order one by one.
-For each package the documentation usually describes what the dependencies are, how to setup the environment to build the package as well as how to setup the environment afterwards to use the package.
```
dirk-thomas, line 28: This important information what is involved is lost in the modification.

articles/101_build_tool.md
```diff
@@ -21,19 +21,29 @@ Original Author: {{ page.author }}
...
-From an efficiency point of view that manual process does not scale well.
+The ROS1 ecosystem has catkin (and the deprecated rosbuild) for this purpose.
+The ROS2 ecosystem has ament.
+
+Both catkin and ament toolsets have 2 conceptually separate parts, a build tool part that deals with a single package at a time, and a build system part that deals with invoking build tools in the right order.
```
dirk-thomas, line 32: > a build system part that deals with invoking build tools

This seems to be swapped.

articles/101_build_tool.md
```diff
@@ -55,16 +65,19 @@ In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
 The build tool also does not create binary packages (e.g. a Debian package).
 In the ROS ecosystem [bloom](http://wiki.ros.org/bloom) is used to generate the required metadata and then platform dependent tools like `dpkg-buildpackage` build binary packages.
 
-## Build Tool vs. Build System
+## Build Tool vs. Build System vs. Package manager
```
dirk-thomas, line 68: Since this article isn't about package managers I don't see why this should be added here. Maybe move it to the "Out of scope" section instead.

articles/101_build_tool.md
```diff
@@ -55,16 +65,19 @@ In the ROS ecosystem [rosdep](http://wiki.ros.org/rosdep) can be used for this.
...
-A build tool operates on a set of packages.
+A build system schedules and executes atomic tasks for creating runnable software from sources.
 It determines the dependency graph and invokes the specific build tool for each package in topological order.
-The build tool itself should know as little as possible about the build system used for a specific package.
-Just enough in order to know how to setup the environment for it, invoke the build, and setup the environment to use the built package.
+Examples are Gnu Make, cmake, python setuptools.
 
-The build system on the other hand operates on a single package.
-In the case a package uses e.g. `CMake` the responsibility of the tool comes down to invoke the common steps `cmake`, `make`, `make install` for this package.
-As another example for a package using `Autotools` the steps could look like `configure`, `make`, `make install`.
```
dirk-thomas, line 67: The example steps for invoking these build systems is lost in the modification.

articles/101_build_tool.md
```diff
@@ -119,30 +122,39 @@ In ROS 2 a package can be a "vanilla" Python package whereas in ROS 1 any Python
...
 ### catkin_make
 
-`catkin_make` is provided by the ROS package `catkin` which contains the build system for ROS 1.
-It has been designed as the successor of `rosbuild` for ROS Fuerte.
+`catkin_make` replaced rosbuild for ROS1, it is provided by the ROS package `catkin`.
+It has been designed as the successor of `rosbuild` since ROS Fuerte.
 
 The tool invokes CMake only a single time and uses CMake's `add_subdirectory` function to process all packages in a single context.
+
+The tool uses CMake's `include` function to process all packages in a single cmake context.
```
dirk-thomas, line 138: This sentence directly contradicts the previous one.

articles/101_build_tool.md
```diff
@@ -119,30 +122,39 @@ In ROS 2 a package can be a "vanilla" Python package whereas in ROS 1 any Python
...
+`catkin_make_isolated` is included with `catkin`.
 It was developed after `catkin_make` to address the problems involved with building multiple packages in a single CMake context.
 
 The tool only supports CMake-based packages and builds each package in topological order using the command sequence common for CMake packages: `cmake`, `make`, `make install`.
 While each package can parallelize the build of its targets the packages are processed sequentially even if they are not (recursive) dependencies of each other.
 
 ### catkin_tools
 
-[catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build ROS 1 packages.
+[catkin_tools](https://catkin-tools.readthedocs.io/) is provided by a standalone Python package used to build catkin packages.
```
dirk-thomas, line 153: I think the statement that it builds ROS 1 packages is more accurate since it also supports plain CMake packages.

articles/101_build_tool.md
```diff
@@ -183,32 +195,50 @@ After cloning the repositories containing Gazebo and all its dependencies (exclu
...
 
-The build tool will support using different build systems within a single workspace.
-If these packages inter-operate with each other correctly depends also to a large degree on the build system.
-The build tool should ensure that it doesn't prevent that use case.
+The build tool will support using different build systems within a single workspace, as ament_tools or catkin_tools already do.
+But there may be restrictions on which build systems can be combined in which kind of workspace (e.g. building catkin packages may not be possible in an ament workspace).
+Removing such restrictions will be an ongoing process.
 
-### Software Criteria
+#### Build Environment Setup
```
dirk-thomas, line 208: This is not an individual feature but a strict requirement in order to build package automatically on-top of each other. Therefore I don't think it should be moved that far down.

dirk-thomas (commented):
In general please use the same spelling / case for names like "ROS 1", "ROS 2", "CMake", Python, etc.

I am not sure how to proceed with this PR. Every iteration other things are added / changed / moved which makes it very difficult to iterate on this. I am open for suggestions how this should be reviewed and merged (either as a whole or in smaller pieces).

tkruse:
Since I suggest large changes, but those are mostly about style, I would be okay with closing this PR and you cherry-pick whatever you like (which might be nothing). This is quicker than talking back and forth.

dirk-thomas:
I tried to integrate some of the changes and feedback from this. Please see https://github.com/ros2/design/pull/115/commits/a92be6ee655105b1c5b02e122f90ebab3d69bf34 I will close this PR. Please make further suggestions and contributions in a separate PR.

PR #120, opened 2017-03-14
Update 101_build_tool.md

wjwwood:
See: https://github.com/ros2/design/pull/115#discussion_r105791922


dirk-thomas:
I added some minor changes. Thank you for the improvements.

PR #121, opened 2017-03-18
Layered diagrams

tkruse:
In the spirit of a picture saying more than a thousand words, this table art might help.

Quick link to rendered result:
https://github.com/tkruse/design/blob/build_tool2/articles/101_build_tool.md

Because there is confusion about homogenous workspaces as a goal, I split up goal into 2 phases. Labelling could of course be different, such as humble/ambitious, or primary/secondary, ... I am dispassionate about that labelling.

### DIFF
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -65,6 +65,14 @@ Additionally this makes new features developed for one ROS version / build syste
 
 This phase seems realistic to achieve in predictable time because both catkin_tools and ament_tools already have a large overlap.
 
+A coarse pseudo-code looks like this
+
+1. Detect workspace folder to use (e.g. current folder, or parent folder with marker file, ...)
+2. Load workspace adapter for this workspaces buildsystem
+3. Ask plugin/wrapper for package declarations (name, folder and dependencies of each package)
+4. Create DAG of packages for building the workspace
+5. Invoke buildsystem plugin command for each package in build order
+
 ### Phase 2 goal: unify workspaces
 
 | Concept Layer         | All ROS / non-ROS                         |
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -42,6 +42,42 @@ As a consequence of being developed separately certain features are only availab
 The reason to work on a single universal build tool comes down to reducing the effort necessary for development and maintenance.
 Additionally this makes new features developed for one ROS version / build system available to the other supported ROS versions / build systems.
 
+### Current Workspace and tool combinations
+
+| Concept Layer | ROS2                               | ROS1 (catkin)                  | ROS1 (rosbuild)         |
+|--------------|------------------------------------|--------------------------------|-------------------------|
+| Workspace    | ROS2 ament workspace               | ROS1 catkin workspace          | ROS1 rosbuild workspace |
+| Build tool   | ament_tools                        | catkin_tools, catkin_make, cmi | rosbuild                |
+| Build system | ament_cmake, cmake, python         | catkin, cmake                  | rosbuild                |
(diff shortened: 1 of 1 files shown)


articles/101_build_tool.md
```diff
@@ -42,6 +42,50 @@ As a consequence of being developed separately certain features are only availab
...
 Additionally this makes new features developed for one ROS version / build system available to the other supported ROS versions / build systems.
 
+### Current Workspace and tool combinations
+
+| Concept Layer | ROS2                               | ROS1 (catkin)                  | ROS1 (rosbuild)         |
+|--------------|------------------------------------|--------------------------------|-------------------------|
+| Workspace    | ROS2 ament workspace               | ROS1 catkin workspace          | ROS1 rosbuild workspace |
+| Build tool   | ament_tools                        | catkin_tools, catkin_make, cmi | rosbuild                |
+| Build system | ament_cmake, cmake, python         | catkin, cmake                  | rosbuild                |
+| Package      | ROS2 package, cmake/python project | ROS1 package, cmake project    | ros1 rosbuild package   |
```
dirk-thomas, line 52: The visual presentation is certainly helpful. But I think the information is presented too early in the document. Especially since at that point the terms of build tools and build systems haven't been clearly defined. And because the table mixes the terminology and doesn't distinguish between those which imo confused the reader.

articles/101_build_tool.md
```diff
@@ -42,6 +42,50 @@ As a consequence of being developed separately certain features are only availab
...
+
+(Note catkin_make cannot build pure cmake projects)
+
+### Phase 1 goal: unify build systems (not rosbuild)
```
dirk-thomas, line 56: If someone would like to implement an extension to the unified build tool to support rosbuild that would be absolutely feasible. There this shouldn't be excluded like this.

articles/101_build_tool.md
```diff
@@ -42,6 +42,50 @@ As a consequence of being developed separately certain features are only availab
...
+
+A coarse pseudo-code looks like this
+
+1. Detect workspace folder to use (e.g. current folder, or parent folder with marker file, ...)
+2. Load workspace adapter for this workspaces buildsystem
+3. Ask plugin/wrapper for package declarations (name, folder and dependencies of each package)
+4. Create DAG of packages for building the workspace
+5. Invoke buildsystem plugin command for each package in build order
+
+### Phase 2 goal: unify workspaces
```
dirk-thomas, line 76: I don't think this article to outline such a phase. The current draft briefly mentions the intrinsic complexity of mixing workspaces like that which are beyond the build tools control: see https://github.com/ros2/design/pull/115/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR186 Also even if all build tools and build systems do the technical part of mixing them well there are other aspects like name spacing of ROS 1 and ROS 2 packages, their rosdep keys, etc. Therefore I think this is exceeding the scope of this article.

articles/101_build_tool.md
tkruse: Feel free to move to another section or suggest one

articles/101_build_tool.md
tkruse: In the comments of the main PR @wjwwood seems to argue for such a goal https://github.com/ros2/design/pull/115#issuecomment-287242822, that's why I added it.

articles/101_build_tool.md
tkruse: would "rosbuild not shown" be ok for you?

dirk-thomas:
I added some information from this in https://github.com/ros2/design/pull/115/commits/7ae99afe94d43701af314a5fb5ab5cbd6fefcb13 Please comment / open another PR for additional changes.

PR #122, opened 2017-03-18
table comparison

tkruse:
again, to make it easier for readers, a table instead of text. This should help readers understand the decision and the effort involved.

Once the table is complete (not just marks, but also rows), an agile estimation of effort could be done to have a little more objective prediction of efforts.


dirk-thomas:
Currently the table is completely empty so this would need to be updated before it can be merged. I also don't think that this level of details is necessary for the goal of the article.

PR #123, opened 2017-04-03
Correct examples for valid rostopics with underscores.

miguelprada:
The examples in the _Topic and Service name mapping to DDS_ article imply that trailing underscores are not allowed in ROS name tokens (i.e. `foo/_bar` is valid while `foo_/bar` is not).

### DIFF
--- a/articles/140_topic_and_service_name_mapping.md
+++ b/articles/140_topic_and_service_name_mapping.md
@@ -67,8 +67,6 @@ For convenience here is a summary of all rules for topic and service names in RO
 - must not end with a forward slash (`/`)
 - must not contain any number of repeated forward slashes (`/`)
 - must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`
-- must not end with an underscore (`_`)
-- must not contain an uderscore (`_`) preceding a forward slash (`/`)
 - must have balanced curly braces (`{}`) when used, i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`
 
 The content of substitutions, i.e. the string in side of balanced curly braces (`{}`), follow very similar rules to names.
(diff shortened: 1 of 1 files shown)


wjwwood:
@miguelprada thanks for the pr, but I think we need to opposite, i.e. we need to update the examples. We dropped this restriction when we changed how we implemented this in DDS.

@Karsten1987 does that sound right to you too?

miguelprada:
I was too fast to assume which side of the _discrepancy_ was right then :wink: 

I'll leave you guys to close the PR one @Karsten1987 confirms.

Karsten1987:
The argument whether trailing or leading underscores are allowed may be based on this ROS1 comparison: http://design.ros2.org/articles/topic_and_service_names.html#compare-and-contrast-with-ros-1

```
may be indicated as “hidden” by using a leading underscore (_) in one of the namespaces
This is used to hide common but infrequently introspected topics and services.
```

Within this refactoring or the design page, we didn't take that explicitly as a rule for now. 
@miguelprada In any case, I do believe it does make sense to have a statement about the underscores. So thanks for that. 

IMO, I would stick to the ROS1 rational of declaring _leading_ underscores for ROS Topics as reserved for internal purposes.

Karsten1987:
So after a quick discussion with @wjwwood we came to the conclusion that the rules and contraints for topics are fine as is. Underscores are allowed and don't make a topic invalid. 
@miguelprada You are correct that there is a mismatch in the given examples. Would you mind updating your PR for correcting these examples? Specifically, I mean removing `foo_/bar` from the invalid section and add it as well as `foo_` to the valid topics.
Does that make sense?

miguelprada:
Done, @Karsten1987.

Let me know if you want me to squash.

Karsten1987:
that looks good to me 👍 
I'll squash and merge this here via github.

I hope you don't mind that I changed the title or this PR to make it more expressive on what was changed. 
Thanks a lot for iterating on this.

PR #127, opened 2017-04-26
simplify and clarify rules for topic substitutions

wjwwood:
I would like to simplify the way topic name substitutions are parsed in topic names. This proposed change would:

- evaluate substitutions after `~` whereas they were previously said to happen before
- prevent recursive substitutions allowing for a single pass when expanding substitutions
  - e.g. prevent something like `{sub}` -> `foo/{sub2}` -> `foo/bar`
- prevent nested substitutions, i.e. substitutions within other substitution names
  - e.g. prevent something like `{{foo}_bar}` -> `{foo_bar}` -> `foo/bar`

I don't think these features are necessary and in fact will help rein in abuse of this feature. It also happens to greatly simplify the implementation which I'm currently working on.

If anyone sees a problem with this, please speak up.

Also if you know more precise terminology to use here, please suggest a change or let me know.

### DIFF
--- a/articles/140_topic_and_service_name_mapping.md
+++ b/articles/140_topic_and_service_name_mapping.md
@@ -182,9 +182,16 @@ Note the private namespace substitution character makes the name absolute, and t
 The bracket syntax (`{substitution_name}`) may be used in non-fully qualified names to substitute useful contextual information into the name.
 The set of substitution keys (names) are not set in this document, but some reasonable examples might be: `{node}` expands to the current node's name or `{ns}` expands to the current node's namespace.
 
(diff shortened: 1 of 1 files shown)


wjwwood:
I just fixed the rendering issues. I'm still looking for feedback on this. I've already implemented what is described here in https://github.com/ros2/rcl/pull/132. I'd really appreciate some people reviewing it.

dirk-thomas (approved):
Looks good to me.

Better start narrow and make it more powerful / complex in the future.

mikaelarguedas (approved):
this reads clearly to me :+1: . I suggested changes for what appeared to be typos in https://github.com/ros2/design/pull/129

PR #128, opened 2017-04-27
enhanced time article to include realtime factor

BrannonKing:
This pull request modifies the ROS2 time design to include a clock message with a realtime factor. It also further defines "ROS time". At the bottom of the document are some open questions. At ASI we presently uses our own time source. I find this painful; I don't want to have to modify each of the standard nodes to include this. I have some time that I can address this issue in the ROS2 core libraries. Hence, please make an effort to resolve the questions in the design soon so that I can put in pull requests to rcl and rclcpp that everyone is happy with. Tagging @tfoote @wjwwood @clalancette and @mikaelarguedas .

### DIFF
--- a/articles/130_ros_time.md
+++ b/articles/130_ros_time.md
@@ -47,14 +47,19 @@ Additionally if the simulation is paused the system can also pause using the sam
 
 A third situation arises with hardware that doesn't have a system clock chip available. 
 Nodes running there need to be notified of the current time so that they can utilize that information to time-stamp their output. 
-(We will assume that nodes running on this type of hardware have accurate frequency counters available that can assist in computing the current time when they have some base time to compound.)
+We will assume that nodes running on this type of hardware have accurate frequency counters available that can assist in computing the current time when they have some base time to compound.
+However, we strongly recommend against broadcasting time over multiple machines.
+Clock synchronization is not the intent of the ROS clock message; it is outside the scope of ROS.
+This is not a viable use case because of network latency.
+Instead, use an existing mechanism to synchronize clocks between systems.
+Even if your mixed system has node-hosting parts with no real-time clock chip, those can still host fake a clock that is synchronized to some other real clock.
+Use quality time synchronizers like chrony or TICSync.
 
 ### Challenges in using abstracted time
 
 There are many algorithms for synchronization and they can typically achieve accuracies which are better than the latency of the network communications between devices on the network.
 However, these algorithms take advantage of assumptions about the constant and continuous nature of time.
-
-Again, this is not the primary intent of synchronizing time between ROS nodes. 
+Again, this is not the primary intent of publishing time between ROS nodes. 
 We need to ensure the ability to manipulate time.
 In some cases, speeding up, slowing down, or pausing time entirely is important for debugging.
 And the ability to support pausing time requires that we not assume that the time values are always increasing.
@@ -63,7 +68,7 @@ When communicating the changes in time propagation, the latencies in the communi
 Any change in the time abstraction must be communicated to the other nodes in the graph, but will be subject to normal network communication latency.
 This inaccuracy is proportional to the latency of communications and also proportional to the increase in the real time factor.
 If very accurate timestamping is required when using the time abstraction, it can be achieved by slowing down the real time factor such that the communication latency is comparatively small.
-Of course we recommend using synchronized system clock sources for live work on nodes spread across multiple devices.
+We recommend using synchronized system clock sources for live work on nodes spread across multiple devices.
 
 The final challenge is that the time abstraction must be able to jump backwards in time, a feature that is useful for log file playback.
 This behavior is similar to a system clock after a negative date change, and requires developers using the time abstraction to make sure their algorithm can deal with the discontinuity.
--- a/articles/130_ros_time.md
+++ b/articles/130_ros_time.md
@@ -24,13 +24,19 @@ Original Author: {{ page.author }}
 Many robotics algorithms inherently rely on timing as well as synchronization.
 To this end we require that nodes running in the ROS network have a synchronized clock such that they can accurately report timestamps for events.
 
-There are, however, several other use cases where being able to control the progress of the system is important. We review those next.
+There are, however, several other use cases where being able to control the progress of the system is important; we review those next.
 
 ## Use cases requiring time abstraction
 
+What are the situations where `std::chrono` or other system time mechanisms are insufficient?
+
 When playing back logged data it is often very valuable to support accelerated, slowed, or stepped control over the progress of time.
 This control can allow you to get to a specific time and pause the system so that you can debug it in depth.
-It is possible to do this with a log of the sensor data. However, if the sensor data is out of synchronization with the rest of the system, it will break many algorithms.
+Data playback does not just affect a single node.
+All nodes involved in the playback would need the same clock source.
+It is possible to do this with a log of the sensor data. 
+However, if the sensor data is out of synchronization with the rest of the system, it will break many algorithms.
+For example, the transform system (tf) timestamps messages that contain data computed between two different time/data combos.
 
 Another important use case for using an abstracted time source is when you are running logged data against a simulated robot instead of a real robot.
 Depending on the simulation characteristics, the simulator may be able to run much faster than real time or it may need to run much slower.
@@ -39,18 +45,21 @@ Slower than real time simulation is necessary for complicated systems where accu
 Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
 Additionally if the simulation is paused the system can also pause using the same mechanism.
 
-A third situation arises with hardware that doesn't have a system clock chip available. Nodes running there need to be notified of the current time so that they can utilize that information to time-stamp their output. We will assume that nodes running on this type of hardware have accurate timers available.
+A third situation arises with hardware that doesn't have a system clock chip available. 
+Nodes running there need to be notified of the current time so that they can utilize that information to time-stamp their output. 
+(We will assume that nodes running on this type of hardware have accurate frequency counters available that can assist in computing the current time when they have some base time to compound.)
 
 ### Challenges in using abstracted time
 
 There are many algorithms for synchronization and they can typically achieve accuracies which are better than the latency of the network communications between devices on the network.
 However, these algorithms take advantage of assumptions about the constant and continuous nature of time.
 
-Again, this is not the primary intent of synchronizing time between ROS nodes. We need to ensure the ability to manipulate time.
+Again, this is not the primary intent of synchronizing time between ROS nodes. 
+We need to ensure the ability to manipulate time.
 In some cases, speeding up, slowing down, or pausing time entirely is important for debugging.
 And the ability to support pausing time requires that we not assume that the time values are always increasing.
 
-When communicating the changes in time propagation, the latencies in the communication network becomes a challenge.
+When communicating the changes in time propagation, the latencies in the communication network become a challenge.
 Any change in the time abstraction must be communicated to the other nodes in the graph, but will be subject to normal network communication latency.
 This inaccuracy is proportional to the latency of communications and also proportional to the increase in the real time factor.
 If very accurate timestamping is required when using the time abstraction, it can be achieved by slowing down the real time factor such that the communication latency is comparatively small.
@@ -58,47 +67,57 @@ Of course we recommend using synchronized system clock sources for live work on 
 
 The final challenge is that the time abstraction must be able to jump backwards in time, a feature that is useful for log file playback.
 This behavior is similar to a system clock after a negative date change, and requires developers using the time abstraction to make sure their algorithm can deal with the discontinuity.
-Appropriate APIs must be provided for to the developer API to enable notifications of jumps in time, both forward and backwards.
+Appropriate APIs must be provided for to the developer API to enable or handle notifications of jumps in time, both forward and backwards.
 
 ### Time Abstractions
 
 There will be at least three versions of these abstractions with the following types, `ROSTime`, `SystemTime`, and `SteadyTime`.
+All abstractions provide/utilize the same API.
 The latter two choices are designed to parallel the [`std::chrono`][] [`system_clock`][] and [`steady_clock`][].
 It is expected that the default choice of time will be to use the `ROSTime` source, however the parallel implementations supporting `steady_clock` and `system_clock` will be maintained for use cases where the alternate time source is required.
 
+#### System Time
+
+System Time pulls the current time from the system clock chip. 
+It provides benefit over Steady Time when all nodes in the system have synchronized system time available or when all nodes run on the same host that contains a system clock chip.
+
+We expect a System Time publisher to throw an error if the system clock resource is not available.
+We expect methods returning the current time to fall back to Steady Time if System Time is not available.
+
 #### Steady Time
 
-Steady Time implies a realtime incrementing source. There is no modern hardware that does not support this option. Steady Time typically begins at system bootup and measures increments using hardware frequency counters.
+Steady Time implies a real-time incrementing source.
+There is no modern hardware or operating system that does not support this option.
+Steady Time typically begins at system bootup and measures increments using hardware frequency counters.
 
 Example use cases for this include hardware drivers which are interacting with peripherals with hardware timeouts and systems lacking a realtime clock chip.
 
-In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
-
-#### System Time
-
-System Time pulls the current time from the system clock chip. It provides benefit over Steady Time when all nodes in the system have synchronized system time available.
-
-For convenience in these cases we will also provide the same API as above, but use the name `SystemTime`.
-
-We expect a System Time publisher to throw an error if the system clock resource is not available.
-
+In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals, it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
 
 #### ROS Time
 
 ROS Time derives the current time from the most-recently received clock broadcast message published on the `/clock` topic.
 
-When the ROS time source is active `ROSTime` will return the latest value reported by the Time Source plus the elapsed time since that report multiplied by the real time factor.
+When the ROS Time source is active `ROSTime` will return the latest value reported by the Time Source plus the elapsed time since that report multiplied by the real time factor.
 
 The ROS Time broadcast (clock message) needs to include this information:
-- The current time in nanoseconds
-- The current time multiplier (also called "factor", typically one, but could zero for paused or negative for going backwards)
-- If possible, the system clock time (in nanoseconds) when the message was published
+- The current time (in nanoseconds, epoch depends upon source and should not be needed)
+- The current real time factor (also called the "multiplier", typically a value of one, but could be zero for paused or negative when going backwards).
+- The system clock time (in nanoseconds since the Unix epoch) when the message was created (or -1 if not available)
+
+This message cannot be latched; to use the real time factor you will need to grab a frequency count (aka, start a stopwatch) when the message is received.
+Stale data would make this mechanism fail.
+An exception to this is when all nodes have a system clock available and the message contains the system clock stamp.
+This sytem timestamp could be used to determine the staleness of the "current time" in the message.
+Nodes that publish data before they receive the current time would also be problematic.
+It will be helpful to have the clock publisher publish a new message whenever there is a new subscription added to reduce this period of unknown time.
 
 ### ROS Time Message Sources
 
 - An included (where?) Steady Clock Publisher.
 - An included (where?) System Clock Publisher.
 - An included (where?) Gazebo Plugin.
+- The ROSBag playback engine.
 
 #### Default Time Source
 
@@ -112,16 +131,19 @@ Users can also add callbacks to `Node::time_reversing` and `Node::time_jumping_b
 The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
 If the time on the clock jumps backwards, a callback handler will be invoked and be required to complete before any calls to the ROS time abstraction report the new time.
 
-The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific. Assuming they latch one message, they shouldn't need to publish more often than the multiplier changes. If the realtime multiplier cannot be computed, the message will need to be published very frequently. 
+The frequency of publishing the `/clock` as well as the granularity are not specified as they are application-specific. 
+If the realtime multiplier cannot be computed, the message will need to be published very frequently. 
 
 #### Custom Time Source
 
 It is possible that the user may have access to an out-of-band-time source which can provide better performance than the default source the `/clock` topic.
-It might be possible that for their use case a more advanced algorithm would be needed to propagate the simulated time with adequate precision or latency with restricted bandwidth or connectivity.
-The user will be able to switch out the time source for either each instance of their Time object as well as have the ability to override the default for the process.
+It might be possible that, for their particular use case, a more advanced algorithm is needed to propagate the simulated time with adequate precision for latency arising with restricted bandwidth or connectivity.
+The user will be able to switch out the time source for each instance of their Time object as well as have the ability to override the default for the process or node.
 This can be acheived in C++ using the template on the Time object. The RCL package also contains a method to disable `ROSTime`.
 
-It is possible to use an external time source such as GPS as a ROSTime source. When writing this publisher, ensure that you handle jumps in time (both forward and back) and set the realtime multipleir correctly in those situations.
+It is possible to use an external time source such as GPS as a ROS Time source. 
+When writing this publisher, ensure that you handle jumps in time (both forward and back) and set the real time factor correctly in those situations.
+The preferred option is to use the GPS as an TICSync (or IEEE 1588 PTP or, worst case, NTP) time source instead of a ROS Time source; synchronize your system clocks to the GPS unit's atomic clock.
 
 The `SystemTime`, `SteadyTime`, and `ROSTime` API's will be provided by each client library in an idiomatic way, but they may share a common implementation, e.g. provided by `rcl`.
 However, if a client library chooses to not use the shared implementation then it must implement the functionality itself.
@@ -130,7 +152,7 @@ However, if a client library chooses to not use the shared implementation then i
 
 In each implementation will provide `Time`, `Duration`, and `Rate` datatypes, for all three time source abstraction..
 The `Duration` will support a `sleep_for` function as well as a `sleep_until` method.
-The implementation will also provide a `Timer` object which will provide periodic callback functionality for all the abstractions. It will be microsecond-accurate.
+The implementation will also provide a `Timer` object which will provide periodic callback functionality for all the abstractions.
 
 #### RCL implementation
 
@@ -149,9 +171,23 @@ The underlying datatypes will also provide ways to register notifications, howev
 - Decide where and when to autosubscribe to the `/clock` message. Built in to Node? A Node addon or service? Part of the first call to `rcl_node_init`?
 - Decide how to deal with multiple nodes in a single process that all update the one static time (or decide to ditch the static time methods).
 - Decide where to put the out-of-the-box clock publishers.
+- Decide how to deal with nodes publishing data before they receive the current `ROSTime` update.
 - Ensure that there is no reason we can't enable `ROSTime` (with fallbacks) by default.
-- Change the RCL time implementation to call a method on the time source to return the current time (rather than using a field). This is necessary to compute the current time from the realtime factor. Fix the `rcl_get_ros_time` method to use the newly-added current time method.
-- Add an `rcl_time_source_t` (to the RCL project) for `ROSTime`.
+- Change the RCL time implementation (in `rcl_get_ros_time`) to call a method on the time source to return the current time (rather than using a field). This is necessary to compute the current time from the real time factor. 
+- Add an `rcl_time_source_t` (to the RCL project) for `ROSTime` that includes a method to compute the current time.
+
+
+If we decide to opt-in to ROS Time, you might use an interface like this in C++:
+```cpp
+class ROSTimeManager {
+  public:
+    ROSTimeManager(Node::shared_ptr node, bool overrideStaticTimeNow = true, bool allowFallbackToSystemClock = true, milliseconds initialBlockWaitingForClockMsg = 50ms);
+    Time now()
+    void register_clock_message_received_handler(std::function ...)
+    void register_time_about_to_reverse_handler(std::function ...)
+    void register_time_about_to_jump_backwards_handler(std::function ...)
+    ...
+```
 
 ## References
 
--- a/articles/130_ros_time.md
+++ b/articles/130_ros_time.md
@@ -22,15 +22,15 @@ Original Author: {{ page.author }}
 ## Background
 
 Many robotics algorithms inherently rely on timing as well as synchronization.
-To this end we require that nodes running in the ROS network have a synchronized system clock such that they can accurately report timestamps for events.
+To this end we require that nodes running in the ROS network have a synchronized clock such that they can accurately report timestamps for events.
 
-There are however several use cases where being able to control the progress of the system is important.
+There are, however, several other use cases where being able to control the progress of the system is important. We review those next.
 
 ## Use cases requiring time abstraction
 
 When playing back logged data it is often very valuable to support accelerated, slowed, or stepped control over the progress of time.
 This control can allow you to get to a specific time and pause the system so that you can debug it in depth.
-It is possible to do this with a log of the sensor data, however if the sensor data is out of synchronization with the rest of the system it will break many algorithms.
+It is possible to do this with a log of the sensor data. However, if the sensor data is out of synchronization with the rest of the system, it will break many algorithms.
 
 Another important use case for using an abstracted time source is when you are running logged data against a simulated robot instead of a real robot.
 Depending on the simulation characteristics, the simulator may be able to run much faster than real time or it may need to run much slower.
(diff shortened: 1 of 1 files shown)


articles/130_ros_time.md
```diff
@@ -22,15 +22,15 @@ Original Author: {{ page.author }}
 ## Background
 
 Many robotics algorithms inherently rely on timing as well as synchronization.
-To this end we require that nodes running in the ROS network have a synchronized system clock such that they can accurately report timestamps for events.
+To this end we require that nodes running in the ROS network have a synchronized clock such that they can accurately report timestamps for events.
 
-There are however several use cases where being able to control the progress of the system is important.
+There are, however, several other use cases where being able to control the progress of the system is important. We review those next.
```
wjwwood: Nitpick: please put one sentence per line

See: https://github.com/ros2/ros2/wiki/Developer-Guide#markdown

articles/130_ros_time.md
```diff
@@ -39,85 +39,89 @@ Slower than real time simulation is necessary for complicated systems where accu
 Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
 Additionally if the simulation is paused the system can also pause using the same mechanism.
 
+A third situation arises with hardware that doesn't have a system clock chip available. Nodes running there need to be notified of the current time so that they can utilize that information to time-stamp their output. We will assume that nodes running on this type of hardware have accurate timers available.
```
wjwwood: I think you should clarify here that this would use a different mechanism from what ROS Time uses traditionally (i.e. the `/clock` topic), but instead is generally a rationale for why we have a Time abstraction rather than directly using `std::chrono`, for example.

articles/130_ros_time.md
```diff
@@ -39,85 +39,89 @@ Slower than real time simulation is necessary for complicated systems where accu
...
 ### Time Abstractions
 
-There will be at least three versions of these abstractions with the following types, `SystemTime`, `SteadyTime` and `ROSTime`.
-These choices are designed to parallel the [`std::chrono`][] [`system_clock`][] and [`steady_clock`][].
+There will be at least three versions of these abstractions with the following types, `ROSTime`, `SystemTime`, and `SteadyTime`.
+The latter two choices are designed to parallel the [`std::chrono`][] [`system_clock`][] and [`steady_clock`][].
 It is expected that the default choice of time will be to use the `ROSTime` source, however the parallel implementations supporting `steady_clock` and `system_clock` will be maintained for use cases where the alternate time source is required.
 
-#### System Time
+#### Steady Time
```
wjwwood: Why reorder these? It now no longer matches the order in which they are mentioned above.

wjwwood: If anything ROSTime would be moved ahead of these, but the order changed here would have been left alone.

There was also a rationale for introducing them in this order:

- System time is what people are used to using
- Steady time is an alternative that's less commonly used
- ROS Time is a weird and exotic new concept, which builds on both system time and steady time

wjwwood: The reorganization also makes it hard to read this diff. If you still think the reorganization of the text makes sense, please try to move the content in one commit and change it in another.

articles/130_ros_time.md
```diff
@@ -39,85 +39,89 @@ Slower than real time simulation is necessary for complicated systems where accu
...
 It might be possible that for their use case a more advanced algorithm would be needed to propagate the simulated time with adequate precision or latency with restricted bandwidth or connectivity.
 The user will be able to switch out the time source for either each instance of their Time object as well as have the ability to override the default for the process.
+This can be acheived in C++ using the template on the Time object. The RCL package also contains a method to disable `ROSTime`.
 
-It is possible to use an external time source such as GPS in as a ROSTime source, but it is recommended to integrate a time source like that using standard ntp integrations with the system clock since that is already an established mechanism and will not need to deal with more complicated changes such as time jumps.
-
-## Implementation
+It is possible to use an external time source such as GPS as a ROSTime source. When writing this publisher, ensure that you handle jumps in time (both forward and back) and set the realtime multipleir correctly in those situations.
```
wjwwood: Spelling: `multiplier`

wjwwood: Also, I feel like this is an abuse of ROS Time. Why not integrate those sources into the system time? This is the same argument for how to integrate PTP (IEEE 1588) or something like TICSync (http://ieeexplore.ieee.org/document/5980112/?reload=true).

I'm curious to know why you would want to have a separate ROS specific mechanism for using those sources rather than just integrating those into the system time of the machine.

wjwwood: Maybe we need something related to, but in addition to, ROS Time for providing access to more exoctic time synchronization systems. But I still feel pretty strongly that we shouldn't conflate "time simulation" with "time synchronization".

articles/130_ros_time.md
```diff
@@ -126,9 +130,9 @@ However, if a client library chooses to not use the shared implementation then i
 
 In each implementation will provide `Time`, `Duration`, and `Rate` datatypes, for all three time source abstraction..
 The `Duration` will support a `sleep_for` function as well as a `sleep_until` method.
-The implementation will also provide a `Timer` object which will provide periodic callback functionality for all the abstractions.
+The implementation will also provide a `Timer` object which will provide periodic callback functionality for all the abstractions. It will be microsecond-accurate.
```
wjwwood: Why microsecond? At the moment our time structure uses nanoseconds as storage. I'm hesitant to try and make any guarantees about the accuracy of the timers, since they are often affected by other timers and/or work by default.

articles/130_ros_time.md
```diff
@@ -140,6 +144,15 @@ The underlying datatypes will also provide ways to register notifications, howev
   <b>TODO:</b> Enumerate the <code>rcl</code> datastructures and methods here.
 </div>
 
+## Work Needing To Be Done (Apr 2017)
+
+- Decide where and when to autosubscribe to the `/clock` message. Built in to Node? A Node addon or service? Part of the first call to `rcl_node_init`?
```
wjwwood, line 176: I think this builds on the comment you made elsewhere (can't find it right now), that Time needs to be somehow connected to a Node so it can subscribe to this topic. I think you're right about that, and it's something we have to consider now because in ROS 1 there was an omnipresent node which could be used for this purpose (the node singleton). We'll have to find a way to connect ROS Time to a node when sim time is desired.

I'm not sure how that should work atm. But it is definitely related to the mapping between ROS nodes and DDS participants and also how we manage multiple nodes in a single process, i.e. how the "container node" which loads and runs all the other nodes is implemented.

articles/130_ros_time.md
```diff
@@ -140,6 +144,15 @@ The underlying datatypes will also provide ways to register notifications, howev
...
+- Decide how to deal with multiple nodes in a single process that all update the one static time (or decide to ditch the static time methods).
+- Decide where to put the out-of-the-box clock publishers.
```
wjwwood, line 178: This is driven by the use case, at least in ROS 1. For instance, rosbag would publish it when playing back data and gazebo will also publish it when running a simulation.

articles/130_ros_time.md
```diff
@@ -140,6 +144,15 @@ The underlying datatypes will also provide ways to register notifications, howev
...
+- Ensure that there is no reason we can't enable `ROSTime` (with fallbacks) by default.
+- Change the RCL time implementation to call a method on the time source to return the current time (rather than using a field). This is necessary to compute the current time from the realtime factor. Fix the `rcl_get_ros_time` method to use the newly-added current time method.
+- Add an `rcl_time_source_t` (to the RCL project) for `ROSTime`.
```
wjwwood: I'm not sure I follow these last two bullets, can you expand on them and use exact type names and function names wherever possible? For example, "Change the RCL time implementation" is ambiguous to me. Which types and functions need to be changed.

> This is necessary to compute the current time from the realtime factor.

What is the realtime factor and where does it come from? I'm guessing it is described some in part of this pr, but I lost it in the diff. I'll try to come back and read it in the proposed state rather than just the diff if I have time.

wjwwood (commented):
Thanks for the feedback on the article, I tried to reply inline as best I could.

articles/130_ros_time.md
BrannonKing: Fixed.

articles/130_ros_time.md
BrannonKing: Fixed.

articles/130_ros_time.md
BrannonKing: Fixed to original order. I shouldn't have reordered them. I was thinking of ordering in default -> fallback 1 -> fallback 2, but things didn't come out that way regardless.

articles/130_ros_time.md
BrannonKing: Fixed.

articles/130_ros_time.md
BrannonKing: Fixed. I also renamed some things to avoid the word "timer" when I meant "frequency counter". Timer implies periodic events, which wasn't what I was talking about.

articles/130_ros_time.md
BrannonKing: You're probably thinking of my trailing comments here:
https://discourse.ros.org/t/of-clocks-and-simulation-betimes-and-otherwise/1587

The decision boils down to this: build it in to ROS node or opt-in to ROS Time by programmatically constructing an additional object? The same question goes for the ROS 1 node and system parameters, or maybe we say use command line parameters for all those usages. Maybe we want to use the command line parameters passed to rclcpp::init to change the default clock subscription mechanism. I'm okay with any of those options. The critical point is this: if we can make a decision on this very soon, I will have some time to implement it as part of my work for a June software release. I'd be happy to set up a conference call for discussing these options.

articles/130_ros_time.md
BrannonKing: I added ROSBag to the list of sources.

articles/130_ros_time.md
BrannonKing: I attempted to clarify the API change.

The real time factor has to be included in the clock message. I've tried to clarify that in the document as well.

tfoote (commented):
This looks to me like the proposal is trying to solve a separate problem than what the `/clock` topic is designed for.

One a live system `/clock` is not expected to be published. However if you had a processor which does not have system time in the current implementation would not help it so this change would require the `/clock` be published all the time.

Also in the proposal the realtime rate adds significant complications to propogating the time information in the case that the rate ever changes. This can happen both if you turn the knob on playback or simulation higher or lower. Or else pause the playback. There are also negative effects if the `/clock` messages get queued or lost.

Looking at the implications when you pause can be quite informative. In particular if the pause message is delayed due to delivery and the local node continues to interpolate into the future, when the pause or lower rate signal arrives. Suddenly your time is now just a little bit in the past. Do you jump it back? Do you wait? Do you slew? Do you invoke the callback to let the user know that time jumped backwards, because it didn't for the system, only for your node?

Adding a realtime factor basically assumes that the realtime factor will remain constant which is only the case if it's actually tracking "real" time. Which is not true for most use cases for using `/clock` at the moment, such as bag playback/looping, and simulation which often runs as fast at the simulator can handle.

Only sending the latest time, allows an imprecise but stepwise consistent time function that only goes backwards if the `/clock` signal goes backwards.


From the use case you cite of a computer w/o a system clock chip I think that there are better solutions than trying to use the `/clock` and ROS Time abstraction. There are many system time synchronization algorithms that are very robust and can provide a much more accurate and precise solution than just adding a realtime factor to the clock message. In particular these algorithms have to take into account the latency of communication, and deal with network dropouts etc. By far the most common one is Network Time Protocol (NTP) [Protocol RFC](https://tools.ietf.org/html/rfc5905). An implementation I know works well on robots is Chrony: https://chrony.tuxfamily.org/ since it's designed to work for intermittent network connections.

The ROS Time abstraction is not designed to replace the clock synchronization but to allow for other use cases like simultion and log file playback with support for looping, pausing, and variable rates. And all of this is turned on by the `using_sim_time` parameter. I think trying to add clock synchronization to the capabilities will actually bring down the performance of the system for all components unless we can achieve the same level of accuracy as the existing clock synchronization algorithms available. And clock synchronization is a very specialized field that's been well developed for a long time and has many robust implementations. 

So for your use case of having nodes that do not have a system clock chip. I'm assuming that means they do not persist the system time across reboot? If you simply add a NTP server to any computer on they system and start up ntp or chrony on the computer without persistent system time you'll get a more accurate time than if we enhanced the `/clock` messages and published it all the time.

For an understanding of the performance we'd be comparing to see: https://chrony.tuxfamily.org/comparison.html

- Calling out a few features such as oscillator rate correction/adjustment (assumed not necessary in the proposal)
- Works when network drops or is conjested.
- Can leverage hardware timestamping.

If we do want to provide a system clock synchronization I'd suggest that it be a completely separate channel than the `/clock` abstraction. And it would actually update the system time instead of trying to provide full time synchronization in the ROS implementation.

Also of note that there are times when avoiding the ROS Time abstraction is appropriate such as hardware drivers with hardware timeouts etc.

BrannonKing:
Thanks @tfoote .

> This looks to me like the proposal is trying to solve a separate problem than what the /clock topic is designed for.

Sorry for the confusion. That was not at all what I was intending. Your arguments against using ROS Time in place of clock synchronization are spot-on. What I want is the `using_sim_time` feature added to ROS2. That's what this is all about. I need it sooner rather than later. I'm proposing some enhancements to it, which I believe would make transform computations more accurate while using fewer resources. I will modify the document to be clearly recommending against cross-system clock broadcasts for real-time processing. 

> in the proposal the realtime rate adds significant complications to propagating the time information in the case that the rate ever changes. 

The complications are not any more sever with a real time factor than they are when dealing with an expired clock timestamp. The real time factor changes less frequently than time itself, which means that it should be more robust than relying on a continuously-published clock tick. It's not clear to me what real time rate the bag publisher would use or how that works in ROS1; if it is 0 for that and simulation, meaning all messages pushed out between clock messages are intended to happen concurrently, then the factor is useless. I don't see anything enforcing that in Gazebo, but maybe that was the intent -- publish one message per message type per world update (or less).

> Suddenly your time is now just a little bit in the past. Do you jump it back? ... or fire the callbacks?

Great question. I've been racking my brain to understand the need of the callbacks. Did they exist in ROS1? In everything I work with, we compute/acquire the time when a message leaves for the timestamp. When messages come in, we subtract/differentiate their timestamp with either the current clock or the time on the previous message. It's a math function that can take positive or negative numbers. The problem comes when `newMsg.time - oldMsg.time == 0`. That's worse than projecting a transform slightly into the future and sending that out.

> Only sending the latest time allows an imprecise but stepwise consistent time function...

It's exactly the "step" that is the problem. When my laser data capture comes in between time steps, I'm giving up accuracy to floor to the recent step. In testing the Cartographer on a nice i7 system, while simultaneously publish the clock at 1000Hz, I still see occasions where I go 50ms without getting a clock update. The loopback stalls in conjunction with a heavy processing load, and the time between messages gets too big. Adding in the stopwatch time since the last clock message would be much more accurate in this scenario.

> The ROS Time abstraction is not designed to replace the clock synchronization... And all of this is turned on by the using_sim_time parameter. 

So help me with the design on this. Should the sim-time be on by default? IOW, we use it if we get the message? Do you have any other arguments against including the real time factor for more accurate transformations?

tfoote:
> > in the proposal the realtime rate adds significant complications to propagating the time information in the case that the rate ever changes.
> 
> The complications are not any more sever with a real time factor than they are when dealing with an expired clock timestamp. The real time factor changes less frequently than time itself, which means that it should be more robust than relying on a continuously-published clock tick. It's not clear to me what real time rate the bag publisher would use or how that works in ROS1; if it is 0 for that and simulation, meaning all messages pushed out between clock messages are intended to happen concurrently, then the factor is useless. I don't see anything enforcing that in Gazebo, but maybe that was the intent -- publish one message per message type per world update (or less).

With respect to gazebo or rosbag publishing, one thing that I think is important is that it's not important when the data was published but when it was generated/collected etc. If you get a batch of messages all published at the same moment that is ok. It's just like they were queued up due to retransmission on a lossy network. The timestamps are related to the data source not the transport.

> > Suddenly your time is now just a little bit in the past. Do you jump it back? ... or fire the callbacks?
> 
> 
> Great question. I've been racking my brain to understand the need of the callbacks. Did they exist in ROS1?
We don't have callbacks for time changes so you get logic like this: https://github.com/ros/geometry2/blob/indigo-devel/tf2_ros/src/transform_listener.cpp#L105 That's part of our vision to clean up the API and support registering callback to handle time jumps more systematically.

> In everything I work with, we compute/acquire the time when a message leaves for the timestamp. When messages come in, we subtract/differentiate their timestamp with either the current clock or the time on the previous message. It's a math function that can take positive or negative numbers. The problem comes when newMsg.time - oldMsg.time == 0. That's worse than projecting a transform slightly into the future and sending that out.

I think that this is exactly where we're having a miss communication about simulation time vs synchronization. In addition to the above point about publish time vs generation time. In simulation or when playing back a bag. You never have an "expired clock timestamp" because if you haven't received a new clock message, time has not progressed. In simulation, until the next time step comes along, there will be no new images, or measurements etc. Likewise in bag playback when you pause there is no new data. And since the clock has stopped advancing all times and loops are also paused since their timeout has not been reached since time is not progressing.

Whereas you're talking about having a sensor that's still generating data despite the fact that time has stopped. If it's using simulated time, the sensor should be effectively paused as well. But if you're talking about a real hardware sensor it does not obey simulated time, so there's an inherent discrepancy that must be solved on a case by case basis how to connect a live clock to simulation time.

> > Only sending the latest time allows an imprecise but stepwise consistent time function...
> 
> It's exactly the "step" that is the problem. When my laser data capture comes in between time steps, I'm giving up accuracy to floor to the recent step. In testing the Cartographer on a nice i7 system, while simultaneously publish the clock at 1000Hz, I still see occasions where I go 50ms without getting a clock update. The loopback stalls in conjunction with a heavy processing load, and the time between messages gets too big. Adding in the stopwatch time since the last clock message would be much more accurate in this scenario.

Again this looks like you're trying to connect a simulated time system to a real time system and having trouble keeping them in sync. Thus it's really clock synchronization not simulated time.

> > The ROS Time abstraction is not designed to replace the clock synchronization... And all of this is turned on by the using_sim_time parameter.
> 
> So help me with the design on this. Should the sim-time be on by default? IOW, we use it if we get the message? 

Sim time should definitely not be on by default. If you do that and there's no `/clock` publisher nothing will happen.

In ROS 1 we've used the `/using_sim_time` parameter. However, since we don't have a global parameter server that won't work well. We haven't fleshed this out, but I think that it might make sense to simply trigger simulation time if there is an active publisher on `/clock`.

In ROS1 it was important to use the parameter and get that set before nodes were started so that they did not first jump to WallTime and then jump back to SimTime.  However if we have the callbacks that notify when time has jumped backwards and give nodes a change to respond, then that transition is possible to do if the `/clock` publisher is detected later.

So when there's a publisher of `/clock` then the latest value received will be provided by the ROSTime abstraction.

If the publisher goes away we will return to falling through to the system time.

> Do you have any other arguments against including the real time factor for more accurate transformations?

The main reason for not including the realtime factor is that it can lead to divergence which would require corrections.

Taking the simple example of pausing the simulated time at t=10. If you're stepping forward with a realtime factor of 1 and you publish every second. Half a second later you report 10.5 and then as you get closer to 1 second later you're approaching reporting 11. However you then receive 10.5x0.5. You're now ~ half a second ahead of the clock signal. But did the message come in late or is the signal just self inconsistent? It's ambiguous. We could conservatively call the callback function to tell the process that time jumped backwards, but that's starting to risk triggering the backwards time callbacks any time we get network congestion. 

Now we could escalate and start encoding a maximum extrapolation distance. But then what do you do after you've reached the limit and you've received no new data? And if there was a change in the propagation that was not communicated quickly again we could get out of sync.

So we can add an offset from the system time, with the simulated time and the rate. But there's still issues with jitter and extrapolation limits and possible inconsistency. And to pause the system we need to communicate ahead to shorten the extrapolation limits or set the rate to zero. Which means any time you pause the system you're basically guaranteed to have time overshoot. Or you cannot pause instantly, but need to let the system wind down to the next extrapolation limit.

These problems are particularly an issue if you're using simulation and you want it to run as fast as possible so the clock is ticking at a rate that's not consistent. Some time steps are 95% realtime some are 90% realtime. We can avoid changes in the time rate by limiting the rate to something that it's guarenteed the simulator can keep up with. But if you're doing things like turning on and off cameras that can have a 25% change in the rate of simulation you really don't want to always force the minimum rate.

Whereas if we just make the promise that the time will be non-decreasing we can simply publish the current clock step and as long as delivery is in order the clock will remain consistent. Any given node that's over a latent like may be as far behind as the latency of the link. There are cases where this is not close enough, however those are relatively few high performance control loops etc that are very closely tied to time passing. You do have to make sure your code is robust to having two now() calls return the same time. But that's true of most clocks besides specific MONOTONIC clocks.

We can relatively easily add more complexity to the system, but until shown that the simpler system is inadequate I don't think we should make it more complicated. And to that point the current proposal is effectively the current ROS1 clock system with a plan for an improved API. But structurally the current simple system works for the large majority of use cases and I'm not aware of complaints or requests for higher fidelity simulation time.

BrannonKing:
This question still remains: where and when in the codebase do we subscribe to `/clock`? My earlier proposal was that we hijack rcl's `ros_node_init` for a one-time initialization (aka, for the first node initialized in a process).

BrannonKing:
Suppose we make the clock use explicit through environment variables:
`ROS_TIME_SOURCE=SYSTEM_CLOCK or STEADY_CLOCK or BROADCAST_CLOCK_MSG`

We will default to SYSTEM_CLOCK, and throw an error if there is no system clock. If using the clock message, we will block if you try to acquire time before we get the message. 

I was thinking earlier that we would fall back to the steady clock if there was no system clock, and that we would fall back to the system clock if there was no clock message broadcast, but I think now that is a mistake. I think the auto-fallbacks would result in messages published with incorrect times; it would make debugging difficult. I hate to block a program indefinitely while waiting for a clock message that is never coming; I suppose we would block for a second and then chuck an exception.

I suppose we could have an invalid time instead of blocking. We would probably use a seconds == -1 && nanoseconds == 0 for that.

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
 
 #### System Time
 
-For convenience in these cases we will also provide the same API as above, but use the name `SystemTime`.
+System Time pulls the current time from the system clock chip. 
+It provides benefit over Steady Time when all nodes in the system have synchronized system time available or when all nodes run on the same host that contains a system clock chip.
 
-`SystemTime` will be directly tied to the system clock.
+We expect a System Time publisher to throw an error if the system clock resource is not available.
+We expect methods returning the current time to fall back to Steady Time if System Time is not available.
```
wjwwood, line 90: Unfortunately I'm concerned that this is not possible, because most steady clocks reference some arbitrary point in the past, like time since system boot, whereas system time usually assumes a known point in history, such that it can be converted to a date.

The system clock, as described by the C++ standard needs to convertible to a (`std::time_t`):

> It [`std::chrono::system_clock`] is the only C++ clock that has the ability to map its time points to C-style time, and, therefore, to be displayed [implication is displayed as a date].

-- http://en.cppreference.com/w/cpp/chrono/system_clock

And since it must be able to be mapped to the "C-style time" (`std::time_t`), which the reference has this to say about it:

> Although not defined, this is almost always an integral value holding the number of seconds (not counting leap seconds) since 00:00, Jan 1 1970 UTC, corresponding to POSIX time.

-- http://en.cppreference.com/w/cpp/chrono/c/time_t

So strictly speaking, falling back to steady time should be ok, but in practice if one node has to fall back to steady time and another node (on a PC) is using its own system time, they will not be talking about the same time in history.

So the system time doesn't necessarily need to return a time since the unix epoch, but everyone in the system needs to agree on what that arbitrary starting point is, so the fallback behavior seems dangerous and possible not useful in practice.

My gut reaction is to say we require a system clock or some service to emulate it (not necessarily using ROS Time's `/clock` to do so, but required nonetheless), but I'm interested to see what others think about it.

wjwwood: I also think it would be useful to have a convention for what time is returned by ROS Time, like "POSIX time" (since unix epoch) rather than something like [`CLOCK_TAI`](http://stackoverflow.com/a/32674960/671658) or January 1, 1601 UTC (I think Windows uses that, see [this](https://support.microsoft.com/en-us/help/167296/how-to-convert-a-unix-time-t-to-a-win32-filetime-or-systemtime)).

The obvious answer seems like POSIX time since that's what is implied in ROS 1, but I'm not sure what the ramifications are, especially if someone wants to use `CLOCK_TAI` for example, which actually probably makes more sense for our applications, since we care about unique timestamps and distances in time more than representing them as dates.

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
+Steady Time implies a real-time incrementing source.
+There is no modern hardware or operating system that does not support this option.
+Steady Time typically begins at system bootup and measures increments using hardware frequency counters.
 
-In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
+Example use cases for this include hardware drivers which are interacting with peripherals with hardware timeouts and systems lacking a realtime clock chip.
+
+In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals, it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
```
wjwwood, line 100: The other use case (in my opinion the main use case) for this is timers that should not be interrupted by jumps in system time, which might be implied by "interacting with hardware or other peripherals". But systems lacking a real-time clock (btw we should try to be consistent with real-time vs realtime) don't really benefit by steady time, as they still need system clock (provided by the system or externally) to exchange information with other computers or to be correlated to data taken elsewhere. Though it is true they should always be able to provide it.

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
 
 #### ROS Time
 
-The `ROSTime` will report the same as `SystemTime` when an ROS Time Source is not active.
-When the ROS time source is active `ROSTime` will return the latest value reported by the Time Source.
+ROS Time derives the current time from the most-recently received clock broadcast message published on the `/clock` topic.
 
-### ROS Time Sources
+When the ROS Time source is active `ROSTime` will return the latest value reported by the Time Source plus the elapsed time since that report multiplied by the real time factor.
```
wjwwood, line 106: I'm still not convinced of this. Because individual systems will be extrapolating from the most recently broadcasted time, I'm concerned that pausing the flow of time will possibly leave multiple processes on the same machine disagreeing about what the current time is.

Additionally, I think it would also introduce frequent (but small) jumps backwards in time. I can also imagine timers that should wake up after t=2 with code that comes after it seeing current time as t=1.9 which would be odd. Note that this _could_ happen normally, if the system date was adjusted between the timer waking up and checking the current time, but it usually doesn't happen very frequently.

@tfoote maybe you can explain this better than me (he and I talked about this on a whiteboard a day or so ago).

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
 
-#### Default Time Source
+The ROS Time broadcast (clock message) needs to include this information:
+- The current time (in nanoseconds, epoch depends upon source and should not be needed)
+- The current real time factor (also called the "multiplier", typically a value of one, but could be zero for paused or negative when going backwards).
+- The system clock time (in nanoseconds since the Unix epoch) when the message was created (or -1 if not available)
```
wjwwood, line 111: I do not think we can assume the system clock is not available. Also, `-1` is not a special value in system time. It would just be 1 second before 1970 if the seconds field is signed. I'd recommend `0` be the go to value if system time is not yet available (still not sure that make sense).

This is the behavior in ROS 1 (btw):

```
% python
Python 2.7.11 (default, Jan 18 2016, 12:41:04)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import rospy
>>> rospy.Time.now()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/william/indigo/install/lib/python2.7/site-packages/rospy/rostime.py", line 155, in now
    return get_rostime()
  File "/Users/william/indigo/install/lib/python2.7/site-packages/rospy/rostime.py", line 209, in get_rostime
    raise rospy.exceptions.ROSInitException("time is not initialized. Have you called init_node()?")
rospy.exceptions.ROSInitException: time is not initialized. Have you called init_node()?
>>> rospy.init_node('foo')
>>> rospy.Time.now()
rospy.Time[1493933142113349914]
>>> ^D

% rosparam set use_sim_time true

% python
Python 2.7.11 (default, Jan 18 2016, 12:41:04)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import rospy
>>> rospy.Time.now()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/william/indigo/install/lib/python2.7/site-packages/rospy/rostime.py", line 155, in now
    return get_rostime()
  File "/Users/william/indigo/install/lib/python2.7/site-packages/rospy/rostime.py", line 209, in get_rostime
    raise rospy.exceptions.ROSInitException("time is not initialized. Have you called init_node()?")
rospy.exceptions.ROSInitException: time is not initialized. Have you called init_node()?
>>> rospy.init_node('foo')
>>> rospy.Time.now()
rospy.Time[0]
>>>
```

As you can see it also raises an exception if a node has not been created yet. If I then start a rosbag I will start getting some sim time (with a big initial jump from 0 to the first time in the bag):

```
>>> # started rosbag play with --clock
...
>>> rospy.Time.now()
rospy.Time[1481249541938720020]
```

Not to say that this is the only way to do it, but I think it works ok in ROS 1 and I don't see any alternatives with advantages atm.

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
+- An included (where?) System Clock Publisher.
+- An included (where?) Gazebo Plugin.
+- The ROSBag playback engine.
 
-The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific.
+#### Default Time Source
 
-##### No Advanced Estimating Clock
+Users can call `Node::now()` for the current time.
+With the assumption of a single Node instance per process, you can also get the current time from a static method: `Time::now()`.
```
wjwwood, line 130: Unfortunately I don't think this assumption will ever work. We've been talking at OSRF about the skeleton of the code so that others can help implement it (like you 😀), and I think we've concluded that the "ROS Time" class will only be available from the node class. This ensures that it has the resources it needs to get the "broadcast time" if necessary.

I'm working on a new issue on ros2/ros2 to help guide this effort and to collect details about the detailed code design. hopefully I can finish that this weekend. After that you, and anyone else who is interested, can join us in a hangout or something to discuss it.

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
+Nodes will automatically subscribe for the `/clock` message unless it is programmatically disabled. (Do we need to disable it? We can simply "not publish" to use the system clock.)
+If a publisher exists for the `/clock` topic, it will override the system time when using the ROS time abstraction.
+The `ROSTime` and the `now()` methods will fall back to `SystemTime` (and then `SteadyTime`) when an ROS Time Source is not active (aka, it has been disabled purposely or hasn't received a clock broadcast).
```
wjwwood, line 133: Also from talking about this offline we thought that keeping the `use_sim_time` parameter (like in ROS 1) would be the best way to control it. We thought about using a heuristic to enable it, like "is /clock being published to", but there were so many issues we didn't see how to solve, like the fact that there would be a startup race condition and knowing how and when to swap back to non-simulated time.

I think the the parameter to control it should be used.

As to being able to programmatically avoid subscribing to it, I'd again say this can be covered by the `use_sim_time` parameter.

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
 
-There are more advanced techniques which could be included to attempt to estimate the propagation properties and extrapolate between time ticks.
-However all of these techniques will require making assumptions about the future behavior of the time abstraction.
-And in the case that playback or simulation is instantaneously paused, it will break any of these assumptions.
-There are techniques which would allow potential interpolation, however to make these possible it would require providing guarantees about the continuity of time into the future.
-For more accuracy the progress of time can be slowed, or the frequency of publishing can be increased.
-Tuning the parameters for the `/clock` topic lets you trade off time for computational effort and/or bandwidth.
+Users can also add callbacks to `Node::time_reversing` and `Node::time_jumping_back`.
+The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
+If the time on the clock jumps backwards, a callback handler will be invoked and be required to complete before any calls to the ROS time abstraction report the new time.
```
wjwwood, line 137: What's the purpose of `Node::time_reversing`?

articles/130_ros_time.md
```diff
@@ -39,85 +45,110 @@ Slower than real time simulation is necessary for complicated systems where accu
...
-The user will be able to switch out the time source for either each instance of their Time object as well as have the ability to override the default for the process.
+#### Custom Time Source
 
-It is possible to use an external time source such as GPS in as a ROSTime source, but it is recommended to integrate a time source like that using standard ntp integrations with the system clock since that is already an established mechanism and will not need to deal with more complicated changes such as time jumps.
+It is possible that the user may have access to an out-of-band-time source which can provide better performance than the default source the `/clock` topic.
+It might be possible that, for their particular use case, a more advanced algorithm is needed to propagate the simulated time with adequate precision for latency arising with restricted bandwidth or connectivity.
+The user will be able to switch out the time source for each instance of their Time object as well as have the ability to override the default for the process or node.
+This can be acheived in C++ using the template on the Time object. The RCL package also contains a method to disable `ROSTime`.
```
wjwwood, line 147: This might be too implementation specific. It might be better to just say, a user will be able to control the time source for ROS Time programmatically in C++.

Nitpick: one sentence per line

articles/130_ros_time.md
wjwwood: As I mentioned above I think that the "ROS Time" object will need to be created from the Node or by getting a node in from the constructor.

The "System" time object and the "Steady" time object can be independent of the node. The only reason to still have a custom version of those, rather than use `std::chrono` directly, is to present a similar enough interface to the "ROS Time" object so that other classes like Timers can use any of them interchangeably.

wjwwood (commented):
I added a few more comments. Also I'm working on an issue on the ros2/ros2 repository that can coordinate the work on this issue. It's not quite ready yet, but I think we're getting closer to an idea of what the actual code will look like thanks to your initiative here.

Thanks for iterating with me and being patient. I am trying to remain responsive on the issue, but I have a lot of other obligations too 😀.

articles/130_ros_time.md
BrannonKing: I too have come to the conclusion that opting into a broadcast clock should be explicit.

articles/130_ros_time.md
BrannonKing: The purpose of `time_reversing` was for special handling when the real time factor went negative (aka, time was going backwards). However, it sounds like the real time factor complicates too many scenarios; we aren't going with that plan.

BrannonKing:
I'm planning to close this pull request soon. It sounds like the "real time factor" parameter is not where we want to go. Here are my conclusions:

1. We want to opt in to the broadcast time. Do this through a command line parameter (`--use_sim_time`) and/or an environment variable (`ROS_USE_SIM_TIME`). We will return `system_clock::now()` (or some OS-independent clock) if you don't opt in.
2. The preferred way to get time will involve a node instance. We can have an additional time service that takes a node in its constructor, or we can add it into the node itself. We need to do this soon so people aren't in the habit of using the static time method.
3. We won't use the real time factor and we will keep the existing time message format for the `/clock` topic.

articles/130_ros_time.md
BrannonKing: Yes, we definitely want to ensure that our clock code returns the same numbers independently of the operating system.

wjwwood:
I created https://github.com/ros2/ros2/issues/350 to try and track the progress on this issue.

I don't think this one needs to be closed. I think there are some improvements to be taken from it that are unrelated to the suggestions about better propagation of the simulated clock and how to extend the rationale for a custom time object to possibly include a custom source of synchronized time.

Btw, I think both of those points may still have value, just in addition to the ROS 1 style `/clock` topic rather than as a replacement (right now).

tfoote:
@BrannonKing I've just merged #144 where instead of changing the default time source I've left it open to providing alternative time sources to clocks. There's currently the default TImeSource but that should be overridable to provide an advanced time source. https://github.com/ros2/rclcpp/pull/371/files 

Related implementation: https://github.com/ros2/rcl/pull/164

These are still under review but could you look at implementing the timesource you're describing on top of this new structure such that it will pump the clocks will support interpolation as you've described.

tfoote:
The related APIs have been merged with a default time source implementation. An enhanced time source could be implemented too.

PR #130, opened 2017-05-01
remap char and byte to int8_t and uint8_t

mikaelarguedas:
This is a draft of what deprecating char and byte would look like.

What could be done on top: On the python side we could relax the constraint by allowing users to pass `builtin.bytes of length 1` and make the conversion in the setter functions

Connects to ros2/rosidl#190

### DIFF
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -54,8 +54,6 @@ Each of these can optionally be a dynamically or statically sized array.
 The following primitive types are defined:
 
 - `bool`
-- `byte`
-- `char`
 - `float32`, `float64`
 - `int8`, `uint8`
 - `int16`, `uint16`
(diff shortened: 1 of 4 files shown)


wjwwood (approved):
lgtm, the rationale makes sense

We may need to (re)consider how we map int types of various sizes to types in Python. Int is fine for now, but in ROS 1 there is an option to use `numpy` and map them to more efficient types.

I only bring this up because if you imagine transmitting image data, defined as a `uint8[]` in the message file would become a list of `Int` in Python, which is quite a bit larger. If we used `numpy` we could map the `[u]int{8,16,32,64}` types to their exact size. But even without `numpy` perhaps there should be a special exception for `uint8[]` being mapped to `bytes` or a `bytearray` in Python (since they are between 0 and 256 too):

> Return a new “bytes” object, which is an immutable sequence of integers in the range 0 <= x < 256.

--
 https://docs.python.org/3.1/library/functions.html#bytes

dirk-thomas (changes requested):
When I am looking at this from the point of view of the user I see the following case. A user wants to send and receive a sequence of bytes in Python. So I would expect the Python API to write `bytes` and read `bytes`. That type is the user supposed to use in the `.msg` file? I would think `byte[]` would be the natural choice (which would probably need an update in `rosidl_generator_py`).

Until this is being figured out I would rather not merge this change,

dirk-thomas:
This has been addressed differently in the meantime: see https://github.com/ros2/design/pull/213/files#diff-0307e01fdcc93b6ad5b2241e0bde127fR284

PR #136, opened 2017-08-23
Update the CPP article to talk about initialization.

clalancette:
connects to ros2/ros2#396

@ros2/team Despite our long conversations elsewhere, this ends up being a short addition to the design document.  I'll put some more notes inline; feedback is welcome.

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -105,7 +105,7 @@ All other constants are declared as `static const` members in the struct and the
 
 The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
 In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
-Therefore, the constructor takes an optional directive of type `rosidl_runtime_cpp::MessageInitialization` to control how initialization is done:
+Therefore, the constructor takes an optional directive of type `rosidl_generator_cpp::MessageInitialization` to control how initialization is done:
 
 - `MessageInitialization::ALL` - Initialize all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization)
   - The safest option, and also the default (used if not passing any argument to the constructor).
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -103,10 +103,20 @@ All other constants are declared as `static const` members in the struct and the
 
 ### Constructors
 
-Currently the *default constructor* does no initialization of member fields.
-This means that the user is responsible for ensuring that all fields are initialized before use, otherwise undefined behavior may result.
-In a future release, the *default constructor* will initialize all fields by default.
-Optionally the constructor can be invoked with an allocator as discussed [above](#messages).
+The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
+In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
+Therefore, the constructor takes an optional directive of type `rosidl_runtime_cpp::MessageInitialization` to control how initialization is done:
+
+- `MessageInitialization::ALL` - Initialize all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization)
+  - The safest option, and also the default (used if not passing any argument to the constructor).
+- `MessageInitialization::SKIP` - Don't initialize any members; it is the user's responsibility to ensure that all fields get initialized with some value, otherwise undefined behavior may result
+  - Used for maximum performance if the user is setting all of the members themselves.
+- `MessageInitialization::ZERO` - Zero initialize all members; this differs from `MessageInitialization::ALL` in that all members will be [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization), and default values from the message definition will be ignored
+  - Used when the user doesn't want the overhead of initializing potentially complex or large default values, but still wants to ensure that all variables are properly initialized.
+- `MessageInitialization::DEFAULTS_ONLY` - Default initialize only fields that have default values assigned to individual members; all other fields will be left uninitialized
+  - Minimal initialization which ensures that existing code has correctly initialized members when a new field with a default value is added to the IDL later.
+
+Optionally the constructor can be invoked with an allocator.
 
 The struct has no constructor with positional arguments for the members.
 The short reason for this is that if code would rely on positional arguments to construct data objects changing a message definition would break existing code in subtle ways.


articles/112_generated_interfaces_cpp.md
```diff
@@ -104,6 +104,13 @@ All other constants are declared as `static const` members in the struct and the
 ### Constructors
 
 The *default constructor* initializes all members with their default value.
+In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
+Therefore, the constructor takes an optional directive of type `enum rcl_msg_init_type` to control how initialization is done:
```
clalancette: rcl seems to me a decent place to put this enum, but I'm certainly open to moving it elsewhere.

articles/112_generated_interfaces_cpp.md
```diff
@@ -104,6 +104,13 @@ All other constants are declared as `static const` members in the struct and the
...
+
+- `MSG_INIT_DEFAULT_INITIALIZE_ALL` - Default initialize all members; any fields that have default values assigned to individual members will be set to the default values, all other values will be assigned to C++ defaults (generally 0 or the empty string).  Equivalent to not passing any argument to the constructor
+- `MSG_INIT_DEFAULT_INITIALIZE_NONE` - Don't initialize any members; it is the user's responsibility to ensure that all fields get initialized with some value, otherwise undefined behavior results
+- `MSG_INIT_ZERO_INITIALIZE_ALL` - Zero initialize all members; this differs from `MSG_INIT_DEFAULT_INITIALIZE_ALL` in that all members will be set to their C++ defaults (generally 0 or the empty string), and default values from the message definition will be ignored
+
```
clalancette: When we discussed this before, @dirk-thomas had a fourth type he mentioned which was "all-zero+defaults".  However, I don't quite see how that is different than DEFAULT_INITIALIZE_ALL, so I haven't added it here.  If we need another type, please let me know and I'll add it.

clalancette: Also, the `MSG_INIT_*` names seem OK to me, but feel free to suggest alternative names.

articles/112_generated_interfaces_cpp.md
dirk-thomas: The data structures must not be coupled with the ROS client library. One goal is that they are usable as plain data classes without requiring the ROS client libraries.

Such an enum needs to be defined in a package which is a runtime dependency of any message package. Since we don't want message packages to runtime depend on the generator package it needs to be in a separate package (similar as `roscpp_core` in ROS 1).

articles/112_generated_interfaces_cpp.md
dirk-thomas: The four options we discussed in the meeting are:

* `SKIP_INITIALIZE` - Does neither set members which have default values defined in the IDL nor set member to a type-specific default value.
  * This is used for maximum performance if the user sets all members on his own anyway.
* `INITIALIZE_DEFAULTS_ONLY` - Only initializes the members which have default values defined in the IDL. The other member are not initialized and are expected to be set by user land code.
  * This is the "minimal" initialization that ensures that existing code continues to have correctly initialized member when a new field with a default value is added later to the IDL.
* `INITIALIZE_ALL` - Initializes the members which have default values defined in the IDL. All other members are initialized with their type-specific default value.
  * This is the safest options and should be the "default".
* `ZERO_INITIALIZE` - Initializes all members with their type-specific default value not considering any default values specified in the IDL.
  * This could be used when the user doesn't want the overhead of potentially complex / large default values (since they are overridden anyway) but still wants to ensure that all variables are properly initialized.

articles/112_generated_interfaces_cpp.md
clalancette: Right, understood.  Thanks for the explanation.  Given that we want this to be from the C layer, what do you think about a new package `rcl_core`?  At this point it would just contain a single header file with this enum, we could expand it over time.  Other ideas for this welcome.

articles/112_generated_interfaces_cpp.md
clalancette: Got it.  I was missing the `INITIALIZE_DEFAULTS_ONLY`.  I'll update the description to change to these names and add in that 4th option.

articles/112_generated_interfaces_cpp.md
dirk-thomas: Since the symbol should be independent of the ROS client library I don't think `rcl` is a good prefix. Maybe `rosidl_runtime_cpp`? `rosidl_cpp_runtime` sounds nicer but all the generator packages are named `rosidl_generator_*`...

dirk-thomas (commented):
I think it would be beneficial to include the mentioned use cases for each option.

articles/112_generated_interfaces_cpp.md
```diff
@@ -103,7 +103,15 @@ All other constants are declared as `static const` members in the struct and the
...
+In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
+Therefore, the constructor takes an optional directive of type `enum rosidl_runtime_cpp_msg_init_type` to control how initialization is done:
+
+- `MSG_INIT_INITIALIZE_ALL` - Initialize all members; any fields that have default values assigned to individual members will be set to the default values, all other values will be assigned to C++ defaults (generally 0 or the empty string).  Equivalent to not passing any argument to the constructor
+- `MSG_INIT_SKIP_INITIALIZE` - Don't initialize any members; it is the user's responsibility to ensure that all fields get initialized with some value, otherwise undefined behavior results
+- `MSG_INIT_ZERO_INITIALIZE` - Zero initialize all members; this differs from `MSG_INIT_INITIALIZE_ALL` in that all members will be set to their C++ defaults (generally 0 or the empty string), and default values from the message definition will be ignored
+- `MSG_INIT_INITIALIZE_DEFAULTS_ONLY` - Default initialize only fields that have default values assigned to individual members; all other fields will be left uninitialized
+
```
dirk-thomas: Why do all the symbols start with `MSG_INIT_`? Why not `enum class MessageInitialization { ALL, SKIP, ZERO, DEFAULTS_ONLY };`?

clalancette: That is mostly because I wanted to put this in C header file, rather than C++.  That way, it could conceptually be used by both C initialization and C++, though it would obviously be different and more awkward in C (I'd guess we'd need to use a factory-style function there, but the idea would be the same).  If we don't care about that requirement anymore, then we could use an enum class instead.

dirk-thomas: Since the article this PR is editing is for C++ I thought the proposed symbol names are supposed to be used for the C++ message types. I would rather like a C++ specific enum class for the C++ message classes.

Of course the C message classes need C-type arguments instead. The C article should probably be updated as well in this PR. Than both documents clearly define there language specific symbols.

clalancette: It is true, if we restrict this to C++, then we can use the more advanced constructs.  However, we would then have to effectively duplicate this enum into C as well, and make sure to keep them in sync as we change them.  We have precedent for sharing C-style enums between C and C++ elsewhere; `rmw/include/rmw/types.h` comes to mind.

articles/112_generated_interfaces_cpp.md
```diff
@@ -103,7 +103,19 @@ All other constants are declared as `static const` members in the struct and the
 
 ### Constructors
 
-The *default constructor* initializes all members with their default value.
+The *default constructor* initializes all members with their default value; in case the field doesn't have a default value, the C++ default is used (generally 0 or the empty string).
```
wjwwood: Is there a reference for the "C++ default" for a value? If so perhaps we should link to it, if not maybe it could be added to the table above where there is a mapping between our idl primitive types and the matching C++ type.

wjwwood (approved):
Other than a comment, lgtm.

articles/112_generated_interfaces_cpp.md
clalancette: Yeah, you are right, this language is actually too vague.  I'm going to update it to talk about value initialization instead, since that is what we want to emulate.

articles/112_generated_interfaces_cpp.md
```diff
@@ -103,7 +103,19 @@ All other constants are declared as `static const` members in the struct and the
 
 ### Constructors
 
-The *default constructor* initializes all members with their default value.
+The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
+In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
+Therefore, the constructor takes an optional directive of type `enum rosidl_runtime_cpp_msg_init_type` to control how initialization is done:
```
dirk-thomas: I still don't see why a C++ specific type should have such a long type name (`rosidl_runtime_cpp_msg_init_type`). Same for the enums below.

articles/112_generated_interfaces_cpp.md
```diff
@@ -103,7 +103,19 @@ All other constants are declared as `static const` members in the struct and the
...
+
+- `MSG_INIT_INITIALIZE_ALL` - Initialize all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
+  - The safest option, and also the default (used if not passing any argument to the constructor).
+- `MSG_INIT_SKIP_INITIALIZE` - Don't initialize any members; it is the user's responsibility to ensure that all fields get initialized with some value, otherwise undefined behavior results
```
dirk-thomas: > otherwise undefined behavior results

I don't think this is accurate. Undefined behavior doesn't necessarily happen - it could be though. I think it would be good to clarify that.

clalancette: Right.  Will clarify that point.

articles/112_generated_interfaces_cpp.md
clalancette: Yeah, I agree it is a bit unwieldy, though I think you do want to encapsulate both a prefix and a description.  How about `rosidl_runtime_cpp_msg_init`?  Or `rosidl_msg_init_type`? Or `rosidl_msg_init`?

As for the name of the enums themselves, that comes down to whether we use the C++-specific enum class or C-style enum.  If we use the latter, then they go into the global namespace, so you want something unique.  If we use the former, then we can get rid of a lot of that prefix.  I think this is the one unresolved question here; do we stick with C-style enums or go with C++-style ones?  C-style enums will be better for maintenance across C and C++ longterm, but C++ ones are easier to work with.  Further thoughts?

dirk-thomas: For a C++ interface I expect C++ enum class (as proposed in https://github.com/ros2/design/pull/136#discussion_r134885656). Internally the enum values can be defined based on the C enum values - no need to duplicate the constants.

clalancette: Got it, will do it that way.

clalancette:
All right, updated to use C++-style enum.  I also shortened the name to `rosidl_msg_init_type`, which is a lot less unwieldy.  I'm not at all wedded to that name though, so I'll happily take other suggestions.

dirk-thomas:
> I'm not at all wedded to that name though, so I'll happily take other suggestions.

I have made a suggestion twice already. I also mentioned why I think the type name should be shorter.

Maybe someone else can comment what names they prefer.

dirk-thomas (approved):
LGTM.

We should probably merge this once the implementation has been updated accordingly.

clalancette:
@dirk-thomas Great, thanks.  I'll wait to merge this as you say.

clalancette:
This was previously approved, and now that the code has been merged, I'm going to merge this.

PR #137, opened 2017-08-25
update middleware interface article

dirk-thomas:
* Mention `rmw_implementation` with the selection at compile time as well as runtime.
* Mention optional native handles.
* Mention `rmw_fastrtps`.

---

* Remove obsolete link time selection.

### DIFF
--- a/articles/060_ros_middleware_interface.md
+++ b/articles/060_ros_middleware_interface.md
@@ -309,40 +309,37 @@ While using such a feature would make the user land code specific to the middlew
 
 The described concept has been implemented in the following packages:
 
-- the package [rmw](https://github.com/ros2/rmw/tree/master/rmw) defines the middleware interface
+- The package [rmw](https://github.com/ros2/rmw/tree/master/rmw) defines the middleware interface.
 
-  - the functions are declared in [rms/rmw.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h)
-  - the handles are defined in [rmw/types.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h)
+  - The functions are declared in [rms/rmw.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h).
+  - The handles are defined in [rmw/types.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h).
 
-- the package [rmw_fastrtps_cpp](https://github.com/ros2/rmw_fastrtps/tree/master/rmw_fastrtps_cpp) implements the middleware interface using [eProsima Fast-RTPS](http://www.eprosima.com/index.php/products-all/eprosima-fast-rtps) based on the introspection type support
+- The package [rmw_fastrtps_cpp](https://github.com/ros2/rmw_fastrtps/tree/master/rmw_fastrtps_cpp) implements the middleware interface using [eProsima Fast-RTPS](http://www.eprosima.com/index.php/products-all/eprosima-fast-rtps) based on the introspection type support.
 
-- the package [rmw_connext_cpp](https://github.com/ros2/rmw_connext/tree/master/rmw_connext_cpp) implements the middleware interface using [RTI Connext DDS](http://www.rti.com/products/dds/index.html) based on statically generated code
+- The package [rosidl_generator_dds_idl](https://github.com/ros2/rosidl_dds/tree/master/rosidl_generator_dds_idl) generates DDS IDL files based on ROS msg files which are being used by all DDS-based RMW implementations which use static / compile type message types.
 
-  - the package [rosidl_typesupport_connext_cpp](https://github.com/ros2/rmw_connext/tree/master/rosidl_typesupport_connext_cpp) generates
+- The package [rmw_connext_cpp](https://github.com/ros2/rmw_connext/tree/master/rmw_connext_cpp) implements the middleware interface using [RTI Connext DDS](http://www.rti.com/products/dds/index.html) based on statically generated code.
 
-    - the DDS specific code based on IDL files for each message
-
-      - the package [rosidl_generator_dds_idl](https://github.com/ros2/rosidl_dds/tree/master/rosidl_generator_dds_idl) generates DDS IDL files based on ROS msg files
-
-    - additional code to enable invoking the register/create/convert/write functions for each message type
-
-- the package [rmw_connext_dynamic_cpp](https://github.com/ros2/rmw_connext/tree/master/rmw_connext_dynamic_cpp) implements the middleware interface using *RTI Connext DDS* based on the DynamicData API
-
-  - the package [rosidl_typesupport_introspection_cpp](https://github.com/ros2/rosidl_dds/tree/master/rosidl_typesupport_introspection_cpp)
-
-    - generates code which encapsulated the information from each ROS msg file in a way which makes the data structures introspectable
-
-- the package [rmw_opensplice_cpp](https://github.com/ros2/rmw_opensplice/tree/master/rmw_opensplice_cpp) implements the middleware interface using [PrismTech OpenSplice DDS](http://www.prismtech.com/opensplice) based on statically generated code
-
-  - the package [rosidl_typesupport_opensplice_cpp](https://github.com/ros2/rmw_opensplice/tree/master/rosidl_typesupport_opensplice_cpp) generates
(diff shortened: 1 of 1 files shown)


articles/060_ros_middleware_interface.md
```diff
@@ -330,6 +314,8 @@ The described concept has been implemented in the following packages:
   - the functions are declared in [rms/rmw.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h)
   - the handles are defined in [rmw/types.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h)
 
+- the package [rmw_fastrtps_cpp](https://github.com/ros2/rmw_fastrtps/tree/master/rmw_fastrtps_cpp) implements the middleware interface using [eProsima Fast-RTPS](http://www.eprosima.com/index.php/products-all/eprosima-fast-rtps) based on the introspection type support
+
```
mikaelarguedas: given that we're shuffling sections around, I wonder if the list below should be updated as well:
- rosidl_generator_dds_idl is used by all rmw implementations based on static code (e.g. opensplice, coredx) and not only Connext

- not sure if we should flag non supported implementations one way or another because we could expect that rmw_connext_dynamic and rmw_opensplice are at feature parity and receive the same amount of effort as the other ones

articles/060_ros_middleware_interface.md
```diff
@@ -353,6 +339,11 @@ The described concept has been implemented in the following packages:
     - the DDS specific code based on IDL files for each message
     - additional code to enable invoking the register/create/convert/write functions for each message type
 
+- the package [rmw_implementation](https://github.com/ros2/rmw_implementation/tree/master/rmw_implementation) provides the mechanism to switch between compile time and runtime selection of the middleware implementation
+
+  - if only one implementation is available at compile time it links directly against it
+  - if multiple implementations are available it implements the middleware interface itself and acts according to the strategy pattern by loading the shared library of a specific middleware implementation identified by an environment variable at runtime and pass all calls along
```
mikaelarguedas: feel free to discard if you think it's too repetitive "if multiple implementations are available **at compile time**"

articles/060_ros_middleware_interface.md
```diff
@@ -170,44 +171,19 @@ However the performance using the DynamicData API will likely always be lower co
...
 When ROS supports different middleware implementations it should be as easy and low effort as possible for users to switch between them.
 
-One obvious way will be for a user to rebuild all ROS packages from source selecting a different middleware implementation.
+#### Decide at compile time
+
+One obvious way will be for a user to build all ROS packages from source selecting a specific middleware implementation.
 While the workflow won't be too difficult (probably half a dozen command-line invocations), it still requires quite some build time.
 
-To avoid the overhead of rebuilding everything a different set of binary packages could be provided.
+To avoid the need to build from source a set of binary packages could be provided which chose the middleware implementation at build time.
```
mikaelarguedas, line 179: are provided ? looks like that's what we ship in our binaries already

articles/060_ros_middleware_interface.md
```diff
@@ -170,44 +171,19 @@ However the performance using the DynamicData API will likely always be lower co
...
+#### Decide at runtime
 
-#### Generate "fat" binary packages
-
-Another way to enable the user to switch between middleware implementations without the need to use the DynamicData API is to embed the middleware specific generated code for all supported implementations into each binary package.
-The specific middleware implementation would then be selected, for example, at link time and only the corresponding generated code of that middleware implementation would be used.
-
-### Using single middleware implementation only
+The alternative is to support selecting a specific middleware implementation at runtime.
+This will require that the to be selected middleware implementation was available at compile time in order for the middleware specific type support to be generated for every message package.
```
mikaelarguedas: this requires ? that's how our current code base behaves right?

mikaelarguedas (commented):
I made a few comments, mostly phrasing but that I wanted you to asses rather than pushing them directly to this branch

articles/060_ros_middleware_interface.md
dirk-thomas: While that might be the case right now we are aiming to provide binaries for multiple RMW implementations in the future. Therefore I would claim (and promise) here that we do.

articles/060_ros_middleware_interface.md
dirk-thomas: Indeed. Fixed in 104e93ba9da14e04dba2e60ad4ea96a5e5ba09d0.

articles/060_ros_middleware_interface.md
dirk-thomas: > rosidl_generator_dds_idl is used by all rmw implementations based on static code (e.g. opensplice, coredx) and not only Connext

Fixed in 15b76d52b004f2d4e1f804dee743b605c639ce10 by moving it above the vendor specific bullets.

> not sure if we should flag non supported implementations one way or another because we could expect that rmw_connext_dynamic and rmw_opensplice are at feature parity and receive the same amount of effort as the other ones

Each repository should already mention that. I don't see a need to duplicate that information here (and require it to be kept up-to-date in the future).

articles/060_ros_middleware_interface.md
dirk-thomas: Added in 9ada2feede26253276287e533020ad1659a9695f.

articles/060_ros_middleware_interface.md
```diff
@@ -321,38 +297,50 @@ In order to ensure that these information are passed back to the same middleware
...
-- the package [rmw](https://github.com/ros2/rmw/tree/master/rmw) defines the middleware interface
+- The package [rmw](https://github.com/ros2/rmw/tree/master/rmw) defines the middleware interface.
 
-  - the functions are declared in [rms/rmw.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h)
-  - the handles are defined in [rmw/types.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h)
+  - The functions are declared in [rms/rmw.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/rmw.h).
+  - The handles are defined in [rmw/types.h](https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h).
 
-- the package [rmw_connext_cpp](https://github.com/ros2/rmw_connext/tree/master/rmw_connext_cpp) implements the middleware interface using [RTI Connext DDS](http://www.rti.com/products/dds/index.html) based on statically generated code
+- The package [rmw_fastrtps_cpp](https://github.com/ros2/rmw_fastrtps/tree/master/rmw_fastrtps_cpp) implements the middleware interface using [eProsima Fast-RTPS](http://www.eprosima.com/index.php/products-all/eprosima-fast-rtps) based on the introspection type support.
```
mikaelarguedas: This is the first time the term introspection is being used. Maybe moving out `rosidl_typesupport_introspection_cpp` from the connext dynamic sub-section would make more sense given that it's used by this package as well

dirk-thomas: Done in e3f9be6c2075351b3ebd7f6c162744edb08a3dd0.

articles/060_ros_middleware_interface.md
mikaelarguedas: good point, let's leave it as is then

articles/060_ros_middleware_interface.md
```diff
@@ -170,44 +171,19 @@ However the performance using the DynamicData API will likely always be lower co
...
 
-#### Generate "fat" binary packages
+#### Decide at runtime
 
-Another way to enable the user to switch between middleware implementations without the need to use the DynamicData API is to embed the middleware specific generated code for all supported implementations into each binary package.
-The specific middleware implementation would then be selected, for example, at link time and only the corresponding generated code of that middleware implementation would be used.
-
-### Using single middleware implementation only
+The alternative is to support selecting a specific middleware implementation at runtime.
+This requires that the to be selected middleware implementation was available at compile time in order for the middleware specific type support to be generated for every message package.
```
mikaelarguedas, line 186: not sure that sentence is valid english, but we would need a native speaker to infirm/confirm it

mikaelarguedas (approved):
thanks for iterating, looks good to me now. except one question that I can't really answer myself

articles/060_ros_middleware_interface.md
dirk-thomas: I will go ahead and merge this. Please feel free to make a follow up PR if the sentence (or anything else) needs refinement / fixing.

PR #139, opened 2017-08-28
add article about coexistence with and migration from ROS 1

dirk-thomas:
While not complete yet this draft describes various aspects relevant for coexistence and migration.

Any feedback is appreciated.

### DIFF
--- a/articles/200_coexistence_and_migration.md
+++ b/articles/200_coexistence_and_migration.md
@@ -103,12 +103,12 @@ Since it is not desired to change existing `catkin` / `ament_cmake` packages to 
 ### CMake version
 
 ROS 2 requires CMake version 3.5+.
-It should be easily possible to update existing CMake code to be compatible with various versions of CMake.
+It should be easily possible to update existing ROS 1 CMake code to be compatible with various versions of CMake.
 
 ### Python version
 
 ROS 2 requires Python version 3.5+.
-It should be easily possible to update existing Python code to be compatible with Python 2.7 as well as Python 3.5+.
+It should be easily possible to update existing ROS 1 Python code to be compatible with Python 2.7 as well as Python 3.5+.
 
 ### CMake API
 
--- a/articles/200_coexistence_and_migration.md
+++ b/articles/200_coexistence_and_migration.md
@@ -77,7 +77,7 @@ If this problem would be address within `rosdep` to consider the Python version 
 This would also allow building / using / testing ROS 1 with Python 3.
 
 Another example is the mapping of `rosdep` key to the Qt 4 or Qt 5 version of a dependency.
-A package may try to avoid changing the rosdep key used for their dependencies but rosdep only considers the platform without the context of the ROS distribution.
+Package try to avoid changing the rosdep key used for their dependencies but rosdep only considers the platform without the context of the ROS distribution.
 
 **Proposal:** Change request for `rosdep` to consider information beyond the platform to determine the mapping.
 
@@ -119,6 +119,8 @@ The rational for the changes and the different name are described in [this artic
 **Idea:** Create a ROS 2 package names `catkin` which provides the same CMake API as the ROS 1 package (minus unavailable features like the devel space) by mapping the calls to `ament_cmake` API.
 In addition it is expected that existing packages need to call a new function at the end of their CMake code to account for the fact that `ament_package` needs to be invoked *after* all targets where `catkin_package` needs to be invoked *before* them.
 
+In addition to `catkin` the CMake of ROS 1 packages also uses CMake API of other package, e.g. `genmsg`, `dynamic_reconfigure`, `rostest`, etc.
+
 ### Message / service definitions
 
 In ROS 2 message / service definitions are expected to be in the subdirectory `msg` / `srv`.
@@ -147,7 +149,7 @@ At the same time updating the subdirectory name of the C++ headers would make se
 
 For backward compatibility the same API as for ROS 1 could be generated into a separate location.
 The user could opt-in to use these to reduce the migration effort but by sacrificing the dual-home support.
-In order to use those ROS 1 shim APIs the user would have to add another path to the include directories / `PYTHONPATH`.
+In order to use those ROS 1 shim APIs the user would to add another path to the include directories / `PYTHONPATH`.
 
 **Idea:** Implement these additional message generators and provide a mechanism to easily use those interfaces.
 
@@ -210,7 +212,7 @@ If yes, these two variations of the packages can be managed in different ways:
   * (-) The separation makes the packages appear less correlated.
   * (+) Changes between the two branches can be easily ported both ways.
   * (+) More obvious for users (assuming there is comparable infrastructure like distributions file, wiki, etc.).
-  * (+) It might be unclear which version reported issues are referring to.
+  * (+) Reported issues are clearly related to a specific ROS version.
 
 * Both versions are stored in separate unrelated repositories.
 
@@ -224,7 +226,7 @@ In any case it is important to try getting the current maintainers involved in t
 ### API shims
 
 As described in the above section about the [CMake API](#cmake-api) it is possible to create ROS 2 packages mimicking the same name and API as existing ROS 1 packages.
-Depending on how different the ROS 1 interface is compared to the ROS 2 interface to which it should be mapped, that becomes more or less feasible.
+Depending how different the ROS 1 interface is compared to the ROS 2 interface it should be mapped to that becomes more or less feasible.
 In the case of `catkin` and `ament_cmake` the functionality is very close and therefore an API shim is realistic.
 But even in that case some changes might be necessary as mentioned for a new required call at the end of the CMake code.
 
@@ -244,4 +246,4 @@ For each case it needs to be carefully evaluated if the effort necessary to impl
 * Apply ROS 2 specific patches.
 
 * In the future port relevant fixes from either version to the other.
-  The more significant the two branches / forks diverge the more effort / less likely these ports will become.
+  The more significant the two branches/ forks diverge the more effort / less likely these ports will become.
--- a/articles/200_coexistence_and_migration.md
+++ b/articles/200_coexistence_and_migration.md
@@ -1,0 +1,247 @@
+---
+layout: default
+title: Coexistence with and Migration from ROS 1
+permalink: articles/coexistence_and_migration.html
+abstract:
+  This article summarizes how the two major versions of ROS can coexist and be used together.
+  Additionally it covers the different aspects of migrating ROS packages from ROS 1 to ROS 2.
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+published: true
+categories: Overview
+---
+
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Goals
+
+### Separately usable
+
+* One version should not be affected by other versions choices / drawbacks.
+
+* ROS 1 users shouldn't be worried about ROS 2 development potentially negatively affecting ROS 1.
+
+  * No changes to existing ROS 1 packages should be necessary to continue to work.
+
+* ROS 2 users shouldn't need to make compromises due to ROS 1 choices / limitations / design decisions.
+
+### Support dual-homed applications
+
+To support creating dual-homed applications like the [ros1_bridge](https://github.com/ros2/ros1_bridge) ROS 1 and ROS 2 must be usable within a single application.
+
+### Minimize migration effort
+
+To ease the migration path the goal is to minimize the effort necessary where possible and when this goal doesn't conflict with the previously mentioned goals.
+
+## Technical Constraints
+
+### Namespace of ROS packages
+
+The first and very important question is about the namespace of ROS packages.
+Should ROS 1 and ROS 2 packages use separate namespaces or should they share the same namespace?
+For example can there be two packages names `foo` - one in ROS 1 and the other in ROS2 or do they have to be named differently.
+
+In the case that they use separate namespaces that does not prevent the release of a package `bar` into both, ROS 1 as well as ROS 2.
+
+Having them share a single namespace makes it difficult for users to distinguish them.
+
+Using separate namespaces allows e.g. to release a package `foo` in ROS 2 which provides the same API (but different implementation) as the ROS 1 package with the same name.
+This way downstream packages don't need to change when switching between the ROS 1 and ROS 2 version of the package.
+
+**Decision:** Use separate namespaces for packages in ROS 1 and ROS 2 (this is currently already the case).
+
+### Namespace of rosdep keys
+
+The next question is very similar and is about the namespace of the `rosdep` keys.
+Should ROS 1 and ROS 2 use a different set of `rosdep` rules or should they share the same database?
+
+#### Same key vs. different key
+
+If a ROS package declares a dependency on the `rosdep` key `python-foo` that maps on e.g. Ubuntu Xenial to the Python 2 package `foo`.
+In order to work for ROS 2 it needs to depend on the Python 3 version of the package instead.
+
+Currently the mapping in `rosdep` only depends on the platform and doesn't consider which ROS distribution is being used.
+
+If the migration to ROS 1 requires the key to be changed to `python3-foo` that implies that any package which depends on Python packages does need a modification in the manifest.
+
+The same conceptional problems already exist in ROS 1.
+When being built on platforms which use Python 3 they already map the keys to Python 3 packages.
+If this problem would be address within `rosdep` to consider the Python version as part of the platform that would allow Python 2 as well as Python 3 packages to be resolved using the same `rosdep` key.
+This would also allow building / using / testing ROS 1 with Python 3.
+
+Another example is the mapping of `rosdep` key to the Qt 4 or Qt 5 version of a dependency.
+Package try to avoid changing the rosdep key used for their dependencies but rosdep only considers the platform without the context of the ROS distribution.
+
+**Proposal:** Change request for `rosdep` to consider information beyond the platform to determine the mapping.
+
+**Decision:** With that change the `rosdep` rules should share a single namespace (this is currently already the case).
+
+### Package manifest
+
+In ROS 1 two versions of the package manifest are supported: format 1 specified in [REP 127](http://www.ros.org/reps/rep-0127.html) and format 2 specified in [REP 140](http://www.ros.org/reps/rep-0140.html).
+While ROS 2 only supports format 2 there is no technical limitation behind that design decision.
+It simply reduces the effort to distinguish the different semantics of some depend tags in various parts of the software.
+
+It might be the case that a to-be-specified format 3 will be created to add support for the desired features to describe [dependency groups]() as well as declaring an [ABI version]().
+
+#### ROS dependencies
+
+While a ROS 1 package might depend on `roscpp` and `message_generation` a ROS 2 package depends on `rclcpp` and `rosidl_message_generation` instead.
+
+#### Build type
+
+Almost all build types for ROS 1 and ROS 2 are disjoint except the build type `cmake` which can be processed by both build tools.
+Since it is not desired to change existing `catkin` / `ament_cmake` packages to use the build type `cmake` those are currently different for each ROS version.
+
+### CMake version
+
+ROS 2 requires CMake version 3.5+.
+It should be easily possible to update existing CMake code to be compatible with various versions of CMake.
+
+### Python version
+
+ROS 2 requires Python version 3.5+.
+It should be easily possible to update existing Python code to be compatible with Python 2.7 as well as Python 3.5+.
+
+### CMake API
+
+ROS 1 packages use the CMake API provided by [catkin](https://github.com/ros/catkin).
+ROS 2 packages use a slightly different CMake API provided by [ament_cmake](https://github.com/ament/ament_cmake).
+The rational for the changes and the different name are described in [this article](http://design.ros2.org/articles/ament.html).
+
+**Idea:** Create a ROS 2 package names `catkin` which provides the same CMake API as the ROS 1 package (minus unavailable features like the devel space) by mapping the calls to `ament_cmake` API.
+In addition it is expected that existing packages need to call a new function at the end of their CMake code to account for the fact that `ament_package` needs to be invoked *after* all targets where `catkin_package` needs to be invoked *before* them.
+
+### Message / service definitions
+
+In ROS 2 message / service definitions are expected to be in the subdirectory `msg` / `srv`.
+This implicitly ensures that message and service names are unique.
+There is no technical reason behind that design decision rather than simplifying the implementation.
+
+**Idea:** Allow arbitrary locations in the source space.
+
+Currently ROS 2 already supports additional feature in these files (e.g. default values) which are not available in ROS 1.
+In the future ROS 2 might support a different file format to satisfy the requested feature requests (like grouping constants in enums, optional fields, file / constant / field comments, etc.).
+If that is the case a backward compatible parser to read existing `.msg` / `.srv` files will likely be available.
+
+### Generated message / service API
+
+In order to satisfy the goal of dual-homed applications the generated code of ROS 1 and ROS 2 need to be side-by-side usable.
+
+For C++ that is already the case.
+In ROS 1 the header files are generated in namespace specific to the package name.
+In ROS 2 the header files are placed into a sub-namespace names `msg` / `srv`.
+Beside the different path / namespace the data structure also uses different types for e.g. the shared pointers (Boost vs. C++11).
+
+For Python that is not the case.
+
+**Proposal:** Change the location of the generated Python modules from using `msg` / `srv` as the subdirectory to `message` / `service`.
+At the same time updating the subdirectory name of the C++ headers would make sense to be similar.
+
+For backward compatibility the same API as for ROS 1 could be generated into a separate location.
+The user could opt-in to use these to reduce the migration effort but by sacrificing the dual-home support.
+In order to use those ROS 1 shim APIs the user would to add another path to the include directories / `PYTHONPATH`.
+
+**Idea:** Implement these additional message generators and provide a mechanism to easily use those interfaces.
+
+## Different categories of changes
+
+For the migration two different categories of changes need to be considered separately:
+
+* Changes which can be applied to ROS 1 packages to either make it compatible with ROS 2 or minimize the necessary changes.
+* Changes which are necessary to make the package work for ROS 2 but which are not compatible with ROS 1.
+
+If the second set of changes would be empty a package could be used as a ROS 1 package as well as ROS 2 packages using the same sources.
+Currently the previous sections mention a few different items which fall into the second category therefore it doesn't seem realistic to achieve compatibility using the exact same sources.
+The following sections assume that is not feasible to maintain a single source for the ROS 1 as well as the ROS 2 version of a package.
+
+### Changes applicable to ROS 1 packages
+
+* Update the package manifest to format 2.
+
+* In C++ introduce typedefs for all message / service classes as well as their pointer types.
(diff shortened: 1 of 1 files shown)


articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+Should ROS 1 and ROS 2 use a different set of `rosdep` rules or should they share the same database?
+
+#### Same key vs. different key
+
+If a ROS package declares a dependency on the `rosdep` key `python-foo` that maps on e.g. Ubuntu Xenial to the Python 2 package `foo`.
+In order to work for ROS 2 it needs to depend on the Python 3 version of the package instead.
+
+Currently the mapping in `rosdep` only depends on the platform and doesn't consider which ROS distribution is being used.
+
+If the migration to ROS 1 requires the key to be changed to `python3-foo` that implies that any package which depends on Python packages does need a modification in the manifest.
```
tkruse: "migration to ROS *2*". And the elimination of the possibility for ROS1 to ever migrate to Python3 should be stated if necessary for this document.

tkruse: "migration to ROS *2*"
And if you want to eliminate the possibility of ever moving ROS1 to Python3, that should be stated.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+**Idea:** Create a ROS 2 package names `catkin` which provides the same CMake API as the ROS 1 package (minus unavailable features like the devel space) by mapping the calls to `ament_cmake` API.
+In addition it is expected that existing packages need to call a new function at the end of their CMake code to account for the fact that `ament_package` needs to be invoked *after* all targets where `catkin_package` needs to be invoked *before* them.
+
+### Message / service definitions
+
+In ROS 2 message / service definitions are expected to be in the subdirectory `msg` / `srv`.
+This implicitly ensures that message and service names are unique.
+There is no technical reason behind that design decision rather than simplifying the implementation.
+
+**Idea:** Allow arbitrary locations in the source space.
```
tkruse, line 130: I see no mention of a benefit.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+In order to satisfy the goal of dual-homed applications the generated code of ROS 1 and ROS 2 need to be side-by-side usable.
+
+For C++ that is already the case.
+In ROS 1 the header files are generated in namespace specific to the package name.
+In ROS 2 the header files are placed into a sub-namespace names `msg` / `srv`.
+Beside the different path / namespace the data structure also uses different types for e.g. the shared pointers (Boost vs. C++11).
+
+For Python that is not the case.
+
+**Proposal:** Change the location of the generated Python modules from using `msg` / `srv` as the subdirectory to `message` / `service`.
```
tkruse: I see no mention of a benefit for the rename.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+### CMake API
+
+ROS 1 packages use the CMake API provided by [catkin](https://github.com/ros/catkin).
+ROS 2 packages use a slightly different CMake API provided by [ament_cmake](https://github.com/ament/ament_cmake).
+The rational for the changes and the different name are described in [this article](http://design.ros2.org/articles/ament.html).
+
+**Idea:** Create a ROS 2 package names `catkin` which provides the same CMake API as the ROS 1 package (minus unavailable features like the devel space) by mapping the calls to `ament_cmake` API.
+In addition it is expected that existing packages need to call a new function at the end of their CMake code to account for the fact that `ament_package` needs to be invoked *after* all targets where `catkin_package` needs to be invoked *before* them.
+
+### Message / service definitions
```
tkruse: maybe should be level 2 header?

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+## Different categories of changes
+
+For the migration two different categories of changes need to be considered separately:
+
+* Changes which can be applied to ROS 1 packages to either make it compatible with ROS 2 or minimize the necessary changes.
+* Changes which are necessary to make the package work for ROS 2 but which are not compatible with ROS 1.
+
+If the second set of changes would be empty a package could be used as a ROS 1 package as well as ROS 2 packages using the same sources.
+Currently the previous sections mention a few different items which fall into the second category therefore it doesn't seem realistic to achieve compatibility using the exact same sources.
+The following sections assume that is not feasible to maintain a single source for the ROS 1 as well as the ROS 2 version of a package.
```
tkruse: "that *it* is not"

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+* One of the versions is stored in a forked repository.
+
+  * (-) The separation makes the packages appear less correlated.
+  * (+) Changes between the two branches can be easily ported both ways.
+  * (+) More obvious for users (assuming there is comparable infrastructure like distributions file, wiki, etc.).
+  * (+) It might be unclear which version reported issues are referring to.
+
+* Both versions are stored in separate unrelated repositories.
+
+  * No obvious advantages over the other twp approaches.
```
tkruse: typo: two

tkruse:
my comments can be deleted without further discussion (with or without fix).

articles/200_coexistence_and_migration.md
dirk-thomas: Thanks. Fixed in c908ff0a18d10be6b5d979a37684a92c4eaf1966.

Regarding Python 3 support for ROS 1: this document only aims for ROS 2 specific topic. Python 3 support in ROS 1 is out of scope for this article and it doesn't try to make any recommendation for it or "elimination the possibility" in any way. If in order to support the use case of ROS 2 rosdep keys we can make improvements to rosdep itself to make it easier to use Python 3 with ROS 1 that would be a great side effect.

dirk-thomas: This looks like a duplicate of https://github.com/ros2/design/pull/139#discussion_r135758429 ?

articles/200_coexistence_and_migration.md
dirk-thomas: While arbitrary locations are possible in ROS 1 in ROS 2 we simplified the implementation by just allowing one location. It has been requested before to make this more flexible again (ros2/rosidl#213).

articles/200_coexistence_and_migration.md
dirk-thomas: The benefit would be the ability to support dual-homed applications in Python (as mentioned a few lines above). Currently the generated Python files of ROS 1 and ROS 2 use the exact same path and can therefore not be used at the same time within a single application.

articles/200_coexistence_and_migration.md
dirk-thomas: The intention was to have this subsection under `Technical Constraints`

articles/200_coexistence_and_migration.md
dirk-thomas: Thanks. Fixed in c908ff0a18d10be6b5d979a37684a92c4eaf1966.

articles/200_coexistence_and_migration.md
dirk-thomas: Thanks. Fixed in c908ff0a18d10be6b5d979a37684a92c4eaf1966.

dirk-thomas:
> my comments can be deleted without further discussion (with or without fix).

Thank you for your feedback. I have addressed the mistakes you pointed out and replied to all your comments inline.

articles/200_coexistence_and_migration.md
sloretz: Keeping the restriction on message and service locations may make it easier for new users to learn ROS 2 because it means there can be a tutorial with a canonical ROS2 package layout.

Even though it's not the case, it feels like the same restrictions are in ROS 1. `genmsg` `add_message_files` expects messages to be in `msg` by default. It allows changing the directory, but I don't think that's easy to discover. For a long time I thought this was a restriction and thought the beginner tutorials should say the messages must be placed there.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+For example can there be two packages names `foo` - one in ROS 1 and the other in ROS2 or do they have to be named differently.
+
+In the case that they use separate namespaces that does not prevent the release of a package `bar` into both, ROS 1 as well as ROS 2.
+
+Having them share a single namespace makes it difficult for users to distinguish them.
+
+Using separate namespaces allows e.g. to release a package `foo` in ROS 2 which provides the same API (but different implementation) as the ROS 1 package with the same name.
+This way downstream packages don't need to change when switching between the ROS 1 and ROS 2 version of the package.
+
+**Decision:** Use separate namespaces for packages in ROS 1 and ROS 2 (this is currently already the case).
```
sloretz, line 58: I'm a little confused by the term namespace. Does it mean there is some tool where `namespace`*delimiter*`package_name` is used? Can I do `ros2 pkg ros2/foo` and `ros2 pkg ros1/foo`?

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+It might be the case that a to-be-specified format 3 will be created to add support for the desired features to describe [dependency groups]() as well as declaring an [ABI version]().
+
+#### ROS dependencies
+
+While a ROS 1 package might depend on `roscpp` and `message_generation` a ROS 2 package depends on `rclcpp` and `rosidl_message_generation` instead.
+
+#### Build type
+
+Almost all build types for ROS 1 and ROS 2 are disjoint except the build type `cmake` which can be processed by both build tools.
+Since it is not desired to change existing `catkin` / `ament_cmake` packages to use the build type `cmake` those are currently different for each ROS version.
```
sloretz, line 101: Thinking out loud here. Say I wanted to use the same code for ROS 1 and ROS 2, and assume the package.xml format was extended to support something like

```
<oneof>
  <buildtool_depend>ament_cmake</buildtool_depend>
  <buildtool_depend>catkin</buildtool_depend>
</oneof>
```

And then the CMakeLists.txt used
```
if(COMMAND catkin_package)
  ...
else()
  ...
endif()
```

What other obstacles would there be?

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+It should be easily possible to update existing CMake code to be compatible with various versions of CMake.
+
+### Python version
+
+ROS 2 requires Python version 3.5+.
+It should be easily possible to update existing Python code to be compatible with Python 2.7 as well as Python 3.5+.
+
+### CMake API
+
+ROS 1 packages use the CMake API provided by [catkin](https://github.com/ros/catkin).
```
sloretz: Is it also worth mentioning `genmsg` since it's cmake macros are so widely used?

articles/200_coexistence_and_migration.md
dirk-thomas: I think the ROS 1 API is reasonable for this: use a sane default which is recommended in tutorials but allow the user to provide a custom location if necessary.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+In order to satisfy the goal of dual-homed applications the generated code of ROS 1 and ROS 2 need to be side-by-side usable.
+
+For C++ that is already the case.
+In ROS 1 the header files are generated in namespace specific to the package name.
+In ROS 2 the header files are placed into a sub-namespace names `msg` / `srv`.
+Beside the different path / namespace the data structure also uses different types for e.g. the shared pointers (Boost vs. C++11).
+
+For Python that is not the case.
+
+**Proposal:** Change the location of the generated Python modules from using `msg` / `srv` as the subdirectory to `message` / `service`.
```
sloretz: This sounds fine to me.

I remember there were some comments a while back to the effect of changing the include paths increases the migration effort. It may be possible to create a ROS2 package that uses `importlib` to import only from ROS 1 packages. This would mean import statements would be unchanged when porting from ros1 to ros2, but dual homed applications could use something extra to get ROS 1 python packages.

Dual homed code might look something like

```
from my_pkg.msg import MyMsg as MyMsgRos2
MyMsgRos1 = ros1_bridge.import('my_pkg.msg', 'MyMsg')
```

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,247 @@
...
+  * (+) Changes between the two branches can be easily ported both ways.
+  * (-) Likely less discoverable for users.
+  * (-) It might be unclear which version reported issues are referring to.
+
+* One of the versions is stored in a forked repository.
+
+  * (-) The separation makes the packages appear less correlated.
+  * (+) Changes between the two branches can be easily ported both ways.
+  * (+) More obvious for users (assuming there is comparable infrastructure like distributions file, wiki, etc.).
+  * (+) It might be unclear which version reported issues are referring to.
```
sloretz: Nitpick: Is this supposed to be a negative effect? `(-)`

sloretz (commented):
Is this document intended to be used as a guide by developers who are porting from ROS 1 to ROS 2?

articles/200_coexistence_and_migration.md
mikaelarguedas: I think in this case we're referring to the debian package namespace. for example ros2 packages are released under `ros-r2b2-...` and ros1 package released under `ros-lunar-...`
This way if a package depends on `joy` and a user sourced a ROS2 workspace, he/she will get the headers from `ros-r2b2-joy`,  if he/she sourced a Lunar workspace he/she will get them from `ros-lunar-joy` without having to change the name of your dependencies or the code (if API is consistent)

dirk-thomas: The term `namespace` isn't related to the Debian package names because that name is different from distro to distro.

If the ROS 1 and ROS 2 packages would share the same namespace that would imply that they are being listed in a single `distribution.yaml` file in the `rosdistro` repo.

Currently we have a separate `rosdistro` repo so they have "separate namespace". The same would be the case if the ROS 2  packages are always listed in a separate distro (like `r2b2`) even when they are stored in the same repo.

mikaelarguedas: :+1: right thanks for clarifying

articles/200_coexistence_and_migration.md
dirk-thomas: Those are listed in this article:

* the rosdep keys for the non-ROS dependencies
* ROS dependencies like `roscpp` and `message_generation`
* message definitions must only use the ROS 1 feature set
* the include paths for message headers
* the includes and types of API which is different in ROS 1 and ROS 2
* the code base using Boost vs. C++11 types

How is `COMMAND catkin_package` going to work? What happens before? How does a package know if it should build for ROS 1 or ROS 2?

articles/200_coexistence_and_migration.md
dirk-thomas: Absolutely: a48d1b6c21b09576bcfeede75bd77ead6775db2f

articles/200_coexistence_and_migration.md
dirk-thomas: Oh, this should actually say:

> Reported issues are clearly related to a specific ROS version.

Fixed in 979051ea29bc6528119d108288d33ad77f157871

wjwwood:
Looks like my commit may have undone some things, not sure how that happened. Let me try to fix it.

articles/200_coexistence_and_migration.md
sloretz: This isn't something I think should be done, I'm just trying to understand why it can't be done. Feel free to ignore this.

I was assuming the developer would have either sourced the ros2 workspace and used ament, or sourced a ros1 workspace and used catkin.

On those bullets It seems like the package.xml dependencies being different for ROS1 and ROS2 is really the only thing that can't be worked around. C++ API incompatibilities can be preprocessor macros and `configure_file`. Message definitions being restricted is inconvenient, but since ROS1 .msg files can be anywhere a package could have a ros1 message directory and use `msg/` for ros2 messages

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+### CMake version
+
+ROS 2 requires CMake version 3.5+.
+It should be easily possible to update existing CMake code to be compatible with various versions of CMake.
+
+### Python version
+
+ROS 2 requires Python version 3.5+.
+It should be easily possible to update existing Python code to be compatible with Python 2.7 as well as Python 3.5+.
```
nuclearsandwich: Is this proposing backporting all ROS 2 packages to support both python 2.7. I have only limited and very recent experience doing this with code that started life in Python 3, but it's a hefty amount of legwork. 

Off the top of my head

- Strings vs unicodes vs bytes don't appear to have a satisfying mutual API which means there will be a scattering of conditionals or wrapper functions required to support this.
- A bunch of changes were just made to ROS 2 code dropping things that were required in python 2 and optional in python 3. That work would all need to be reverted and future code would need to be written against 2/3 mutual python.

- How much did the Python C API change between Py2 and Py3? How much work would be involved in making rclpy py2 compatible?

All of this is doable but it's a significant amount of effort when Python 2.7 has a planned EOL in 2020. Even if the date gets extended, it seems like effort would be better spent determining the path forward for ROS 1 Python rather than backporting ROS 2 Python.

I guess the core question is: What is the benefit of making ROS 2 Python 2 compatible and is it worth the effort when it expires after two years?

dirk-thomas: > What is the benefit of making ROS 2 Python 2 compatible and is it worth the effort when it expires after two years?

ROS 2 is only supporting Python 3.5+ and won't support older versions. The sentence was referring to ROS 1 which currently only support Python 2.7 but could be updated to support both.

dirk-thomas: Clarified in ffc1b26f1098a948f882a3118d7b8426b7202b14.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,249 @@
...
+If a ROS package declares a dependency on the `rosdep` key `python-foo` that maps on e.g. Ubuntu Xenial to the Python 2 package `foo`.
+In order to work for ROS 2 it needs to depend on the Python 3 version of the package instead.
+
+Currently the mapping in `rosdep` only depends on the platform and doesn't consider which ROS distribution is being used.
+
+If the migration to ROS 2 requires the key to be changed to `python3-foo` that implies that any package which depends on Python packages does need a modification in the manifest.
+
+The same conceptional problems already exist in ROS 1.
+When being built on platforms which use Python 3 they already map the keys to Python 3 packages.
+If this problem would be address within `rosdep` to consider the Python version as part of the platform that would allow Python 2 as well as Python 3 packages to be resolved using the same `rosdep` key.
```
dhood, line 76: Is what you're saying that it would be all-or-nothing python3? Or would I still be able to specify explicitly if I want my ROS 1 package to depend on the python2 or python3 version of a dependency? If not, choosing it automatically would break a lot of ROS 1 packages that don't support python3, right? that seems to conflict with the goal of not requiring changes to existing ROS 1 packages

articles/200_coexistence_and_migration.md
dhood: I also remember seeing that people are bothered by the change in C++ header namespaces, and we explained that it was a requirement for the dual-homed applications. However, in [that conversation](https://github.com/ros2/design/pull/50) I didn't see an explanation of why it would be technically necessary in python. If it's true that it could be worked around in python (perhaps by what @sloretz has suggested, or something else), it seems that this proposal conflicts with the 'minimise migration effort' goal, and the namespaces should be left as is in python

articles/200_coexistence_and_migration.md
nuclearsandwich: Okay. 😌 phew. 👍

articles/200_coexistence_and_migration.md
mikaelarguedas: Currently in rosdep you can specify that you want the python3 or the python2 version using the same key (defining an environment variable on your system). We don't have many rosdep keys providing the python3 version, mostly because it's cumbersome to define them. We discussed with @tfoote recently ways to make this more convenient (https://github.com/ros-infrastructure/rosdep/issues/547).
One of the suggested ideas is to allow rosdep rules like:
```
python-foo:
  ubuntu: [python-foo]
  ubuntu_python3: [python3-foo]
```

the current system forces you to define:
```
python-foo:
  ubuntu:
    trusty: [python-foo]
    trusty_python3: [python3-foo]
    xenial: [python-foo]
    xenial_python3: [python3-foo]
    zesty: [python-foo]
    zesty_python3: [python3-foo]
    ...
```

This way there is no need to change the package.xml because they use the same key `python-foo`
While currently for ros2 we need another key for each package
```
python3-foo:
  ubuntu: [python3-foo]
```
> If this problem would be address within `rosdep` to consider the Python version as part of the platform

Maybe you can clarify @dirk-thomas what you mean by this exactly.
 do you mean "`arch` should always map to python3 and `ubuntu` should always map to python2"
or "allow user to define on their platform that they want to map to the python3 version of the packages and not the python2 one"

> Or would I still be able to specify explicitly if I want my ROS 1 package to depend on the python2 or python3 version of a dependency? 

My reading of this is that you won't be able to specify this in the `package.xml` and that will be configured globally as a trade-of between not changing your package.xml and python3 users able to use python 3 dependencies for their for packages.
I think it's a valid point being raised. How does that work for people using dual-home systems ? if they "have to" define that they want to use python3 for their platform (because using ROS2), will they be able to use ROS1 packages that are compatible only with the python2 version of their dependencies?

If that's the case, how can we mitigate it ?
- by inciting people to either be compatible with both python2 and python3 ?
- by asking them to create python2 only keys and use those if they dont want to upgrade their package?
- something else?

articles/200_coexistence_and_migration.md
dirk-thomas: > Dual homed code might look something like

```
from my_pkg.msg import MyMsg as MyMsgRos2
MyMsgRos1 = ros1_bridge.import('my_pkg.msg', 'MyMsg')
```

@sloretz This snippet would not work since the imported file identified by the arguments `my_pkg.msg` and `MyMsg` might also call import and would therefore resolve to the ROS 2 version of the dependent modules (assuming the message generation hasn't been altered).

sloretz: @dirk-thomas How about registering/unregistering an [import hook](https://www.python.org/dev/peps/pep-0302/) using a `with` statement?

```
from my_pkg.msg import MyMsg as MyMsgRos2
with ros1_bridge.Ros1Import:
    from my_pkg.msg import MyMsg as MyMsgRos1
```

articles/200_coexistence_and_migration.md
dirk-thomas: The question is "how could rosdep know to resolve Python 2 packages in ROS 1 and Python 3 packages in ROS 2 for the same key"? If the user provides a custom configuration like `xenial_python3` that is already possible but for the ROS 2 case it is desired that it is not necessary to do so.

The Debian packages created for a single ROS distribution have to agree on a Python version. Currently for ROS 1 that is Python 2. But it should be possible to create binary package like `ros-lunar3-*` which choose Python 3 instead. Just mixing isn't supported. Nothing prevents a user to bypass the default Python version and embed whatever is desired in e.g. a shebang line (as long as the line isn't rewritten by setuptools). Currently that is being achieved by having a `*_python3` mapping for some rosdep keys and the user choosing this explicitly.

For "dual-homed" application this is not relevant imo. The user will first install the ROS 1 dependencies in a ROS 1-only environment, then does the same in a ROS 2-only environment. For afterwards running the dual-homed application it won't interact with rosdep anymore.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+* Update the dependencies in the manifest.
+
+* Update the CMake code to find the updated dependencies.
+
+* The API of the message / service classes has changed:
+
+  * Different headers / modules to include / import.
+  * Different classes / methods / types.
+  * C++11 interface using e.g. `std::shared_ptr` rather then relying on `Boost`.
```
Karsten1987, line 186: I would like to see a little note about Windows compatibility, i.e. exporting headers and/or compiling into shared libraries.

articles/200_coexistence_and_migration.md
```diff
@@ -0,0 +1,249 @@
...
+It should be easily possible to update existing ROS 1 Python code to be compatible with Python 2.7 as well as Python 3.5+.
+
+### CMake API
+
+ROS 1 packages use the CMake API provided by [catkin](https://github.com/ros/catkin).
+ROS 2 packages use a slightly different CMake API provided by [ament_cmake](https://github.com/ament/ament_cmake).
+The rational for the changes and the different name are described in [this article](http://design.ros2.org/articles/ament.html).
+
+**Idea:** Create a ROS 2 package names `catkin` which provides the same CMake API as the ROS 1 package (minus unavailable features like the devel space) by mapping the calls to `ament_cmake` API.
+In addition it is expected that existing packages need to call a new function at the end of their CMake code to account for the fact that `ament_package` needs to be invoked *after* all targets where `catkin_package` needs to be invoked *before* them.
```
Karsten1987, line 120: can I build ROS1 packages with ament? If so, one possible step could be also to migrate to ament within the ROS1 scope.

articles/200_coexistence_and_migration.md
wjwwood: What would you like to see? I don't think this has anything to do with coexistence. Or at least boost vs. std doesn't make a difference here.

articles/200_coexistence_and_migration.md
wjwwood: Not at the moment, though I'm only guessing that by "build ROS 1 packages with ament" you mean build with the ament build tool.

articles/200_coexistence_and_migration.md
tkruse: I get it with your explanation. If it said: 'For python that is not the case, preventing packages from containing both ROS1 and ROS2 packages.' I would have understood the first time.

articles/200_coexistence_and_migration.md
Karsten1987: It's not related to coexistence, but certainly with migration and maintenance. It's not related to boost or std, I wanted to add another bullet point. Because if you want to migrate your existing ROS1 code to ROS2 and want to have it run on Windows, you'll have to export the symbols.

dirk-thomas: I would argue that this is not related to ROS 2. If you want your ROS 1 package to compile on Windows you already have to do the same. And if you migrate to ROS 2 but don't care about Windows support that is not an issue either. Therefore I would prefer not to mention it in this article.

articles/200_coexistence_and_migration.md
dirk-thomas: Clarified in 937363a1969adc8a257d06c3c4eb6fc82e555a7a.

dirk-thomas:
> Is this document intended to be used as a guide by developers who are porting from ROS 1 to ROS 2?

I would say no. We already have a wiki page with that content: https://github.com/ros2/ros2/wiki/Migration-Guide

articles/200_coexistence_and_migration.md
dirk-thomas: @sloretz An import hook might help with this but I haven't used it. Please consider providing an example which demonstrates loading of a ROS 1 as well as ROS 2 message in the same Python script, e.g. `geometry_msgs/PointStamped` to cover the case of fields being of other message types.

articles/200_coexistence_and_migration.md
gbiggs: I'm in favour of having a canonical package layout in ROS 2 that is taught in tutorials and highly recommended to be followed. It makes developing tools easier.

I don't think it needs to be absolutely enforced; having the ability to provide a custom location as @dirk-thomas said is sensible. I think it should be implied that putting them there is the best approach and you put them elsewhere at your own (hopefully informed) risk of someone's tool not working.

dirk-thomas:
Closing due to inactivity.

PR #144, opened 2017-09-06
Update the Time article

tfoote:
This is an update of the time article.

The update primarily inverts the logic so that the Clock which is pumped by a TimeSource (if active) provides all blocking calls etc. And the pure datatypes no longer have internal references to the TimeSource. 

Note that this is a WIP looking for comments on the structure not for merge at the moment.

### DIFF
--- a/articles/130_ros_time.md
+++ b/articles/130_ros_time.md
@@ -39,6 +39,16 @@ Slower than real time simulation is necessary for complicated systems where accu
 Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
 Additionally if the simulation is paused the system can also pause using the same mechanism.
 
+## Approach
+
+To provide a simplified time interface we will provide a ROS time and duration datatype.
+To query for the latest time a ROS Clock interface will be provided.
+A TimeSource can manage one or more Clock instances.
+
+
+## Clock
+
+
 ### Challenges in using abstracted time
 
 There are many algorithms for synchronization and they can typically achieve accuracies which are better than the latency of the network communications between devices on the network.
@@ -70,18 +80,19 @@ For convenience in these cases we will also provide the same API as above, but u
 
 `SystemTime` will be directly tied to the system clock.
 
+#### ROS Time
+
+The `ROSTime` will report the same as `SystemTime` when an ROS Time Source is not active.
+When the ROS time source is active `ROSTime` will return the latest value reported by the Time Source.
+`ROSTime` is considered active when the parameter `using_sim_time` is set on the node.
+
 #### Steady Time
 
 Example use cases for this include hardware drivers which are interacting with peripherals with hardware timeouts.
 
 In nodes which require the use of `SteadyTime` or `SystemTime` for interacting with hardware or other peripherals it is expected that they do a best effort to isolate any `SystemTime` or `SteadyTime` information inside their implementation and translate external interfaces to use the ROS time abstraction when communicating over the ROS network.
 
-#### ROS Time
-
-The `ROSTime` will report the same as `SystemTime` when an ROS Time Source is not active.
-When the ROS time source is active `ROSTime` will return the latest value reported by the Time Source.
-
-### ROS Time Sources
+### ROS Time Source
 
 #### Default Time Source
 
@@ -100,7 +111,7 @@ The developer has the opportunity to register callbacks with the handler to clea
 
 The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific.
 
-##### No Advanced Estimating Clock
+##### No Advanced Estimating Clock By Default
 
 There are more advanced techniques which could be included to attempt to estimate the propagation properties and extrapolate between time ticks.
 However all of these techniques will require making assumptions about the future behavior of the time abstraction.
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@ros2/team to give feedback

articles/130_ros_time.md
```diff
@@ -70,18 +81,19 @@ For convenience in these cases we will also provide the same API as above, but u
 
 `SystemTime` will be directly tied to the system clock.
 
+#### ROS Time
+
+The `ROSTime` will report the same as `SystemTime` when an ROS Time Source is not active.
+When the ROS time source is active `ROSTime` will return the latest value reported by the Time Source.
+`ROSTime` is considered active when the parameter `using_sim_time` is set on the node.
```
dirk-thomas, line 87: Compared to a central parameter in ROS 1 this will be more challenging to achieve when using a node-specific parameter. How would it be configured / ensured that all nodes use sim time?

articles/130_ros_time.md
```diff
@@ -39,6 +39,17 @@ Slower than real time simulation is necessary for complicated systems where accu
 Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
 Additionally if the simulation is paused the system can also pause using the same mechanism.
 
+
+## Approach
+
+To provide a simplified time interface we will provide a ROS time and duration datatype.
+To query for the latest time a ROS Clock interface will be provided.
+The Clock will be populated by one or more TimeSource mechanisms.
```
dirk-thomas: Not sure I understand the sentence. How can the clock be populated by more than one time source?

articles/130_ros_time.md
```diff
@@ -115,23 +127,38 @@ It is possible that the user may have access to an out of band time source which
...
+
+For the current implementation a `TimeSource` API will be defined such that it can be overridden in code.
+If in the future a common implementation is found that would be generally useful it could be extended to optionally dynamically select the alternative TimeSource via a parameter similar to enabling the simulated time.
 
 ## Implementation
 
 The `SystemTime`, `SteadyTime`, and `ROSTime` API's will be provided by each client library in an idiomatic way, but they may share a common implementation, e.g. provided by `rcl`.
 However, if a client library chooses to not use the shared implementation then it must implement the functionality itself.
 
+`SteadyTime` will be typed differently than the interchangable `SystemTime` and `ROSTime`.
```
dirk-thomas, line 139: Can you elaborate how these types will be different. Do they share a common base in order to use them interchangeably? Is the API each of them offers different?

articles/130_ros_time.md
```diff
@@ -115,23 +127,38 @@ It is possible that the user may have access to an out of band time source which
...
+
+The `Clock` will support a `sleep_for` function as well as a `sleep_until` method using a `Duration` or `Time` argument.
 The implementation will also provide a `Timer` object which will provide periodic callback functionality for all the abstractions.
 
+It will also support registering callback before and after a time jump.
+The first callback will be to allow proper preparations for a time jump.
+The latter will allow code to respond to the change in time and include the new time specifically as well as a quantification of the jump.
+
+Any API which is blocking will allow a set of flags to indicate the appropriate behavior in case of time jump.
+This will allow the user to choose to error immediately on a time jump or choose to ignore.
```
dirk-thomas, line 156: What is considered to be a time jump in this context? When the clock moves backward? Or also when it moves forward "by a lot" - what would that threshold be?

articles/130_ros_time.md
```diff
@@ -39,6 +39,17 @@ Slower than real time simulation is necessary for complicated systems where accu
 Often the simulation is the limiting factor for the system and as such the simulator can be a time source for faster or slower playback.
 Additionally if the simulation is paused the system can also pause using the same mechanism.
 
+
```
dirk-thomas: Nitpick: no double empty lines.

test_local_deploy.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-docker run --rm -v `pwd`:/srv/jekyll -i -t -p 127.0.0.1:4000:4000 jekyll/jekyll:pages
+docker run --rm -v `pwd`:/srv/jekyll -i -t -p 127.0.0.1:4000:4000 jekyll/jekyll:pages jekyll serve --watch
```
dirk-thomas: I would suggest to separate this change since it is unrelated to the time article.

articles/130_ros_time.md
tfoote: Yeah, it's supposed to be one to many the other way. I inverted the logic and used the old language here.

articles/130_ros_time.md
tfoote: I think likely we'll need a cannonical way to assert these sort of things. Whether it be a node that just stands by and listens for new nodes to come online and sets the parameter on them. As well as some syntax for launch files etc to say apply this parameter to "all" nodes launched.

articles/130_ros_time.md
tfoote: I plan to define a struct that will allow you to define what you consider a "jump" when you register the callback. Such that you will only get the callback when a jump exceeds your threshold.

Right now I see 3 dimensions. There's a backwards jump of size N, a forwards jump of size N, and a change in clock type (aka ROSTime is disable or enabled)

I would expect the default to trigger if the clock changes or there's a backwards jump of any size.

test_local_deploy.sh
tfoote: Yeah, I'll do that. #145

articles/130_ros_time.md
tfoote: This was from discussions with @wjwwood. It makes sense to have SystemTime and ROSTime have a common base class with runtime checks that they are correctly comparable. However SteadyTime is never comparable to SystemTime or SteadyTime so it would actually have a separate implementation of the same API. Thus you could get protection from misusing them at compile time instead of only catching it at runtime. 

There could also potentially be some ways to implement this same protection using template logic for the C++ API.

articles/130_ros_time.md
dirk-thomas: I would say until that is the case I wouldn't "lock in" the way this configuration flag is defined. Maybe just mention that the boolean flag if sim time should be used needs to be retrieved from "somewhere" e.g. a global parameter or a node specific parameter which needs to be kept in sync across multiple nodes.

articles/130_ros_time.md
dirk-thomas: The rational makes sense to me. Can you please add this to the article.

articles/130_ros_time.md
dirk-thomas: That sounds reasonable. Can you please add this concept to the article.

mikaelarguedas:
@tfoote to confirm all comments have been addressed and to tag the ros2/team for new review

articles/130_ros_time.md
tfoote: I specifically scoped this to call out the local requirement. As we don't have any guaranteed global connectivity relying on global state before initializing Time is a big problem on startup. Otherwise we'd have to implement some sort of try to attempt to access a global parameter with a non-trivial timeout to support a full discovery cycle. And behavior would change based on potential network topology/connectivity events.

This is both the current implementation and I believe what we should expect for the future.

The API has significantly more interactions for callbacks for changes in time so that if the coordination of setting the parameter is not exactly synchronized as to when it's notified to use sim time it can deal with the changes in a principled manner.

articles/130_ros_time.md
tfoote: added in fa75254

tfoote:
@ros2/team please take another pass

articles/130_ros_time.md
wjwwood: I think it's ok to say the local parameter needs to be set. We could additionally initialize it from command line parameters, an environment variable, some global config file, or a global parameter server. In any case, we would read it from the local parameter, most likely.

articles/130_ros_time.md
wjwwood: +1 for adding to the article, at least that they should not be compared and by having separate types we avoid that at compile time.

articles/130_ros_time.md
```diff
@@ -115,23 +126,39 @@ It is possible that the user may have access to an out of band time source which
...
+`SteadyTime` will be typed differently than the interchangable `SystemTime` and `ROSTime`.
+
+
 ### Public API
 
-In each implementation will provide `Time`, `Duration`, and `Rate` datatypes, for all three time source abstraction..
-The `Duration` will support a `sleep_for` function as well as a `sleep_until` method.
+In each implementation will provide `Time`, `Duration`, and `Rate` datatypes, for all three time source abstractions.
+
+The `Clock` will support a `sleep_for` function as well as a `sleep_until` method using a `Duration` or `Time` argument.
```
wjwwood: I'd add "respectively" to the end of the sentence because it sounds like `sleep_for` and `sleep_until` could both take either `Duration` or `Time`.

articles/130_ros_time.md
wjwwood: Also, in the case of sim time, you can know any time the time jumps backwards or irregularly regardless of the size of the jump.

test_local_deploy.sh
wjwwood: Rebase since #145 is merged?

wjwwood (approved):

PR #147, opened 2017-09-11
moving analytics into the head block

tfoote:
This is as per google search console email.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -13,17 +13,5 @@
     <script src="/js/bootstrap.min.js"></script>
     <script src="/js/github.js"></script>
     <script src="/js/ros2-design.js"></script>
-    <!-- Google Analytics -->
-    <script>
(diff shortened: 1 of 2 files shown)


wjwwood (approved):

dirk-thomas:
This is commonly done at the end of the body to avoid slowing down the page loading.

wjwwood:
That's why I had it there to begin with, but @tfoote said he needed it to move it based on a recommendation from some Google search panel or something like that.

tfoote:
From the search console verification steps: 
```
Recommended: Google Analytics
Use your Google Analytics account.
You must be using the asynchronous tracking code .
Your tracking code should be in the <head> section of your page.
You must have the "edit" permission for the Analytics web property.
The Google Analytics tracking code is used only to verify site ownership. No Google Analytics data will be accessed.
```

I believe that the new asynchronous tracking snippets also allow much of the work to be done in parallel with the main render path. 

However we could probably revert this since it also appears to work find at the end of the body.

wjwwood:
I don't care either way. Probably won't impact performance too much even in the header.

PR #148, opened 2017-09-11
Update the cpp article.

### LINKED ISSUE #146: Temporarily update the CPP article to say that message field initialization is not done
In http://design.ros2.org/articles/generated_interfaces_cpp.html , the "Constructors" section says that all message fields in the generated C++ code are initialized to zero by default.  While we do want this to be the case, and we have code to make this the case (ros2/ros2#396), that code is not going to make Beta3.  Therefore, we should update the CPP article to represent the current state of the code and say that we are working towards making all of the fields initialized by default.

clalancette:
Make sure to mention that the default constructor doesn't
(currently) initialize member variables.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Fixes #146

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -41,7 +41,7 @@ To separate the generated code from other code within the package it is defined 
 Following the C++ style guide of ROS 2 the namespace hierarchy is mapped to a folder structure.
 The filenames use lowercase alphanumeric characters with underscores for separating words and end with either `.hpp` or `.cpp`.
 
-## <a name="Messages"></a>Messages
+## Messages
 
 For a message a templated `struct` with the same name followed by an underscore is generated.
 The single template argument is the allocator for the data structure.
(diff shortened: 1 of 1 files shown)


articles/112_generated_interfaces_cpp.md
```diff
@@ -41,7 +41,7 @@ To separate the generated code from other code within the package it is defined
 Following the C++ style guide of ROS 2 the namespace hierarchy is mapped to a folder structure.
 The filenames use lowercase alphanumeric characters with underscores for separating words and end with either `.hpp` or `.cpp`.
 
-## Messages
+## <a name="Messages"></a>Messages
```
dirk-thomas: This is not necessary. Each subsection already has an `id` and can be referenced, see http://design.ros2.org/articles/generated_interfaces_cpp.html#messages.

articles/112_generated_interfaces_cpp.md
```diff
@@ -103,8 +103,10 @@ All other constants are declared as `static const` members in the struct and the
 
 ### Constructors
 
-The *default constructor* initializes all members with their default value.
-Optionally the constructor can be invoked with an allocator.
+Currently the *default constructor* does no initialization of member fields.
+This means that the user is responsible for ensuring that all fields are initialized before use, otherwise undefined behavior may result.
+In a future release, the *default constructor* will initialize all fields.
```
dirk-thomas: Append: "by default".

articles/112_generated_interfaces_cpp.md
clalancette: I wasn't aware of that, thanks.  Fixed in 9b18608

articles/112_generated_interfaces_cpp.md
clalancette: Also fixed in 9b18608

dirk-thomas (approved):

clalancette:
Thanks, squashing and merging.

PR #149, opened 2017-09-13
fix link in 060_ros_middleware_interface.md

scpeters:
fix link to rosidl_typesupport_introspection_cpp
moved in https://github.com/ros2/rosidl/pull/53


dirk-thomas:
Thanks for the fix!

PR #153, opened 2017-09-27
Broken Link

ruffsl:
Previous link style had a redundant `articles`
http://design.ros2.org/articles/articles/interface_definition.html
perhaps just omitting `articles/` would also work.


dirk-thomas:
This is already being addressed in https://github.com/ros2/design/pull/152/files#r141025609 which also uses a relative link which is preferred.

PR #154, opened 2017-10-19
lifecycle: change noun to verb for 'deactivate'

gavanderhoorn:
As per subject.

Reading the rest of the article I'm pretty sure the transition is actually named `deactivate`, not `deactive` (which could be a state name).


Karsten1987:
thanks for the patch. The transition is indeed called `deactivate` and the corresponding state `inactive`

https://github.com/ros2/rcl_interfaces/blob/master/lifecycle_msgs/msg/Transition.msg#L12
https://github.com/ros2/rcl_interfaces/blob/master/lifecycle_msgs/msg/State.msg#L10

PR #157, opened 2017-12-04
Clarify the language around zero initialization.

### LINKED ISSUE #156: [C++ message initialization] ZEROS option clarification
Clarify C++ message initialization, in particular how `ZEROS` currently performs and what the expected behavior (not yet implemented) will be.

See https://github.com/ros2/rosidl/issues/251 for more details

clalancette:
Once ros2/rosidl#254 goes in, this documentation update can be merged to clarify what happens with bounded or unbounded array types.

fixes #156

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -103,6 +103,8 @@ All other constants are declared as `static const` members in the struct and the
 
 ### Constructors
 
+In the following discussion, "member" refers to the class member in the C++ class while "field" refers to the field definition in the IDL file.
+
 The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the member is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
 In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
 Therefore, the constructor takes an optional directive of type `rosidl_generator_cpp::MessageInitialization` to control how initialization is done:
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -103,17 +103,17 @@ All other constants are declared as `static const` members in the struct and the
 
 ### Constructors
 
-The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
+The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the member is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
 In some cases this may not be desirable, since these fields will often be immediately overwritten with user-provided values.
 Therefore, the constructor takes an optional directive of type `rosidl_generator_cpp::MessageInitialization` to control how initialization is done:
 
-- `MessageInitialization::ALL` - Initialize all members with their default value; if a field doesn't have a default value, then the field is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization)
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Can you clarify the difference being made here between "member" and "field" ?

clalancette:
Yeah, it is a bit subtle.  A `field` is one of the types that comes out of the IDL file, while a `member` is the actual member variable in the C++ class.  I think if you read it with that in mind, the current wording makes sense, but you do need to know those definitions.  Do you think I should add a section to the article stating the above?

mikaelarguedas:
Thanks for clarifying. Yeah making this explicit would be useful for clarity, we can either define these terms before using them or use "class member" instead of just "member" ( I think I prefer the first option).

`Initialize all members with their default value` becomes `Initialize all members with the field default value`. That's a bit weird to me, `Initialize each member with their fields default value` maybe ? not sure

"([dynamic size](interface_definition.html#arrays-with-dynamic-size) or [upper boundary](interface_definition.html#upper-boundaries) arrays will have zero elements)"

I dont think this is necessary as we say that these field maps to an `std::vector` and that these are zero-initialized with 0 elements (I think ?). If we want to be clearer we can specify (in the type mapping table) that the "custom_class" used for bounded array behaves the same / is wrapped around `std::vector`

clalancette:
> Thanks for clarifying. Yeah making this explicit would be useful for clarity, we can either define these terms before using them or use "class member" instead of just "member" ( I think I prefer the first option).

I ended up adding a sentence at the beginning defining this.

> Initialize all members with their default value becomes Initialize all members with the field default value. That's a bit weird to me, Initialize each member with their fields default value maybe ? not sure

I particularly didn't like the possessive "their" in the sentence, which is why I ended up rewriting it.  I'm not sure your alternate wording is less wordy; I'm going to leave it for now, let me know if you really think we should change it.

> I dont think this is necessary as we say that these field maps to an std::vector and that these are zero-initialized with 0 elements (I think ?). If we want to be clearer we can specify (in the type mapping table) that the "custom_class" used for bounded array behaves the same / is wrapped around std::vector

The default value initialization for `std::vector` does indeed give it 0 elements, but you can also value-initialize it  like `std::vector<int> foo(3)`, which would give it 3 elements.  Given that this was a point of confusion for me when originally writing it, I thought it would be better to be explicit here.

mikaelarguedas:
> > Initialize all members with their default value becomes Initialize all members with the field default value. That's a bit weird to me, Initialize each member with their fields default value maybe ? not sure

> I particularly didn't like the possessive "their" in the sentence, which is why I ended up rewriting it. I'm not sure your alternate wording is less wordy; I'm going to leave it for now, let me know if you really think we should change it.

Why I think it deserves a change is that now it seems like **all** members are initialized based on a **single field** default value.  I'd like to see this plural / singular mismatch be addressed for clarity even if it ends up being slightly more wordy. I don't have a great suggestion as english phrasing is not my strong suit and I trust you'd do a better job than me.
> The default value initialization for std::vector does indeed give it 0 elements, but you can also value-initialize it like std::vector<int> foo(3), which would give it 3 elements. Given that this was a point of confusion for me when originally writing it, I thought it would be better to be explicit here.

I guess being more explicit doesn't hurt if you think there is room for misinterpretation :+1:

clalancette:
> Why I think it deserves a change is that now it seems like all members are initialized based on a single field default value. I'd like to see this plural / singular mismatch be addressed for clarity even if it ends up being slightly more wordy. I don't have a great suggestion as english phrasing is not my strong suit and I trust you'd do a better job than me.

Got it.  Let me try to rephrase it again.

dirk-thomas (approved):

clalancette:
Thanks, merging.

PR #161, opened 2018-01-16
Ban consecutive underscores in topic names

sloretz:
As discussed in the meeting today this PR forbids a topic name from having consecutive underscores. The original motivation is to disambiguate the remapping grammar where `__ns:=/foo/bar` can be parsed both as namespace replacement or a relative topic name remap.

Banning leading double underscores would be enough, but this PR makes a broader restriction because: 
* the ROS topic to dds topic mapping may return to substituting double underscores for slashes, 
* a use case requiring consecutive underscores in a topic name is unlikely, 
* and it is easier to relax a restriction in the future than it is to make the rules more strict.

### DIFF
--- a/articles/140_topic_and_service_name_mapping.md
+++ b/articles/140_topic_and_service_name_mapping.md
@@ -66,6 +66,7 @@ For convenience here is a summary of all rules for topic and service names in RO
 - must not start with a numeric character (`[0-9]`)
 - must not end with a forward slash (`/`)
 - must not contain any number of repeated forward slashes (`/`)
+- must not contain any number of repeated underscores (`_`)
 - must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`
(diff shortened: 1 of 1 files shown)


tfoote (approved):

wjwwood (approved):
This lgtm, as we discussed in the meeting. I would, however, recommend circulating this on discourse as an RFC, even if we don't expect anyone to have a problem with it and we start moving forward as-if it will be merged.

sloretz:
@wjwwood discourse topic [here](https://discourse.ros.org/t/requesting-comments-on-forbidding-consecutive-underscores-in-topic-names/3718).

gbiggs:
I don't have any use cases that this would conflict with. I think it's a sensible restriction.

dirk-thomas (approved):

IanTheEngineer (approved):

ros-discourse:

https://discourse.ros.org/t/requesting-comments-on-forbidding-consecutive-underscores-in-topic-names/3718/1

PR #163, opened 2018-02-13
ROS 2 Launch System

wjwwood:
This is a WIP. It's not ready for review or comment right now (not even in a readable state I would say).

When it's ready for an early review I'll post a comment. When it's ready for wider review I'll post an RFC on discourse.ros.org.

### DIFF
--- a/articles/150_roslaunch.md
+++ b/articles/150_roslaunch.md
@@ -1,0 +1,170 @@
+---
+layout: default
+title: ROS 2 Launch System
+permalink: articles/roslaunch.html
+abstract:
+  The launch system in ROS is responsible for helping the user describe the configuration of their system and then execute it as described. The configuration of the system includes what programs to run, what arguments to pass them, and ROS specific conventions which make it easy to reuse components throughout the system through configuration. Also, because the launch system is the entity that executes processes, it is also responsible for monitoring the state of processes in the system, then reporting and/or reacting to changes in the state of those processes.
+author: '[William Woodall](https://github.com/wjwwood)'
+published: true
+---
+
+- This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Context
+
+<div class="alert alert-warning" markdown="1">
+TODO: Fill out some context with respect to ROS 1's `roslaunch` and other similar tools which exist in OS's.
+</div>
+
+## Separation of Concern
+
+The launch system can be considered in parts, separated by concern.
+The coarse breakdown is like so:
+
+- Calling Conventions for Processes and Various Styles of Nodes
+- System Description
+- Execution of the System Description
+- Reporting System for Event
+
+The purpose of this section is to enumerate what the launch system could interact with or do, but is not the requirements list for the launch system in ROS 2.
+The requirements for the launch system will be enumerated in section below based on what's possible in this section.
+
+### Calling Conventions
+
+In order for the launch system to execute a described system, it needs to understand how it can achieve the description.
+In this case, the phrase "calling conventions" is meant to describe the "interface" or "contract" the launch system has with the entities it is executing and monitoring.
+This contract covers initial execution, activity during runtime, signal handling and behavior of the launch system, and shutdown.
+
+#### Operating System Processes
+
+The most basic version of these entities, and the foundation for the other entities, are operating system processes.
+
+##### Execution
+
+For these, the launch system needs to know how to execute them, and to do that it needs:
+
+- name of the executable (just the name, relative path, or absolute path)
+- environment variables (`PATH`, `LD_LIBRARY_PATH`, `DL_PRELOAD`, etc...)
+- command line arguments
+- working directory (directory from which to execute the process)
+- launch prefix (used to inject things like `gdb`, `valgrind`, etc...)
+
+<div class="alert alert-warning" markdown="1">
+RFC:
+
+Missing from this list is the user which should be used to execute the process.
+It's possible that it would be necessary or at least useful to change the user based on the launch description.
+However, it can always be done in a user written script and supporting it in our Python implementation in a portable way looks to be difficult.
+</div>
+
+With this information the launch system can execute any arbitrary operating system process on the local machine.
+
+##### Runtime
+
+During runtime, the launch system may monitor all operating system process's:
+
+- `stdout` pipe
+- `stderr` pipe
+
+The launch system may choose to either capture these pipes, for logging or suppressing output to the console, or it can connect the pipes to an existing `pty`, like the terminal's `stdout` and/or `stderr` pipes or a null pipe (e.g. `/dev/null`).
+
+When capturing the output pipes of a process, the launch system could report this data in a way that the user may process them in real-time or could pass the data through user defined filters, generating a user-handled event when the filter matches.
+
+In addition, the launch system may interact with, or allow the user to interact with, an operating system process's:
+
+- `stdin` pipe
+- signals (`SIGINT`, `SIGTERM`, `SIGUSR1`, etc...)
+
+Regardless of how the user uses the launch system to interact with these items, they should be exposed by the launch system, which is the only entity which can interact with them directly.
+
+##### Termination
+
+If the operating system process terminates, and therefore returns a return code, the launch system will report this event and it can be handled in a user defined way.
+Termination covers expected termination (e.g. return from `main()` or use `exit()`) and unexpected termination (e.g. the abort trap or a segmentation fault or bus error).
+
+Historically, ROS 1's `roslaunch` allowed a few common exit handling cases:
+
+- `require=true`: if this process exits (any reason) shutdown everything else, as it's "required"
+- `respawn=true`: if this process exits (any reason) restart it with the same settings as startup
+  - `respawn_delay=N`: if restarting it, delay a number of seconds between attempts
+
+#### Shell Evaluation
+
+A special case of operating system processes, shell evaluation would simply be passing shell script code as an argument to the default system shell.
+This is a problematic thing to support because it is hard/messy to make it portable to all operating systems.
+
+An in between entity is an operating system process which uses shell evaluation to expand a relative executable name to an absolute path using the PATH environment variable.
+
+#### Remote Operating System Processes
+
+<div class="alert alert-warning" markdown="1">
+TODO: Describe this as a "mix-in" which can convert anything based on an "operating system process" into a remote one by adding ssh/user/setup/etc... information needed to do so, see how ROS 1's roslaunch does it.
+</div>
+
+#### Process with a Single Node
+
+This is a ROS specific entity which contains a single ROS node.
+This was likely the most commonly used type of entity launched in ROS 1, as you could only have one node per process in ROS 1.
+In ROS 2, this will likely be less common, but will still be used quite a bit in the form of quickly developed scripts and drivers or GUI tools which might require control over the main thread.
+
+In 
+
+#### Process with Multiple Nodes
+
+#### Dynamically loaded Nodes
+
+##### By Configuration
+
+##### By Proxy
+
+### System Description
+
+<div class="alert alert-warning" markdown="1">
+TODO: Restructure notes on this and put them here.
+
+Temporary summary:
+
+This is the section where the Python API in `ros2/launch` and the XML file format from ROS 1 overlap.
+Basically how do you describe the system in a way that is flexible, but also verifiable and ideally can also be statically analyzed (at least for something like the XML format).
+
+I have compare and contrast like notes for other systems like upstart, systemd, and launchd. 
+</div>
+
+### Event System
+
+<div class="alert alert-warning" markdown="1">
+TODO: Restructure notes on this and put them here.
+
+Temporary summary:
+
+The launch system is responsible for executing processes (according to calling convention below) and exposing information that only it, as the parent process, has access to via events, e.g. things like stdout/stderr, return code.
+The "reporting responsibility" could also extend to ROS specific things in the form of aggregation, i.e. allow the user to say "do X when nodes A, B, and C all reach the inactive state", but this could also be done on top of roslaunch.
+This event system would also be reused (in combination with other sources of information) by the launch system to implement things like verification.
+</div>
+
+## Requirements
+
+<div class="alert alert-warning" markdown="1">
+TODO: Reformat requirements list, possibly combine/reconcile with "separation of concerns section"
+</div>
+
+## Reference Implementation Proposal
+
+<div class="alert alert-warning" markdown="1">
+TODO: This will outline what we have and what we need to build and how it should be separated.
+</div>
+
+## Alternatives
+
+<div class="alert alert-warning" markdown="1">
+TODO: Anything we choose not to support in the requirements vs. the "separation of concern section", and also any alternatives which we considered but rejected in the reference implementation proposal.
+</div>
--- a/articles/150_roslaunch.md
+++ b/articles/150_roslaunch.md
@@ -530,8 +530,11 @@ Again, like many other event systems, the events should have a type (either as a
 #### Event Handlers
 
 Events can be handled by registering an event handler with the launch system.
-This could be something like a callback within the launch system itself, a "lambda" defined in the description of the launch file, or even a built-in handler which publishes the events as ROS messages
-In the last case, it could be either be a subscription to a topic or a service call which was registered with the launch system a priori.
+The only required form of event handler is one that is a function, registered locally with the launch system.
+
+Other kinds of event handlers could be supported by building on a locally defined function.
+They could be something like a user-defined "lambda" defined in the description of the launch file, or even a built-in event handler function which just publishes the events as ROS messages.
+In the latter case, it could be either be a subscription to a topic (which needs no a priori registration with the launch system) or a service call (which was registered with the launch system a priori).
 So if it is a topic, the subscription could be considered an event handler.
 In the case of a service, which would be called by the launch system and handled by a user defined service server, the service server (and it's response) would be considered the event handler.
 
@@ -541,6 +544,12 @@ By default, events are passed to all event handlers and there is no way for an e
 
 While event handlers have no comparison operators between one another (so no sorting), the order of delivery of events to event handlers should be deterministic and should be in the reverse order of registration, i.e. "first registered, last delivered".
 Note that delivery to asynchronous event handlers (e.g. a subscription to a ROS topic for events, sent via a ROS publisher), will be sent in order, but not necessarily delivered in order.
+
+<div class="alert alert-warning" markdown="1">
+RFC:
+
+With respect to the delivery order, I think first registered, last delivered makes the most sense, but I could be convinced otherwise.
+</div>
 
 ##### Event Filters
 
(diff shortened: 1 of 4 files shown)


articles/150_roslaunch.md
```diff
@@ -0,0 +1,170 @@
...
+TODO: Fill out some context with respect to ROS 1's `roslaunch` and other similar tools which exist in OS's.
+</div>
+
+## Separation of Concern
+
+The launch system can be considered in parts, separated by concern.
+The coarse breakdown is like so:
+
+- Calling Conventions for Processes and Various Styles of Nodes
+- System Description
```
jack-oquin: Is the system description static?  Or, will there be ways to launch specific components following dynamic device discovery?

articles/150_roslaunch.md
```diff
@@ -0,0 +1,170 @@
...
+
+<div class="alert alert-warning" markdown="1">
+TODO: Restructure notes on this and put them here.
+
+Temporary summary:
+
+This is the section where the Python API in `ros2/launch` and the XML file format from ROS 1 overlap.
+Basically how do you describe the system in a way that is flexible, but also verifiable and ideally can also be statically analyzed (at least for something like the XML format).
+
+I have compare and contrast like notes for other systems like upstart, systemd, and launchd.
```
jack-oquin: Those are good things to compare. 

Is there some thought of integrating with your [capabilities package](http://wiki.ros.org/capabilities) at some point? Does that even make sense?

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+- Testing
+
+The purpose of the following sections is to enumerate what the launch system could interact with or do, but is not the requirements list for the launch system in ROS 2.
+The requirements for the launch system will be enumerated in section below based on what's possible in these sections.
+
+## Calling Conventions
+
+In order for the launch system to execute a described system, it needs to understand how it can achieve the description.
+This is an existing phrase in Computer Science[^calling_convention_wikipedia], but this section is not talking specifically about the compiler defined calling convention, through it is appropriating the term to describe a similar relationship.
+In this case, the phrase "calling conventions" is meant to describe the "interface" or "contract" the launch system has with the entities it is executing and monitoring.
```
sloretz: Might be a good idea to define what an entity is before this section. I assume `entities` means the types of nodes: a process with one node, a process manually composed with multiple nodes, a library containing a dynamically composible node, or all of these but with a lifecycle node.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+<br/>
+</wjwwod's opinion>
+
+This will come up again in the "System Description" section, because I believe the mind set around the "Execution of the System Description" (basically the python library that runs things) will be "personal responsibility".
+</div>
+
+#### Termination
+
+Managed ROS Nodes have some additional observable effects when terminating (the node, not necessarily the process containing it).
+A managed node enters the `Finalized` state after passing through the `ShuttingDown` transition state on termination.
```
sloretz, line 344: During termination will ROS 2 launch try to transition to the Finalized state, or just send signals and hope handling code in the node takes care of it?

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+
+The first is a single process with a single ROS node within it.
+This was likely the most commonly used type of entity launched in ROS 1, as you could only have one node per process in ROS 1.
+In ROS 2, this will likely be less common because you can have one to many nodes per process, but will may still be used quite a bit in the form of quickly developed scripts and drivers or GUI tools which might require control over the main thread.
+
+Since there is only one ROS node, the command line arguments do not need to be explicit about to which node they apply.
+For example, changing the namespace of the single node could be expressed with the command line argument `__ns:=new_namespace`.
+
+Even though there is only one node in the process, that node does not need to start with the process starts, nor does the process need to end when the node is shutdown and/or destroyed.
+If it is a managed node, the lifecycle of the node is best tracked using the lifecycle events.
```
sloretz, line 364: Is ROS 2 launch responsible for transitioning a node to `Active`? If so, how would it know if a process contained a normal node or a managed one? If not, how would the user manage the node? By launching a separate node that triggers the transitions?

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+  <node package="my_package" executable="listener" />
+
+</node_container_process>
+```
+
+Whether or not the above results in three processes (one container and two proxy processes) or a single process, probably doesn't matter to the user or at least would be fine with the user in exchange for less overhead.
+
+It would be more efficient because you don't need to have a node in each proxy to actually make the Service call to the load that node the user wants to run, and you don't need to maintain a bond between the container and the proxy processes.
+
+So I'm interested in what others think about coming up with a standardized executable which can load and configure many nodes at once from command line arguments or a config file.
```
sloretz: The above XML could be syntactic sugar for a system similar to nodelets. However, I think that means ROS2 launch would have a dependency on the container process/package, which would have a dependency on rclpy/rclcpp. Third parties implementing different language client APIs may desire a way to be launched into a container process without making ros 2 launch depend on them. If that's supported then rclcpp/rclpy should probably use the same API.

articles/150_roslaunch.md
wjwwood: I was hoping this sentence would imply "entities are anything that the launch system executes and/or monitors", rather than me specifically defining entities. I used "entities" because "things" sounded unsophisticated.

Perhaps I could reword it to be "... the launch system has with entities, which are anything it executes or monitors."? If not, do you have a suggestion as how to reword it?

articles/150_roslaunch.md
wjwwood: A good question, to which I do not know the answer. My gut reaction would be to say it sends a signal and the launch system tries not to send state transition requests itself, unless directed to do so by the user. So maybe, defaults to signals, the user can ask the launch system to transition first, then signal. I'll mention it and add an RFC block about the topic.

articles/150_roslaunch.md
wjwwood: > Is ROS 2 launch responsible for transitioning a node to Active?

Again, not sure. My initial reaction is "no, the node should transition itself unless directed to do otherwise". But I haven't fleshed out how that would work. We (me, Karsten, perhaps others) have spoken in the past about lifecycle profiles which would be basically an option to lifecycle nodes which determines what they do automatically and what they wait to do based on external requests, e.g. a default profile might transition all the way to active, then stay there until shutdown (sigint or something else), or a managed profile that goes automatically until inactive then waits for external transition requests or shutdown.

> If so, how would it know if a process contained a normal node or a managed one?

I believe the user will already have to inform the launch system if a process contains a node and if so whether or not it is a managed node. So, the user would have to tell it, otherwise it will assume it's a normal node or a normal process.

> If not, how would the user manage the node? By launching a separate node that triggers the transitions?

Yes, this gets into one of the trickier issues for which I've been trying to come up with the right taxonomy (meronomy?) relationships. Basically, where does the launch system end and general purpose lifecycle managed begin? The launch system could use the lifecycle to verify the system and/or do more sophisticated startup procedures (e.g. when nodes A and B get to state inactive, launch node C...), but how is that different from (if node A goes to error state, move node B to inactive). Should it do both, or neither, or some grey line in between?

If the answer is neither or some grey line in between, then yes the user would simply start another node to manage the nodes that need external transitions.

articles/150_roslaunch.md
wjwwood: That's a fair point, in ROS 1 this was worked around because `<node_container_process name="my_container_process">` was more like `<node pkg="nodelet" type="nodelet_manager" name="my_container_process" ...>`, i.e. it was just another node to run. If we remove the syntatic sugar, this would just be another process and if someone wanted for e.g. a rust based container for rust plugins then they would just make a new package and a new executable.

A separate concern is what happens if you put a node which is implemented as a Python script within the "node_container_process" tag? I suppose it will try to load it and fail since it's not a plugin, but it brings up the, point should we have a way to express that a `<node>` tag refers to a "process" or a dynamically loadable thing, or both?

articles/150_roslaunch.md
sloretz: Oops I assumed entity was going to be a special term. If it's not going to be used later on then I suggest shortening to "... `the launch system has with anything it executes or monitors.`".

articles/150_roslaunch.md
sloretz: It sort of sounds like you're saying your gut reaction is to separate the concern of launching nodes from that of managing their lifecycle.

Assuming a separate process/project for managing lifecycle nodes, it seems like that project would need an API into ROS 2 launch rather than the other way around. E.g. a managing process might ask ROS 2 launch to start some unmanaged log-backup node when a managed node transitions to `cleanup`.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+- command line arguments
+- working directory (directory from which to execute the process)
+- launch prefix (used to inject things like `gdb`, `valgrind`, etc...)
+
+<div class="alert alert-warning" markdown="1">
+RFC:
+
+Missing from this list is the user which should be used to execute the process.
+It's possible that it would be necessary or at least useful to change the user based on the launch description.
+However, it can always be done in a user written script and supporting it in our Python implementation in a portable way looks to be difficult.
```
jack-oquin, line 211: Adding a user name would also raise the complex question of how to authenticate. 

I would avoid that here, unless ROS 2 somehow evolves a system-wide security design which depends on it.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+The latter two steps can be skipped, or the time until escalation can be adjusted, on a per process basis.
+
+The launch system will initiate this process when an event (built-in or user generated) initiates shutdown, e.g. when a process with the equivalent of the `require=true` exit handler terminates, or when the launch system itself receives the `SIGINT` signal.
+
+If the launch system itself receives the `SIGTERM` signal it will send the `SIGKILL` signal to all child processes and exit immediately.
+
+<div class="alert alert-warning" markdown="1">
+RFC:
+
+There are several small decisions here that I made somewhat arbitrarily, e.g. the default time until escalation and the propagation of `SIGKILL` when the launch system receives `SIGTERM`.
```
jack-oquin: Sending SIGKILL on receipt of SIGTERM feels a bit abrupt. Clients have no opportunity to catch it a clean up anything. Are you assuming that SIGINT was already used, and the user got impatient?

Doing the SIGTERM; wait 10 sec.; SIGKILL sequence might work better, but would not satisfy impatient users.

articles/150_roslaunch.md
jack-oquin: I wonder if these things should be recursively nestable, with the roslauch itself serving as a managed ROS node?

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+
+<div class="alert alert-warning" markdown="1">
+TODO: Restructure notes on this and put them here.
+
+Temporary summary:
+
+This is the section where the Python API in `ros2/launch` and the XML file format from ROS 1 overlap.
+Basically how do you describe the system in a way that is flexible, but also verifiable and ideally can also be statically analyzed (at least for something like the XML format).
+
+I have compare and contrast like notes for other systems like upstart, systemd, and launchd.
```
jack-oquin: Is there any notion of supporting something like the ROS 1 [capabilities package](http://wiki.ros.org/capabilities)?

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+</div>
+
+## Requirements
+
+<div class="alert alert-warning" markdown="1">
+TODO: Reformat requirements list, possibly combine/reconcile with "separation of concerns section"
```
jack-oquin: Normally, requirements would go near the beginning of a design document.

articles/150_roslaunch.md
jack-oquin: I like the idea of a standardized executable for loading and configuring each supported language. 

That assumes we're not expecting to mix C++ and Python nodes in a single process. Whether other languages provide similar support would be up to them.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+#### By Proxy
+
+The other way to implement a process which dynamically instantiates ROS nodes, is to send the request to instantiate a node asynchronously from another process.
+So the "container" process would start up and provide a service (ROS Service or otherwise) that lets external processes request a node be instantiated with a given set of configurations.
+
+One form this can take is by providing a "proxy" process which doesn't actually run the node, but instead loads it into an already running container and just stays running until the node is shutdown in the remote container.
+This pattern is used by the "nodelet's" in ROS 1.
+It's a useful pattern when you're running nodes by hand but you want them to share a process and don't want to write your own program to do that.
+
+There are a few other forms this can take, but the common thread between them is that a process instantiates nodes dynamically based on asynchronous input from external actors (proxy's), and that the configuration for those nodes is communicated through something other than command line arguments and environment variables.
```
jack-oquin: While this looks useful for porting existing ROS 1 nodelet usages, I prefer the ``<node_container_process>`` approach for native ROS 2 programming. I don't see the advantage of the proxy approach, but perhaps there are some and they should be mentioned explicitly.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,364 @@
...
+Alternatively, we could say "all event handling is opaque to the launch system and is user defined" which removes the ambiguity.
+
+\<wjwwod's opinion>
+<br/>
+I think the idea that saying all "supervision" or "reactionary launching" like activities are not part of the launch system is the simplest way forward, but then you'd still like to include in the launch system some convenient ways to express the most common cases, e.g. something like ROS 1's `roslaunch`'s `require=true` and `respawn=true`.
+In my opinion, it would be unacceptably inconvenient to have users write a lambda for each of those cases, and I also think it would lead to everyone doing it slightly different and making it impossible to understand with tools.
+<br/>
+</wjwwod's opinion>
+
+This will come up again in the "System Description" section, because I believe the mind set around the "Execution of the System Description" (basically the python library that runs things) will be "personal responsibility".
```
jack-oquin: Is there a general ROS 2 design decision to move away from the ROS 1 concept that each node should be written in a way that is independent of the order in which other nodes start?

Many people question that approach, but experience suggests that it was a good way to write reusable nodes.

If not, then the use case of "wait for A to get to Active then launch B" may not be very important. It plays into a simple-minded system design paradigm that we may not want to encourage.

jack-oquin: To me, the more important problem is providing recovery, availability and serviceability in the face of node or process failures. That could involve:
* if a node fails, restart it
  - what do other nodes do in the interim? Just wait?
* if a node cannot be restarted (device failure perhaps?), run something else as a fallback
* if everything is FUBAR, capture as much relevant information as possible for off-line analysis

jack-oquin:
The current description abstracts the actual syntax of the "system configuration". That's nice at this design stage.

However, I find myself intensely curious whether XML, Python or some other syntax is under consideration.

gbiggs:
Are we reviewing this now? I was holding off until @wjwwood said go for it, but would be happy to start commenting. :smile:

jack-oquin:
I am commenting because I find the document and "RFC" sections worth discussing, and probably because I don't know any better. :smile:

wjwwood:
> Are we reviewing this now? I was holding off until @wjwwood said go for it, but would be happy to start commenting. 😄

I don't mind comments now, but it's not "ready" for review yet. I'm trying to push out a completed document as soon as possible, at which point I will solicit feedback actively, first on the pr and then on discourse. But until then, feel free to discuss, though I might not get back to you right away.

wjwwood:
Good morning :), I pushed a set of changes for the "context" section where it compares it to ROS 1 and considers what might be different. I'm close but not quite finished with the system description and event sections, at which I think it will be in a relatively good place to start discussion on the pr (not quite ready for discourse still though).

@jack-oquin I'll try to respond to your comments asap, but I'll have to switch gears a bit before I have time to do that.

jack-oquin:
No hurry on my account.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,562 @@
+---
+layout: default
+title: ROS 2 Launch System
+permalink: articles/roslaunch.html
+abstract:
+  The launch system in ROS is responsible for helping the user describe the configuration of their system and then execute it as described. The configuration of the system includes what programs to run, what arguments to pass them, and ROS specific conventions which make it easy to reuse components throughout the system by giving them each different configurations. Also, because the launch system is the process (or the set of processes) which executes the user's processes, it is responsible for monitoring the state of the processes it launched, as well as reporting and/or reacting to changes in the state of those processes.
```
clalancette: (edit: oh, you talk about this below, sorry; maybe we should add a little something to the abstract anyway?)

I hate to make a complicated subject even more complicated, but what do we think about executing processes across multiple different compute modules?  After all, a reasonably complex robot will have multiple processing modules, and hence successful launching of the "system" really involves launching processes on all of the modules.  There's also the related, even larger question about launching swarms of robots, but that gets a little pie-in-the-sky.

If it is not our intention to handle either of those cases here, I think that is fine, but maybe we want to explicitly say that it is a non-goal of this, and larger orchestration may come in a later design document?

articles/150_roslaunch.md
stonier: Per node parameterisation of those parameters would be nice. We were at times 1) impatient, but also at times 2) requiring more shutdown time to be made available to some processes

articles/150_roslaunch.md
wjwwood: I think it is a goal of the overall system and therefore this design doc (you've noticed that it's mentioned below), but not a feature we have to implement immediately if time doesn't allow, so long as we design for it.

articles/150_roslaunch.md
wjwwood: > Sending SIGKILL on receipt of SIGTERM feels a bit abrupt.

So to be clear, if you SIGINT the launch system, then processes run by the launch system will get SIGINT too, before being escalated to SIGTERM and then finally SIGKILL.

If you SIGTERM the launch system, it will immediately SIGKILL child processes and try to exit as quickly as possible. So only do that if you're impatient or already tried SIGINT.

If you SIGKILL the launch system, I don't know what happens to the child processes, it might be OS dependent, but it could result in "zombie" processes.

So, given that graduation of signals the user can send the launch system, I felt that these were appropriate.

> Clients have no opportunity to catch it a clean up anything.

Well, that's the idea with SIGTERM to the launch system. If you want the child processes to be able to react, then SIGINT the launch system and let it do the normal SIGINT->SIGTERM->SIGKILL escallation.

That all being said, I'm open to suggestions (and rationale) about what the launch system should do instead when receiving SIGTERM.

> Are you assuming that SIGINT was already used, and the user got impatient?

Yes, or that something is wrong in the launch system itself, so I was aiming for an "as simple as possible" SIGTERM signal handler for the launch system, which basically sends SIGKILL to all child processes and then exits itself immediately.

Doing a SIGTERM to child processes, waiting 10 seconds or until all child processes exit, and then sending SIGKILL to all child process still up might be more complicated and therefore more likely to also hang.

Basically, you'd want to remove any incentive to SIGKILL the launch system by making SIGTERM reliable and quick to exit.

> Doing the SIGTERM; wait 10 sec.; SIGKILL sequence might work better, but would not satisfy impatient users.

That might be better, I honestly don't know. Anyone else want to weigh in on that?

----

> Per node parameterisation of those parameters would be nice. We were at times 1) impatient, but also at times 2) requiring more shutdown time to be made available to some processes

Yeah, that's a good idea, and I was thinking of that myself, though I'll have to double check the document to see if that's called out clearly.

However, one concern is that it might break some expected behavior of the launch system for users. For example, if a user expects the launch system to exit and escalate at set time intervals, but some nodes aren't affected by this and instead have different schedules, it might be very confusing to them why the launch system is suddenly behaving differently or taking so long to shutdown. So we'll have to be careful about logging what's happening when extra shutdown constraints are made on a per process basis.

articles/150_roslaunch.md
wjwwood: That's a good question. That's what has been consuming my time when working on this document is finding the right partition between the value of a static format that can be edited with tools and introspected without running it, versus a fully dynamic system that can do anything, but also is unpredictable without executing it (something like the halting problem, but not quite). It could be addressed by having a limited, declarative style for the description but being implemented within a more expressive syntax than xml or yaml. There are whole projects for this topic, basically using something like Python or Ruby to make a DSL for a specific topic or domain.

I've collected a lot of information on this subject, but I'm still working out how to present it most clearly and come up with a suggestion on how to proceed before discussing it.

To summarize, I think it's clear we need dynamic behavior (as you said "ways to launch specific components following dynamic device discovery"), but the question for me is "does the static description need to support that?" or "should that use case be solved by using the launch system's API, completely outside of the static description?".

articles/150_roslaunch.md
wjwwood: > Is there some thought of integrating with your capabilities package at some point? Does that even make sense?

I don't know, I'll have to think about it more, but I think it would be good to mention in this document, if only to better describe how the launch system is different from a system like that.

I think it makes sense to consider it, but I don't know off hand if it makes sense to integrate something like the capabilities server into the launch system, or to instead build it on top.

articles/150_roslaunch.md
wjwwood: I tend to agree, but I do think it would be a useful feature at some point. However, I'm concerned it would make launch files less reusable, e.g. you don't want a driver to try to run as "camera_user" when that might be specific to the machine it was originally developed on. In this way, it would need to be used more cautiously and at integration time, like the remote machine functionality.

articles/150_roslaunch.md
wjwwood: > That assumes we're not expecting to mix C++ and Python nodes in a single process. Whether other languages provide similar support would be up to them.

Yes, for now, this only helps C++, but it would in principle be possible to have an executable which could combine multiple Python nodes into one interpreter, or even run C++ nodes and Python nodes in a manually instantiated Python interpreter within the same process (that doesn't necessarily imply intra-process communication cross language). But I don't plan to do that right now.

articles/150_roslaunch.md
wjwwood: > Assuming a separate process/project for managing lifecycle nodes, it seems like that project would need an API into ROS 2 launch rather than the other way around. E.g. a managing process might ask ROS 2 launch to start some unmanaged log-backup node when a managed node transitions to cleanup.

Yeah, I expect we'll have that interface regardless. It should be possible to interact with a running launch system via API (probably Python), or via ROS interfaces like services, or maybe even via a different SOA like REST or XMLRPC (just ideas, no concrete plans to do that).

articles/150_roslaunch.md
wjwwood: It's possible, and I do expect the launch system will be a Python ROS node itself, though it might not be a managed one to start with, partially because there's no lifecycle in Python atm and because it's simpler to start without that.

But I can imagine calling the launch system from the launch system in a nesting fashion. This just has to be balanced with including other launch files instead. I think that might be more transparent than calling the launch systems executable multiple times.

wjwwood: For instance, in ROS 1, the capability server is often called by roslaunch and itself in turn calls roslaunch multiple times. If I were to reimplement that in ROS 2, I would hope to instead have the capability server ask an existing launch system to add something dynamically, rather than starting a new instance. That is if the capability server is not an instance of the launch system itself or rolled into it entirely.

articles/150_roslaunch.md
stonier: Aye, +1 to clear logging on shutdown. I'd be unconfused if roslaunch printed a table every few seconds on shutdown reporting out on what processes remain and when the next escalation is due to occur.

articles/150_roslaunch.md
stonier: +1 for focusing on recovery, availability and serviceability in the face of surviving post-apocolyptic failures.

One item that would have been very useful is for roslaunch to report out (with relevant stdout/stderr) on launch failures or crashes, either to a watchdog node or a graphical interface to the robot. We used to track a node's existence from watchdog programs, but were not able to extract stdout/stderr.

Perhaps for this, a useful feature would be the ability to provide callbacks for node launch failures/crashes? This could enable a variety of patterns, such as some of those suggested by Jack above. Common use cases could be implemented as extensions later.

articles/150_roslaunch.md
stonier: I like this idea also - capabilities can greatly simplify the launch for applications further up the stack, but it can probably go in as an extension and only then after capabilities have been reworked for ROS2.

articles/150_roslaunch.md
jack-oquin: > > Are you assuming that SIGINT was already used, and the user got impatient?

> Yes, or that something is wrong in the launch system itself, so I was aiming for an "as simple as possible" SIGTERM signal handler for the launch system, which basically sends SIGKILL to all child processes and then exits itself immediately.
>
> Basically, you'd want to remove any incentive to SIGKILL the launch system by making SIGTERM reliable and quick to exit.

That seems reasonable to me, and the rationale you give would help other readers.

stonier:
**Configuring the Quality of Service for Connections (reliable, unreliable, ...)**

In ROS1 you could only do this in the code itself. It would have been awesome if you could configure this at the launch level. You could rewire connections, but could not reconfigure the *kind* of connection. That always struck me as odd.

We did however, manage to get by on workarounds. If it was our code, we parameterised the node to provide QoS configuration. If it wasn't, we got by with relays. Though for the product, we wrote specialised nodes to handle a conglomeration of relays with various capabilities (subsampling, QoS, ...) to minimise the impact of so many extra processes. We even had ghosts on the server to manage the unreliable connections to the robot and present a an api to the rest of the server that was always 'up'.

How hard would this be to do for ROS2's launch system? In scope?

wjwwood:
> In ROS1 you could only do this in the code itself. It would have been awesome if you could configure this at the launch level. You could rewire connections, but could not reconfigure the kind of connection. That always struck me as odd.

So, DDS does have a way to do this with XML files which can affect the QoS of entities in a program or globally. And we could do something similar.

However, I personally dislike this kind of runtime configuration, because when the person writing the code originally is not the person doing the system integration, changing QoS settings might break expectations.

As an example, a developer might create a node with a publisher which expects certain blocking behavior from the `publish` call (e.g. non-blocking), but if you change the QoS settings that might cause the `publish` call's blocking behavior to change, which breaks the expectation of the original programmer and thus subtly breaking the program itself.

Another example of this might be that a developer creates a program which pre-allocates 10 messages and sets the publish history to keep last and depth 10. If you externally change that to say 100, then maybe that breaks the program because now the history is larger than the resources pre-allocated by the developer.

Remapping of topic names is similar, and so the previous paragraphs might sound hypocritical, but I would argue that the important difference is that changing a topic name doesn't change the behavior of any of the API's in ways that might cause the program to break (as far as I can imagine).

> We did however, manage to get by on workarounds. If it was our code, we parameterised the node to provide QoS configuration. If it wasn't, we got by with relays. Though for the product, we wrote specialised nodes to handle a conglomeration of relays with various capabilities (subsampling, QoS, ...) to minimise the impact of so many extra processes.

I actually very much prefer this approach of having developer defined configurations which system integrators can tweak. By only allowing changes to settings the developer exposes, the developer can assume the others are not changing under their feet (so it avoids the subtle changes in behavior when QoS changes) and for the ones they expose then they can also put constraints on them, e.g. if the developer wanted to make the depth configurable, then they could expose a ROS parameter for that, and then constrain it to be within the range 10-100.

As for relays which rename, throttle, or downsample, I think the approach of having these are reusable nodes which can be added to the process of other user defined nodes is a better approach. In ROS 1 you had to have a separate process for each and that caused a lot of overhead.

So that's a bit opinionated from me about the way systems should be constructed, but I'm open to being convinced that runtime configuration of QoS is a good idea.

> How hard would this be to do for ROS2's launch system? In scope?

It would be hard as described because we have no way to control this right now in the C++/Python API. Like if a developer hard codes the history depth to 10 with keep last, then we have no way to intercept that and change it as it is being created. We could add that, it just isn't planned as far as I know.

I do think this out of scope for the launch system because the only impact it would have on the launch system is whether or not it's represented in the launch description and if so how it is communicated to the process containing the node. So I'd mention here after we decide to support that use case, but I'd put it in the same bucket as static remapping and parameters which are mentioned here but not part of the actual launch system itself.

stonier:
Going to take a step back and say I completely agree with you. That feature thought was from a few years ago, yet when working on the product, we aggressively found ways to very carefully manage robot-server connections using a variety of methods, only some of which I have mentioned above. If I think to where we might apply roslaunch QoS configuration, it would be for almost trivial situations and not for where the critical problems lie.

articles/150_roslaunch.md
iluetkeb: At the moment (i.e., roslaunch 1) the system specifies several things
 - what to run
   - including launch-prefixes
 - how to run (largely implicitly, but worth mentioning)
 - placement (where to run)
   - including remote-node access 
 - configuration
   - parameters (in several different ways)
   - command-line arguments
   - env-handlers
 - data connectivity (remap, group)
 - error handling (respawn, required)
 - test definition

Arguably, this was not really necessary, but just came about because it was a convenient place to stick all the stuff.

This definitely caused usability issues already in ROS1. For example, I have rewritten countless launch files, instead of just including them, because one of these things was defined in a way that did not match requirements. Also, grouping is problematic -- you could group, but not ungroup. The various ways to set parameters, and the point in time when they apply, is also often unclear.

As you eloquently discuss in the remainder of  the document, each of these issues becomes even more complex with ROS2.

Thus, we might want to revisit whether we can break this up, and create a combination of simpler tools.

articles/150_roslaunch.md
```diff
@@ -0,0 +1,821 @@
...
+Also, since there can be multiple nodes per process, shutting down a node no longer always means sending a unix signal to a single process.
+Other mechanisms might need to be used to have more granular shutdown control in multi-node processes.
+
+#### Launching Nodes (Processes) Remotely and Portability
+
+The launch system in ROS 1 only really ever was supported on Linux and other Unix-like operating systems like BSD and macOS.
+These machines all have SSH, which is the mechanism which is specifically called out to be used when launching processes on remote machines.
+It also played a role in defining what you specified and how when configuring `roslaunch` from ROS 1 to be able to launch processes on remote machines.
+
+In ROS 2, Windows has been added to the list of targeted platforms, and as of the writing of this document it does not support SSH natively.
```
josephduchesne, line 88: Would requiring the execution of a launch agent process on each additional available machine be a good solution for this? The agent could register its hostname, ip etc. (via a normal ros topic maybe?), which could then be used by the primary launch processes to delegate launches on other machines.

Would people be OK with each additional computer requiring some sort of ros2 launch client running? There could be some system that allows waiting for worker clients, handling events on an expected 2nd(+) computer failing to be present (e.g. timeout, connection loss).

Another option would be to allow launching the remote launch agent remotely (like Jenkins does with its agents). That way we could have a ssh launched worker on linux/mac, psexec launched processes on Windows. This does raise further questions about cross platform support.

If the default behavior is that the agent is launched somehow(daemon, cron, service, manually, etc.) on each secondary machine, cross-platform behavior would be identical. This would be portable but a departure from the convenience of ros1's ability to launch everything at once (despite the shortcomings of that specific implementation in error handling and recovery).

If there is a system to automatically launch such agents from the primary machine via an OS specific handler, things might get very complicated to implement and test (windows->linux launches for example).

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,255 @@
...
+</group>
+```
+
+## Dynamic Configuration
+
+### Substitution Syntax
+
+All substitutions are enclosed by `$(...)`.
+
+### Built-in Substitutions
```
dirk-thomas, line 240: How about other substitutions available in ROS 1? Have they been skipped intentionally or just to reduce the scope? If it is the latter it might make sense to add a list of substitutions which are planned to be added.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,255 @@
...
+
+`$(find-package <pkg-name>)`
+: Substituted by the path to package installation directory in the local filesystem.
```
dirk-thomas: What does "package installation directory" mean here? If it I the install prefix path of the package then I would suggest to use that term.

articles/151_roslaunch_xml.md
ivanpauno: I added some comments about that in https://github.com/ros2/ros2_documentation/pull/302.

articles/151_roslaunch_xml.md
wjwwood: It used this method, so yes @dirk-thomas's terminology is correct:

https://github.com/ament/ament_index/blob/0dba2f42c7f9b18a676c96d5643f26d5213513dd/ament_index_python/ament_index_python/packages.py#L38

It would be nice to also have a `$(find-package-share <pkg-name>)` as well as it would prevent the users from writing the package name twice, e.g.:

```
$(find-package really_long_package_name)/share/really_long_package_name/...
```

versus

```
$(find-package-share really_long_package_name)/...
```

josephduchesne: What are your thoughts on checking for file existance as part of a substitution? I've come across numerous instances where this would have been helpful over the years (e.g. large multi-robot project configuration where some robots override configuration and others don't).

josephduchesne: What are your thoughts on checking for file existance as part of a substitution? I've come across numerous instances where this would have been helpful over the years (e.g. large multi-robot project configuration where some robots override configuration and others don't).

wjwwood: We could have a `$(find-file-in-package[-share] <pkg> <relative_path> [<default>])`, though we don't have one now. Or we could have a few small, sharp tools instead, like `$(file-exists $(find-package-share my_pkg)/some_launch.xml)`.

articles/150_roslaunch.md
wjwwood: 3473687

articles/150_roslaunch.md
wjwwood: I added a note that this is not specified in this document, basically kicking the can down the road to some other lifecycle related documentation: 7d6c5fa

articles/150_roslaunch.md
wjwwood: I think there's already a note that lifecycle stuff is out of scope for this document and how it is actually achieved and by who (launch, user, autonomously?) will be determined by some other lifecycle specific documentation.

articles/150_roslaunch.md
wjwwood: This has been resolved by our implementation of components in a separate package `launch_ros` and also due to the introduction of the concept of launch frontend and the XML frontend specifically.

articles/150_roslaunch.md
wjwwood: I tried to wrap this up in 2964a67.

articles/150_roslaunch.md
wjwwood: I don't think the intention is to discourage writing nodes which are decoupled and asynchronous, just that in a few situations having a coordinated startup is very important and in ROS 1 the system worked against you trying to do that. Hopefully the new launch system provides the ability to have some more sophisticated startup behavior without changing the common practice or having nodes which are decoupled from their peers.

It is quite a bit harder to write launch files (in Python or the new XML frontend) which operate in an "if this then that" manner, so I don't think people will default to that pattern, instead choosing only to use it if they need to do so.

ivanpauno:
I don't know if this is ready to be merged or not, but it would be good if we merge `ros2 launch` related design documents.
Ongoing discussions could continue here after being merged, or in follow-up issues.

@wjwwood if you think a review is needed before merging, let me know and I will do it.

wjwwood:
Let me have a once more over it. I was planning on merging it a few weeks ago and it got lost under the avalanche.

wjwwood:
Thanks for the feedback everyone. I'm going to merge this so we can continue to build on it with more pull requests. Feel free to continue discussion here or open new pull requests as needed to extend this document.

KenwoodFox:
Did https://github.com/ros2/launch/issues/31 ever get looked at here? I noticed it had a few comments linking back, robot_upstart was for ros1 iirc, is it still the recomended course of action to leverage things like initd or systemd for running launch files at boot?

PR #168, opened 2018-04-16
update universal build tool article to reflect decision to use colcon

dirk-thomas:
Follow up of #115 which created the first version of the article.

This update describes the new alternative [colcon](https://github.com/colcon/), the decision process we followed as well as the decision to continue with `colcon` as the universal build tool in ROS (2).

### DIFF
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -39,7 +39,7 @@ This article describes the steps to unify these build tools as well as extend th
 
 The goal of a unified build tool is to build a set of packages with a single invocation.
 It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
-It should also work with packages that do not provide manifest files themselves, given that the necessary meta information is provided externally.
+It should also work with packages that do not provide manifest files themselves, given that the necessary meta information can be inferred and/or is provided externally.
 This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its ignition dependencies, sdformat, etc.).
 
 In the ROS ecosystems several tools already exist which support this use case (see below).
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -40,7 +40,7 @@ This article describes the steps to unify these build tools as well as extend th
 The goal of a unified build tool is to build a set of packages with a single invocation.
 It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
 It should also work with packages that do not provide manifest files themselves, given that the necessary meta information is provided externally.
-This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its ignition dependencies).
+This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its ignition dependencies, sdformat, etc.).
 
 In the ROS ecosystems several tools already exist which support this use case (see below).
 Each of the existing tools performs similar tasks and duplicates a significant amount of the logic.
@@ -381,12 +381,12 @@ For option **A)** the follow items should be considered:
 
 - Support for Python packages using a `setup.cfg` file.
 - Support for `PowerShell` to work around length limitations for environment variable on Windows.
+- Support for `Pytest` to run Python tests (instead of using `nose`).
 - Support to pass package specific argument.
-- Remove support for the *devel space* concept in ROS 1.
 - Update code base to Python 3.5+.
 - Refactor code base to reduce coupling (e.g. separate [API](https://github.com/catkin/catkin_tools/blob/2cae17f8f32b0193384d2c7734afee1c60c4add2/catkin_tools/execution/controllers.py#L183-L205) for output handling).
 - Additional functionality to build Gazebo including its dependencies.
-- Whether or not to continue supporting the *devel space*.
+- Whether or not to continue supporting the *devel space* concept in ROS 1.
 
 For option **B)** the follow items should be considered:
 
@@ -420,9 +420,9 @@ The separation of the build tool name from the supported build systems as well a
 
 The following next steps will happen before the next ROS 2 release *Bouncy*.
 
+- The ROS 2 buildfarm(s) will be updated to use `colcon` and provide devel / PR / prerelease jobs for the *Bouncy* release.
 - The instructions in the ROS 2 wiki to build from source will be updated to use `colcon` instead.
 - The `ament_tools` repository will be archived, removed from the `ros2.repos` file, and won't be released into *Bouncy*.
-- The ROS 2 buildfarm(s) will be updated to use `colcon` and provide devel / PR / prerelease jobs for the *Bouncy* release.
 
 ### Implications
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -334,7 +334,7 @@ Both of the considered options have unique and valuable features and there are g
 Since both are written in Python either of the two tools could be "transformed" to cover the pros of the other one.
 So the two important criteria for the decision are:
 
-- the effort it takes to so (in the short term as well as in the long term) and
+- the effort it takes to do (in the short term as well as in the long term) and
 - the difference of the resulting code base after the "transformation" is completed.
 
 ### Immediate goals
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -394,7 +394,7 @@ For option **B)** the follow items should be considered:
 - Support `DESTDIR`.
 - Support a feature similar to the `profile` verb of `catkin_tools`.
 
-## Decision
+## Summary and Decision
 
 Based on the above information we made the not easy decision to pick `colcon` as the universal build tool.
 
@@ -407,3 +407,26 @@ The necessary up front development effort for `catkin_tools` to achieve the goal
 While the short term advantages are certainly a main reason why we decided in favor of `colcon` they are not the only ones.
 The cleaner architecture, modularity and extensibility as well as Python 3.5 code base will be valuable long term benefits when developing this tool in the future.
 The separation of the build tool name from the supported build systems as well as the separation from being a "ROS-only" tool will hopefully also help users to understand the difference and attract new users and potential contributors.
+
+### Next steps
+
+The following next steps will happen before the next ROS 2 release *Bouncy*.
+
+- The instructions in the ROS 2 wiki to build from source will be updated to use `colcon` instead.
+- The `ament_tools` repository will be archive, removed from the `ros2.repos` file, and won't be released into *Bouncy*.
+- The ROS 2 buildfarm(s) will be updated to use `colcon` and provide devel / PR / prerelease jobs for the *Bouncy* release.
+
+### Implications
+
+The following items briefly enumerate what This means for ROS developers and users:
+
+- **No CMake code** of any ROS 2 (or ROS 1) package **needs to be changed** for this transition.
+- When building and testing ROS 2 the command `colcon build` / `colcon test` will be used instead of `ament build` / `ament test`.
+  Please see the [documentation](http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) how to map `ament_tools` command line arguments to `colcon` arguments.
+- For ROS 1 nothing is changing at this point in time.
+
+### Outlook
+
+- Since `colcon` can be used to build ROS 1 early adopters can try to use it to build ROS 1 from source.
+  While there is documentation how to migrate from [catkin_make_isolated](http://colcon.readthedocs.io/en/latest/migration/catkin_make_isolated.html) and [catkin_tools](http://colcon.readthedocs.io/en/latest/migration/catkin_tools.html) `colcon` won't be the recommended build tool in ROS 1 for the foreseeable future.
+- If a test buildfarm using `colcon` proofs to deliver the exact same results as the ROS 1 buildfarm using `catkin_make_isolated` it might be changed to use `colcon` in the future to benefit from the features `colcon` provides (like non-interleaved output per package when building in parallel, per package log files, etc.).
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -135,7 +135,7 @@ The latest format of the manifest file is specified in the [ROS REP 149](http://
 The main difference between `ament_cmake` and `catkin` is described in [another article](http://design.ros2.org/articles/ament.html).
 In the context of the build tool the biggest difference is that `ament_cmake` generates package-specific files to setup the environment to use the package after it has been built and installed.
 
-A package using `ament_cmake` uses the same manifest file as `catkin` (except that it format version 2 or higher).
+A package using `ament_cmake` uses the same manifest file as `catkin` (except that it requires format version 2 or higher).
 
 ### Python setuptools
 
@@ -192,7 +192,7 @@ The tool supports building CMake packages and builds them in isolation as well a
 It was developed to bootstrap the ROS 2 project, is therefore only targeting Python 3, and works on Linux, MacOS and Windows.
 In addition to CMake packages it also supports building Python packages and can infer meta information without requiring an explicit package manifest (which is e.g. used for the FastRTPS package).
 The tool performs an "isolated" build like `catkin_make_isolated` and `catkin_tools` (one CMake invocation per package) and also parallelizes the build of packages which have no (recursive) dependencies on each other (like `catkin_tools`).
-While is coveres more build systems and platforms than `catkin_tools` it doesn't have any of `catkin_tools`s usability features like profiles, output handling, etc.
+While it covers more build systems and platforms than `catkin_tools` it doesn't have any of `catkin_tools`s usability features like profiles, output handling, etc.
 
 `ament_tools` supports building the following packages:
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -249,7 +249,7 @@ After the build a single file can be sourced / invoked to setup the environment 
 Invoking a build system for a package implies also setting up environment variables before the process, e.g. the `CMAKE_PREFIX_PATH`.
 It should be possible for developers to manually invoke the build system for one package.
 The environment variable might be partially different from the environment variables necessary to use a package after it has been built.
-To make that convenient the tool should provide an easy to use mechanism to setup the development environment necessary to manually invoke the build system.
+To make that convenient the tool should provide an easy-to-use mechanism to setup the development environment necessary to manually invoke the build system.
 
 ### Beyond Building
 
@@ -327,7 +327,7 @@ B.  Use `colcon` as a starting point
 
 If this topic would have been addressed earlier some of the duplicate effort could have likely been avoided.
 When the work towards a universal build tool was suspended over a year ago it was a conscious decision based on available resources.
-Nevertheless move forward with a decision now will at least avoid further uncertainty and duplication.
+Nevertheless moving forward with a decision now will at least avoid further uncertainty and effort duplication.
 
 Both of the considered options have unique and valuable features and there are good arguments to build our future development on either of the two tools.
 Since both are written in Python either of the two tools could be "transformed" to cover the pros of the other one.
@@ -426,7 +426,7 @@ The following next steps will happen before the next ROS 2 release *Bouncy*.
 
 ### Implications
 
-The following items briefly enumerate what This means for ROS developers and users:
+The following items briefly enumerate what this decision means for ROS developers and users:
 
 - **No CMake code** of any ROS 2 (or ROS 1) package **needs to be changed** for this transition.
 - When building and testing ROS 2 the command `colcon build` / `colcon test` will be used instead of `ament build` / `ament test`.
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -346,7 +346,7 @@ For the last ROS 2 release *Ardent* the buildfarm [build.ros.org](http://build.r
 Neither *devel* jobs or *pull request* jobs are available nor is it supported to build a local *prerelease*.
 For the coming ROS 2 release *Bouncy* these job types should be available to support maintainers.
 
-In ROS 2 *Bouncy* the universal build tool will be the only supported option and `ament_tools` will be archived.
+In ROS 2 *Bouncy* the universal build tool will be the recommended option.
 
 #### Necessary work
 
@@ -422,7 +422,8 @@ The following next steps will happen before the next ROS 2 release *Bouncy*.
 
 - The ROS 2 buildfarm(s) will be updated to use `colcon` and provide devel / PR / prerelease jobs for the *Bouncy* release.
 - The instructions in the ROS 2 wiki to build from source will be updated to use `colcon` instead.
-- The `ament_tools` repository will be archived, removed from the `ros2.repos` file, and won't be released into *Bouncy*.
+
+After the Bouncy release the `ament_tools` repository will be archived, removed from the `ros2.repos` file, and won't be released into future ROS 2 distributions.
 
 ### Implications
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -233,7 +233,6 @@ It is up to the implementation to decide if it only supports the standard CMake 
 
 In ROS 2 the concept of the *devel space* has intentionally been removed.
 In the future it might be feasible to provide the concept of *symlinked installs* in ROS 1 to provide a similar benefit without the downsides.
-Therefore not supporting this concept in the universal build tool is considered a viable path forward.
 
 #### Build ROS 2 workspaces
 
@@ -347,7 +346,7 @@ For the last ROS 2 release *Ardent* the buildfarm [build.ros.org](http://build.r
 Neither *devel* jobs or *pull request* jobs are available nor is it supported to build a local *prerelease*.
 For the coming ROS 2 release *Bouncy* these job types should be available to support maintainers.
 
-In ROS 2 *Bouncy* the univeral build tool will be the only supported option and `ament_tools` will be archived.
+In ROS 2 *Bouncy* the universal build tool will be the only supported option and `ament_tools` will be archived.
 
 #### Necessary work
 
@@ -365,7 +364,7 @@ For option **A)** the follow items would need to be addressed:
 - Support for Windows using `.bat` files.
 - Support for the package manifest format version 3.
 
-For option **B)** the follow items would need to be addressed.
+For option **B)** the follow items would need to be addressed:
 
 - Address user feedback when the tool is being used by a broader audience.
 
@@ -387,16 +386,26 @@ For option **A)** the follow items should be considered:
 - Update code base to Python 3.5+.
 - Refactor code base to reduce coupling (e.g. separate [API](https://github.com/catkin/catkin_tools/blob/2cae17f8f32b0193384d2c7734afee1c60c4add2/catkin_tools/execution/controllers.py#L183-L205) for output handling).
 - Additional functionality to build Gazebo including its dependencies.
+- Whether or not to continue supporting the *devel space*.
 
 For option **B)** the follow items should be considered:
 
 - Support cross compilation.
 - Support `DESTDIR`.
 - Support a feature similar to the `profile` verb of `catkin_tools`.
+- Support for a shared GNU Make job sever.
+- Support for `GNUInstallDirs`
+  - Not sure about the status of this, it would be in `colcon`'s generated shell files if anywhere.
+  - Should have a test for this case.
+- Test for, and fix if necessary, correct topological order with dependencies across workspaces.
+  - See: [https://github.com/ros/catkin/pull/590](https://github.com/ros/catkin/pull/590)
 
 ## Summary and Decision
 
-Based on the above information a decision has been made to pick `colcon` as the universal build tool which was not an easy one.
+Based on the above information a decision has been made to pick `colcon` as the universal build tool.
+
+The decision was made after considering the input of ROS 2 team members and some ROS 1 users.
+The decision was not easy, as it was not unanimous, but the vast majority of input was either pro `colcon` or ambivalent.
 
 To elaborate on the rationale one significant advantage of `colcon` is that it is ready to be deployed for ROS 2 right now and it covers our current use cases.
 Another argument leaning towards `colcon` is the expected little effort to provide devel / PR / prerelease jobs on build.ros2.org across all targeted platforms for the upcoming *Bouncy* release.
@@ -424,6 +433,11 @@ The following items briefly enumerate what This means for ROS developers and use
 - When building and testing ROS 2 the command `colcon build` / `colcon test` will be used instead of `ament build` / `ament test`.
   Please see the [documentation](http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) how to map `ament_tools` command line arguments to `colcon` arguments.
 - For ROS 1 nothing is changing at this point in time.
+- In the future `colcon` will replace `catkin_make_isolated` and `catkin_make` as the recommended build tool for ROS 1.
+  - `colcon` will not support the *devel space* and will require packages to have install rules
+  - `catkin` will likely still support the *devel space*, though it might be removed at some point (that has not been decided yet)
+  - Therefore, it is possible that the default build tool for ROS 1 may not support the *devel space*, though legacy tools will continue to support it.
+  - Note that it is already the case that individual ROS 1 catkin packages may either not have installation rules but support the *devel space*, or they might have installation rules but not properly support the *devel space*.
 
 ### Outlook
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -453,6 +453,6 @@ The following items briefly enumerate what this decision means for ROS developer
 
 ### Outlook
 
-- Since `colcon` can be used to build ROS 1 early adopters can try to use it to build ROS 1 from source.
+- Since `colcon` can be used to build ROS 1, early adopters can try to use it to build ROS 1 from source.
   While there is documentation how to migrate from [catkin_make_isolated](http://colcon.readthedocs.io/en/latest/migration/catkin_make_isolated.html) and [catkin_tools](http://colcon.readthedocs.io/en/latest/migration/catkin_tools.html) `colcon` won't be the recommended build tool in ROS 1 for the foreseeable future.
 - If a test buildfarm using `colcon` proofs to deliver the exact same results as the ROS 1 buildfarm using `catkin_make_isolated` it might be changed to use `colcon` in the future to benefit from the features `colcon` provides (like non-interleaved output per package when building in parallel, per package log files, etc.).
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -3,7 +3,7 @@ layout: default
 title: A universal build tool
 permalink: articles/build_tool.html
 abstract:
-  This article describes a universal build tool for ROS 1 and ROS 2.
+  This article describes the rationale for a universal build tool.
 published: true
 author: '[Dirk Thomas](https://github.com/dirk-thomas)'
 categories: Overview
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -3,7 +3,7 @@ layout: default
 title: A universal build tool
 permalink: articles/build_tool.html
 abstract:
-  This article describes the rationale for a universal build tool.
+  This article describes a universal build tool for ROS 1 and ROS 2.
 published: true
 author: '[Dirk Thomas](https://github.com/dirk-thomas)'
 categories: Overview
@@ -19,15 +19,6 @@ categories: Overview
 
 Original Author: {{ page.author }}
 
-<div class="alert alert-info" markdown="1">
-  <b>NOTE:</b>
-  The work towards a unified build tool have been reprioritized.
-  Nobody is currently planning to spend any time towards the described goals.
-
-  While the article in its current state describes the use cases, goals as well as the rational behind it, and mentions possible approaches moving forward, it does not propose a specific path forward at the moment.
-
-</div>
-
 ## Preface
 
 In the ROS ecosystem the software is separated into numerous packages.
@@ -40,7 +31,7 @@ Such a workflow is impracticable at scale without a tool that automates that pro
 
 A build tool performs the task of building a set of packages with a single invocation.
 For ROS 1 multiple different tools provide support for this, namely `catkin_make`, `catkin_make_isolated`, and `catkin_tools`.
-For ROS 2 the build tool providing this functionality is called `ament_tools`.
+For ROS 2 up to the Ardent release the build tool providing this functionality is called `ament_tools`.
 
 This article describes the steps to unify these build tools as well as extend the field of application.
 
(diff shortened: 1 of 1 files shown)


articles/101_build_tool.md
```diff
@@ -40,23 +31,23 @@ Such a workflow is impracticable at scale without a tool that automates that pro
...
 This article describes the steps to unify these build tools as well as extend the field of application.
 
 ## Goal
 
 The goal of a unified build tool is to build a set of packages with a single invocation.
 It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
-It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
-This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its dependencies).
+It should also work with packages that do not provide manifest files themselves, given that the necessary meta information is provided externally.
+This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its ignition dependencies).
```
mikaelarguedas: Why is this now restricted to ignition deps? this wouldnt work for e.g. sdformat?

articles/101_build_tool.md
```diff
@@ -136,15 +127,15 @@ It automates the generation of CMake config files as well as pkg-config files.
...
+The latest format of the manifest file is specified in the [ROS REP 149](http://www.ros.org/reps/rep-0149.html).
 
 ### ament_cmake
 
 [ament_cmake](https://github.com/ament/ament_cmake) is an evolution of `catkin` and is also based on CMake.
 The main difference between `ament_cmake` and `catkin` is described in [another article](http://design.ros2.org/articles/ament.html).
 In the context of the build tool the biggest difference is that `ament_cmake` generates package-specific files to setup the environment to use the package after it has been built and installed.
 
-A package using `ament_cmake` uses the same manifest file as `catkin` (except that it only allows the newer format version 2).
+A package using `ament_cmake` uses the same manifest file as `catkin` (except that it format version 2 or higher).
```
mikaelarguedas: missing words in refactoring ?

articles/101_build_tool.md
```diff
@@ -199,8 +190,9 @@ The tool supports building CMake packages and builds them in isolation as well a
 
 `ament_tools` is provided by a standalone Python 3 package used to build ROS 2 packages.
 It was developed to bootstrap the ROS 2 project, is therefore only targeting Python 3, and works on Linux, MacOS and Windows.
-In addition to CMake packages it also supports building Python packages and can infer meta information without requiring an explicit package manifest.
+In addition to CMake packages it also supports building Python packages and can infer meta information without requiring an explicit package manifest (which is e.g. used for the FastRTPS package).
 The tool performs an "isolated" build like `catkin_make_isolated` and `catkin_tools` (one CMake invocation per package) and also parallelizes the build of packages which have no (recursive) dependencies on each other (like `catkin_tools`).
+While is coveres more build systems and platforms than `catkin_tools` it doesn't have any of `catkin_tools`s usability features like profiles, output handling, etc.
```
mikaelarguedas: not sure what you meant here, "while it covers" maybe?

articles/101_build_tool.md
```diff
@@ -210,9 +202,15 @@ The tool performs an "isolated" build like `catkin_make_isolated` and `catkin_to
 - Python package with a `package.xml` file.
 - Python package without a manifest file (extracting the package name and dependencies from the `setup.py` file).
 
+### colcon
+
+When the first draft of this article was written the conclusion was to not to spend any resources towards a universal build tool.
+As a consequence the author of this article went ahead and developed [colcon](https://github.com/colcon/) as a personal project.
+Therefore its feature set is closely aligned with the following requirements.
```
mikaelarguedas: Should we highlight here that this doesn't have usability features of catkin_tools either such as profiles (like it's been mentioned above for ament_tools)?

articles/101_build_tool.md
dirk-thomas: It is not "restricted". It is just an example:

> e.g. ...

articles/101_build_tool.md
dirk-thomas: It is already mentioned [above](https://github.com/ros2/design/pull/168/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR195) and [below](https://github.com/ros2/design/pull/168/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR395).

articles/101_build_tool.md
dirk-thomas: Fixed in 553d408a0762581565bc0aacd4e2032bbde864cd.

articles/101_build_tool.md
dirk-thomas: Fixed in 553d408a0762581565bc0aacd4e2032bbde864cd.

articles/101_build_tool.md
mikaelarguedas: I think that's my point... You added it above for `ament_tools` and not here while it applies to both.

The fact the selecting colcon requires that feature to be listed in the "Future" section doesn't justify not mentioning it here in my opinion.

dirk-thomas: Please feel free to commit any proposal you think makes it better.

gbiggs:
Something important that is missing from this discussion (not necessarily the document) is information about what the change to colcon actually means. In particular, what happens to ament? Is it replaced entirely? Or is colcon expected to sit on top of it? Perhaps a mixture, with some native colcon packages and some using ament, especially the ament CMake macros? I did not find this clearly in the document.

How much work is expected for users to change to using colcon?

dirk-thomas:
> In particular, what happens to ament? Is it replaced entirely? Or is colcon expected to sit on top of it? Perhaps a mixture, with some native colcon packages and some using ament, especially the ament CMake macros?

The article already tries to describe clearly the difference between the build system and the build tool. The former one is `CMake`, `catkin`, `ament_cmake` or `Python setuptools`. Those stay as they are - no ROS package needs to be changed. The later one is being replaced by `colcon` and `ament_tools` will be obsolete (mentioned [here](https://github.com/ros2/design/pull/168/files#diff-361c5cb5b98ec5e7f690f2dbce2cd7fdR350)).

Since I am the author of the article as well as the tool it is difficult for me to "see" what needs to be described better. Please feel free to propose changes to the article to improve it and clarify where necessary.

> How much work is expected for users to change to using colcon?

That means the user "only" has to learn how to invoke the new build tool. [This page](http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) gives some advice how to transition from `ament_tools` arguments to `colcon` arguments. Beside that `colcon --help` / `colcon <verb> --help` aim to provide very detailed information.

articles/101_build_tool.md
```diff
@@ -210,9 +202,15 @@ The tool performs an "isolated" build like `catkin_make_isolated` and `catkin_to
 - Python package with a `package.xml` file.
 - Python package without a manifest file (extracting the package name and dependencies from the `setup.py` file).
 
+### colcon
+
+When the first draft of this article was written the conclusion was to not to spend any resources towards a universal build tool.
+As a consequence the author of this article went ahead and developed [colcon](https://github.com/colcon/) as a personal project.
```
tkruse: While it is interesting for me to see how decisions are made at OSRF, maybe this part could still just state the relevant technical facts, such as: "The discussion about the first draft was inconclusive. Colcon was created as a PoC to demonstrate the viability of the approach of a unified build tool layer above respective build-system libraries."

articles/101_build_tool.md
```diff
@@ -289,59 +291,119 @@ The following items are possible extension points to provide custom functionalit
...
 
-Since `catkin_tools` is in many aspects the most complete build tool it should be the one being evolved.
+Since `catkin_tools` is in many aspects the most complete ROS build tool it should be the one being evolved.
 While `ament_tools` has a few features `catkin_tools` currently lacks (e.g. plain CMake support without a manifest, Windows support) the feature richness of `catkin_tools` makes it a better starting point.
 
-The following items are highlighting some of the necessary efforts (not a complete list):
+### Start "from scratch" / colcon
+
+Since the first draft of this article the `colcon` project has been developed with the goals and requirements of a universal build tool in mind.
+In its current form it is already able to build ROS 1 workspaces, ROS 2 workspaces, as well as Gazebo including its ignition dependencies.
```
tkruse, line 316: I assume that colcon cannot build workspaces with a mix of ros1 and ros2 packages, since this was a limitation discussed originally. I that true? I did not see it mentioned in the changes or the colcon manuals. Might just help managing expectations to state it explicitly.

articles/101_build_tool.md
```diff
@@ -289,59 +291,119 @@ The following items are possible extension points to provide custom functionalit
...
+- Support cross compilation.
+- Support `DESTDIR`.
+- Support a feature similar to the `profile` verb of `catkin_tools`.
 
-- Port / reimplement many of the features existing in the other build tools.
-  It will take a non trivial amount of time to reach the feature level of e.g. `catkin_tools`
+## Decision
 
-- Thorough test the functionality and write documentation for developers as well as users.
+Based on the above information we made the not easy decision to pick `colcon` as the universal build tool.
```
tkruse: Who is "we"?

gbiggs:
> Since I am the author of the article as well as the tool it is difficult for me to "see" what needs to be described better. Please feel free to propose changes to the article to improve it and clarify where necessary.

I think it needs a section right at the end, after the decision is discussed, explaining the specific consequences of the decision in a clear format (a list would work fine; items like "ament_tools will be retired", "ament_cmake will be supported by colcon and kept", "you will need to replace execution of ament with execution of colcon", etc.). It doesn't need to be in detail, but unless you clearly state in one place "these are the impacts on ROS 2 development, and these are the impacts on you, as a ROS 2 user", you are going to get this question a lot - and probably a lot of complaints about yet another build tool. I can see that the relevant information is present in the article, now that it's been clarified by you, but it's spread out and not easy to accumulate for the reader. I did actually read the article first, but most users will skim it and miss that information.

dirk-thomas:
> I think it needs a section right at the end ...

Ok, I agree that a summary at the end would help. I tried to capture the "next steps", "implications" and "outlook" in 9974be391470c2b360805a34631c60b57298d830. Please let me know if that addresses your point or suggest additional changes.

articles/101_build_tool.md
```diff
@@ -233,6 +231,10 @@ The following uses cases should be satisfied by the unified build tool.
 The tool needs to be able to build ROS 1 workspaces which can already be built using `catkin_make_isolated` / `catkin_tools`.
 It is up to the implementation to decide if it only supports the standard CMake workflow or also the *custom devel space concept* of `catkin`.
 
+In ROS 2 the concept of the *devel space* has intentionally been removed.
+In the future it might be feasible to provide the concept of *symlinked installs* in ROS 1 to provide a similar benefit without the downsides.
+Therefore not supporting this concept in the universal build tool is considered a viable path forward.
```
wjwwood: For ROS 1 or just ROS 2? If this includes ROS 1 (only thing that make sense from the text), then I don't feel that this was properly resolved.

Though I'm in favor of getting rid of the devel space in ROS 1, I think we cannot commit to doing that (and therefore not supporting it in the build tool) until we propose how to deprecate and then remove the devel space. In my opinion, It should still be a requirement that the default build tool for ROS 1 needs to support the devel space, at least until it is in the process of being removed.

Did others (other than @dirk-thomas) have a different conclusion based on our discussions?

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+Since it hasn't been used by many people yet more advanced features like cross compilation, `DESTDIR`, etc. hasn't been tested (and will therefore likely not work yet).
+
+## Decision process
+
+For the decision process we are only considering the following two options based on the rational described above:
+
+A.  Use `catkin_tools` as a starting point
+B.  Use `colcon` as a starting point
+
+We acknowledge that if this topic would have been addressed earlier that some of the duplicate effort could have been avoided.
```
wjwwood: What duplication of effort could have been avoided by making a decision earlier?

It was always possible (whether officially or in spare time) to either build something new (duplicating effort) or contribute to existing software, or a fork of existing software (avoiding some redundant work).

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+When the work towards a universal build tool was suspended over a year ago it was a conscious decision based on available resources.
+Nevertheless we have to move forward with a decision in order to avoid further uncertainty and duplication.
```
wjwwood: Avoiding uncertainty, I agree with, however.

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+
+Both of the considered options have unique and valuable features and there are good arguments to build our future development on either of the two tools.
+Since both are written in Python either of the two tools could be "transformed" to cover the pros of the other one.
+So the two important criteria for the decision are:
+
+- the effort it takes to so (in the short term as well as in the long term) and
```
wjwwood: This sentence doesn't make sense to me, maybe you mean `to do`, but not sure?

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+A.  Use `catkin_tools` as a starting point
+B.  Use `colcon` as a starting point
+
+We acknowledge that if this topic would have been addressed earlier that some of the duplicate effort could have been avoided.
+When the work towards a universal build tool was suspended over a year ago it was a conscious decision based on available resources.
+Nevertheless we have to move forward with a decision in order to avoid further uncertainty and duplication.
+
+Both of the considered options have unique and valuable features and there are good arguments to build our future development on either of the two tools.
+Since both are written in Python either of the two tools could be "transformed" to cover the pros of the other one.
+So the two important criteria for the decision are:
```
wjwwood: Are these the only criteria? If not, how did we down-select to these being the most important criteria?

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+- Update code base to Python 3.5+.
+- Refactor code base to reduce coupling (e.g. separate [API](https://github.com/catkin/catkin_tools/blob/2cae17f8f32b0193384d2c7734afee1c60c4add2/catkin_tools/execution/controllers.py#L183-L205) for output handling).
+- Additional functionality to build Gazebo including its dependencies.
+
+For option **B)** the follow items should be considered:
 
-- Refactor the software architecture of the existing code base to support the flexibility sketched by the extension points listed above.
+- Support cross compilation.
+- Support `DESTDIR`.
+- Support a feature similar to the `profile` verb of `catkin_tools`.
```
wjwwood: More things:

- job server
- support for `GNUInstallDirs` (not sure if `colcon` already supports this or not, e.g. https://github.com/catkin/catkin_tools/issues/154)
- tests to cover corner cases already found and addressed before, e.g. correct topological order with underlays (https://github.com/ros/catkin/pull/590)

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+  Please see the [documentation](http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) how to map `ament_tools` command line arguments to `colcon` arguments.
+- For ROS 1 nothing is changing at this point in time.
 
-The decision of which approach should be selected is deferred at the moment since nobody is likely able to spend any time on this in the foreseeable future.
-It could also be considered an implementation detail - as long as the described goals are reached it doesn't matter how the unified build tool was getting there.
+### Outlook
 
-*Hopefully to be continued in the future...*
+- Since `colcon` can be used to build ROS 1 early adopters can try to use it to build ROS 1 from source.
+  While there is documentation how to migrate from [catkin_make_isolated](http://colcon.readthedocs.io/en/latest/migration/catkin_make_isolated.html) and [catkin_tools](http://colcon.readthedocs.io/en/latest/migration/catkin_tools.html) `colcon` won't be the recommended build tool in ROS 1 for the foreseeable future.
```
wjwwood: But the implication is that `colcon` will eventually be recommended instead of `catkin_make_isolated` and `catkin_make`, and that `catkin_tools` will not be receiving attention with respect to ROS 2. And because `colcon` will eventually be the recommendation, we will therefore not spend any extra time on `catkin_tools` for ROS 1 either.  Is that not the case?

articles/101_build_tool.md
dirk-thomas: I completely agree with you points. Therefore I don't see `colcon` becoming the recommended tool for ROS 1 until that has been worked out.

dirk-thomas: > For ROS 1 or just ROS 2?

`colcon` doesn't aim to support the concept of a devel space - so that applies for ROS 1 as well as ROS 2.

articles/101_build_tool.md
dirk-thomas: Fixed in 783e1aeeaccef594f55d7de10dc684084097f67d.

articles/101_build_tool.md
dirk-thomas: Imo the amount of time necessary to reach the stated goal is of highest importance. Simply because we are already very thin on resources and progress on feature development is rather slow. So any extra time spend on other items like this will slow our progress down. The second point is the quality of the end result since that will immediately effect the effort necessary to maintain and develop the software in the future. Therefore these are the "two important criteria".

There are certainly more but I didn't enumerate them here. Please feel free to add more from our internal documentation.

articles/101_build_tool.md
dirk-thomas: Yes, that is the case.

articles/101_build_tool.md
dirk-thomas: Please feel free to add them to the article.

articles/101_build_tool.md
dirk-thomas: From my point of view no involved party wanted to commit resources on this topic for a long time. If that would have been different everyone involved could have influenced the direction of the development. Since that was not the case I spend my time on `colcon` because I saw the advantages of starting from scratch. I decided not to contribute to `catkin_tools` since I didn't consider the architecture of the code base to be a good foundation for the goals (the reference API for output handling in the article is just one example). And refactoring the whole code base wasn't in the scope of time I was willing to spend on this.

articles/101_build_tool.md
dirk-thomas: > While it is interesting for me to see how decisions are made at OSRF, maybe this part could still just state the relevant technical facts ...

As any other company OSRF has limited resources. And with that constraint comes the difficult decision what project / part to spend them on. I think it is important to acknowledge that here to provide context to readers why there was no significant development in this area in the past year.

> The discussion about the first draft was inconclusive

From my point of view that does not reflect the facts. The previous draft came up with a clear rational why it would be useful to unify the different tools. At the end of the day no party was willing to put the necessary effort behind the idea to move in any of the proposed directions. That is why the topic was basically "on hold" for over a year. I think it is valuable to include the history of this article to help readers follow how we got into the current state.

> Colcon was created as a PoC to demonstrate the viability of the approach of a unified build tool layer ...

I don't think this is the case either. I have implemented build systems and build tools in the past as well as wrote the initial draft of this article describing that there is a strong need for unifying the development. The technical feasibility wasn't really a concern.

`colcon` was pretty much developed by myself for two reasons:
* curiosity to learn and use new stuff (e.g. `setup.cfg` files), the challenge to design a complex piece of software in a highly modular and extensible way and
* as a "side effect" improve my own workflow to build Gazebo, ROS 2 and ROS 1 which I am using in my day job more efficiently.> While it is interesting for me to see how decisions are made at OSRF, maybe this part could still just state the relevant technical facts ...

As any other company OSRF has limited resources. And with that constraint comes the difficult decision what project / part to spend them on. I think it is important to acknowledge that here to provide context to readers where there was no significant development in this area in the past year.

> The discussion about the first draft was inconclusive

From my point of view that does not reflect the facts. The previous draft came up with a clear rational why it would be useful to unify the different tools. At the end of the day no party was willing to put the necessary effort behind the idea to move in any of the proposed directions. That is why the topic was basically "on hold" for over a year. I think it is valuable to include the history of this article to help readers follow how we got into the current state.

> Colcon was created as a PoC to demonstrate the viability of the approach of a unified build tool layer ...

I don't think this is the case either. I have implemented build systems and build tools in the past as well as wrote the initial draft of this article describing that there is a strong need for unifying the development. The technical feasibility wasn't really a concern.

`colcon` was pretty much developed by myself for two reasons:
* curiosity to learn and use new stuff (e.g. `setup.cfg` files), the challenge to design a complex piece of software in a highly modular and extensible way and
* as a "side effect" to improve my own workflow to build Gazebo, ROS 2 and ROS 1 which I am using in my day job more efficiently.

articles/101_build_tool.md
dirk-thomas: > I assume that colcon cannot build workspaces with a mix of ros1 and ros2 packages, since this was a limitation discussed originally. I that true?

I don't think that is the case. `colcon` doesn't mind what kind of packages it processes in a workspace. Technically I don't see anything within the build tool prohibiting to process a heterogeneous workspace.

But there are other things affecting how useful or how feasible a mixed workspace is. Commonly ROS packages depend on other ROS packages by name, `find_package` them, use their API etc. all of that needs to be "compatible" too in order to successfully build a heterogeneous workspace.

For a simple example: a library exported by a `catkin` package could transparently be used and linked in an `ament_cmake` package. That is more related to the compatibility between both CMake packages than to the build tool. Please feel free to give that use case a try.> I assume that colcon cannot build workspaces with a mix of ros1 and ros2 packages, since this was a limitation discussed originally. I that true?

I don't think that is the case. `colcon` doesn't mind what kind of packages it processes in a workspace. Technically I don't see anything within the build tool prohibiting to do so.

But there are other things affecting how useful or how feasible a mixed workspace is. Commonly ROS package depend on other ROS packages by name, `find_package` them, use their API etc. all of that needs to be "compatible" too in order to successfully build a heterogeneous workspace.

For a simple example: a library exported by a `catkin` package could transparently be used and linked in an `ament_cmake` package. That is more related to the compatibility between both CMake packages then to the build tool. Please feel free to give that use case a try.

articles/101_build_tool.md
wjwwood: > Therefore I don't see colcon becoming the recommended tool for ROS 1 until that has been worked out.

Perhaps I'm the only one (still looking for others input here), but I feel that's an important thing to figure out now, because if the conclusion (at some later date) is that ROS 1 cannot live without the devel space and colon will never support it, I think that reality should be captured here.

If we were to later surmise that "it's unacceptable for the default ROS 1 build tool to not support devel space", then we would have a situation which is at odds with the decision we have now.

I don't suppose we have to sort it out now, but it was a surprise to me when I realized it had not been resolved.

articles/101_build_tool.md
wjwwood: My intention was not to criticize the approach, though I do think iterating on `catkin_tools` would have been a better end result, instead I was just curious how you thought making the decision earlier would have saved code duplication.

articles/101_build_tool.md
dirk-thomas: Let's summarize the "facts" first then:

* The devel space requires ROS packages to do "special" logic (not native to CMake) to support the devel space correctly (e.g. different CMake extra files for devel and install).
* The fact that a package might only work in one of the cases (devel vs. install) if it the code is incorrect is surprising to users (since they might only use one of the two).
* The advantage of not copying resources can be achieved with symlinked installations which doesn't come with any of the above cons.
* The only disadvantage of using symlink install over devel space is that it requires packages to have install rules.

Based on these we decided in ROS 2 to not support the concept of devel space. Imo the same rational applies for a new build tool in ROS 1 (since the facts haven't changed). There would be an easy transition path: to add install rules if they are not already there.

That doesn't imply that we would ever remove the devel space logic from `catkin`. If users would like to continue using it they are fine to do it using the existing build tools. I don't think we will ever remove those since we don't want to break existing workflows. Instead we would just recommend something different. Users just shouldn't expect active development on the not anymore recommended build tools - but they should continue to work.

wjwwood: I think you might be missing my point.

I'm not arguing for or against having the devel space or the build tool support for it.

Instead, my point is that the intended future state ought to be decided on, it sounds like it has already been, and that in the implications section should say something like "therefore the default ROS 1 build tool will not support a devel space only workspaces, i.e. the default build tool will require install rules".

gbiggs: I agree with @wjwwood. We need to have certainty about the intended final result.

I have no problems with the idea of ROS 1 eventually not supporting a devel space. I personally dislike the idea and much prefer ROS 2's approach to supporting a rapid development workspace.

However, the idea that ROS 1 could continue to work with both catkin and the new build tool sounds infeasible to me, if they have different requirements on the CMakeLists.txt files, which is what I think will be the case (please correct me if I'm wrong). Users of the new build tool would probably be fine, but as more and more packages support *only* the new build tool, those still using catkin for their work flow might find it increasingly difficult to stay that way. This seems like it would lead to friction.

gbiggs:
> Ok, I agree that a summary at the end would help. I tried to capture the "next steps", "implications" and "outlook" in 9974be3. Please let me know if that addresses your point or suggest additional changes.

:+1:

articles/101_build_tool.md
tkruse: Maybe replace "As a consequence" with "Meanwhile", to reduce the risk of sounding accusatory.

gbiggs:
This may be tangential to the discussion, but reading the bit about inferring meta information from things like the `CMakeLists.txt` made me wonder: is it going to be possible to reduce or even eliminate the duplication of information between the `package.xml` file and other build-related files? I'm guessing that some things, like maintainer name and email, are going to need the package.xml file, but it would be nice not to have to duplicate dependency information, as an example.

wjwwood:
> This may be tangential to the discussion, but reading the bit about inferring meta information from things like the `CMakeLists.txt` made me wonder: is it going to be possible to reduce or even eliminate the duplication of information between the `package.xml` file and other build-related files? I'm guessing that some things, like maintainer name and email, are going to need the `package.xml` file, but it would be nice not to have to duplicate dependency information, as an example.

I think that needs to be almost completely within the build system (not the build tool), because ideally it should build the same way with or without the build tool.

tkruse:
@wjwwood , @gbiggs is talking about this text about amend_tools : 

> `ament_tools` is ... In addition to CMake packages it also supports building Python packages and can infer meta information without requiring an explicit package manifest

So that seems to relate to the build tool (which makes sense, if this is about inferring the build-sytem invocation order of packages based on their dependencies). But I guess this is limited to evaluating the setup.py, not about the CMakeLists.txt

articles/101_build_tool.md
dirk-thomas: > my point is that the intended future state ought to be decided on, it sounds like it has already been

I didn't intend to decide anything for ROS 1 in this article. The idea was to describe how a path forward could look like. A decision about ROS 1 should be made in an REP and/or discourse discussion.

> the idea that ROS 1 could continue to work with both catkin and the new build tool sounds infeasible to me, if they have different requirements on the CMakeLists.txt files, which is what I think will be the case (please correct me if I'm wrong). Users of the new build tool would probably be fine, but as more and more packages support only the new build tool, those still using catkin for their work flow might find it increasingly difficult to stay that way. This seems like it would lead to friction.

I am not sure a new recommended build tool not supporting the devel space has a big impact here. As I mentioned in [this comment](https://github.com/ros2/design/pull/169#discussion_r182474264) this kind of problem already exists in ROS 1 today. Some packages do only work in the devel space and some do only work in the install space (since these packages haven't been developed with both in mind and the developers / maintainers / users only use one of the two). The choice of build tool doesn't change that. The only difference would be that recommending the new build tool without support for devel spaces will likely shift the ratio to more packages working in install space.

articles/101_build_tool.md
dirk-thomas: Just for ref: #169.

dirk-thomas:
> But I guess this is limited to evaluating the setup.py, not about the CMakeLists.txt

That guess is incorrect. Even `ament_tools` extracts the project / package name as well as dependencies from the CMake code in case no manifest is present (see https://github.com/ament/ament_tools/blob/669f335018341097f282097fba9627c560af43f0/ament_tools/package_types/cmake.py#L75-L100).

But lets exclude that topic form this discussion since it is not related to the actual question of the build tool choice.

articles/101_build_tool.md
dirk-thomas: > Maybe replace "As a consequence" with "Meanwhile", ...

Updated in d326e110396d2640aad629d41cfa713598cf0cb5.

articles/101_build_tool.md
gbiggs: Is there any data available on what the ratio is between one or the other style?

dirk-thomas: > Is there any data available on what the ratio is between one or the other style?

Not that I know of.

wjwwood:
I see what you mean now @gbiggs, thanks @tkruse. While that feature is convenient, I personally feel that it is fragile because there are so many cases where automatically extracting all the dependencies, especially in CMake, is not possible, and the `package.xml` allows you to be explicit about what you want to do. Also, there's a lot of value in just parsing the `package.xml`, otherwise every tool needs the more complex logic to extract the requisite information from multiple places.

gbiggs:
Fair enough. We'll just have to continue accepting it as the price to pay for managing the packages properly.

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+
+Since the first draft of this article the `colcon` project has been developed with the goals and requirements of a universal build tool in mind.
+In its current form it is already able to build ROS 1 workspaces, ROS 2 workspaces, as well as Gazebo including its ignition dependencies.
+It uses Python 3.5+ and targets all platforms supported by ROS: Linux, macOS, and Windows.
+
+Since it hasn't been used by many people yet more advanced features like cross compilation, `DESTDIR`, etc. hasn't been tested (and will therefore likely not work yet).
+
+## Decision process
+
+For the decision process only the following two options are being considering based on the rational described above:
```
esteve: ... based on the rational**e** ...

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
-- Move `catkin` specific concepts out of the core of the build tool into a catkin specific extension (e.g. manifest format, *devel space*).
+## Summary and Decision
 
-- Support for ROS 2 which includes:
+Based on the above information a decision has been made to pick `colcon` as the universal build tool which was not an easy one.
 
-  - Support for Python 3 and Windows.
-  - Support for pure Python packages as well as packages without an in-source manifest file.
-  - Environment setup of `ament` packages
+To elaborate on the rational one significant advantage of `colcon` is that it is ready to be deployed for ROS 2 right now and it covers our current use cases.
```
esteve: ... on the rational**e** ...

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+### Next steps
 
-Another approach is to implement the necessary software architecture to enable the desired flexibility and modularity "from scratch".
-Then fill in the features step-by-step by porting existing building blocks from the existing solutions.
-Some items to highlight the necessary efforts (not a complete list):
+The following next steps will happen before the next ROS 2 release *Bouncy*.
 
-- Create the software architecture to support the flexibility sketched by the extension points listed above which will be easier "from scratch" than for an existing code base.
+- The instructions in the ROS 2 wiki to build from source will be updated to use `colcon` instead.
+- The `ament_tools` repository will be archive, removed from the `ros2.repos` file, and won't be released into *Bouncy*.
```
esteve: **archived**

articles/101_build_tool.md
dirk-thomas: Fixed in 2e9804e00fc8775f8c10d80cb5f647f3533078b9.

articles/101_build_tool.md
dirk-thomas: Fixed in 2e9804e00fc8775f8c10d80cb5f647f3533078b9.

articles/101_build_tool.md
dirk-thomas: Fixed in 2e9804e00fc8775f8c10d80cb5f647f3533078b9.

articles/101_build_tool.md
```diff
@@ -289,59 +291,142 @@ The following items are possible extension points to provide custom functionalit
...
+- Support to pass package specific argument.
+- Remove support for the *devel space* concept in ROS 1.
+- Update code base to Python 3.5+.
+- Refactor code base to reduce coupling (e.g. separate [API](https://github.com/catkin/catkin_tools/blob/2cae17f8f32b0193384d2c7734afee1c60c4add2/catkin_tools/execution/controllers.py#L183-L205) for output handling).
+- Additional functionality to build Gazebo including its dependencies.
+
+For option **B)** the follow items should be considered:
 
-- Refactor the software architecture of the existing code base to support the flexibility sketched by the extension points listed above.
+- Support cross compilation.
```
mikaelarguedas: I remember from offline discussion that there was no feature to be added here but just a lack of testing.
I recommend splitting the "missing features" from the "untested features" as it gives a better idea of how much work is still needed.

As a data point I cross-compiled ardent following [these instructions](https://github.com/ros2-for-arm/ros2/wiki/ROS2-on-arm-architecture) with colcon and it built successfully. I didnt have a board around to do runtime testing though

dirk-thomas: Thank you for trying it. I removed the bullet for now (3b1925a282d3bf128e875ab17713ac591ca82729) since there seems to be no problem. If it turns out that there is still a problem it can be addressed as a bug fix.

dirk-thomas:
In the meantime `colcon` has been used for daily development and many issues have been addressed as well as improvements implemented. Python packages are available from [PyPI](https://pypi.org/search/?q=colcon) as well Debian packages from the ROS apt repos. ci.ros2.org has been migrated some time ago and devel / PR jobs on build.ros2.org are about to be deployed. I added a few links in 50a158913e1e691eea7a9283ed9a95b28e5603fd.

Since it is the plan to use `colcon` for the Bouncy release (~~with a decision still pending if we should not release `ament_tools` or to keep it around for this release~~ we will still release `ament_tools` into Bouncy even though not "using" it in the instructions) I put the PR back in review.

articles/101_build_tool.md
```diff
@@ -3,7 +3,7 @@ layout: default
 title: A universal build tool
 permalink: articles/build_tool.html
 abstract:
-  This article describes the rationale for a universal build tool.
+  This article describes a universal build tool for ROS 1 and ROS 2.
```
mikaelarguedas: as part of the goals is to have this build tool ROS independent. Do we need to narrow it to ROS 1 and ROS 2 here?

articles/101_build_tool.md
```diff
@@ -40,23 +31,23 @@ Such a workflow is impracticable at scale without a tool that automates that pro
...
 This article describes the steps to unify these build tools as well as extend the field of application.
 
 ## Goal
 
 The goal of a unified build tool is to build a set of packages with a single invocation.
 It should work with ROS 1 packages as well as ROS 2 packages which provide the necessary information in their manifest files.
-It should also work with packages that do not provide manifest files themselves, given that the meta information is externally provided.
-This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its dependencies).
+It should also work with packages that do not provide manifest files themselves, given that the necessary meta information is provided externally.
+This will allow the build tool to be utilized for non-ROS packages (e.g. Gazebo including its ignition dependencies, sdformat, etc.).
```
mikaelarguedas: Not sure if it's part of the goals but I'm mentionning it as its part of the ament_tools and colcon featureset:
- infer dependencies form the CMakeLists for packages that don't provide manifests and no information is provided externally

articles/101_build_tool.md
mikaelarguedas: Should we add a sentence here to highlight the state of colcon regarding support of heterogeneous workspaces?

articles/101_build_tool.md
```diff
@@ -289,59 +290,168 @@ The following items are possible extension points to provide custom functionalit
...
 While `ament_tools` has a few features `catkin_tools` currently lacks (e.g. plain CMake support without a manifest, Windows support) the feature richness of `catkin_tools` makes it a better starting point.
 
-The following items are highlighting some of the necessary efforts (not a complete list):
+### Start "from scratch" / colcon
+
+Since the first draft of this article the `colcon` project has been developed with the goals and requirements of a universal build tool in mind.
+In its current form it is already able to build ROS 1 workspaces, ROS 2 workspaces, as well as Gazebo including its ignition dependencies.
+It uses Python 3.5+ and targets all platforms supported by ROS: Linux, macOS, and Windows.
+
+Since it hasn't been used by many people yet more advanced features like cross compilation, `DESTDIR`, etc. hasn't been tested (and will therefore likely not work yet).
```
mikaelarguedas, line 319: I didnt try cross-compilation since my original review but last time I tried it seemed to  work as expected

articles/101_build_tool.md
```diff
@@ -289,59 +290,168 @@ The following items are possible extension points to provide custom functionalit
...
+
+A ROS 2 developer currently builds a steadily growing workspace with ROS 2 packages.
+The same is happening in the monolithic Jenkins jobs on [ci.ros2.org](https://ci.ros2.org) (with the advantage to test changes across repositories easily).
+Therefore features to easily filter the packages which need to be build are eagerly awaited to improve the development process.
+
+For the last ROS 2 release *Ardent* the buildfarm [build.ros.org](http://build.ros2.org) only provides jobs to generate Debian packages.
+Neither *devel* jobs or *pull request* jobs are available nor is it supported to build a local *prerelease*.
+For the coming ROS 2 release *Bouncy* these job types should be available to support maintainers.
+
+In ROS 2 *Bouncy* the universal build tool will be the only supported option and `ament_tools` will be archived.
```
mikaelarguedas: In Bouncy, while encouraging users to use the universal build tool, `ament_tools` will still be released in Bouncy to ease user transition and the repository cannot be archived until the distributions using it are all EOL (in this case June 2019)

articles/101_build_tool.md
```diff
@@ -289,59 +290,168 @@ The following items are possible extension points to provide custom functionalit
...
 
-- Support for ROS 2 which includes:
+The following next steps will happen before the next ROS 2 release *Bouncy*.
 
-  - Support for Python 3 and Windows.
-  - Support for pure Python packages as well as packages without an in-source manifest file.
-  - Environment setup of `ament` packages
+- The ROS 2 buildfarm(s) will be updated to use `colcon` and provide devel / PR / prerelease jobs for the *Bouncy* release.
+- The instructions in the ROS 2 wiki to build from source will be updated to use `colcon` instead.
+- The `ament_tools` repository will be archived, removed from the `ros2.repos` file, and won't be released into *Bouncy*.
```
mikaelarguedas: Same remark as above

mikaelarguedas (commented):
pushed a new nitpicks in https://github.com/ros2/design/pull/168/commits/6f79a7a982b1c49313c562a7889e9be660aa82ab, feel free to revert any of them if they are not appropriate.

This looks good to me overall, I made a couple comments below

articles/101_build_tool.md
dirk-thomas: Updated in 81b8dbd43ef64c29dcdfc2a6e11246ff4775eb44.

articles/101_build_tool.md
dirk-thomas: Updated in 81b8dbd43ef64c29dcdfc2a6e11246ff4775eb44.

articles/101_build_tool.md
dirk-thomas: Until we have exercised this I think mentioning this is the "defensive" approach.

articles/101_build_tool.md
dirk-thomas: Mentioned the option to infer metadata in 242c139f62a4f969491aa69dc84ff0838678f9a9.

articles/101_build_tool.md
dirk-thomas: I don't mind either or. I reverted the change to the abstract sentence to keep it as is: c57a4682e388f0605bc4a8cc8f93937636d62e8b.

dirk-thomas:
> pushed a new nitpicks in 6f79a7a, feel free to revert any of them if they are not appropriate.

:+1:

> I made a couple comments below

I replied to / addressed all your recent comments. Ready for review again.

mikaelarguedas (approved):
reads well to me !

ros-discourse:

https://discourse.ros.org/t/colcon-ament-tools/4685/1

PR #169, opened 2018-04-18
Wjwwood/build tool colcon

wjwwood:
Request and recommended changes.

### DIFF
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -233,7 +233,7 @@ It is up to the implementation to decide if it only supports the standard CMake 
 
 In ROS 2 the concept of the *devel space* has intentionally been removed.
 In the future it might be feasible to provide the concept of *symlinked installs* in ROS 1 to provide a similar benefit without the downsides.
-Therefore not supporting this concept in the universal build tool is considered a viable path forward.
+However, this design document will assume that the *devel space* will remain in ROS 1.
 
 #### Build ROS 2 workspaces
 
@@ -387,6 +387,7 @@ For option **A)** the follow items should be considered:
 - Update code base to Python 3.5+.
 - Refactor code base to reduce coupling (e.g. separate [API](https://github.com/catkin/catkin_tools/blob/2cae17f8f32b0193384d2c7734afee1c60c4add2/catkin_tools/execution/controllers.py#L183-L205) for output handling).
 - Additional functionality to build Gazebo including its dependencies.
+- Whether or not to continue supporting the *devel space*.
 
 For option **B)** the follow items should be considered:
 
@@ -433,6 +434,10 @@ The following items briefly enumerate what This means for ROS developers and use
 - When building and testing ROS 2 the command `colcon build` / `colcon test` will be used instead of `ament build` / `ament test`.
   Please see the [documentation](http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) how to map `ament_tools` command line arguments to `colcon` arguments.
 - For ROS 1 nothing is changing at this point in time.
+- In the future `colcon` will replace `catkin_make_isolated` and `catkin_make` as the recommended build tool for ROS 1.
+  - `colcon` will not support the *devel space* and will require packages to have install rules
+  - `catkin` will likely still support the *devel space*, though it might be removed at some point (that has not been decided yet)
+  - Therefore, it is possible that the default build tool for ROS 1 may not support the *devel space*, though legacy tools will continue to support it.
 
 ### Outlook
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -233,7 +233,6 @@ It is up to the implementation to decide if it only supports the standard CMake 
 
 In ROS 2 the concept of the *devel space* has intentionally been removed.
 In the future it might be feasible to provide the concept of *symlinked installs* in ROS 1 to provide a similar benefit without the downsides.
-However, this design document will assume that the *devel space* will remain in ROS 1.
 
 #### Build ROS 2 workspaces
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -347,7 +347,7 @@ For the last ROS 2 release *Ardent* the buildfarm [build.ros.org](http://build.r
 Neither *devel* jobs or *pull request* jobs are available nor is it supported to build a local *prerelease*.
 For the coming ROS 2 release *Bouncy* these job types should be available to support maintainers.
 
-In ROS 2 *Bouncy* the univeral build tool will be the only supported option and `ament_tools` will be archived.
+In ROS 2 *Bouncy* the universal build tool will be the only supported option and `ament_tools` will be archived.
 
 #### Necessary work
 
--- a/articles/101_build_tool.md
+++ b/articles/101_build_tool.md
@@ -365,7 +365,7 @@ For option **A)** the follow items would need to be addressed:
 - Support for Windows using `.bat` files.
 - Support for the package manifest format version 3.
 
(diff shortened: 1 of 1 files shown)


articles/101_build_tool.md
```diff
@@ -387,16 +387,26 @@ For option **A)** the follow items should be considered:
...
+  - Should have a test for this case.
+- Test for, and fix if necessary, correct topological order with dependencies across workspaces.
+  - See: [https://github.com/ros/catkin/pull/590](https://github.com/ros/catkin/pull/590)
 
 ## Summary and Decision
 
-Based on the above information a decision has been made to pick `colcon` as the universal build tool which was not an easy one.
+Based on the above information a decision has been made to pick `colcon` as the universal build tool.
+
+The decision was made after considering the input of ROS 2 team members and some ROS 1 users.
```
gbiggs, line 407: What happened to consulting ROS 2 users?

tkruse: And "some ROS1 users" is terribly unspecific. ROS1 has been working by REPs, and OSRF religiously decreed "ROS1 will not be disrupted". So now you want to disrupt ROS1 without a REP?

dirk-thomas: > What happened to consulting ROS 2 users?

I am not sure what kind of feedback we would expect from this. `colcon` basically works the same way as `ament_tools`. Some of the command line options are named slightly different. If that is a major concern it would be easy to provide an executable named `ament` which provides the exact same CLI (I have a prototype of that locally from when I was exploring such options). But since the difference is really small (see http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) I am not sure if that would even be valuable or more confusing.

Maybe you can clarify what other kind of feedback you would expect?

articles/101_build_tool.md
```diff
@@ -387,16 +387,26 @@ For option **A)** the follow items should be considered:
...
+The decision was not easy, as it was not unanimous, but the vast majority of input was either pro `colcon` or ambivalent.
```
dirk-thomas, line 408: Sounds good to me. :+1:

articles/101_build_tool.md
```diff
@@ -233,7 +233,7 @@ It is up to the implementation to decide if it only supports the standard CMake
 
 In ROS 2 the concept of the *devel space* has intentionally been removed.
 In the future it might be feasible to provide the concept of *symlinked installs* in ROS 1 to provide a similar benefit without the downsides.
-Therefore not supporting this concept in the universal build tool is considered a viable path forward.
+However, this design document will assume that the *devel space* will remain in ROS 1.
```
dirk-thomas: I don't this this article should that claim either. Maybe it would be better to just state the fact that ROS 1 currently supports the devel space and the universal build tool doesn't. But exclude any future decision and defer that to be dealt with on the ROS 1 side (maybe in a REP).

articles/101_build_tool.md
```diff
@@ -424,6 +434,10 @@ The following items briefly enumerate what This means for ROS developers and use
 - When building and testing ROS 2 the command `colcon build` / `colcon test` will be used instead of `ament build` / `ament test`.
   Please see the [documentation](http://colcon.readthedocs.io/en/latest/migration/ament_tools.html) how to map `ament_tools` command line arguments to `colcon` arguments.
 - For ROS 1 nothing is changing at this point in time.
+- In the future `colcon` will replace `catkin_make_isolated` and `catkin_make` as the recommended build tool for ROS 1.
+  - `colcon` will not support the *devel space* and will require packages to have install rules
+  - `catkin` will likely still support the *devel space*, though it might be removed at some point (that has not been decided yet)
+  - Therefore, it is possible that the default build tool for ROS 1 may not support the *devel space*, though legacy tools will continue to support it.
```
dirk-thomas, line 439: If we want to mention that explicitly we should probably also clarify that a similar problem already exists. Some packages do only work in the devel space and some do only work in the install space (since these packages haven't been developed with both in mind and the developers / maintainers / users only use one of the two).

articles/101_build_tool.md
wjwwood: > What happened to consulting ROS 2 users?

I think feedback on this design document update _is_ the solicitation for feedback from ROS 2 users and others. Though I don't know if it would impact the decision that's been made one way or the other.

> And "some ROS1 users" is terribly unspecific.

That was intentional, because I didn't take time to ask if I could mention them directly, also I didn't feel it was material to the point.

> So now you want to disrupt ROS1 without a REP?

Well, there isn't a REP to update or replace w.r.t. the build tool in ROS. It's more of a recommendation than a requirement, and not the same in my mind as the switch from `rosbuild` to `catkin`.

Other things which would actually be disruptive (require changes in existing code) would require REP's, like getting rid of the devel space, and deciding on whether or not to pursue those or not is apparently also blocking roll out of the universal build tool in ROS 1, so I don't it will be disruption without REP's personally, but I suppose we'll have to see.

To be fair, had we decided to use catkin_tools, roll out in ROS 1 would have just been fix a few final things and change documentation to recommend catkin_tools instead of catkin_make, but there would not have been a need for a REP in that case, in my opinion.

articles/101_build_tool.md
wjwwood: It wasn't my intention to claim either. I was just taking the "status quo" as the most likely future. I think that's the best assumption to take given the circumstances. Note an assumption is not an assertion or claim. Just what we have to assume until it changes.

I'm not sure how I would change this. Would you like me to remove it, or modify it some specific way?

articles/101_build_tool.md
wjwwood: But currently that's not driven by the build tool (the build tool allows you to do either), so that's why I didn't include it. But I guess I can add a note about that.

wjwwood: 4717cbf

articles/101_build_tool.md
gbiggs: I don't think there is much feedback to provide from the point of view of a ROS 2 user. I'm actually in favour of the change in tool, especially because it is so non-disruptive. But I'm disappointed that I only get to provide that opinion after the decision has been made. I find it annoying that the OSRF assumes that existing ROS 2 users will have no valid or useful input into a high level decision such as this. We went from a state we were in a year ago to suddenly having a new tool and a decision made to use it without any notice.

articles/101_build_tool.md
dirk-thomas: I read the assumption slightly different. Like "we have thought about it and for now think it will stay that way". I think it should more like "at this point we are not making a decision how to move forward with this in ROS 1".

Maybe:
> The direction forward in ROS 1 - if the new build tool needs to support the devel space or if the concept of the devel space is deprecated at some point (maybe even implicitly if the new build tool doesn't support it) - has not been decided yet and is out of scope of this article.

wjwwood: > Like "we have thought about it and for now think it will stay that way". I think it should more like "at this point we are not making a decision how to move forward with this in ROS 1".

Those are the same thing... basically "at this point we are not making a decision how to move forward with this in ROS 1, therefore it will stay that way until we do make a decision".

I'm just removing the sentence. The previous lines describe the situation sufficiently.

dirk-thomas (approved):

wjwwood:
I didn’t mean to let this linger, I guess I’ll merge since the approval.

articles/101_build_tool.md
wjwwood: Yeah, I agree that community feedback before a decision would be nice, but as a member of the team I also understand the desire to get things moving. I’m certain your feedback is considered but it would have to be a significant amount of negative feedback to change the direction. And that was the case before the decision was made as well.

PR #170, opened 2018-04-25
Remove topic partitions

rohitsalem:
connects to ros2/ros2#476
Update design to specify the usage of the topic/service name directly without using the partitions.

### DIFF
--- a/articles/140_topic_and_service_name_mapping.md
+++ b/articles/140_topic_and_service_name_mapping.md
@@ -198,51 +198,17 @@ Any topic or service name that contains any tokens (either namespaces or a topic
 
 ## Mapping of ROS 2 Topic and Service Names to DDS Concepts
 
-The ROS topic and service name constraints allow more types of characters than the DDS topic names because ROS additionally allows the forward slash (`/`), the tilde (`~`), and the balanced curly braces (`{}`).
+The ROS topic and service name constraints allow more types of characters than the DDS topic names because ROS additionally allows the, the tilde (`~`), and the balanced curly braces (`{}`).
 These must be substituted or otherwise removed during the process of converting the topic or service name to DDS concepts.
 Since ROS 2 topic and service names are expanded to fully qualified names, any balanced bracket (`{}`) substitutions and tildes (`~`) will have been expanded.
 Additionally any URL related syntax, e.g. the `rostopic://` prefix, will be removed once parsed.
-Therefore only forward slashes (`/`) need to be addressed when converting to DDS concepts.
-
-### ROS Namespaces with DDS Partitions
-
-The proposed strategy to address the forward slashes (`/`) which are present in ROS names, is to separate the ROS name into the "namespace" and the "base name", and then place the namespace, stripped of leading and trailing forward slashes (`/`), into a single DDS partition entry and the remaining base name into the DDS topic name.
-This addresses the issue because the ROS name's base name will not contain any forward slashes (`/`) by definition and so there are no longer any disallowed characters in the DDS topic name.
-The DDS partition will contain the ROS name's namespace, including any forward slashes (`/`) that make up the namespace and were not at the beginning or the end of the namespace.
-That is acceptable because DDS partitions are allowed to contain forward slashes (`/`) unlike the DDS topics.
-
-This strategy also avoids DDS partitions from being exposed to the ROS API's.
-This is because the splitting of the ROS name into namespace and base name can occur inside the implementation of the ROS middleware interface, and so the ROS topic or service name will remain intact throughout the middleware agnostic code.
-This means that if someone chooses to implement ROS 2 on top of something other than DDS, then this strategy will not impact that implementation, as it applies only to DDS implementations.
-
-#### DDS Partitions Characteristics
-
-DDS partitions are implemented as an array of strings within the `DDS::Publisher` and `DDS::Subscriber` QoS settings.
-The array allows up to 64 items, and each string in the array is an individual partition.
-This means that every string of the array is an independently matchable entity when paired with the topic name and the other QoS settings.
-Put another way, a DDS DataWriter and DataReader will not connect to one another unless at least one of the partition strings in the array match, even if the topic names match.
-Sometimes special characters are used in the partition strings to express regular expressions matching (such as `+`, `*`, `^`), but they are not officially supported by the standard.
-The maximum size of each string in the array may depend on the DDS or RTPS implementation.
-You can read more about partitions in RTI's documentation:
-
-- [PARTITION_QosPolicy](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/PARTITION_QosPolicy.htm)
-
-The array of strings have no hierarchy and the order does not matter.
-Each entry in the partition array is directly combined with the DDS topic and they are not sequentially combined.
-If a publisher has two partition entries, e.g. `foo` and `bar` with a base name of `baz`, this would be equivalent to having two different publishers on these topics: `/foo/baz` and `/bar/baz`.
-Therefore we cannot use the array of strings to represent our implied namespace hierarchy and we have to continue to use the forward slash to accomplish this.
-
-#### Hierarchy with DDS Partitions
-
-Since DDS partitions have no hierarchy or order to the array of strings, this proposal would use only one of the strings in the array to hold the entire ROS name's namespace.
-Therefore we preserve the forward slashes (`/`) that exist in the ROS name's namespace already.
+Previously forward slashes (`/`) were disallowed in DDS topic names, now the restriction has been lifted and therefore the ROS topic names are first prefixed with ROS Specific Namespace prefix (described below) are then mapped directly into DDS topic names.  
 
 ### ROS Specific Namespace Prefix
 
 In order to differentiate ROS topics easily, all DDS topics created by ROS shall be automatically prefixed with a namespace like `/rX`, where `X` is a single character that indicates to which subsystem of ROS the topic belongs.
-At this point it's important to remember that when the namespace part of the ROS name is put into the DDS partition it is stripped of leading and trailing forward slashes (`/`).
-For example, a plain topic called `/foo` would translate to a DDS partition `rt` and a DDS topic `foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name `foo`, then the namespace getting the leading and trailing forward slashes (`/`) stripped.
-As another example, a topic called `/left/image_raw` would translate to a DDS partition `rt/left` and a DDS topic `image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name `image_raw`, again stripping the leading and trailing forward slashes (`/`) from the namespace.
+For example, a plain topic called `/foo` would translate to a DDS topic `rt/foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name `foo`.
+As another example, a topic called `/left/image_raw` would translate to a DDS topic `/rt/left/image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name `image_raw`.
 
 For systems where Services are implemented with topics (like with OpenSplice), a different subsystem character can be used: `rq` for the request topic and `rr` for the response topic.
 On systems where the implementation is handled for us by DDS (like with Connext), we use `rs` as the common prefix.
@@ -265,12 +231,13 @@ The standard reserves the right to use up to 8 characters for the prefix in case
 
 Here are some examples of how a fully qualified ROS name would be broken down into DDS concepts:
 
-| ROS Name                        | DDS Topic   | DDS Partition                 |
-|---------------------------------|-------------|-------------------------------|
-| `/foo`                          | `foo`       | \[`rt`\]                      |
-| `rostopic:///foo/bar`           | `bar`       | \[`rt/foo`\]                  |
-| `/robot1/camera_left/image_raw` | `image_raw` | \[`rt/robot1/camera_left`\]   |
+| ROS Name                        | DDS Topic                         |
+|---------------------------------|-----------------------------------|
+| `/foo`                          | `rt/foo`                          |
+| `rostopic:///foo/bar`           | `rt/foo/bar`                      |
+| `/robot1/camera_left/image_raw` | `rt/robot1/camera_left/image_raw` |
 
+(TODO)
 ### Changing ROS Names During Runtime
 
 If the ROS name needs to be changed during runtime (after the publisher or subscriber has already been created and is in use), then some details about DDS partitions might be useful to consider.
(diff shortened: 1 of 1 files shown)


articles/140_topic_and_service_name_mapping.md
```diff
@@ -198,51 +198,17 @@ Any topic or service name that contains any tokens (either namespaces or a topic
...
-For example, a plain topic called `/foo` would translate to a DDS partition `rt` and a DDS topic `foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name `foo`, then the namespace getting the leading and trailing forward slashes (`/`) stripped.
-As another example, a topic called `/left/image_raw` would translate to a DDS partition `rt/left` and a DDS topic `image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name `image_raw`, again stripping the leading and trailing forward slashes (`/`) from the namespace.
+For example, a plain topic called `/foo` would translate to a DDS topic `rt/foo`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the root namespace `/` and has a base name `foo`.
+As another example, a topic called `/left/image_raw` would translate to a DDS topic `/rt/left/image_raw`, which is the result of implicitly adding `/rt` to the namespace of a ROS topic which is in the namespace `/left` and has a base name `image_raw`.
```
mikaelarguedas: Do we add a slash in front of the prefix?
in this example  `/foo` becomes `rt/foo` whereas `/left/image_raw` becomes `/rt/left/image_raw`

Based on the examples below, it looks like we don't add the leading `/` and `/left/image_raw` becomes `rt/left/image_raw`, is that correct?

rohitsalem: Yes, we don't add the / to the prefixes, I'll change it

articles/140_topic_and_service_name_mapping.md
```diff
@@ -198,51 +198,17 @@ Any topic or service name that contains any tokens (either namespaces or a topic
 
 ## Mapping of ROS 2 Topic and Service Names to DDS Concepts
 
-The ROS topic and service name constraints allow more types of characters than the DDS topic names because ROS additionally allows the forward slash (`/`), the tilde (`~`), and the balanced curly braces (`{}`).
+The ROS topic and service name constraints allow more types of characters than the DDS topic names because ROS additionally allows the, the tilde (`~`), and the balanced curly braces (`{}`).
```
Karsten1987: is the `the ,` a typo or do you want to highlight the comma specifically?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -198,51 +198,17 @@ Any topic or service name that contains any tokens (either namespaces or a topic
...
-The array of strings have no hierarchy and the order does not matter.
-Each entry in the partition array is directly combined with the DDS topic and they are not sequentially combined.
-If a publisher has two partition entries, e.g. `foo` and `bar` with a base name of `baz`, this would be equivalent to having two different publishers on these topics: `/foo/baz` and `/bar/baz`.
-Therefore we cannot use the array of strings to represent our implied namespace hierarchy and we have to continue to use the forward slash to accomplish this.
-
-#### Hierarchy with DDS Partitions
-
-Since DDS partitions have no hierarchy or order to the array of strings, this proposal would use only one of the strings in the array to hold the entire ROS name's namespace.
-Therefore we preserve the forward slashes (`/`) that exist in the ROS name's namespace already.
+Previously forward slashes (`/`) were disallowed in DDS topic names, now the restriction has been lifted and therefore the ROS topic names are first prefixed with ROS Specific Namespace prefix (described below) are then mapped directly into DDS topic names.
```
Karsten1987: Can you reference any official statement or something indicating that it's now allowed?

Maybe the complete sentence can actually be part of the next section `ROS Specific Namespace Prefix`.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -265,29 +231,16 @@ The standard reserves the right to use up to 8 characters for the prefix in case
...
 
 ### ROS Topic and Service Name Length Limit
 
-The length of the DDS topic must not exceed 256 characters.
-The actual length of a partition field may be limited to 256 characters, however this varies drastically depending on the vendor.
-RTI Connext does not allow a creation of a publisher/subscription with a partition length of 248 characters, whereas FastRTPS does not have any limitation in length.
-Please bare in mind, that the length of the partition gets further diminished due to the introduction of a ROS specific prefix.
-The actual length of a ROS Topic, including the namespace hierarchy and the base name of the topic, may thus be varying in length as well.
-Yet, the base name token must not exceed the length of 256 characters as this is getting mapped directly as the DDS topic.
+The length of the DDS topic must not exceed 256 characters. Therefore the length of a ROS Topic, including the namespace hierarchy, the base name of the topic and any ros specific prefixes must not exceed 256 characters since this is mapped directly as DDS topic.
```
Karsten1987: I think you should give some example service names.
Also make sure to reference the changes specifically for RTI here (contentfilter etc).
We should highlight that RTI add the GUID additional to the reply topic.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -298,35 +251,29 @@ Therefore to allow ROS programs to interoperate with "native" DDS topic names th
 #### Examples and Ideas for Communicating with Non-ROS Topics
 
 Note that these are not part of the proposal, but only possible solutions to the issue of communicating with "native" DDS topics.
-This section should be update when/if a complete solution is found.
+This section should be updated when/if a complete solution is found.
```
Karsten1987: I think you can actually update the design doc here as well, since we have the option to ignore the ROS conventions.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -391,6 +338,31 @@ This compromise was made so that when more work is done on substitutions, it wou
...
+
+- Splitting the ROS name into "namespace" and "base name", and placing the complete namespace into a single partition seemed un-natural.
+- In general partitions are recommended to be used as a spare, but using partitions for all ROS names suggested otherwise.
+- Major concern was reported in this [issue](https://github.com/ros2/rmw_connext/issues/234), Where having two topics with same base name, although different namespace and different ypes caused problem. For example: topicA is `/camera/data` of type `Image` and topicB is `/imu/data` of type `Imu`. The base names for both topicA and topicB is `data`, generated errors as described in the [issue](https://github.com/ros2/rmw_connext/issues/234).
+- Using the complete ROS name in the later strategy will cause a tighter length limit on base name because the DDS topic name would contain ROS prefix, namespace along with the base name which should not exceed DDS topic name limitation which is  256 characters.
+
+Rationale:
+- With the decision form the DDS vendors to allow forward slashes (`/`) in DDS topic names, using the complete ROS name seemed simple and more intuitive than using partitions.
```
Karsten1987: typo: `from`

articles/140_topic_and_service_name_mapping.md
```diff
@@ -391,6 +338,31 @@ This compromise was made so that when more work is done on substitutions, it wou
...
+If a publisher has two partition entries, e.g. `foo` and `bar` with a base name of `baz`, this would be equivalent to having two different publishers on these topics: `/foo/baz` and `/bar/baz`.
+Therefore this proposal used only one of the strings in the partitions array to hold the entire ROS name's namespace.
+
+You can read more about partitions in RTI's documentation:
+
+- [PARTITION_QosPolicy](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/PARTITION_QosPolicy.htm)
+
+Trade-offs (in comparison to using the whole ROS name along with the namespaces):
+
+- Splitting the ROS name into "namespace" and "base name", and placing the complete namespace into a single partition seemed un-natural.
```
Karsten1987: why unnatural?
Another reason for this is that newer standards such as XRCE DDS might not have partitions at all.

articles/140_topic_and_service_name_mapping.md
```diff
@@ -391,6 +338,31 @@ This compromise was made so that when more work is done on substitutions, it wou
...
+- In general partitions are recommended to be used as a spare, but using partitions for all ROS names suggested otherwise.
+- Major concern was reported in this [issue](https://github.com/ros2/rmw_connext/issues/234), Where having two topics with same base name, although different namespace and different ypes caused problem. For example: topicA is `/camera/data` of type `Image` and topicB is `/imu/data` of type `Imu`. The base names for both topicA and topicB is `data`, generated errors as described in the [issue](https://github.com/ros2/rmw_connext/issues/234).
```
Karsten1987: typoc: `Where` in caps

Karsten1987 (commented):
a few comments

also make sure to put one sentence per line. This makes reviewing easier.

articles/140_topic_and_service_name_mapping.md
rohitsalem: Regarding the official statement is there any issue or thread that refers to this? I think this is a general statement which says that ROS topics are mapped directly to DDS topics and therefore is here.

articles/140_topic_and_service_name_mapping.md
rohitsalem: Its a typo

articles/140_topic_and_service_name_mapping.md
rohitsalem: yes, should we also explain that we add Request/Reply to the service names in general (or is it mentioned somewhere else in the design doc) along with the special cases for RTI?

articles/140_topic_and_service_name_mapping.md
rohitsalem: maybe there is a better word that un-natural. Placing the whole namespace with the prefix in a 'single' partition sounds like a workaround and therefore I added the term un-natural

articles/140_topic_and_service_name_mapping.md
mikaelarguedas: There is a DDS issue opened on this but hasn't been marked as "approved"/"resolved": https://issues.omg.org/issues/lists/dds-rtf5#issue-42236

rohitsalem: If that's the only pointer we have I'll go ahead and include it

articles/140_topic_and_service_name_mapping.md
rohitsalem: Now that we can use `avoid_ros_namespace_convention` Should I remove the alternatives of using the 'to have some markup in the scheme name' in the same section below? or leave it as is?

wjwwood: Having a markup to express the exact topic name is separate from having the ability to avoid applying our ROS specific mangling in our API. So I think it should _not_ be removed because we have `avoid_ros_namespace_convention`.

rohitsalem: Thanks, I'll have it listed as an alternative in the same section

articles/140_topic_and_service_name_mapping.md
Karsten1987: I don't think it's mentioned explicitly here, but I would actually be in favor of it. Also, it important to mention that RTI adds the GUID to it.

articles/140_topic_and_service_name_mapping.md
Karsten1987: this typo is still there.

articles/140_topic_and_service_name_mapping.md
rohitsalem: yes I added about RTI adding the GUID below

articles/140_topic_and_service_name_mapping.md
rohitsalem: I think it is fixed [here](https://github.com/ros2/design/pull/170/files#diff-51ea0dd0254ef288483a09be8b28379dR369), this is an outdated version

articles/140_topic_and_service_name_mapping.md
```diff
@@ -265,68 +231,53 @@ The standard reserves the right to use up to 8 characters for the prefix in case
...
-Please bare in mind, that the length of the partition gets further diminished due to the introduction of a ROS specific prefix.
-The actual length of a ROS Topic, including the namespace hierarchy and the base name of the topic, may thus be varying in length as well.
-Yet, the base name token must not exceed the length of 256 characters as this is getting mapped directly as the DDS topic.
+The length of the DDS topic must not exceed 256 characters. Therefore the length of a ROS Topic, including the namespace hierarchy, the base name of the topic and any ros specific prefixes must not exceed 256 characters since this is mapped directly as DDS topic.
+
+The length of service name has tighter limits than the length of the ROS Topics.
+In underlying RMW-DDS implementation, because the way in which services are implemented vary for different DDS vendors, and hence imposes the limit on actual service names that can be used.
+For example, in RTI connext, the service names are suffixed with the GUID value of the DDS participant, and a content filtered topic(max length 256 characters) is created which is mapped from the suffixed service name.
```
ruffsl: That's odd. Where can I read more up on this about RTI Connext? I'd think this runtime specific GUID suffix would disrupt efforts to statically define topic permissions for DDS security.

mikaelarguedas: I believe that the suffix is not taken into account by Connext's builtin plugins.
That is why the current sros2 generated permission files work for Connext even if they dont have the postfix in them. Not sure how they handle it internally though..

Karsten1987 (approved):
lgtm.

@mikaelarguedas Any further thoughts?

articles/140_topic_and_service_name_mapping.md
```diff
@@ -265,68 +231,56 @@ The standard reserves the right to use up to 8 characters for the prefix in case
...
+There is an option in the API, a boolean `avoid_ros_namespace_convention` in the qos_profile which can be set to `false` to use ROS prefix and `true` to not using ROS namespace prefixing.
 
-It may be possible, through an option in the API or through some name syntax, to get a similar break down but without the ROS specific prefix.
+For example:
 
-For example with an option in the API:
+| ROS Name              | avoid_ros_namespace_conventions    | DDS Topic   |
+|-----------------------|------------------------------------|-------------|
+| `rostopic://image`    | `false`                            | `rt/image`  |
+| `rostopic://image`    | `true`                             | `image`     |
```
mikaelarguedas, line 267: my understanding is that this will result in `/image` (based on discussion from https://github.com/ros2/rmw_fastrtps/pull/192#discussion_r178970046).
We can leave this as is in the design doc if we ticket the missing bit that is to pass the user-specified topic name to rmw instead of the expanded one.
Otherwise we should correct it here

(same below)

rohitsalem: yes that's correct I think it should be `/image` instead of `image`, as of now by default the topic name being passed here will already have a `/` appended to it (if `/` is not already present) be a fully qualified as ros topic isn't it?

mikaelarguedas: https://github.com/ros2/ros2/issues/496 has been created to track what is described here but not yet implemented

ruffsl:
For posterity, was there a online discussion somewhere else that pertains to the reasoning behind this PR. I just want to share with someone else the context for this decision for the removal of partitions, and none of the connected PR here include the motivating description that I can cite to.

dhood:
This has context behind the motivation: https://github.com/ros2/rmw_connext/issues/234

Karsten1987:
@ruffsl The rationale for removing partitions is explained in this section of the design doc, which also links to the original reported issue on github:
http://design.ros2.org/articles/topic_and_service_names.html#alternative-using-dds-partitions

Do you think it's insufficient? We can reiterate over it if it's unclear.

spiderkeys:
@karsten1987 just as an outside opinion from a heavy user of both pure DDS and ROS, the rationale and write up were crystal clear and informative.

ruffsl:
@Karsten1987 , the rationale explained in the design doc is sufficient. I should of have checked back with the updated docs again, but it good that it's linked here now as well.

PR #177, opened 2018-06-30
Use list form instead of normal description

youtalk:
When I read the latest [A universal build tool](http://design.ros2.org/articles/build_tool.html), the following sentence seems to be difficult to read.

```
A. Use catkin_tools as a starting point B. Use colcon as a starting point
```

This PR fixes it to use list form instead.


dirk-thomas:
Thanks.

PR #178, opened 2018-07-05
Messages on /clock must be delivered in order

sloretz:
Add sentence saying messages published on `/clock` must be delivered in order. The design doc requires an API to call user callbacks when time jumps backwards. If messages are received out of order then state could be reset unnecessarily.

### DIFF
--- a/articles/130_ros_time.md
+++ b/articles/130_ros_time.md
@@ -109,7 +109,6 @@ If the time on the clock jumps backwards, a callback handler will be invoked and
 Calls that come in before that must block.
 The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
 
-Messages published on `/clock` must be delivered in order.
 The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific.
 
 ##### No Advanced Estimating Clock By Default
--- a/articles/130_ros_time.md
+++ b/articles/130_ros_time.md
@@ -109,6 +109,7 @@ If the time on the clock jumps backwards, a callback handler will be invoked and
 Calls that come in before that must block.
 The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
 
+Messages published on `/clock` must be delivered in order.
 The frequency of publishing the `/clock` as well as the granularity are not specified as they are application specific.
 
 ##### No Advanced Estimating Clock By Default
--- a/articles/qos.md
+++ b/articles/qos.md
@@ -103,6 +103,14 @@ However, this mechanism will not be added into the common ROS2 API so as to keep
 To honor the QoS settings of the system, developers can use the `rmw_qos_profile_system_default` QoS profile which delegates the responsibility of the QoS machinery to the underlying DDS vendor.
 This allows developers to deploy ROS2 applications and use DDS vendor tools to configure the QoS settings.
 
+## Integration with non DDS RMW Implementations
+
+There is no QoS setting that affects the order messages may be received.
+Instead an RMW implementation must implement the same behavior as DDS.
+Once a subscriber takes a message from the RMW implementation it must not be allowed to take an earlier message from the same publisher.
+If the connection is Reliable then newer messages should be kept in the History until all older messages are received.
+A best effort connection should drop an older message if it has already received a newer one.
+
 ## Open questions
 
 How should the integration of the QoS machinery with intraprocess communication be like.
--- a/articles/qos.md
+++ b/articles/qos.md
@@ -109,7 +109,7 @@ There is no QoS setting that affects the order messages may be received.
 Instead an RMW implementation must implement the same behavior as DDS.
 Once a subscriber takes a message from the RMW implementation it must not be allowed to take an earlier message from the same publisher.
 If the connection is Reliable then newer messages should be kept in the History until all older messages are received.
-A best effort connection should drop an older message if it has already received a newer one.
+A best effort connection should drop an older message if a subscriber has taken a newer one.
 
 ## Open questions
 


articles/130_ros_time.md
```diff
@@ -109,6 +109,7 @@ If the time on the clock jumps backwards, a callback handler will be invoked and
 Calls that come in before that must block.
 The developer has the opportunity to register callbacks with the handler to clear any state from their system if necessary before time will be in the past.
 
+Messages published on `/clock` must be delivered in order.
```
tfoote: I understand the sentiment of this, but in it's current form it is "delivered in order" is not clearly defined. Maybe something like in the order published? Do we want to specify a policy of wait for all vs drop all but the most recent, or leave that to the application specific settings? Are there any settings expected to cause out of order delivery that maybe we should discourage?

sloretz: >"delivered in order" is not clearly defined. Maybe something like in the order published?

How about `delivered in order` -> `received by subscribers in the same order they were published`?

> Do we want to specify a policy of wait for all vs drop all but the most recent, or leave that to the application specific settings?

I'm not sure, it might be application specific. Any dropped messages could be an issue with small jumps back in time. If only one message is a jump back and that message is dropped then a node will not call the time jumped callback when it should have.

On another ticket I saw reference to configuring the duration needed to consider a jump a backwards jump. The number of stored messages has a similar affect. The publisher's (history size / publish frequency * real time factor) is related to the maximum amount of time that messages may be dropped where node is guaranteed to be notified of a backwards jump

> Are there any settings expected to cause out of order delivery that maybe we should discourage?

I'm not sure if the doc should specifically call out DDS settings, or be generic for any middleware. Maybe both make sense given that the first priority is DDS, but an `rmw` implementation could be made for any other protocol.

wjwwood: How should we enforce that messages were delivered in order? What's the difference in jumping back one time period before continuing forward and delivering the messages out of order? Is there a rule that the period between values in a clock message must be consistent, i.e. must all delta's be 1ms for instance?

It seems like we need a sequence number in addition to the time in the clock value. That's the only way to know the difference in a deliberate time jump and an out of order delivery. If we had sequence numbers, I'd suggest saying that consumers of `/clock` should ignore messages received after older instances. This is more easily implemented, but obviously the publishers should strive to deliver them in order, I just don't know if we can guarantee it efficiently.

We deprecated `seq` in the `Header` message type, but we always said that we'd like to be able to get this value from the middleware somehow (like in the message_info struct) on a per publisher basis.

The default in DDS is to deliver them in order (see link below), and DDS has the presentation order QoS setting for different scenarios, but I think it has some additional overhead:

https://community.rti.com/examples/ordered-presentation

sloretz: > How should we enforce that messages were delivered in order? What's the difference in jumping back one time period before continuing forward and delivering the messages out of order?

Originally I assumed `BEST_EFFORT` meant unordered data was possible, and that the design doc should make it clear QoS settings should be chosen to prevent that. However, it looks like DDS 1.4 says unordered data cannot happen in `BEST_EFFORT`. 

```
2.2.3.14 RELIABILITY
[...]
If the RELIABILITY kind is set to BEST_EFFORT, the service will not re-transmit missing data-samples. However for data-
samples originating from any one DataWriter the service will ensure they are stored in the DataReader history in the same
order they originated in the DataWriter. In other words, the DataReader may miss some data-samples but it will never see the
value of a data-object change from a newer value to an order value.
```

If the question is how to enforce that messages are delivered in order even for custom middleware, then maybe this design doc is the wrong place. Instead documentation for `rmw_qos_profile_t` should say that `RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT` still means messages are delivered in order.

tfoote: Really I don't think that this is something that we can enforce or require. But I think we might just want to suggest that we recommend the `/clock` topic be transported with settings that will deliver messages in order to avoid spurrious time jumps due to out of order delivery.

basically say that this is good practice and will provide better performance. But not try to mandate it.

wjwwood: But are there any settings which impact this? It seems like a meaningless recommendation if you can't actually do anything (or avoid anything) that might impact delivery order.

Also I don't think it provides better performance, i.e. the order of delivery shouldn't impact performance, but instead should impact how often the time jump callback is called incorrectly.

wjwwood: Seems to me that either reliable or best effort will result in delivery in the order sent, just above what @sloretz quoted above:

```
If the RELIABILITY kind is set to RELIABLE, data-samples originating from a single DataWriter cannot be made available
to the DataReader if there are previous data-samples that have not been received yet due to a communication error. In other
words, the service will repair the error and re-transmit data-samples as needed in order to re-construct a correct snapshot of the
DataWriter history before it is accessible by the DataReader.
```

sloretz: > But are there any settings which impact this? It seems like a meaningless recommendation if you can't actually do anything (or avoid anything) that might impact delivery order.

It seems like the only thing someone could do is create a non-DDS rmw implementation that doesn't guarantee in-order delivery.

I would still like to document this somewhere in case someone has the same misunderstanding I had, but it looks like the time design doc is the wrong spot. How about I instead add a note to http://design.ros2.org/articles/qos.html saying messages are expected to be delivered in order using any QoS setting?

wjwwood: Yeah, that would make more sense to me -- it should be a property of our middleware, not a recommendation related only to `/clock`. Because if you say "`/clock` should be delivered in order", the immediate next question for a reader of this document is "how do I know that?".

Clarifying this in the QoS design doc or in the documentation of `rcl_take*` (maybe) is better, and if in the future a QoS might affect delivery order, then it would make sense to mention that here and be specific about which settings to watch out for.

sloretz: Removed sentence from time doc 13c97cc and added section to QoS doc 79c9e00

tfoote (dismissed):

articles/qos.md
```diff
@@ -103,6 +103,14 @@ However, this mechanism will not be added into the common ROS2 API so as to keep
 To honor the QoS settings of the system, developers can use the `rmw_qos_profile_system_default` QoS profile which delegates the responsibility of the QoS machinery to the underlying DDS vendor.
 This allows developers to deploy ROS2 applications and use DDS vendor tools to configure the QoS settings.
 
+## Integration with non DDS RMW Implementations
+
+There is no QoS setting that affects the order messages may be received.
+Instead an RMW implementation must implement the same behavior as DDS.
+Once a subscriber takes a message from the RMW implementation it must not be allowed to take an earlier message from the same publisher.
+If the connection is Reliable then newer messages should be kept in the History until all older messages are received.
+A best effort connection should drop an older message if it has already received a newer one.
```
tfoote: Maybe `received` could be clarified as it's more specifically if it's been taken.

sloretz: 785914d0e84a957d8b69221454909252e3127876 `if it has already received` -> `if a subscriber has taken`

tfoote (approved):

PR #179, opened 2018-07-11
Fix an error in ROS Time article.

nuclearsandwich:
This looks like a simple word substitution error. Confirmed in offline discussion with @tfoote.


wjwwood (approved):
Looks like the right fix to me.

PR #180, opened 2018-07-13
Update the simtime parameter name: using_sim_time -> use_sim_time

dhood:
Was previously listed as `using_sim_time` which does not match `rclcpp` implementation or ros1.

This PR includes other improvements I noticed while reading, but 7609588 is the functional change.


mikaelarguedas (approved):
lgtm

tfoote (approved):

PR #181, opened 2018-07-16
Clarify remapping order for __node and __ns

kev-the-dev:
See [this comment](https://github.com/ros2/rcl/pull/263#issuecomment-404905463). The design doc for remapping left the order of __node and __ns ambiguous, but RCL applied them in a specific order.

### DIFF
--- a/articles/141_static_remapping.md
+++ b/articles/141_static_remapping.md
@@ -322,14 +322,31 @@ The replacement must be a single token which will become the node's new name.
 
 #### Order of Applying Remapping Rules
 
-Rules for changing a default namespace or the name of the node get applied first before any other rule.
-All other rules are applied in the order in which the user gave them.
+Remapping rules are applied in the following order:
 
-**Example:**
+1. Node name remapping
+1. Namespace remapping
+1. All other rules
+
+Within each category, the rules are applied in the order in which the user gave them.
+
+**Example of topic/service remapping order:**
 
 - a node uses a name `/foo/bar`
 - a user gives the node a rule `/*/*:=/asdf` and then `/foo/bar:=fizzbuzz`
 - the final name is `/asdf` because the name did not match the second rule after being remapped by the first rule
+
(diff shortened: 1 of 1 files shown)


articles/141_static_remapping.md
```diff
@@ -322,15 +322,33 @@ The replacement must be a single token which will become the node's new name.
...
 - a node uses a name `/foo/bar`
 - a user gives the node a rule `/*/*:=/asdf` and then `/foo/bar:=fizzbuzz`
 - the final name is `/asdf` because the name did not match the second rule after being remapped by the first rule
 
+**Example of node/namespace remapping order:**
+
+- A node has name `talker`
+- A user remaps the node name to foo `talker:__node:=foo`
+- A user remaps the original node's namespace `talker:__ns:=/my_namespace`
+- The namespace is NOT changed because node name remaps are applied BEFORE namespace remaps
```
sloretz: I recommend phrasing to say what happens rather than what doesn't happen: `The final namespace is /ns because the node name was remapped before the namespace`

articles/141_static_remapping.md
```diff
@@ -322,15 +322,33 @@ The replacement must be a single token which will become the node's new name.
...
 
 - a node uses a name `/foo/bar`
 - a user gives the node a rule `/*/*:=/asdf` and then `/foo/bar:=fizzbuzz`
 - the final name is `/asdf` because the name did not match the second rule after being remapped by the first rule
 
+**Example of node/namespace remapping order:**
+
+- A node has name `talker`
+- A user remaps the node name to foo `talker:__node:=foo`
+- A user remaps the original node's namespace `talker:__ns:=/my_namespace`
```
sloretz: The example should show that `__node` happens first even if the rule is given second. How about combining the two bullets to show the order?

```
a user gives the rules `talker:__ns:=/my_namespace' then `__node:=foo`.
```

sloretz (approved):

PR #184, opened 2018-08-10
Clarify ROS default values will be used in default constructor

dhood:
@mikaelarguedas explained this to me in https://github.com/ros-simulation/gazebo_ros_pkgs/pull/793#discussion_r209356781

### DIFF
--- a/articles/112_generated_interfaces_cpp.md
+++ b/articles/112_generated_interfaces_cpp.md
@@ -105,7 +105,8 @@ All other constants are declared as `static const` members in the struct and the
 
 In the following discussion, "member" refers to the class member in the C++ class while "field" refers to the field definition in the IDL file.
 
-The *default constructor* initializes all members with their default value; if a field doesn't have a default value, then the member is [value-initialized](http://en.cppreference.com/w/cpp/language/value_initialization).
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
I think there is still misunderstanding here.

In this article, "default value" always refers to the default value provided in the msg file.

We don't have the concept of "No default value at all" as we define default values for all the types we define in the ROS IDL format (detailed in [this article](http://design.ros2.org/articles/interface_definition.html))

dhood:
since `char` wasn't listed in the article with common defaults I thought it was an example of the 'no default' category. I tried to clarify there's always a default in 93a6c39

mikaelarguedas:
As this paragraph was duplicated below, is it possible to update the following lines as well ?

dhood:
Other sentence updated in 77de52d8fb73f78d89bdb61d3263021d8e4d5309

mikaelarguedas:
lgtm.
Maybe @clalancette wants to review as he has the most context on this

clalancette:
This clarification looks fine to me!  I just corrected some spelling, otherwise, I think this is good to go.

dhood:
thanks to you both!

PR #189, opened 2018-09-13
Add unpublished idl documents to iterate on

sloretz:
~~~This PR combines the 4 IDL design documents into one page.~~~

Edit: PR now is the bare minimum to start parallelizing design doc work

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -1,0 +1,37 @@
+---
+layout: default
+title: IDL - Interface Definition and Language Mapping
+permalink: articles/idl_interface_definition.html
+abstract:
+  This describes defining interfaces using a subset of the [Interface Definition Language](https://www.omg.org/spec/IDL/) (IDL).
+published: false
+author: '[Dirk Thomas](https://github.com/dirk-thomas)'
+categories: Interfaces
+---
+
+{:toc}
+
+# {{ page.title }}
+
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Is this PR the first of a series?

As this page is getting pretty long and doesn't have the information for C generated structures I think we could benefit from more combination in that document.
For example I did found summary tables for types useful in the past: https://github.com/ros2/ros2/wiki/About-ROS-Interfaces#211-field-types (rather than 1 table per language)

sloretz:
@mikaelarguedas Now it's the first of a series. The content of this PR has been replaced to add two unpublished documents. One is a copy of the existing interface definition, the other is a stub for new content

dirk-thomas (dismissed):

dirk-thomas:
> One is a copy of the existing interface definition, the other is a stub for new content

In order to maintain the history please use one commit duplicating the existing file without any changes and a separate commit to update the header. The second commit can also contain the new document (I don't mind about that one). This should be merged without squashing then.

sloretz:
> In order to maintain the history please use one commit duplicating the existing file without any changes 

Done 97c52bb  and ~~~aef9f12~~~ 031af7a

dirk-thomas (approved):
Thanks for iterating on this.

PR #190, opened 2018-09-13
Add type mapping info to documents

sloretz:
This adds mappings between IDL types and other types.

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -197,9 +197,42 @@ When specified `T` is either one of the types above or an IDL struct.
 These array and sequence types have special mappings.
 If a cell is blank then the default mapping is used.
 
-| IDL type                     | C type                            | C++ type       | Python type |
-| ---------------------------- | --------------------------------- | -------------- | ----------- |
-| octet[N]                     |                                   |                | bytes       |
-| sequence<octet>              |                                   |                | bytes       |
-| sequence<octet, N>           |                                   |                | bytes       |
-| array of string              | struct {size\_t, T * }, size\_t N |                |             |
+| IDL type                     | C type                            | C++ type       | Python type                      |
+| ---------------------------- | --------------------------------- | -------------- | -------------------------------- |
+| octet[N]                     |                                   |                | bytes                            |
+| sequence<octet>              |                                   |                | bytes                            |
+| sequence<octet, N>           |                                   |                | bytes                            |
+| array of string              | struct {size\_t, T * }, size\_t N |                |                                  |
+| int8[N]                      |                                   |                | numpy.array(dtype=numpy.int8)    |
+| sequence<int8>               |                                   |                | numpy.array(dtype=numpy.int8)    |
+| sequence<int8, N>            |                                   |                | numpy.array(dtype=numpy.int8)    |
+| uint8[N]                     |                                   |                | numpy.array(dtype=numpy.uint8)   |
+| sequence<uint8>              |                                   |                | numpy.array(dtype=numpy.uint8)   |
+| sequence<uint8, N>           |                                   |                | numpy.array(dtype=numpy.uint8)   |
+| int16[N]                     |                                   |                | numpy.array(dtype=numpy.int16)   |
+| sequence<int16>              |                                   |                | numpy.array(dtype=numpy.int16)   |
+| sequence<int16, N>           |                                   |                | numpy.array(dtype=numpy.int16)   |
+| uint16[N]                    |                                   |                | numpy.array(dtype=numpy.uint16)  |
+| sequence<uint16>             |                                   |                | numpy.array(dtype=numpy.uint16)  |
+| sequence<uint16, N>          |                                   |                | numpy.array(dtype=numpy.uint16)  |
+| int32[N]                     |                                   |                | numpy.array(dtype=numpy.int32)   |
+| sequence<int32>              |                                   |                | numpy.array(dtype=numpy.int32)   |
+| sequence<int32, N>           |                                   |                | numpy.array(dtype=numpy.int32)   |
+| uint32[N]                    |                                   |                | numpy.array(dtype=numpy.uint32)  |
+| sequence<uint32>             |                                   |                | numpy.array(dtype=numpy.uint32)  |
+| sequence<uint32, N>          |                                   |                | numpy.array(dtype=numpy.uint32)  |
+| int64[N]                     |                                   |                | numpy.array(dtype=numpy.int64)   |
+| sequence<int64>              |                                   |                | numpy.array(dtype=numpy.int64)   |
+| sequence<int64, N>           |                                   |                | numpy.array(dtype=numpy.int64)   |
+| uint64[N]                    |                                   |                | numpy.array(dtype=numpy.uint64)  |
+| sequence<uint64>             |                                   |                | numpy.array(dtype=numpy.uint64)  |
+| sequence<uint64, N>          |                                   |                | numpy.array(dtype=numpy.uint64)  |
+| float[N]                     |                                   |                | numpy.array(dtype=numpy.float\_) |
+| sequence<float>              |                                   |                | numpy.array(dtype=numpy.float\_) |
+| sequence<float, N>           |                                   |                | numpy.array(dtype=numpy.float\_) |
+| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
+| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
+| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
+| long double[N]               |                                   |                | numpy.array(dtype=numpy.float64) |
+| sequence<long double>        |                                   |                | numpy.array(dtype=numpy.float64) |
+| sequence<long double, N>     |                                   |                | numpy.array(dtype=numpy.float64) |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -55,9 +55,9 @@ ROS 2 supports a subset of the [OMG IDL 4.2 specification](https://www.omg.org/s
 
 #### 7.4.1.4.4.2.3 Char Type
 
-| IDL type | Value range                   |
-| -------- | ----------------------------- |
-| char     | A 8-bit single-byte character |
+| IDL type | Value range                                                                            |
+| -------- | -------------------------------------------------------------------------------------- |
+| char     | A 8-bit single-byte character with a numerical value between 0 and 255 (see 7.2.6.2.1) |
 
 The type can store a single-byte character from any byte-oriented code set, or
 when used in an array, encode a multi-byte character from a multi-byte code set.
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -168,7 +168,7 @@ The next table defines how IDL types are mapped to the following programming lan
 | char        | signed char | char                           | str with length 1    |
 | wchar       | char16\_t   | char16\_t                      | str with length 1    |
 | bool        | \_Bool      | bool                           | bool                 |
-| octet       | uint8\_t    | std::byte[<sub>1</sub>](#byte) | bytes with length 1  |
+| octet       | uint8\_t    | std::byte[<sup>1</sup>](#byte) | bytes with length 1  |
 | int8        | int8\_t     | int8\_t                        | int                  |
 | uint8       | uint8\_t    | uint8\_t                       | int                  |
 | int16       | int16\_t    | int16\_t                       | int                  |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -182,15 +182,15 @@ The next table defines how IDL types are mapped to the following programming lan
 
 The next table defines how IDL templated and constructed types are mapped to the programming languages (unless specified otherwise below).
 
-| IDL type       | C type                 | C++ type            | Python type |
-| -------------- | ---------------------- | ------------------- | ----------- |
-| T[N]           | T[N]                   | std::array<T, N>    | list        |
-| sequence<T>    | struct {size\_t, T * } | std::vector<T>      | list        |
-| sequence<T, N> | struct {size\_t, T * } | custom\_class<T, N> | list        |
-| string         | char *                 | std::string         | str         |
-| string<N>      | char *                 | std::string         | str         |
-| wstring        | char16\_t *            | std::u16string      | str         |
-| wstring<N>     | char16\_t *            | std::u16string      | str         |
+| IDL type       | C type                 | C++ type          | Python type |
+| -------------- | ---------------------- | ----------------- | ----------- |
+| T[N]           | T[N]                   | std::array<T, N>  | list        |
+| sequence<T>    | struct {size\_t, T * } | std::vector<T>    | list        |
+| sequence<T, N> | struct {size\_t, T * } | std::vector<T, N> | list        |
+| string         | char *                 | std::string       | str         |
+| string<N>      | char *                 | std::string       | str         |
+| wstring        | char16\_t *            | std::u16string    | str         |
+| wstring<N>     | char16\_t *            | std::u16string    | str         |
 
 These array and sequence types have special mappings.
 If a cell is blank then the default mapping is used.
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -188,9 +188,11 @@ If a cell is blank then the default mapping is used.
 
 | IDL type                     | C type                 | C++ type       | Python type |
 | ---------------------------- | ---------------------- | -------------- | ----------- |
+| octet[N]                     |                        |                | bytes       |
+| sequence<octet>              |                        |                | bytes       |
+| sequence<octet, N>           |                        |                | bytes       |
 | array of string              | rcutils\_string\_array |                |             |
 | unbounded sequence of string | rcutils\_string\_array |                |             |
 | bounded sequence of string   | rcutils\_string\_array |                |             |
-| array of octet               |                        |                | bytes       |
 | array of char                | char *                 | std::string    | str         |
 | array of wchar               | char16\_t *            | std::u16string | str         |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -181,16 +181,18 @@ The next table defines how IDL types are mapped to the following programming lan
 1. <a name="byte"></a>If std::byte is not available then uint8\_t is used instead.
 
 The next table defines how IDL templated and constructed types are mapped to the programming languages (unless specified otherwise below).
+When specified `T` is either one of the types above or an IDL struct.
+`N` is the upper limit of a bounded type.
 
-| IDL type       | C type                 | C++ type          | Python type |
-| -------------- | ---------------------- | ----------------- | ----------- |
-| T[N]           | T[N]                   | std::array<T, N>  | list        |
-| sequence<T>    | struct {size\_t, T * } | std::vector<T>    | list        |
-| sequence<T, N> | struct {size\_t, T * } | std::vector<T, N> | list        |
-| string         | char *                 | std::string       | str         |
-| string<N>      | char *                 | std::string       | str         |
-| wstring        | char16\_t *            | std::u16string    | str         |
-| wstring<N>     | char16\_t *            | std::u16string    | str         |
+| IDL type       | C type                            | C++ type          | Python type |
+| -------------- | --------------------------------- | ----------------- | ----------- |
+| T[N]           | T[N]                              | std::array<T, N>  | list        |
+| sequence<T>    | struct {size\_t, T * }            | std::vector<T>    | list        |
+| sequence<T, N> | struct {size\_t, T * }, size\_t N | std::vector<T, N> | list        |
+| string         | char *                            | std::string       | str         |
+| string<N>      | char *                            | std::string       | str         |
+| wstring        | char16\_t *                       | std::u16string    | str         |
+| wstring<N>     | char16\_t *                       | std::u16string    | str         |
 
 These array and sequence types have special mappings.
 If a cell is blank then the default mapping is used.
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -160,23 +160,25 @@ The next table defines how IDL types are mapped to the following programming lan
 * C++11 (or higher)
 * Python 3 (or higher)
 
-| IDL type    | C type      | C++ type    | Python type          |
-| --------    | ----------- | ----------- | -------------------- |
-| float       | float       | float       | float                |
-| double      | double      | double      | float                |
-| long double | long double | long double | float                |
-| char        | signed char | char        | str with length 1    |
-| wchar       | char16\_t   | char16\_t   | str with length 1    |
-| bool        | \_Bool      | bool        | bool                 |
-| octet       | uint8\_t    | uint8\_t    | bytes with length 1  |
-| int8        | int8\_t     | int8\_t     | int                  |
-| uint8       | uint8\_t    | uint8\_t    | int                  |
-| int16       | int16\_t    | int16\_t    | int                  |
-| uint16      | uint16\_t   | uint16\_t   | int                  |
-| int32       | int32\_t    | int32\_t    | int                  |
-| uint32      | uint32\_t   | uint32\_t   | int                  |
-| int64       | int64\_t    | int64\_t    | int                  |
-| uint64      | uint64\_t   | uint64\_t   | int                  |
+| IDL type    | C type      | C++ type                       | Python type          |
+| --------    | ----------- | ------------------------------ | -------------------- |
+| float       | float       | float                          | float                |
+| double      | double      | double                         | float                |
+| long double | long double | long double                    | float                |
+| char        | signed char | char                           | str with length 1    |
+| wchar       | char16\_t   | char16\_t                      | str with length 1    |
+| bool        | \_Bool      | bool                           | bool                 |
+| octet       | uint8\_t    | std::byte[<sub>1</sub>](#byte) | bytes with length 1  |
+| int8        | int8\_t     | int8\_t                        | int                  |
+| uint8       | uint8\_t    | uint8\_t                       | int                  |
+| int16       | int16\_t    | int16\_t                       | int                  |
+| uint16      | uint16\_t   | uint16\_t                      | int                  |
+| int32       | int32\_t    | int32\_t                       | int                  |
+| uint32      | uint32\_t   | uint32\_t                      | int                  |
+| int64       | int64\_t    | int64\_t                       | int                  |
+| uint64      | uint64\_t   | uint64\_t                      | int                  |
+
+1. <a name="byte"></a>If std::byte is not available then uint8\_t is used instead.
 
 The next table defines how IDL templated and constructed types are mapped to the programming languages (unless specified otherwise below).
 
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -194,5 +194,3 @@ If a cell is blank then the default mapping is used.
 | array of string              | rcutils\_string\_array |                |             |
 | unbounded sequence of string | rcutils\_string\_array |                |             |
 | bounded sequence of string   | rcutils\_string\_array |                |             |
-| array of char                | char *                 | std::string    | str         |
-| array of wchar               | char16\_t *            | std::u16string | str         |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -199,10 +199,18 @@ If a cell is blank then the default mapping is used.
 
 | IDL type                     | C type                            | C++ type       | Python type                      |
 | ---------------------------- | --------------------------------- | -------------- | -------------------------------- |
+| float[N]                     |                                   |                | numpy.array(dtype=numpy.float\_) |
+| sequence<float>              |                                   |                | numpy.array(dtype=numpy.float\_) |
+| sequence<float, N>           |                                   |                | numpy.array(dtype=numpy.float\_) |
+| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
+| sequence<double>             |                                   |                | numpy.array(dtype=numpy.float32) |
+| sequence<double, N>          |                                   |                | numpy.array(dtype=numpy.float32) |
+| long double[N]               |                                   |                | numpy.array(dtype=numpy.float64) |
+| sequence<long double>        |                                   |                | numpy.array(dtype=numpy.float64) |
+| sequence<long double, N>     |                                   |                | numpy.array(dtype=numpy.float64) |
 | octet[N]                     |                                   |                | bytes                            |
 | sequence<octet>              |                                   |                | bytes                            |
 | sequence<octet, N>           |                                   |                | bytes                            |
-| array of string              | struct {size\_t, T * }, size\_t N |                |                                  |
 | int8[N]                      |                                   |                | numpy.array(dtype=numpy.int8)    |
 | sequence<int8>               |                                   |                | numpy.array(dtype=numpy.int8)    |
 | sequence<int8, N>            |                                   |                | numpy.array(dtype=numpy.int8)    |
@@ -227,12 +235,4 @@ If a cell is blank then the default mapping is used.
 | uint64[N]                    |                                   |                | numpy.array(dtype=numpy.uint64)  |
 | sequence<uint64>             |                                   |                | numpy.array(dtype=numpy.uint64)  |
 | sequence<uint64, N>          |                                   |                | numpy.array(dtype=numpy.uint64)  |
-| float[N]                     |                                   |                | numpy.array(dtype=numpy.float\_) |
-| sequence<float>              |                                   |                | numpy.array(dtype=numpy.float\_) |
-| sequence<float, N>           |                                   |                | numpy.array(dtype=numpy.float\_) |
-| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
-| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
-| double[N]                    |                                   |                | numpy.array(dtype=numpy.float32) |
-| long double[N]               |                                   |                | numpy.array(dtype=numpy.float64) |
-| sequence<long double>        |                                   |                | numpy.array(dtype=numpy.float64) |
-| sequence<long double, N>     |                                   |                | numpy.array(dtype=numpy.float64) |
+| array of string              | struct {size\_t, T * }, size\_t N |                |                                  |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -166,7 +166,7 @@ The next table defines how IDL types are mapped to the following programming lan
 | float       | float            | float                          | float                |
 | double      | double           | double                         | float                |
 | long double | long double      | long double[<sup>2</sup>](#ld) | float                |
-| char        | signed char      | char                           | str with length 1    |
+| char        | unsigned char    | unsigned char                  | str with length 1    |
 | wchar       | char16\_t        | char16\_t                      | str with length 1    |
 | bool        | \_Bool           | bool                           | bool                 |
 | octet       | unsigned char    | std::byte[<sup>1</sup>](#byte) | bytes with length 1  |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -149,8 +149,9 @@ For now only a one-dimensional, fixed-size array is supported though.
 
 Generated code for messages is only guaranteed to enforce constraints when a message is published.
 For example, there might not be an error when a field constrained to be a fixed size array is assigned an array with the wrong size.
-In the future constraints might be checked when a field is set.
-Users of generated message code should assume constraints are enforced when fields are set or modified.
+This could be inconsistent accross client library implementations depending on language features and performance cost.
+In the future any client library may add additional constraint checking when a field is set or modified.
+Users of generated message code should assume constraints could be enforced at any time to be compatible with such a change.
 
 ### Type Mapping
 
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -154,8 +154,8 @@ IDL types mapped to different types in different languages.
 | --------    | ----------- | ----------- | -------------------- |
 | bool        | \_Bool      | bool        | bool                 |
 | octet       | uint8\_t    | uint8\_t    | bytes with length 1  |
-| char        | char16\_t   | char16\_t   | str with length 1    |
-| wchar       | signed char | char        | str with length 1    |
+| char        | signed char | char        | str with length 1    |
+| wchar       | char16\_t   | char16\_t   | str with length 1    |
 | float       | float       | float       | float                |
 | double      | double      | double      | float                |
 | long double | long double | long double | float                |
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -145,6 +145,13 @@ For now only a one-dimensional, fixed-size array is supported though.
   <b>TODO:</b> All of this too
 </div>
 
+### Constraint Checking
+
+Generated code for messages is guaranteed to enforce constraints only when a message is published.
+For example, there might not be an error when a field constrained to be a fixed size array is assigned an array with the wrong size.
+In the future constraints might be checked when a field is set.
+Users of generated message code should assume constraints are enforced when fields are set.
+
 ### Type Mapping
 
 The next table defines how IDL types are mapped to the following programming languages:
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -150,21 +150,21 @@ For now only a one-dimensional, fixed-size array is supported though.
 IDL types mapped to different types in different languages.
 
 
-| IDL type | C type    | C++ type    | Python type          |
-| -------- | --------- | ----------- | -------------------- |
-| bool     | \_Bool    | bool        | bool                 |
-| octet    | uint8\_t  | uint8\_t    | bytes with length 1  |
-| char     | char      | char        | str with length 1    |
-| float    | float     | float       | float                |
-| double   | double    | double      | float                |
-| int8     | int8\_t   | int8\_t     | int                  |
-| uint8    | uint8\_t  | uint8\_t    | int                  |
-| int16    | int16     | int16       | int                  |
-| uint16   | uint16    | uint16      | int                  |
-| int32    | int32     | int32       | int                  |
-| uint32   | uint32    | uint32      | int                  |
-| int64    | int64     | int64       | int                  |
-| uint64   | uint64\_t | uint64\_t   | int                  |
+| IDL type | C type      | C++ type    | Python type          |
+| -------- | ----------- | ----------- | -------------------- |
+| bool     | \_Bool      | bool        | bool                 |
+| octet    | uint8\_t    | uint8\_t    | bytes with length 1  |
+| char     | signed char | char        | str with length 1    |
+| float    | float       | float       | float                |
+| double   | double      | double      | float                |
+| int8     | int8\_t     | int8\_t     | int                  |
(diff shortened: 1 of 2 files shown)


articles/142_idl.md
```diff
@@ -35,3 +35,45 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+### Type Mapping
+
+IDL types mapped to different types in different languages.
+
+
+| IDL type | C type    | C++ type    | Python type          |
+| -------- | --------- | ----------- | -------------------- |
+| bool     | \_Bool    | bool        | bool                 |
+| octet    | uint8\_t  | uint8\_t    | bytes with length 1  |
```
sloretz: `bytes` feels odd since it is a sequence type in python. Thoughts on `ctypes.c_char` or `int` to match `uint8_t` instead?

articles/142_idl.md
```diff
@@ -35,3 +35,45 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| int32    | int32     | int32       | int                  |
+| uint32   | uint32    | uint32      | int                  |
+| int64    | int64     | int64       | int                  |
+| uint64   | uint64\_t | uint64\_t   | int                  |
+
+Unless are the default mappings for array, sequence, and string types unless otherwise specified.
+
+| IDL type                | C type                 | C++ type            | Python type |
+| ----------------------- | ---------------------- | ------------------- | ----------- |
+| array                   | T [FIXED\_SIZE]        | std::array<T, N>    | list        |
```
sloretz: IDL defines array as fixed size (7.4.1.4.4.5 `The array size (in each dimension) is fixed at compile time. The implementation of array indices is language mapping specific.`). `list` is a dynamic type in python. Maybe `tuple` makes more sense here? Only problem is when a user creates a message they can't append to the field, and instead must assign a new tuple too it. Passing a message with a tuple received on a subscription make sense though.

articles/142_idl.md
```diff
@@ -35,3 +35,45 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+### Type Mapping
+
+IDL types mapped to different types in different languages.
+
+
+| IDL type | C type    | C++ type    | Python type          |
+| -------- | --------- | ----------- | -------------------- |
+| bool     | \_Bool    | bool        | bool                 |
+| octet    | uint8\_t  | uint8\_t    | bytes with length 1  |
+| char     | char      | char        | str with length 1    |
```
mikaelarguedas: In C char type maps to signed char: https://github.com/ros2/rosidl/blob/ac4ac643d55cd7fb6c9604d0e67dc86dcd8d8d53/rosidl_generator_c/include/rosidl_generator_c/primitives_array.h#L33

mikaelarguedas: As this is a string, the question of the encoding may propagate here as well

articles/142_idl.md
```diff
@@ -35,3 +35,45 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| float    | float     | float       | float                |
+| double   | double    | double      | float                |
```
mikaelarguedas: should we add long double here ?

articles/142_idl.md
```diff
@@ -35,3 +35,45 @@ Further it describes how the interfaces are used to generate code in C, C++ and
 <div class="alert alert-warning" markdown="1">
   <b>TODO:</b> All of this too
 </div>
+
+### Type Mapping
+
+IDL types mapped to different types in different languages.
+
+
+| IDL type | C type    | C++ type    | Python type          |
```
mikaelarguedas: There seem to be differences between IDL types and DDS types. Would it make sense to have a DDS Type column here?

articles/143_legacy_interface_definition.md
```diff
@@ -241,3 +241,34 @@ Depending on the type the following values are valid:
...
+| uint16   | unsigned short     |
+| int32    | long               |
+| uint32   | unsigned long      |
+| int64    | long long          |
+| uint64   | unsigned long long |
+| string   | string             |
+
+| ROS type                | IDL type       |
+| ----------------------- | -------------- |
+| static array            | T\[N\]         |
```
mikaelarguedas: It looks like these IDL types are named array, unbounded sequence etc in the article above.

articles/142_idl.md
sloretz: By DDS types do you mean DDS-XTYPES 1.2?

articles/143_legacy_interface_definition.md
sloretz: 5e2e31ca1772f7cfb65137c3b62edf1aefe8806b

articles/142_idl.md
sloretz: 7c1d4a45826ffbb9137305e1c8848d69697b10a0 for signed char

articles/142_idl.md
sloretz: 750fec5fd6b3c16d50c41096177421dcc9b95324

articles/142_idl.md
sloretz: I threw in wchar and wstring as well. c9ddd20f951c06d9f5c4f85f038eb6d276ae3b50

I'm not yet sure if this doc should link to a separate unicode support design doc, or if that should be included here.

articles/142_idl.md
sloretz: Might be able to avoid adding DDS-XTYPES since the 1.2 spec has a mapping of DDS to IDL types already.

mikaelarguedas: Yes I was thinking of DDS-XTYPES, to have a reference of what will end up on the wire when using a given IDL type.

articles/142_idl.md
```diff
@@ -35,3 +35,50 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+These array and sequence types have special mappings.
+If a cell is blank then the default mapping is used.
+
+| IDL type                     | C type                 | C++ type       | Python type |
+| ---------------------------- | ---------------------- | -------------- | ----------- |
+| array of string              | rcutils\_string\_array |                |             |
+| unbounded sequence of string | rcutils\_string\_array |                |             |
+| bounded sequence of string   | rcutils\_string\_array |                |             |
+| array of octet               |                        |                | bytes       |
+| array of char                | char *                 | std::string    | str         |
```
sloretz: It may not make sense to have `array of char` map to `char *` in `C` because that means it couldn't contain a value of zero.

articles/142_idl.md
```diff
@@ -35,3 +35,50 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+IDL types mapped to different types in different languages.
+
+
+| IDL type    | C type      | C++ type    | Python type          |
+| --------    | ----------- | ----------- | -------------------- |
+| bool        | \_Bool      | bool        | bool                 |
+| octet       | uint8\_t    | uint8\_t    | bytes with length 1  |
+| char        | char16\_t   | char16\_t   | str with length 1    |
+| wchar       | signed char | char        | str with length 1    |
```
mjcarroll: Swap wchar and char types.

sloretz: Fixed in d296cfb8125387421c99c69d4f50fd6e00821119

articles/142_idl.md
```diff
@@ -35,3 +35,53 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+| IDL type       | C type                 | C++ type            | Python type |
+| -------------- | ---------------------- | ------------------- | ----------- |
+| T[N]           | T[N]                   | std::array<T, N>    | list        |
+| sequence<T>    | struct {size\_t, T * } | std::vector<T>      | list        |
+| sequence<T, N> | struct {size\_t, T * } | custom\_class<T, N> | list        |
+| string         | char *                 | std::string         | str         |
+| string<N>      | char *                 | std::string         | str         |
+| wstring        | char16\_t *            | std::u16string      | str         |
+| wstring<N>     | char16\_t *            | std::u16string      | str         |
```
dirk-thomas: I was wondering about the different handling of bounded sequences compared to bounded (w)strings in C++. For the first we go through the hassle to provide our own data structure (to enforce the upper bound) for the other two we don't (atm).

Maybe we should use the same approach homogeneously - it could be either of the two, The effort and "customness" of types we define vs. the delayed check until the data structure is being passed to the lower levels.

articles/142_idl.md
```diff
@@ -35,3 +35,53 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+| IDL type    | C type      | C++ type    | Python type          |
+| --------    | ----------- | ----------- | -------------------- |
+| float       | float       | float       | float                |
+| double      | double      | double      | float                |
+| long double | long double | long double | float                |
+| char        | signed char | char        | str with length 1    |
+| wchar       | char16\_t   | char16\_t   | str with length 1    |
+| bool        | \_Bool      | bool        | bool                 |
+| octet       | uint8\_t    | uint8\_t    | bytes with length 1  |
```
dirk-thomas: Looking at newer specs I was wondering if the C++ mapping should be `byte` (available as of C++17) instead. To match the semantic C could then be mapped to unsigned `char`.

[We would need to check for availability across all targeted platforms. Even if not all support the type yet we could still make the decision today (using a custom typedef or using `char` directly) and switch to the standard type once available.]

articles/142_idl.md
mikaelarguedas: yeah I was thinking the same for the C part, currently in C we have custom data structure for all the dynamic types (bounded and unbounded). AFAICT in C++ we have a custom container for bounded vectors but nothing for strings ATM (as described in previous comment).

In the delayed check approach, when / how would the user be notified that the data doesnt fit?

articles/142_idl.md
```diff
@@ -35,3 +35,53 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| int64       | int64\_t    | int64\_t    | int                  |
+| uint64      | uint64\_t   | uint64\_t   | int                  |
+
+The next table defines how IDL templated and constructed types are mapped to the programming languages (unless specified otherwise below).
+
+| IDL type       | C type                 | C++ type            | Python type |
+| -------------- | ---------------------- | ------------------- | ----------- |
+| T[N]           | T[N]                   | std::array<T, N>    | list        |
+| sequence<T>    | struct {size\_t, T * } | std::vector<T>      | list        |
+| sequence<T, N> | struct {size\_t, T * } | custom\_class<T, N> | list        |
```
mikaelarguedas: should this be `struct {size\_t, T * , N }` instead? (where N would be the max capacity)

articles/142_idl.md
mikaelarguedas: currently in C we have a [custom data structure](https://github.com/ros2/rosidl/blob/ac4ac643d55cd7fb6c9604d0e67dc86dcd8d8d53/rosidl_generator_c/include/rosidl_generator_c/primitives_array.h#L22) for this that has a size component, so I think it is possible for array elements to have any value in the char range in them, we would just not be able to treat it as a string (strlen etc).

articles/142_idl.md
sloretz: I think max should to be a `const` variable; because if it's non-const then it would need to be in two places. First in the struct that can be modified by a user, and second in code used to validate a message.

If that sounds reasonable then it implies the struct with a field for max capacity would be generated for every field. Alternatively a struct definition could be reused if the max capacity was a separate const variable.

```c
const size_t example_msgs__example_bounded__sensor_readings__max_capacity = 64;
//  ...
struct example_msgs__example_bounded
{
  // ...
  rcutils_float_array * sensor_readings = NULL;
}
```

articles/142_idl.md
dirk-thomas: >  when / how would the user be notified that the data doesnt fit?

The `rclpy_publish` function converts the Python object into a ROS C message. If any of the constraints are violated it can signaled through an exception.

Optionally we can offer a validation function which the user could invoke explicitly.

mikaelarguedas: In Python I would expect the user to not be allowed to place data bigger than the message specification at the Python level already (what we currently do with custom setters).

My question was about the C language that doesn't provide exceptions. If we map to a `char(16_t) *` I am not sure how we would restrict the memory allocation or report the error.

But your point still holds: it can be done when publish is called rather than when filling the message datastructure.

dirk-thomas: > In Python I would expect the user to not be allowed to place data bigger than the message specification at the Python level already (what we currently do with custom setters).

That is true but only covers the case where a field is being "assigned". But e.g. a list is mutable and new items can be added after the initial assignment with the property of the class being able to intercept (assuming we don't mock all method of the list instance).

> My question was about the C language that doesn't provide exceptions. If we map to a char(16_t) * I am not sure how we would restrict the memory allocation or report the error.

In a future `rclc` client library the `publish` call can perform the validation the same way as in Python. While the `publish` in `rcl` could also check the constraints I would rather not do that and rely in the client library on top to do that. All of these functions have return code to signal error already (if exceptions are not available / used).

dirk-thomas:
1c4fa2f7e3fa92ad9145ec8e29fdeb5ad2907821 defines a custom mapping to `bytes` for octet arrays as well as (bounded) sequences.

articles/142_idl.md
```diff
@@ -35,3 +35,55 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| IDL type                     | C type                 | C++ type       | Python type |
+| ---------------------------- | ---------------------- | -------------- | ----------- |
+| octet[N]                     |                        |                | bytes       |
+| sequence<octet>              |                        |                | bytes       |
+| sequence<octet, N>           |                        |                | bytes       |
+| array of string              | rcutils\_string\_array |                |             |
+| unbounded sequence of string | rcutils\_string\_array |                |             |
+| bounded sequence of string   | rcutils\_string\_array |                |             |
+| array of char                | char *                 | std::string    | str         |
+| array of wchar               | char16\_t *            | std::u16string | str         |
```
dirk-thomas: The C and C++ mapping for arrays of (w)char need to be different since both don't allow for null characters which isn't forbidden for these IDL types.

articles/142_idl.md
```diff
@@ -35,3 +35,55 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| string         | char *                 | std::string         | str         |
+| string<N>      | char *                 | std::string         | str         |
+| wstring        | char16\_t *            | std::u16string      | str         |
+| wstring<N>     | char16\_t *            | std::u16string      | str         |
+
+These array and sequence types have special mappings.
+If a cell is blank then the default mapping is used.
+
+| IDL type                     | C type                 | C++ type       | Python type |
+| ---------------------------- | ---------------------- | -------------- | ----------- |
```
dirk-thomas: Do we want to specify the mapping to use numpy for scalar type? E.g.

* `uint32[N]`: `numpy.ndarray(shape=(N, ), dtype=numpy.uint32)`
* `sequence<uint32 [, N]\>`: `numpy.ndarray(shape=(0, ), dtype=numpy.uint32)`

sloretz: +1 for `octet[N]` being `np.array()`. It would make it easier to support multiple dimension arrays in the future. numpy seems like a minimal enough dependency. `pip freeze` after installing numpy in a venv only shows `numpy` and `pkg-resources` (which was in the env by default).

dirk-thomas: How about other scalar types and sequences?

> +1 for octet[N] being np.array()

For `octet[N]` `bytes` already provides a native mapping (without the usual Python overhead) so I am not sure about that case.

dirk-thomas: > +1 for octet[N] being np.array()

For `octet[N]` `bytes` already provides a native mapping (without the usual Python overhead) so I am not sure about that case.

articles/142_idl.md
sloretz: > While the publish in rcl could also check the constraints I would rather not do that and rely in the client library on top to do that.

Just adding it looks like either client libraries or the rmw implementations must check the constraints because `rcl_publish` sees the message as a [`void *`](https://github.com/ros2/rcl/blob/0a6d9186e8fbd133308d0ad4bc3b75d61078b04d/rcl/src/rcl/publisher.c#L234).

> In Python I would expect the user to not be allowed to place data bigger than the message specification at the Python level already (what we currently do with custom setters).

I think ros2/ros2#509 is a hint that ros2 should consider only checking the validity of the message once. From the perspective of ROS 2 as a library the message constraints don't matter until the message is used. From the perspective of a developer writing code constraint checking when a field is set is nice because it errors closer to the bug, but from the perspective of a user running the code it can be a significant performance hit for little gain when most of the time everything is working fine.

articles/142_idl.md
sloretz: array of char and wchar no longer strings in cd1e3b27b1352f83effd43f8b123c23728b04542

articles/142_idl.md
mikaelarguedas: Looking at [cppreference](https://en.cppreference.com/w/cpp/compiler_support) and [Visual Studio conformance announcement](https://blogs.msdn.microsoft.com/vcblog/2018/05/07/announcing-msvc-conforms-to-the-c-standard/) it looks like all compilers support it

articles/142_idl.md
```diff
@@ -35,3 +35,62 @@ Further it describes how the interfaces are used to generate code in C, C++ and
 <div class="alert alert-warning" markdown="1">
   <b>TODO:</b> All of this too
 </div>
+
+### Constraint Checking
+
+Generated code for messages is guaranteed to enforce constraints only when a message is published.
```
dirk-thomas: I think this sounds ambiguous:

> guaranteed to enforce ... **only** when ... published

I don't think we want to indicate that the want to guarantee that we don't check it before. Maybe rephrase to:

> Generated code for messages is only guaranteed to enforce constraints when a message is published.

articles/142_idl.md
```diff
@@ -35,3 +35,62 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+For example, there might not be an error when a field constrained to be a fixed size array is assigned an array with the wrong size.
```
dirk-thomas, line 151: Inject the following?

> For example, *depending on the language* there ...

In Python we could (and probably will) check this trivially in the setter.

articles/142_idl.md
```diff
@@ -35,3 +35,62 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+In the future constraints might be checked when a field is set.
```
dirk-thomas: Maybe append the following?

> ... and/or mutated.

Same for the next line.

articles/142_idl.md
```diff
@@ -35,3 +35,62 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+The next table defines how IDL templated and constructed types are mapped to the programming languages (unless specified otherwise below).
+When specified `T` is either one of the types above or an IDL struct.
+`N` is the upper limit of a bounded type.
+
+| IDL type       | C type                            | C++ type          | Python type |
+| -------------- | --------------------------------- | ----------------- | ----------- |
+| T[N]           | T[N]                              | std::array<T, N>  | list        |
+| sequence<T>    | struct {size\_t, T * }            | std::vector<T>    | list        |
+| sequence<T, N> | struct {size\_t, T * }, size\_t N | std::vector<T, N> | list        |
```
dirk-thomas: `std::vector` doesn't have a second template argument `N`.

articles/142_idl.md
```diff
@@ -35,3 +35,62 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| int8        | int8\_t     | int8\_t                        | int                  |
+| uint8       | uint8\_t    | uint8\_t                       | int                  |
+| int16       | int16\_t    | int16\_t                       | int                  |
+| uint16      | uint16\_t   | uint16\_t                      | int                  |
+| int32       | int32\_t    | int32\_t                       | int                  |
+| uint32      | uint32\_t   | uint32\_t                      | int                  |
+| int64       | int64\_t    | int64\_t                       | int                  |
+| uint64      | uint64\_t   | uint64\_t                      | int                  |
+
+1. <a name="byte"></a>If std::byte is not available then uint8\_t is used instead.
```
dirk-thomas: Instead of `uint8_t` I would suggest `unsigned char` since that is how it is defined in C++17.

articles/142_idl.md
```diff
@@ -35,3 +35,95 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+These array and sequence types have special mappings.
+If a cell is blank then the default mapping is used.
+
+| IDL type                     | C type                            | C++ type       | Python type                      |
+| ---------------------------- | --------------------------------- | -------------- | -------------------------------- |
+| octet[N]                     |                                   |                | bytes                            |
+| sequence<octet>              |                                   |                | bytes                            |
+| sequence<octet, N>           |                                   |                | bytes                            |
+| array of string              | struct {size\_t, T * }, size\_t N |                |                                  |
+| int8[N]                      |                                   |                | numpy.array(dtype=numpy.int8)    |
```
dirk-thomas: Should the fixed-size arrays also have `shape=(N, )`?

articles/142_idl.md
```diff
@@ -35,3 +35,95 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| string         | char *                            | std::string       | str         |
+| string<N>      | char *                            | std::string       | str         |
+| wstring        | char16\_t *                       | std::u16string    | str         |
+| wstring<N>     | char16\_t *                       | std::u16string    | str         |
+
+These array and sequence types have special mappings.
+If a cell is blank then the default mapping is used.
+
+| IDL type                     | C type                            | C++ type       | Python type                      |
+| ---------------------------- | --------------------------------- | -------------- | -------------------------------- |
```
dirk-thomas: I would suggest to use the same order as in the first "Type Mapping" table (currently lines 56-70).

articles/142_idl.md
sloretz: c80995f7704643eed1bd7847c91f69a77f180f2e

articles/142_idl.md
sloretz: aee9a126bcce1b60a9646aa4bc64be8552a898af

articles/142_idl.md
sloretz: 4a49a00896448225eee06cd483ba9484a879e909

Side note, I learned that in `C` and `C++` an `unsigned char` is guaranteed to be 1 byte, but 1 byte may be more than 8 bits. It doesn't look like any modern systems do that, but if they do the number of bits in a byte is defined in `<limits.h>` or `<climits>` as `CHAR_BIT`.

articles/142_idl.md
sloretz: It looks like it doesn't accept a keyword argument of `shape`, and the text of `repr()` doesn't show it either. As far as I can tell the shape is inferred from the data passed in.

articles/142_idl.md
sloretz: 4fdd81a40535e8ef2caeaa97a325149da231671d

articles/142_idl.md
sloretz: Added another sentence saying constraint checks could be inconsistent depending on language features and performance cost 46ec1098fbc09fb93278003db8185ca45d37a72b

articles/142_idl.md
sloretz: 9a56e5d78d69a52fa04f159b7e9683030b1896f0 and 46ec1098fbc09fb93278003db8185ca45d37a72b

articles/142_idl.md
dirk-thomas: The following lines work for me:

```
import numpy
numpy.ndarray(shape=(5, ), dtype=numpy.uint32)
```

sloretz: Ah, I was trying `numpy.array(...)`. It looks like the underlying type is `ndarray` and `array` is just a helper to construct one. I'll change the types.

articles/142_idl.md
```diff
@@ -35,3 +35,96 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+| string<N>      | char *                            | std::string      | str         |
+| wstring        | char16\_t *                       | std::u16string   | str         |
+| wstring<N>     | char16\_t *                       | std::u16string   | str         |
+
+These array and sequence types have special mappings.
+If a cell is blank then the default mapping is used.
+
+| IDL type                     | C type                            | C++ type       | Python type                      |
+| ---------------------------- | --------------------------------- | -------------- | -------------------------------- |
+| float[N]                     |                                   |                | numpy.array(dtype=numpy.float\_) |
```
dirk-thomas: `float_`? Isn't that the same as `float64`?

articles/142_idl.md
sloretz: 479e3a1f8488fbbdb6b8ebc00193e3dd809b9513

articles/142_idl.md
sloretz: Oops it is. 48389920ec41cbe54d9648dafcab4d76d3b37f5e makes `float -> numpy.float32`, `double -> numpy.float64` and `long double -> list(float)`

mikaelarguedas: Should we add special mapping  for `long double`s be arrays of  `numpy.float128` as well?

articles/142_idl.md
```diff
@@ -35,3 +35,93 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+* C11 (or higher)
+* C++11 (or higher)
+* Python 3 (or higher)
+
+| IDL type    | C type           | C++ type                       | Python type          |
+| --------    | -----------      | ------------------------------ | -------------------- |
+| float       | float            | float                          | float                |
+| double      | double           | double                         | float                |
+| long double | long double      | long double                    | float                |
```
sloretz: Thoughts on not initially supporting `long double`? It seems inconsistent across platforms and DDS vendors.

Visual Studio `typedef`'d `long double` to `double` which is 64bit
https://docs.microsoft.com/en-us/cpp/c-language/type-long-double?view=vs-2017

Connext works around this by mapping `DDS_LongDouble` to a `char` array (of size 16?) on some platforms and `long double` on others.
https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Translations_for_IDL_Types.htm#Note2_LongDouble

Opensplice doesn't appear to support it. The grammar on [page 589 / pdf page 617](http://download.prismtech.com/docs/Vortex/pdfs/OpenSplice_refman_CPP.pdf)
 only defines `<floating_pt_type>` as `"float"` and `"double"`.

Fast-CDR uses the C++ type `long double` and serializes it as 16 bytes on the wire. https://github.com/eProsima/Fast-CDR/blob/1df42bbd1ea86a9c99bb8fa72104f19872235986/src/cpp/Cdr.cpp#L528

mikaelarguedas: My understanding from original discussion was that we *should* support all IDL types. @dirk-thomas is that correct?

Looking a bit more into it:
- It looks like Connext and Fast-RTPS agree that it should be a 16-byte thing on the wire

The issues we face right now:
- FastRTPS doesn't have special handling to account for the fact that the type has a different size on windows
- OpenSplice doesn't support this type

Looking at Opensplice it appears that wchar and wstring are not supported either https://github.com/ADLINK-IST/opensplice/blob/36e4f9c250a640ade72d5af92084f5f5243ce96c/src/tools/idlpp/code/idl_unsupported.h.

Not convinced it should be a blocker for supporting it though. We may just need to have implementation specific limitations or workarounds

sloretz: after talking with @mikaelarguedas I think this can be resolved by saying `long double `will be supported, but users will have to be cautious of middleware support for it.

I'm not sure if we should use the C and C++ type `long double` on all platforms, but that seems like the easiest thing to do for now.

sloretz: Added warning about `long double` on some platforms and middleware in 
2082ba0

articles/143_legacy_interface_definition.md
```diff
@@ -241,3 +241,34 @@ Depending on the type the following values are valid:
...
+Interfaces described using the legacy format are first converted to [IDL](articles/idl_interface_definition.html).
+Code generation uses the generated file.
+
+### Mapping to IDL types
+
+| ROS type | IDL type           |
+| -------- | ------------------ |
+| bool     | boolean            |
+| byte     | octet              |
+| char     | char               |
```
sloretz: IDL `7.2.6.2.1` says `A char is an 8-bit quantity with a numerical value between 0 and 255 (decimal)`. The [ROS 2 interface definition says char is `[-128, 127]`](http://design.ros2.org/articles/interface_definition.html). Changing this type mapping to idl type `int8` after discussing with @mjcarroll

articles/142_idl.md
```diff
@@ -144,3 +144,95 @@ For now only a one-dimensional, fixed-size array is supported though.
...
+* C11 (or higher)
+* C++11 (or higher)
+* Python 3 (or higher)
+
+| IDL type    | C type           | C++ type                       | Python type          |
+| --------    | -----------      | ------------------------------ | -------------------- |
+| float       | float            | float                          | float                |
+| double      | double           | double                         | float                |
+| long double | long double      | long double[<sup>2</sup>](#ld) | float                |
+| char        | signed char      | char                           | str with length 1    |
```
sloretz: IDL `7.2.6.2.1` says `A char is an 8-bit quantity with a numerical value between 0 and 255 (decimal)`. Changing to `unsigned char` in c and c++

articles/143_legacy_interface_definition.md
```diff
@@ -253,7 +253,7 @@ Code generation uses the generated file.
 | -------- | ------------------ |
 | bool     | boolean            |
 | byte     | octet              |
-| char     | char               |
+| char     | short              |
```
mikaelarguedas: would an `octet` make more sense than a `short` in this case ?

sloretz: I think `short` is a better fit than `octet`. IDL 4.2 section `7.4.1.4.3` says `Octet literals have integer value in the range 0…255`. `char` is an integer in range [-128, 27] in ros 2 .msg files https://github.com/ros2/design/pull/190#discussion_r220283197

mikaelarguedas: It's still unclear why it is treated differently than the `int8` type.

My understanding was that they both fit in 8-bit and can represent signed values, so I'm not sure why we should use a 16-bit type to store them

articles/143_legacy_interface_definition.md
```diff
@@ -241,3 +241,34 @@ Depending on the type the following values are valid:
...
+### Mapping to IDL types
+
+| ROS type | IDL type           |
+| -------- | ------------------ |
+| bool     | boolean            |
+| byte     | octet              |
+| char     | short              |
+| float32  | float              |
+| float64  | double             |
+| int8     | octet              |
```
sloretz: This should map to idl type `int8`

articles/143_legacy_interface_definition.md
```diff
@@ -241,3 +241,34 @@ Depending on the type the following values are valid:
...
+| uint8    | octet              |
```
sloretz: This should map to idl type `uint8`

articles/143_legacy_interface_definition.md
sloretz: Got the int8 types mixed up. It should be fixed in e7c99eafecd86ca8f41f106047177eb56902a64d

articles/142_idl.md
```diff
@@ -144,3 +144,95 @@ For now only a one-dimensional, fixed-size array is supported though.
...
+
+| IDL type       | C type                            | C++ type         | Python type |
+| -------------- | --------------------------------- | ---------------- | ----------- |
+| T[N]           | T[N]                              | std::array<T, N> | list        |
+| sequence<T>    | struct {size\_t, T * }            | std::vector<T>   | list        |
+| sequence<T, N> | struct {size\_t, T * }, size\_t N | std::vector<T>   | list        |
+| string         | char *                            | std::string      | str         |
+| string<N>      | char *                            | std::string      | str         |
+| wstring        | char16\_t *                       | std::u16string   | str         |
+| wstring<N>     | char16\_t *                       | std::u16string   | str         |
```
mikaelarguedas, line 198: from previous discussion, I think string and wstring would be structs in C (like sequences) and not just pointers, isnt' it ?

deeplearningrobotics:
@dirk-thomas:

As far as I understand the current concept is to generate a ROS 2 specific C++ type from an IDL. This leads to the fact that two types with the same data exist.

* One generated by ROS 2 from the IDL. I will call one such type `ROS2LaserScan` for later use.
* Multiple types generated by each DDS implementation from the IDLs. I will call those types `DDS1LaserScan`, `DDS2LaserScan` etc. for later use.

I have two questions:

   1. **Shared Memory / Zero Copy Shared memory performance**

      A DDS implementation might be able to transmit a type `DDS1LaserScan` with one or zero copies. But 
   it will not understand that structure of `ROS2LaserScan`. So `ROS2LaserScan` must be converted 
   (most likely copied) into a `DDS1LaserScan1`.

      **How do you preserve the single / zero-copy performance?**

   2. **Non ROS 2 DDS Participants***

      If a user wants to use directly DDS data writers and readers to access additional DDS features not 
available in ROS 2 he will be forced to use the DDS types. So an algorithm might using either using 
`DDS1LaserScan` or `ROS2LaserScan`. 

       **How do you plan integration of non-ROS 2 DDS systems without an performance impact due to type conversion?**

Please also note that RTI and OpenSplice have (AFAIK) common IDL to C/C++ etc. type translations available [here](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Translations_for_IDL_Types.htm#Table_DataTypesinIDL_C++PSM).

Thank you!

dirk-thomas:
> This leads to the fact that two types with the same data exist.

This is not a change to how the system already works. The user code always only interacts with a ROS type - it must not be coupled with a vendor specific type.

> A DDS implementation might be able to transmit a type `DDS1LaserScan` with one or zero copies. But
it will not understand that structure of `ROS2LaserScan`. So `ROS2LaserScan` must be converted
(most likely copied) into a `DDS1LaserScan1`.

Even when using shared memory DDS implementations still need to serialize the published sample into the shared memory and on the other end deserialize it into a sample class. In the current typesupport implementations the overhead is that the ROS type is first converted to the vendor specific type which then gets serialized. Nothing should prevent a RMW implementation to skip the inbetween type and implement serialization logic from the ROS type directly.

But again: this is already the same in the current code and doesn't change in any way with the transition to using `.idl` file to describe the data structures.

> How do you plan integration of non-ROS 2 DDS systems without an performance impact due to type conversion?

I think I don't understand the question. DDS systems define the type of exchanged sample and that is the "contract" between both endpoints. Since the data is exchanged in serialized form it is not relevant how the object representation looks like on both sides.

> Please also note that RTI and OpenSplice have (AFAIK) common IDL to C/C++ etc. type translations available here.

Did you notice and differences between these translations and the ones proposed in ROS?

articles/143_legacy_interface_definition.md
dirk-thomas: Mapping the ROS 1 type `char` to the IDL type `int8` seems off to me.

In ROS 1 the `char` type (while deprecated) is defined as a `uint8` in the [docs](http://wiki.ros.org/msg#Fields) as well as in the [code](https://github.com/ros/gencpp/blob/b41ee3060badd660662e21cd9d4f81971c87a420/src/gencpp/__init__.py#L41).

deeplearningrobotics:
@dirk-thomas: Thank you for your reply.

> Did you notice and differences between these translations and the ones proposed in ROS?

No, but I was hoping the IDL transition might enable a better interoperability between ROS 2 and native DDS application. 

A lot of people ask us for all the DDS features. We want to provide that while still offering the simplicity of ROS 2. This is what makes things complicated.

> In the current typesupport implementations the overhead is that the ROS type is first converted to the vendor specific type.

I was hoping for ROS 2 types having the following structure:
```
class ROS2PointCloud<DDSImplType> {
public:
  vector points() {
    return impl_->points();
  }
private:
  DDSImplType impl_;
};
```
This would allow transparent use of DDS and ROS 2 types.

> I think I don't understand the question. DDS systems define the type of exchanged sample and that is the "contract" between both endpoints. Since the data is exchanged in serialized form it is not relevant how the object representation looks like on both sides.

I'll try to elaborate with an example. Assuming I write an algorithm taking in a `ROS2PointCloud` type.

```
void segmet_pc(ROS2PointCloud) {...}
```
Now someone wants to reuse this code in a pure DDS application. To be able to do that smoothly the output of the IDLs from ROS 2 should have the exact same API as DDS types so that I can reuse the code.

Another independent question:
* Do you plan to support inheritance and unions?

dirk-thomas:
> No, but I was hoping the IDL transition might enable a better interoperability between ROS 2 and native DDS application.

The transition to use IDL will allow at least all the supported types to be exchanged transparently (e.g. by consensus on the encoding of strings). Other than that I don't know which exact part you would like to see "better interoperability". Maybe you can describe specific cases.

> I was hoping for ROS 2 types having the following structure: `class ROS2PointCloud<DDSImplType>`

Making every ROS 2 data type a template based on the concrete implementation is certainly undesired. This would require compile time knowledge of the chosen RMW impl as well as increase complexity of every code using ROS messages. One of the design goals is that the ROS data types are *not* tied to a specific implementation and can be used even without any RMW impl. in e.g. library API. Also your example implies that all members of the data structure are only to be accessed by methods. In earlier discussions the community strongly expressed the desire to keep member-based access to all the fields.

> Now someone wants to reuse this code in a pure DDS application. To be able to do that smoothly the output of the IDLs from ROS 2 should have the exact same API as DDS types so that I can reuse the code.

Not even two different DDS implementations offer the very same API for the generated data structures. So how should ROS offer that across DDS vendors (and potentially non-DDS RMW impl)? That being said most types are very close to each other (if not even identical). So nothing prevents you from implementing two signatures of the function which then use a shared function to perform the logic or use a templated function if the content of the functions works for both types.

wjwwood:
> I'll try to elaborate with an example. Assuming I write an algorithm taking in a ROS2PointCloud type.

> `void segmet_pc(ROS2PointCloud) {...}`

> Now someone wants to reuse this code in a pure DDS application. To be able to do that smoothly the output of the IDLs from ROS 2 should have the exact same API as DDS types so that I can reuse the code.

I'd avoid using the ROS type or the DDS type in a library unless you're willing to live with the consequences of that decision (a dependency on one or the other).

So I don't agree with the conclusion that:

- your library interface should be templated on a "Point cloud type" rather than a concrete type or
- that in order to make use of that templated interface the ROS message structure needs to have an identical API to the DDS messages.

In ROS 1 (and planned for ROS 2) we have something where you can pub/sub whatever type you want so long as you tell us how to serialize and deserialize it, see:

https://wiki.ros.org/roscpp/Overview/MessagesSerializationAndAdaptingTypes#Example:_Adapting_a_custom_Vector3

I think this is the right way to address the concern of wanting to use data structures different from ROS's message data structures and without incurring copying costs (for both runtime and development time).

For instance, this allows us to pub/sub PCL point clouds rather than our `sensor_msgs/PointCloud2`:

```c++
typedef pcl::PointCloud<pcl::PointXYZ> PointCloud;

void callback(const PointCloud::ConstPtr& msg)
{
  printf ("Cloud: width = %d, height = %d\n", msg->width, msg->height);
  BOOST_FOREACH (const pcl::PointXYZ& pt, msg->points)
    printf ("\t(%f, %f, %f)\n", pt.x, pt.y, pt.z);
}

int main(int argc, char** argv)
{
  ros::init(argc, argv, "sub_pcl");
  ros::NodeHandle nh;
  ros::Subscriber sub = nh.subscribe<PointCloud>("points2", 1, callback);
  ros::spin();
}
```

-- from: https://wiki.ros.org/pcl_ros#CA-2c92c01374035f603e08bcde0eb24382b1ccd02e_6

So, from your example, if you'd like to use `DDSImplType` rather than the `ROS2PointCloud`, you just need to tell us how to get to the `DDSImplType` type from the wire protocol.

articles/142_idl.md
```diff
@@ -144,3 +144,95 @@ For now only a one-dimensional, fixed-size array is supported though.
...
+| sequence<uint32, N>          |                                   |                | numpy.ndarray(dtype=numpy.uint32)               |
+| int64[N]                     |                                   |                | numpy.ndarray(shape=(N, ), dtype=numpy.int64)   |
+| sequence<int64>              |                                   |                | numpy.ndarray(dtype=numpy.int64)                |
+| sequence<int64, N>           |                                   |                | numpy.ndarray(dtype=numpy.int64)                |
+| uint64[N]                    |                                   |                | numpy.ndarray(shape=(N, ), dtype=numpy.uint64)  |
+| sequence<uint64>             |                                   |                | numpy.ndarray(dtype=numpy.uint64)               |
+| sequence<uint64, N>          |                                   |                | numpy.ndarray(dtype=numpy.uint64)               |
+| array of string              | struct {size\_t, T * }, size\_t N |                |                                                 |
```
dirk-thomas: "array" seems wrong here. The mapping suggests that it is a sequence (non static size).

Also the mapping doesn't seem to be different than the generic definition of arrays and sequences above? Should this line be removed?

sloretz: This line probably should be removed, I think it made sense when the mapping was  `rcutils_string_array`. `array of string` was trying to say `string[N]` from `T[N]` in the generic table, so the default type would be `char * [N]`.

dirk-thomas: :heavy_check_mark: removed in 828db257ea3672c95caefa3f7d0cc2e0d932bc01.

The same commit also consolidates the `numpy` mapping into a more readable "templated" form - while the cells get wider I hope it makes it more readable / comprehensible than the verbose version before.

dirk-thomas (approved):
The current state looks solid to me. I propose to get it merged as is.

dirk-thomas:
@mjcarroll Please review / approve.
@sloretz Please review / merge.

articles/143_legacy_interface_definition.md
```diff
@@ -241,3 +241,34 @@ Depending on the type the following values are valid:
...
+Interfaces described using the legacy format are first converted to [IDL](articles/idl_interface_definition.html).
+Code generation uses the generated file.
+
+### Mapping to IDL types
+
+| ROS type | IDL type           |
+| -------- | ------------------ |
+| bool     | boolean            |
+| byte     | octet              |
+| char     | uint8              |
```
mjcarroll, line 256: Do we want to add a note about IDL generators that don't support `int8` and `uint8` or are we assuming that will be correct everywhere soon?

articles/142_idl.md
mjcarroll: @dirk-thomas and @sloretz Did we ever resolve this?

sloretz: Possibly referring to this: https://github.com/ros2/design/pull/190#discussion_r217768639 where array of char and wchar should be structs because they can contain zero?

I don't remember a discussion about wstring or string being structs instead of pointers though.

articles/143_legacy_interface_definition.md
dirk-thomas: I wouldn't put it in the design doc since this is for the ROS-facing IDL. If a vendor doesn't support "something" we will likely take care of it by temporarily generating a "compatible" variation to feed it to their code generator. So I would call it an "implementation detail".

articles/142_idl.md
dirk-thomas: A `string` / `wstring` must not contain null characters. Therefore a raw pointer is sufficient in this case.

The opposite case is `char[]` / `wchar[]` which doesn't have that restriction and therefore is stored as `data ptr` / `size` pair.

mjcarroll (approved):

sloretz:
LGTM, merging

PR #191, opened 2018-09-14
add supported types from the IDL

dirk-thomas:
To help with #190 I enumerated the types from the IDL we are planning to support.

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -30,6 +30,8 @@ Further it describes how the interfaces are used to generate code in C, C++ and 
   <b>TODO:</b> All of this
 </div>
 
+ROS 2 supports a subset of the [OMG IDL 4.2 specification](https://www.omg.org/spec/IDL/4.2).
+
 ### Basic Types
 
 #### 7.4.1.4.4.2.1 Integer Types
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -36,14 +36,14 @@ ROS 2 supports a subset of the [OMG IDL 4.2 specification](https://www.omg.org/s
 
 #### 7.4.1.4.4.2.1 Integer Types
 
-| IDL type       | Value range        |
-| -------------- | ------------------ |
-| short          | -2^15 ... 2^15 - 1 |
-| unsigned short | 0 ... 2^16 - 1     |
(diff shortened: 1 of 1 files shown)


articles/142_idl.md
```diff
@@ -30,6 +30,114 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+### Basic Types
+
+#### 7.4.1.4.4.2.1 Integer Types
+
+| IDL type       | Value range        |
+| -------------- | ------------------ |
+| short          | -2^15 ... 2^15 - 1 |
+| unsigned short | 0 ... 2^16 - 1     |
+| int            | -2^31 ... 2^31 - 1 |
```
sloretz: In 4.2 it looks like the IDL type with this range is `long`. I don't see a type called `int`.

dirk-thomas: Correct, fixed in 4cdcc0f5879d921b0128c667ba794d94f66e9ad2.

articles/142_idl.md
```diff
@@ -30,6 +30,114 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
+
+A structure is a grouping of at least one member.
+
+#### 7.4.1.4.4.4.3 Enumerations
+
+An enumerated type consist of an ordered list of enumerators.
+
+#### 7.4.1.4.4.5 Arrays
+
+A multidimensional, fixed-size array is defined by the type of each item and the explicit sizes for each dimension.
```
dirk-thomas, line 139: Do we agree on limiting this (at least for now) to only a single dimension?

dirk-thomas: 351e3d9c1fd3d84af9f046c4ba09124eec482014

spiderkeys:
Out of curiosity, is this IDL 4.2 support going to bring with it support for keyed fields, even if ROS2 might not intend to use them?

dirk-thomas:
> is this IDL 4.2 support going to bring with it support for keyed fields, even if ROS2 might not intend to use them?

Yes, we will likely allow / parse / support keys in the `.idl` files. We don't plan to work on any key-specific functionality beyond this though at them moment (e.g. in the rmw layer).

sloretz (approved):

mjcarroll (approved):

PR #192, opened 2018-09-19
Mention there are exceptions to the msg naming rule, intended for internal use only

dhood:
`rosidl_parser` allows some exceptions (currently `_Request` and `_Response` suffixes, and `Sample_` prefixes (for opensplice specifically)). It's not trivial to prevent users from exploiting these exceptions, so instead document that it is not permitted.

### DIFF
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -137,6 +137,7 @@ Each file contains a single message or service.
 Message files use the extension `.msg`, service files use the extension `.srv`.
 
 Both file names must use an upper camel case name and only consist of alphanumeric characters.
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
As we are currently working on an alternative to remove these intermediate .msg files, I don't think we need to call it out in this design document.

@dirk-thomas thoughts?

dhood:
(I considered that but thought that would be in addition to the old process, not instead of)

dirk-thomas:
Yes, there won't be any request / response `.msg` files anymore in the future.

dhood:
OK, nw. Will close this then!

PR #201, opened 2018-11-19
uint64 maps to unsigned long long

sloretz:
Previously article mapped uint64 to `unsigned long` which is synonymous with `unsigned int`.

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -99,7 +99,7 @@ the DDS-XTypes specification 1.2 defines it with 16-bit.
 | int32          | int                 |
 | uint32         | unsigned int        |
 | int64          | long                |
-| uint64         | unsigned long       |
+| uint64         | unsigned long long  |
 
 ### Template Types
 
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

articles/142_idl.md
```diff
@@ -99,7 +99,7 @@ the DDS-XTypes specification 1.2 defines it with 16-bit.
 | int32          | int                 |
 | uint32         | unsigned int        |
 | int64          | long                |
```
mjcarroll: Should `int64` be correspondingly `long long`

dirk-thomas:
The `(u)int32` types should also be changed to `(unsigned) long`.

mjcarroll (changes requested):
Rescinding my previous approval ;)

mjcarroll (approved):

articles/142_idl.md
sloretz: 3b9f617

sloretz:
> The (u)int32 types should also be changed to (unsigned) long.

3b9f617

PR #206, opened 2019-01-18
Proposal for launching dynamically composable nodes

sloretz:
This is a rough draft proposing changes to the roslaunch design document for dynamically composable nodes.

### DIFF
--- a/articles/150_roslaunch.md
+++ b/articles/150_roslaunch.md
@@ -524,6 +524,13 @@ The services are hidden to avoid colliding with user created services.
 
 ####### load_node
 
+If a container process is asked to load a node with a full node name matching an existing node, then it must reject the request.
+This is to avoid conflicts in features that assume node name uniqueness, like parameters.
+
+A container process must assign the node a unique id when it is loaded.
+The id of a loaded node instance never changes.
+Two nodes in the same container process must never have the same id, and there should be a significant time delay before an id is reused.
+
 ```
 # A ROS package the composable node can be found in
 string package_name
@@ -551,13 +558,15 @@ bool success
 string error_messsage
 # Name of the loaded composable node (including namespace)
 string full_node_name
+# A unique identifier for the loaded node
+uint64 unique_id
 ```
 
 ####### unload_node
 
 ```
-# Full node name including namespace
-string full_node_name
+# Container specific unique id of a loaded node
+uint64 unique_id
 ---
 # True if the node existed and was unloaded
 bool success
@@ -571,6 +580,8 @@ string error_messsage
 ---
 # List of full node names including namespace
 string[] full_node_names
+# corresponding unique ids (must have same length as full_node_names)
+uint64[] unique_ids
 ```
 
 ###### Exit Code
@@ -590,10 +601,6 @@ If they are of different types then the launch system may choose to try to load 
 How Composable nodes are registered is not defined by this document.
 Instead, the container process is responsible for knowing how to find nodes it is asked to load.
 For example, a container process might use pluginlib for rclcpp nodes, or python entry points for rclpy nodes.
-
-##### Duplicate Node Handling
-
-If a container process is asked to load a node with a full node name matching an existing node, then it must reject the request.
 
 ## Event Subsystem
 
--- a/articles/150_roslaunch.md
+++ b/articles/150_roslaunch.md
@@ -547,6 +547,8 @@ rcl_interfaces/Parameter[] extra_arguments
 ---
 # True if the node was successfully loaded
 bool success
+# Human readable error message if success is false, else empty string
+string error_messsage
 # Name of the loaded composable node (including namespace)
 string full_node_name
 ```
@@ -559,6 +561,8 @@ string full_node_name
 ---
 # True if the node existed and was unloaded
 bool success
+# Human readable error message if success is false, else empty string
+string error_messsage
 ```
 
 ####### list_nodes
--- a/articles/150_roslaunch.md
+++ b/articles/150_roslaunch.md
@@ -585,7 +585,7 @@ uint64[] unique_ids
 ```
 
 ###### Exit Code
-If the container proces is asked to shutdown due to normal [Termination], then the exit code must be 0.
+If the container process is asked to shutdown due to normal [Termination], then the exit code must be 0.
 If it exits due to an error then exit code must be any other number.
 
 ##### Parallel vs Sequential Loading of Nodes
--- a/articles/150_roslaunch.md
+++ b/articles/150_roslaunch.md
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
 However, as an example of a process with multiple nodes, consider a program that instantiates two camera driver nodes called "camera1" and "camera2" by default.
 You could configure their namespaces separately by doing something like `camera1:__ns:=left camera2:__ns:=right`.
 
-### Dynamically loaded Nodes
+#### Dynamically loading Nodes
 
-A special case of processes with multiple nodes are nodes that do not statically instantiate nodes.
-Two separate ways to do this are described below, but in both cases the typical command line arguments (as described in the remapping design document[^static_remapping]) and environment variables may not work as expected.
-In stead the method for causing the dynamic instantiation of a node needs to communicate these settings separately.
-And because of this, the launch system also needs to know how to do this.
+Dynamically loading a node means spawning it in a container process that does not know about the node until it is asked to load it.
+A container process is a stand alone executable that loads and executes nodes within itself.
 
-#### By Configuration
+##### Container Process API
 
-The first way would be to pass some user define command line arguments to the process which the process would interpret and use to dynamically instantiate some ROS nodes.
-There's no current plan to have a generic tool for this, but that's subject to change.
-However, since there's no standard tool, the launch system cannot know how to convert ROS specific declarations into a format that can be used to invoke it.
+While there will be standard container processes, custom container processes would allow using custom executors or client libraries.
+Therefore, there must be a container process API for the launch system to communicate which nodes should be loaded.
 
-Though the user could always generate the appropriate command line arguments for the process and have the launch system treat it as a normal operating system process.
-In that case, though, it might be useful to indicate to the launch system that a process may contain nodes, but should be otherwise treated like a plain operating system process.
-This would let the launch system still monitor it for ROS specific events and input/output.
+The launch system must be able tell the container process what arguments to give to a dynamically loaded node.
+This includes command line arguments and client library specific options (e.g. `rclcpp` has `use_intra_process_comms`).
+Since the launch system cannot know about all custom containers, the API must include a way to pass unknown arguments (e.g. by passing key-value pairs).
 
-As an example of a "by configuration" dynamic node process, you could have a process that takes a variable number of arguments, where each is a serial port address for a laser scanner, e.g. `/dev/ACM0 /dev/ACM1 ...`, and it would instantiate a laser scanner driver node for each address.
-It would would up to this process to figure out how to allow the user to further configure the nodes though.
-It would not be possible, in this case, for the launch system to "apply a ROS namespace" to the nodes within it, for example.
-For this reason, it's probably not an ideal way to structure nodes until a standardized way to load many nodes into a single process via configuration is defined.
+The API will not include setting environment variables per loaded node.
+Many languages have APIs to get environment variables, and there is no way to isolate them within a process.
 
-<div class="alert alert-warning" markdown="1">
-RFC:
+The following options for an API are being considered.
 
-On the point of a standardized way of having a "dynamic nodes process by configuration", I think it might be worth pursuing this, because it will basically become a more efficient way of launch a bunch of nodes into a single process, more efficient than the "by proxy" way described below.
+###### API using Command Line Configuration File
+One option for a container processes API is to pass a configuration file with nodes to load via the command line.
 
-The "by proxy" way below is essentially the pattern used by nodelets in ROS 1, but what I'm describing here is essentially like a NodeletManager executable which rather than loading nodelets via Service calls would load and run them based on command line arguments or a config file or something.
+Advantages
+* No waiting for an API to become available
 
-The "by proxy" way is more convenient when you're running things by hand, but if the roslaunch description looks something like this (just pseudo code):
+Disadvatages
+* Requires write access to the file system
+* Requires parsing a config file
+* Cannot tell from the outside if a container process supports this interface
+* Cannot tell if and when nodes are loaded or unloaded
 
-```xml
-<node_container_process name="my_container_process">
-  <node package="my_package" executable="talker" />
-  <node package="my_package" executable="listener" />
+This API could have very low latency to launch nodes since it does not require waiting for discovery.
+However, there is no way to get feedback about the success or failure of loaded nodes.
+There is also no way to tell a container process to unload a composable node.
 
-</node_container_process>
+###### API using STDIN
+
+Another option for a container process API is to pass configuration in via STDIN.
+
+Advantages
+* No waiting for an API to become available
+* Works with read-only file systems
+
+Disadvantages
+* Requires parsing a config
+* Cannot tell from the outside if a container process supports this interface
+* Cannot tell if and when nodes are loaded or unloaded
+* Cannot stop dynamically loaded nodes from reading STDIN
+
+This API could also have very low latency to launch nodes.
+However, there also is no way to get feedback about the success or failure of loaded nodes.
+STDOUT cannot be used because a composable node logging messages to STDOUT is assumed to be very common and would conflict.
+Since STDIN is always available, it would be possible to unload a node via this API.
+
+###### API using ROS Services or Topics
+
+Lastly, a container process API may be defined by ROS services or topics.
+
+Advantages
+* No config file parsing
+* Works with read-only file systems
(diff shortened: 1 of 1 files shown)


articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+```
+# A ROS package the composable node can be found in
+string package_name
+# a plugin within that package
+string plugin_name
+
+# Name the composable node should use
+string node_name
+# Namespace the composable node should use
+string namespace
```
dirk-thomas: Is there a way to not set the `node_name` and `namespace` and use the default embedded in the source of the component?

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+# Values from message rcl_interfaces/Log
+uint8 log_level
+# Remap rules
+# TODO(sloretz) rcl_interfaces message for remap rules?
+string[] remap_rules
+# Parameters to set
+rcl_interfaces/Parameter[] parameters
+
+# key/value arguments that are specific to a type of container process
+rcl_interfaces/Parameter extra_arguments
```
dirk-thomas: Should this allow more than one?

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+---
+# True if the node was successfully loaded
+bool success
+# Name of the loaded composable node (including namespace)
+string full_node_name
```
dirk-thomas: When is this value different from the concatenation of the two request fields `namespace` and `node_name`? If never, could it just be removed?

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+# TODO(sloretz) rcl_interfaces message for remap rules?
+string[] remap_rules
+# Parameters to set
+rcl_interfaces/Parameter[] parameters
+
+# key/value arguments that are specific to a type of container process
+rcl_interfaces/Parameter extra_arguments
+---
+# True if the node was successfully loaded
+bool success
```
dirk-thomas: Beside a boolean flag an error message why the loading didn't succeed would be helpful.

articles/150_roslaunch.md
dirk-thomas: Since multiple nodes can have the same `full_node_name` maybe returning a unique id here would be better and would allow to precisely specify which node to unload if the full name is not unique.

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+
+####### unload_node
 
-</node_container_process>
+```
+# Full node name including namespace
+string full_node_name
+---
+# True if the node existed and was unloaded
+bool success
```
dirk-thomas: Related to the above comments:
* identify node by unique id rather than full name
* allow returning error message

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+```
+
+####### list_nodes
+
+```
+---
+# List of full node names including namespace
+string[] full_node_names
```
dirk-thomas: Related to the above comments: return a mapping of unique ids to their full node name

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+Instead, the container process is responsible for knowing how to find nodes it is asked to load.
+For example, a container process might use pluginlib for rclcpp nodes, or python entry points for rclpy nodes.
 
-One form this can take is by providing a "proxy" process which doesn't actually run the node, but instead loads it into an already running container and just stays running until the node is shutdown in the remote container.
-This pattern is used by the "nodelets" in ROS 1.
-It's a useful pattern when you're running nodes by hand but you want them to share a process and don't want to write your own program to do that.
+##### Duplicate Node Handling
 
-There are a few other forms this can take, but the common thread between them is that a process instantiates nodes dynamically based on asynchronous input from external actors (proxy's), and that the configuration for those nodes is communicated through something other than command line arguments and environment variables.
+If a container process is asked to load a node with a full node name matching an existing node, then it must reject the request.
```
dirk-thomas: All comments above about using unique ids rather than full names are of course void if we want to forbid duplicate names. What is the main reason to not allow them? The extra effort of using unique ids seems rather small.

sloretz: > What is the main reason to not allow them? The extra effort of using unique ids seems rather small.

I agree that the effort is small. The enforcement of unique fully qualified node names is for other systems like parameters that create services based on the namespace + node name.

I'll add the unique ID. Even if fully qualified node names are unique, if dynamic remapping allows changing a node's name in the future then a unique id may make it easier to track and unload nodes that change their name.

articles/150_roslaunch.md
wjwwood: If you make a way to optionally set the name and/or namespace, then the full name would be useful to know at this point (because depends on what is in the code).

There are already a few places where we assume unique node names (set/get parameter services for instance), so introducing a unique identifier here may not be the best thing, especially if we have to do a lot of work to generate one (call a UUID lib or something). If we just use the pointer to the node instance, then maybe that's ok, but the real solution is to address the node uniqueness design issue.

A potential compromise is that we continue to assume the node name is unique, but the container fails this service call if the new node's name is already present in the container.

articles/150_roslaunch.md
```diff
@@ -428,62 +428,168 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+string[] full_node_names
 ```
 
-Whether or not the above results in three processes (one container and two proxy processes) or a single process, probably doesn't matter to the user or at least would be fine with the user in exchange for less overhead.
+###### Exit Code
+If the container proces is asked to shutdown due to normal [Termination], then the exit code must be 0.
+If it exits due to an error then exit code must be any other number.
 
-It would be more efficient because you don't need to have a node in each proxy to actually make the Service call to the load that node the user wants to run, and you don't need to maintain a bond between the container and the proxy processes.
+##### Parallel vs Sequential Loading of Node
```
wjwwood:
```suggestion
##### Parallel vs Sequential Loading of Nodes
```

articles/150_roslaunch.md
sloretz: Multiple arguments in c9a4d0386cbe8b7cae6e2efb38b48ca8a1825634

articles/150_roslaunch.md
sloretz: Set to empty string to use default embedded name in 77461c1069036d727be7a3484314c50f3a6f09c7

articles/150_roslaunch.md
sloretz: Human readable error message 77461c1069036d727be7a3484314c50f3a6f09c7

articles/150_roslaunch.md
wjwwood: I'd say (as I did above before seeing this) that we already assume unique node names in other parts of the system, and depending on how we generate the unique id's (are they unique to the container or within the whole system?) they may or may not be simple to generate, so we should avoid duplicate names. What is the main reason to allow them?

wjwwood: > I'll add the unique ID. Even if fully qualified node names are unique, if dynamic remapping allows changing a node's name in the future then a unique id may make it easier to track and unload nodes that change their name.

👍 for the rationale of having the unique id for future dynamic node name changing (but again needs to be clear if it's unique in the container or within the system), and also do you still intend to prevent duplicate node names?

Personally I don't see a reason to allow duplicate names.

sloretz: 90480aeeabb904d2693b3f6e0c4e3ae72567ff88 adds a unique id (unique within the container process, not system) and continues to forbid duplicate node names in the same container process.

articles/150_roslaunch.md
sloretz: Return a mapping of unique ids to full node name in 90480aeeabb904d2693b3f6e0c4e3ae72567ff88

articles/150_roslaunch.md
sloretz: Unique ID in 90480aeeabb904d2693b3f6e0c4e3ae72567ff88 and error message in 
5bd6a5c

articles/150_roslaunch.md
sloretz: Kept full node name, full name uniqueness within a container, and added unique id in 90480aeeabb904d2693b3f6e0c4e3ae72567ff88

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
-On the point of a standardized way of having a "dynamic nodes process by configuration", I think it might be worth pursuing this, because it will basically become a more efficient way of launch a bunch of nodes into a single process, more efficient than the "by proxy" way described below.
+###### API using Command Line Configuration File
+One option for a container processes API is to pass a configuration file with nodes to load via the command line.
 
-The "by proxy" way below is essentially the pattern used by nodelets in ROS 1, but what I'm describing here is essentially like a NodeletManager executable which rather than loading nodelets via Service calls would load and run them based on command line arguments or a config file or something.
+Advantages
+* No waiting for an API to become available
 
-The "by proxy" way is more convenient when you're running things by hand, but if the roslaunch description looks something like this (just pseudo code):
+Disadvatages
```
hidmic:
```suggestion
Disadvantages
```

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+# List of full node names including namespace
+string[] full_node_names
+# corresponding unique ids (must have same length as full_node_names)
+uint64[] unique_ids
+```
 
-So I'm interested in what others think about coming up with a standardized executable which can load and configure many nodes at once from command line arguments or a config file.
-</div>
+###### Exit Code
+If the container proces is asked to shutdown due to normal [Termination], then the exit code must be 0.
```
hidmic:
```suggestion
If the container process is asked to shutdown due to normal [Termination], then the exit code must be 0.
```

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
-#### By Proxy
+##### Parallel vs Sequential Loading of Nodes
 
-The other way to implement a process which dynamically instantiates ROS nodes, is to send the request to instantiate a node asynchronously from another process.
-So the "container" process would start up and provide a service (ROS Service or otherwise) that lets external processes request a node be instantiated with a given set of configurations.
+If it is possible to load multiple nodes in parallel, then it needs to be decided how to load the nodes.
+The container process should load nodes as soon as it is asked.
+It should be up to the launch system to decide whether to load nodes in parallel or sequentially.
+If multiple nodes of the same type are to be launched, then the launch system should load the nodes sequentially so each is able to remap it's name before the next is loaded.
+If they are of different types then the launch system may choose to try to load them in parallel, where the exact order they get loaded is determined by chance or the container process.
```
hidmic, line 597: Hmm I understood the ROS2 launch system was aiming to be deterministic in its execution. @wjwwood ?

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
-</node_container_process>
+* `_launch/load_node`
+* `_launch/unload_node`
+* `_launch/list_nodes`
+
+The services are hidden to avoid colliding with user created services.
+`load_node` will be called by the launch system when a composable node is to be dynamically loaded, and `unload_node` destroys a composable node.
+`list_nodes` is not called by launch system, and is only provided for introspection.
+
+####### load_node
```
hidmic: [There are only 6 header levels](https://guides.github.com/features/mastering-markdown/#syntax). Consider using a bold font instead, here and below (or revisiting the document layout).

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+The container process must offer all of the following services.
 
-</node_container_process>
+* `_launch/load_node`
+* `_launch/unload_node`
+* `_launch/list_nodes`
+
+The services are hidden to avoid colliding with user created services.
+`load_node` will be called by the launch system when a composable node is to be dynamically loaded, and `unload_node` destroys a composable node.
+`list_nodes` is not called by launch system, and is only provided for introspection.
```
hidmic, line 523: Meta: why is this API launch specific? It looks general enough to be that of any composable nodes container.

articles/150_roslaunch.md
mjcarroll: I think that the key phrase here is "may choose to".  If you want to continue to have deterministic behavior, it probably makes the most sense to load them in sequential order (perhaps as defined by the launch file?)

I could still see a case for loading in parallel where rigorous sequential guarantees aren't needed.  It should speed up launching a large set of composed nodes.

articles/150_roslaunch.md
mjcarroll: I never thought of it as a launch-specific API, but I think that's the context that we are currently designing it for, so there's probably a bias towards launch here.

I don't see anything here that would stop people from writing additional node containers that conform to this API, or additional tooling that could spawn nodes in composable containers.

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+string package_name
+# a plugin within that package
+string plugin_name
+
+# Name the composable node should use, or empty to use the node's default name
+string node_name
+# Namespace the composable node should use, or empty to use the node's default namespace
+string namespace
+# Values from message rcl_interfaces/Log
+uint8 log_level
```
jacobperron: Another parameter to a launched node to consider is "log output", selecting where logs are sent: screen, file, or both.
Note, this feature is still a work-in-progress: ros2/launch#104

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+
+```
+# A ROS package the composable node can be found in
+string package_name
+# a plugin within that package
+string plugin_name
+
+# Name the composable node should use, or empty to use the node's default name
+string node_name
+# Namespace the composable node should use, or empty to use the node's default namespace
```
dirk-thomas: What if the node has a default namespace but the caller want's to instantiate it without a namespace?

sloretz: When you say instantiate without a namespace, do you mean instantiate with the namespace `/`?

dirk-thomas: Yeah, I wasn't sure if this would be absolute or relative. Passing `/` would indeed work.

articles/150_roslaunch.md
sloretz: How will launch get logs from a node? A topic interface like `/rosout`? If only `STDOUT` then the "log output" setting that matters is that of the container.

jacobperron: The plan was to redirect `stdout` (and possibly `stderr`) per process using the `OnProcessIO` event.

sloretz: IIUC there's only one STDOUT, so a composed node will share it with all the other composable nodes in the container process. What should be passed to an individual composable node?

articles/150_roslaunch.md
sloretz: Made these part of a list in e21211be0d6484759d036c2069f9434a801dabfb

articles/150_roslaunch.md
jacobperron: Right, I misunderstood the context of this service definition. Unless an alternative implementation (like using the `/rosout` topic interface) is used, we may consider limiting the log destination configuration to the container level. I don't feel strongly about adding "log output" to this API.

articles/150_roslaunch.md
sloretz: I think someone will write a custom container process (i.e. for a custom client library or executor). I made the wording about container processes a little more generic in aa18ecaffc9e7685a22ab9e172c5fbcdbc59d8a0.

articles/150_roslaunch.md
```diff
@@ -428,62 +428,179 @@ The remapping design document[^static_remapping] goes into detail on how you can
...
+    string plugin_name
+
+    # Name the composable node should use, or empty to use the node's default name
+    string node_name
+    # Namespace the composable node should use, or empty to use the node's default namespace
+    string node_namespace
+    # Values from message rcl_interfaces/Log
+    uint8 log_level
+    # Remap rules
+    # TODO(sloretz) rcl_interfaces message for remap rules?
```
hidmic, line 547: @sloretz that'd be nice indeed :)

hidmic (commented):
Gave it another pass, I think it's ready to go.

articles/150_roslaunch.md
sloretz: Changed `_launch` to `~/_containter` in d98bb98b67d3bd5df9af1bb0ec5f2a741e20b80c

mjcarroll (approved):
I believe that this is good to go.

wjwwood (approved):

PR #207, opened 2019-01-22
Adds ROS 2 launch XML format design document

hidmic:
This pull request is the v0.1.0 draft of an XML format for the ROS 2 launch system.

~Currently just mostly keeping portability with ROS 1.~ It includes all relevant features for the first milestone on ROS 2 launch front-ends.


articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            A predicate to condition ROS node launch i.e. do so unless
+            the predicate evaluates to a truthy value.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:attribute>
+  </xs:element>
+</xs:schema> 
+```
```
wjwwood, line 43: I don't know if it's valuable to put the schema, here. I expect it to change over time, and I'm not sure we want to keep this document up-to-date along with the actual schema file and/or documentation and implementation of the XML parsing.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Tags Semantics
+
+#### All Tags
+
+Every tag's execution can be conditioned on a boolean predicate via `if` or `unless` attributes.
+
+#### <launch> Tag
+
+Root tag of any launch file.
```
wjwwood, line 53: Should more than one of these be allowed at a time?

articles/151_roslaunch_xml.md
wjwwood, line 43: I'm not sure that it is good to have the full schema here, just because it will need to be kept up-to-date with an actual schema file at some point. That's not to say that the schema isn't useful, just that here might not be the best place for it.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+
+Authors: {{ page.author }}
+
+
+# ROS 2 Launch XML Format v0.1.0
+
+## Rationale
+
+As an alternative to a programmatic approach to the ROS 2 launch system's API, a declarative description features a WYSIWYG approach, easier to read, audit and maintain.
+This is the preferred approach for ROS 1 `roslaunch` launch files, thus some degree of familiarity is expected (and relied upon).
```
wjwwood, line 28: There's a section in the (still open) pr for launch's design document for this topic. It might require less context if this were part of that document.

Additionally, I'd like to see this in two parts:

- a way for any markup language to map to the Python API in launch
  - and specifically how that could be done for XML, e.g. something explicit but not necessarily convenient like `<launch_action pkg="launch_ros" class="Node"><launch_action_arg type="int">42</launch_action_arg>...</launch_action>`
- syntactic sugar that builds on the aforementioned explicit matching, e.g. `<node ...>` rather than `<launch_action pkg="launch_ros" class="Node"...` etc

I have enough thoughts about this and how it might be implemented to probably warrant a meeting to discuss it. At the very least we should discuss to see if it's worth taking this approach to the implementation versus a more direct implementation which doesn't worry about an automatic mapping of any type.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+The <include> tag allows for bringing a launch file description into another, enabling re-use of hierarchical launch layouts.
+The included launch file description has its own scope for launch configurations.
+The included launch file description is not necessarily written in this format nor a declarative one (i.e. a programmatic description).
+To avoid ROS name clashes, included nodes can be namespaced via the `ns` attribute.
+
+##### Examples
+
+```xml
+<include file="/opt/my_launch_file.py" ns="/my_ns"/>
+<include file="/opt/my_other_launch_file.xml">
```
wjwwood, line 69: Maybe you're trying to avoid introducing the substitution syntax, but otherwise this example exhibits some bad practice, as referring to absolute files is usually a bad idea.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+#### <group> Tag
+
+##### Description
+
+The <group> tag allows for launch actions' grouping as well as optional launch file configuration scoping.
+
+##### Examples
+
+```xml
+<group ns="/dummy_group" scoped="true">
```
wjwwood: ns can be relative here, btw and in most cases should be so that this group could be namespaced by as yet unknown launch files above this one.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+##### Description
+
+The <arg> tag allows for launch file configuration via the command-line or when including it via an <include> tag.
+Arguments are limited to the scope of their definition and thus, have to be explictly passed to included files if any.
+
+##### Examples
+
+```xml
+<arg name="publish_frequency" default="10"/>
+<arg name="output_path" doc="Output path for some processing pipeline"/>
```
wjwwood: I think `description` is the term used in launch, rather than `doc`.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+
+##### Description
+
+The <param> tag allows for setting a ROS parameter of a ROS node.
+
+##### Examples
+
+```xml
+<param name="some_numeric_param" value="100.2"/>
+<param name="some_list_param" value="Some phrase,100.0,true" sep=","/>
```
wjwwood: It does need to be inside of a `<node>` tag, which is maybe implied, but it would be good to have in the example as well. In ROS 1, `<param>` could be outside of a node tag and in that case would set global parameters. We might have something like that for ROS 2 in the future, but right now it would not do anything.

wjwwood: Same comment for `<params>` below.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+#### <remap> Tag
+
+##### Description
+
+The <remap> tag allows for ROS name remapping of a <node> instance.
+
+##### Examples
+
+```xml
+<remap from="chatter" to="my_chatter"/>
```
wjwwood: Again it would be useful to show that this must be in a node tag, even if state otherwise.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+: Substituted by the value of the given environment variable
+  Substitution will fail if the variable is not set, unless default values are provided.
+
+`$(dirname)`
+: Substituted by the current launch file directory name.
+  Substitution will always succeed.
+
+#### User-defined Substitutions
+
+TBD
```
wjwwood, line 255: This may be covered by the launch configuration concept, which you can set with an action and get with a substitution:

https://github.com/ros2/launch/blob/master/launch/launch/actions/set_launch_configuration.py
https://github.com/ros2/launch/blob/master/launch/launch/substitutions/launch_configuration.py

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,744 @@
...
+  Forward and backwards slashes will be resolved to the local filesystem convention.
+  Substitution will fail if the package cannot be found.
+
+`$(find-exec exec-name)`
+: Substituted by the path to the executable in the local filesystem.
+  Lookups make use of the PATH environment variable.
+  Forward and backwards slashes will be resolved to the local filesystem convention.
+  Substitution will fail if the executable cannot be found.
+
+`$(arg name)`
```
wjwwood: These are specifically dropped into a launch configuration, so maybe the substitution for that should be used instead, like `$(conf name)`:

https://github.com/ros2/launch/blob/03d0a0a51193cd6a8e5b80a8f022a3cd2e0f4a90/launch/launch/actions/declare_launch_argument.py#L36-L38

wjwwood: Seeing used in context, renaming this concept to be `launch variables` and using `$(var name)` instead might be better, but that's mostly aesthetic.

articles/151_roslaunch_xml.md
hidmic: Hmm, I guess there could be, but what would be the use case?

articles/151_roslaunch_xml.md
mjcarroll: I agree with this, just for the potential burden of syncing.

articles/151_roslaunch_xml.md
mjcarroll: I'm not sure what would happen in the case that you `included` a file with multiple launch tags.

articles/151_roslaunch_xml.md
hidmic: I fully understand your concerns, and I expect this to change over time as well. But being the up-to-date description of the format is IMHO the sole purpose of this document in the long term.

As we add markup for more actions, substitutions, etc., I think we have to consider the possibility that not all ROS 2 launch implementations will support all of them (no interest, no resources, too difficult to implement, etc.) Also, if we enable users to develop and integrate their own, I think we have to define what's standard and what's not. 

In that context, this becomes an statement: for a ROS 2 launch implementation to support XML launch v0.1.0 it has to implement enough functionality for a description compliant with v0.1.0 to work. Anything outside of it is an extension.

hidmic: So while I agree in principle with not having to maintain multiple copies of it, the format document, or format specification to be precise, **is** the schema. Meaning, I don't see much value on a having format specification without a description of that format. Moreover, the rest of the document (e.g. the examples) revolves around that schema, so you'd have to keep them up to date anyways.

We can discuss where this whole document should live though. Maybe this isn't the best place? What do you think?

articles/151_roslaunch_xml.md
hidmic: The idea was to have a very brief intro, followed by a link to the actual architecture design. That made sense because, in my mind, one thing is the launch's architecture design document and another thing is a format specification for one of its front-ends. I wouldn't collapse them all into a single page. 

In that context, I think that implementation hints for XML parsing should go here, while implementation hints for a generalized approach to mapping declarative descriptions to programmatic ones should go in launch's design document. 

We can meet to talk about the latter, though I do see value on having this format spec regardless of how we end up implementing the mapping (see replies above).

articles/151_roslaunch_xml.md
hidmic: Yes! I've added one more case using a substitution to avoid any confusion.

articles/151_roslaunch_xml.md
hidmic: True that. Changed the example to advocate for best practices.

articles/151_roslaunch_xml.md
hidmic: That's correct. Changed.

articles/151_roslaunch_xml.md
hidmic: Makes sense. Enclosed them in `<node>` tags.

articles/151_roslaunch_xml.md
hidmic: Same as above.

articles/151_roslaunch_xml.md
hidmic: I was thinking of user defined extensions, as in `Substitution` subclasses. 

But still, yours is a valid point. I think we can add a `<let var="" value=""/>` tag for that action and use it with `$(var name)`, just like arguments. PTAL.

articles/151_roslaunch_xml.md
hidmic: That makes a lot of sense. Changed.

articles/151_roslaunch_xml.md
wjwwood: For the record, I don't think more than one should be allowed, but I just wanted to point out that this document doesn't say one way or the other.

articles/151_roslaunch_xml.md
wjwwood: > I fully understand your concerns, and I expect this to change over time as well. But being the up-to-date description of the format is IMHO the sole purpose of this document in the long term.

Well this is perhaps a policy discussion that's out of scope here, but I'm not sure that this should be the responsibility of this document. Basically it comes down to "is this a snapshot of our design intentions used to communicate them before/during development" or is it "a living document that states the standard for all launch implementations"?

I think it's too early for the latter (as the first version hasn't even been made or used yet), but I don't have a strong opinion about what should happen I guess, other than I see this needs lots of revisions over time.

wjwwood (approved):

articles/151_roslaunch_xml.md
hidmic: Good point. Schema [is implicitly stating so](https://www.liquid-technologies.com/xml-schema-tutorial/xsd-elements-attributes#cardinality) though. I just made it explicit.

articles/151_roslaunch_xml.md
hidmic: I hear you. And while I do think this gets closer to the latter, IMHO it'll also steer design e.g. we're committing to a format that a human can write. 

For the sake of cutting revision overhead, we could keep the PR open as a draft, maybe bulk update once we've polished an implementation and only then publish it. Launch architecture design document hasn't been published yet either, so I think it's fair to wait. Thoughts?

ruffsl: I'm in favor of having and maintaining a regersious XML schema definition that enforces the launch specification. Where such a `.xst` document should reside is perhaps the notable question. Having the `.xst` schema file in close proximity/revision control with design description could help ensure they stay in sync, but how does ROS1 related REPs host/track the schema definitions for ros1 launch.xml and package.xml files?

dirk-thomas: The package manifest formats do come with a `.xsd` schema file which is referenced in the REP (e.g. [REP 149](http://www.ros.org/reps/rep-0149.html#schema)), stored in the `rep` repo, and hosted on http://download.ros.org/schema/ to be referenced by `package.xml` files.

I don't think an official schema for ROS 1 launch file exists at the moment.

mjcarroll (approved):

articles/151_roslaunch_xml.md
ruffsl: Perhaps having the schema definition in it own `.xsd` (sorry, I meant this extension originally) would be better than inlining it in this markdown document, so commit history for the Schema Definition itself would be a bit cleaner.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,803 @@
...
+## Static Description
+
+### Schema Definition
+
+```xml
+<?xml version="1.0"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:annotation>
+    <xs:documentation xml:lang="en">
+      ROS 2 launch XML schema v0.1.0
```
ruffsl: I think we may want to require a "version" attribute in the root element of the DOM so that it is easier for launcher applications or a version adapter service to figure out what it is processing. E.g:

``` xml
<launch version="0.1.0">
...
```

https://stackoverflow.com/questions/2014237/what-are-the-best-practices-for-versioning-xml-schemas

hidmic: That makes a lot of sense. Added `version` to `<launch>` tag.

articles/151_roslaunch_xml.md
hidmic: Alright, I moved the schema to its own `.xsd` file and I managed to get Jekyll to include it verbatim in the format specification document. 

@wjwwood @dirk-thomas what do you think of this addition to the site layout I'm making here?

wjwwood: I think this will help avoid needing to keep multiple files in sync, so I'm a fan of it. I don't know how we keep the copy on downloads.ros.org in sync with the one in the rep repostiory, but we'll have to do something similar here, unless we host directly from design.ros2.org, which isn't a great idea I think.

dirk-thomas: The `xsd` file itself should be versioned. So the URL used by launch files to validate is "fixed" and the `xsd` file referenced by it does never change. Therefore it is a one time upload to downloads.tos.org - no need to keep anything in sync since the `xsd` file isn't supposed to change. Only a newer version should be uploaded with a new name / URL.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,234 @@
...
+##### Description
+
+The `<remap>` tag allows for ROS name remapping of a `<node>` instance.
+
+##### Examples
+
+```xml
+<node package="my_pkg" executable="my_node">
+  <remap from="chatter" to="/my_chatter"/>
+  <remap from="*/stuff" to="private_\1/stuff"/>
```
ruffsl, line 166: The globing scope of this is presumably all ros topics sub'ed or pub'ed by `my_node` that end with `/stuff` sould be remapped to a single topic?

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,234 @@
...
+
+Every tag's execution can be conditioned on a boolean predicate via `if` or `unless` attributes.
+All action tags that can contain other action tags can scope ROS entities in a namespace via the `ns` attribute.
+
+#### `<launch>` Tag
+
+Root tag of any launch file.
+There must only be one `<launch>` tag on a given launch file.
+
+#### `<include>` Tag
```
ruffsl, line 56: Can the include element be just a macro around xinclude (that pre-expands the Built-in Substitutions), or does it need to do more than that?

articles/specs/launch.0.1.0.xsd
```diff
@@ -0,0 +1,582 @@
...
+      </xs:annotation>
+    </xs:attribute>
+      <xs:attribute name="args" type="xs:string" use="optional">
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            Additional 'command-line' arguments for the ROS node.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:attribute>
+      <xs:attribute name="ns" type="xs:string" use="optional">
```
ruffsl, line 556: Would it be worthwhile to prevent the use of trailing slashes and perhaps enforce naming constraints? E.g:

``` xml
<xs:simpleType>
  <xs:restriction base="xs:string">
    <xs:pattern value="((\/)?([a-zA-Z_])+([a-zA-Z0-9_])*)+"/>
  </xs:restriction>
</xs:simpleType>
```

https://design.ros2.org/articles/topic_and_service_names.html#ros-2-topic-and-service-name-constraints

Although, I suppose this expression alone doesn't account/guard for the substitution syntax.

hidmic: > Although, I suppose this expression alone doesn't account/guard for the substitution syntax.

Right, that's why I didn't even try to get too specific with the pattern. Same goes everywhere else. Do you think we should?

articles/151_roslaunch_xml.md
hidmic: So, as immediate needs, the `<include>` tag has to be able to do ROS namespacing and take arguments. That I'd think it's achievable by just injecting actions before and after included content. 

But on top of that, when I wrote the draft I thought that allowing inclusion of launch files written in different markups or even programmatically would provide a significant amount of flexibility (e.g. an XML launch including a Python launch).

articles/151_roslaunch_xml.md
hidmic: Following the [name remapping design document](http://design.ros2.org/articles/static_remapping.html), above's rule means that all topics sub'ed and/or pub'ed by `my_node` whose name matches `([^/]+)/stuff` would be mapped to `private_\1/stuff` where `\1` is the first capturing group in the matching pattern.

articles/specs/launch.0.1.0.xsd
ruffsl: Perhaps we could could use an XML Template to pre-process advanced launch file markup, e.g. expand the substitutions, so that the final expanded XML is a smaller subset of launch file primitives. We could use two schemas, the first would be loose, ensuring the integrity of the input structure as anticipated by the template transform. The second schema (overloaded from the first schema) would validate the output from the transform to additionally ensure the integrity of the input values.

Perhaps this isn't be the best way about this, but I'd like to see the launch syntax remain implementation agnostic, so that it's easier to incorporate into other system orchestration tools. Perhap something like [Xproc](http://xproc.org) could be used to formulate the pipeline.

articles/151_roslaunch_xml.md
ruffsl: > That I'd think it's achievable by just injecting actions before and after included content.

What are `actions` in this context, is this an XML spec?

articles/151_roslaunch_xml.md
ruffsl: > Following the name remapping design document ...

Wow, guess it's been awhile since I re-read that document. Would there be any chance to be to initiate a dry run to compute the exact remapping offline, assuming one had a prior-model of the nodes they are initiating? I'd just still like to have a mean of statically analyzing resource and interface usage from launch files for security tooling. I feel like launch files could be more tractable complement for formal verification, rather than having to moc stuff like lifecycles to faithfully inference/execute python launch files.

articles/151_roslaunch_xml.md
hidmic: Sorry, my bad for using launch terminology out of place. [Here](https://github.com/ros2/design/pull/163/files#diff-df7db447d3b09c67c5487a923f56a445R586) is the formal definition, but long story short actions are the way one describes the act of doing something in ROS2 launch and ultimately what makes up most of launch descriptions. In this context, `<include>` maps to an action, `<group>` to another and so on.

articles/specs/launch.0.1.0.xsd
hidmic: Hmm, substitutions are a core part of the launch system. As in, substitutions have entity beyond the markup realm, be it XML, YAML, TOML, etc., and thus there isn't such a thing as an expanded XML. I guess you could do so though, not sure if we have (or plan to have) substitutions whose value is not fixed at configuration time. @wjwwood I fall back to you with the question.

articles/151_roslaunch_xml.md
hidmic: That's a nice idea. And the launch system is intended to be introspectable. But I'm not sure how much of the remapping functionality is exposed to perform such dry runs.

hidmic:
FYI I just collapsed the `<params>` tag into the `<param>` to make things slightly more homogeneous and thus easier to deal with in the implementation (which I'm thinking of and drafting some considerations and hints for).

articles/specs/launch.0.1.0.xsd
wjwwood: I'd start without markup level (or preprocessing level) substitution, and instead endeavor to have the markup language capture the substitutions, or failing that have the markup ignore the substitutions, leaving them for the launch code to interpret.

articles/151_roslaunch_xml.md
wjwwood: @ruffsl I think it will be easier for you to contribute to this discussion if you skim over the architecture document for launch:

https://github.com/ros2/launch/blob/master/launch/doc/source/architecture.rst

It's not that long and it introduces most, if not all, of the concepts that the markup needs to capture, and some of the reasoning why they are captured as they are rather than leaning on language features (whether it be programming languages like Python or markup languages like XML) to accomplish certain tasks.

articles/151_roslaunch_xml.md
wjwwood: You could introspect launch without doing anything, extracting all information you might want, and then use `rclpy`'s functions to expand them "off-line". It should be possible to at least get the exact (and fully expanded) remapping arguments that would be passed to nodes without actually running them.

However, in the specific case you're discussing, it's impossible to know what topics might be created during runtime which may be affected by this, if that's the information you're after.

In order to do that, you'd need a Node IDL which described which topics each node would create if run. This would be easy to implement in most case but is problematic to cover dynamic cases, e.g. move base creates topics based on parameter values (one for each source of obstacle avoidance).

articles/specs/launch.0.1.0.xsd
```diff
@@ -0,0 +1,562 @@
...
+        <xs:element ref="param"/>
+      </xs:choice>
+      <xs:attribute name="name" type="xs:string" use="required">
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            Name of the ROS parameter to be set.
+          </xs:documentation>
+        </xs:annotation>
+      </xs:attribute>
+      <xs:attribute name="value" type="xs:string" use="required">
```
ivanpauno: 'value' attribute should be mutually exclusive with 'from'.
Or maybe better: 'name' and 'value' should be mutually exclusive with 'from'.

articles/specs/launch.0.1.0.xsd
```diff
@@ -0,0 +1,562 @@
...
+      <xs:choice minOccurs="0" maxOccurs="unbounded">
+        <xs:annotation>
+          <xs:documentation xml:lang="en">
+            A collection of ROS parameters, ROS remappings and/or
+            environment variable settings for the launched ROS node.
+          </xs:documentation>
+        </xs:annotation>
+
+        <xs:element ref="env"/>
+        <xs:element ref="param"/>
```
ivanpauno, line 524: When the node has some parameters, it should also have a node name (if not, an error is raised). I don't know if it's possible to add an assert condition to the schema, which checks if node has the 'name' attribute when it contains a 'param' child.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,223 @@
...
+```xml
+{% include_relative specs/launch.0.1.0.xsd %}
+```
+
+### Tags Semantics
+
+#### All Tags
+
+Every tag's execution can be conditioned on a boolean predicate via `if` or `unless` attributes.
+All action tags that can contain other action tags can scope ROS entities in a namespace via the `ns` attribute.
```
ivanpauno: I don't think we will allow that.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,223 @@
...
+There must only be one `<launch>` tag on a given launch file.
+
+#### `<include>` Tag
+
+##### Description
+
+The `<include>` tag allows for bringing a launch file description into another, enabling re-use of hierarchical launch layouts.
+The included launch file description has its own scope for launch configurations.
+The included launch file description is not necessarily written in this format nor a declarative one (i.e. a programmatic description).
+To avoid ROS name clashes, included nodes can be namespaced via the `ns` attribute.
```
ivanpauno: `IncludeLaunchDescription` action doesn't have that feature. As it's part of `launch` package, which should be ros agnostic, I don't think we are going to allow it.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,223 @@
...
+#### `<group>` Tag
+
+##### Description
+
+The `<group>` tag allows for launch actions' grouping as well as optional launch file configuration scoping.
+
+##### Examples
+
+```xml
+<group ns="dummy_group" scoped="true">
```
ivanpauno: I don't think we are going to allow `ns` here. Same rationale as above comment.

ivanpauno (commented):
I did some minimal changes in https://github.com/ros2/design/pull/207/commits/96307f7ae16a654b9cfb0ac937c67a39e016461f and https://github.com/ros2/design/pull/207/commits/44b905470262ee26f9b36bc4dabc46c60a011467.

I also left some comment about namespaces. Basically, include and group action are part of `launch` package, which shouldn't include ros related code.
One possible workaround is always doing the following:
```xml
<group>
    <let name="ros_namespace" value="$(var ros_namespace)/my_namespace"/>
    ... further include or node actions
</group>
```
That doesn't look quite nice.

The other workaround for this is adding alternative versions of `IncludeLaunchDescription` and `GroupAction` in `launch_ros` which handle this. We should also add in `launch.frontend` a way of allowing overriding exposed actions and substitutions (which is a little tricky).

@wjwwood @hidmic any opinions?

I think that we can delete it from the design document at the moment, and address it in a follow up.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,223 @@
...
+The `<let>` tags allows for definition of scoped launch file configuration variables.
+
+##### Examples
+
+```
+<let var="foo" value="$(env BAR)"/>
+<let var="baz" value="false"/>
+```
+
+#### `<arg>` Tag
```
ivanpauno, line 102: I forgot implementing this one, will do soon.

hidmic:
> We should also add in launch.frontend a way of allowing overriding exposed actions and substitutions (which is a little tricky).

Hmm, I see why, but I wonder if what this is actually telling us is that we need namespaced tags e.g. `<ros::group/>` vs. `<group/>` in XML.

> I think that we can delete it from the design document at the moment, and address it in a follow up.

I'm also inclined to defer this till the next iteration, along with event handlers.

ivanpauno:
> Hmm, I see why, but I wonder if what this is actually telling us is that we need namespaced tags e.g. `<ros::group/>` vs. `<group/>` in XML.

That sounds like a great idea. I would like to see a using tag too (`<using name="ros"/>`). Handling overrides will be easy.

> I'm also inclined to defer this till the next iteration, along with event handlers.

Sure, I will delete event handlers and ros-namespaces from the document. We can add them in the next iteration.

wjwwood:
So it just boil down to this:

```xml
<group ns="my_ns">
</group>
```

Versus this:

```xml
<group>
  <let name="ros_namespace" value="$(var ros_namespace)/my_namespace"/>
</group>
```

I think we could have something like this:

```xml
<group>
  <push_ros_namespace namespace="my_namespace"/>
  ...
</group>
```

And that would be sufficient.

If we really want to have it as an attribute of `<group>` then I think you guys have identified one way which is to have `<group>` and `<ros::group>` where the latter can override the first if you do something like `<using name="ros::group"/>` or `<using namespace="ros"/>`. Another option would be to extend `<group>` directly, so that if you have `launch_ros_xml`, it adds acceptable attributes to `<group>`, like the `ns` one, through some extensibility interface.

ivanpauno:
> I think we could have something like this:
> 
> ```
> <group>
>   <push_ros_namespace namespace="my_namespace"/>
>   ...
> </group>
> ```

Yes, that one would be easy to implement, and less verbose than my option.

> If we really want to have it as an attribute of `<group>` then I think you guys have identified one way which is to have `<group>` and `<ros::group>` where the latter can override the first if you do something like `<using name="ros::group"/>` or `<using namespace="ros"/>`. Another option would be to extend `<group>` directly, so that if you have `launch_ros_xml`, it adds acceptable attributes to `<group>`, like the `ns` one, through some extensibility interface.

I think the namespace option would be really nice, because it avoids the possibility of name collisions. But, it's harder to implement.

For the moment I will add a `push_ros_namespace` action. I think we can merge the document as-is and update it later.

hidmic:
+1 to `push_ros_namespace` for now and deferring namespace/extension discussions to the next iteration.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,219 @@
...
+  <remap from="generic_topic_name" to="my_topic"/>
+</node>
+```
+
+#### `<param>` Tag
+
+##### Description
+
+The `<param>` tag allows for setting ROS parameters of a ROS node. These
+can be scalar values or sequences of scalar values, defined directly or
```
wjwwood: Please use one sentence per line, and do not manually wrap if possible.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,219 @@
...
+either nested or brought from a YAML file to make a map.
+
+##### Examples
+
+```xml
+<node package="my_pkg" executable="my_node">
+  <param name="some_numeric_param" value="100.2"/>
+  <param name="some_list_param" value="Some phrase,100.0,true" sep=","/>
```
wjwwood: Will this result in a list of strings? Is it possible to do a list of integers?

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,219 @@
...
+
+##### Examples
+
+```xml
+<node package="my_pkg" executable="my_node">
+  <env name="RMW_IMPLEMENTATION" value="rmw_fastrtps_cpp"/>
+</node>
+
+<executable cmd="ls" cwd="/var/log">
+  <env name="LD_LIBRARY" value="/lib/some.so"/>
```
wjwwood, line 186: Can this be used outside of an executable? I think it should be possible.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Dynamic Configuration
+
+### Substitution Syntax
+
+All substitutions are enclosed by `$(...)`.
+
+### Built-in Substitutions
+
+`$(find-pkg pkg-name)`
```
wjwwood: Should it be `$(find-pkg <pkg-name>)` to indicate that `pkg-name` is a placeholder? This is typical in usage documentation, and I see you using `[]` for optional items below.

articles/151_roslaunch_xml.md
```diff
@@ -0,0 +1,219 @@
...
+  Forward and backwards slashes will be resolved to the local filesystem convention.
+  Substitution will fail if the executable cannot be found.
+
+`$(var name)`
+: Substituted by the value of the launch configuration variable.
+  Substitution will fail if the named argument does not exist.
+
+`$(env env-var [default-value])`
+: Substituted by the value of the given environment variable
+  Substitution will fail if the variable is not set, unless default values are provided.
```
wjwwood:
```suggestion
  Substitution will fail if the variable is not set, unless a default value is provided.
```

wjwwood (changes requested):
lgtm, with a few comments and one requested style change.

articles/151_roslaunch_xml.md
wjwwood: I think this is an important one to do 👍.

ivanpauno: Done in https://github.com/ros2/launch/pull/270

articles/151_roslaunch_xml.md
ivanpauno: Ups, that wasn't updated.

articles/151_roslaunch_xml.md
ivanpauno: For that, I should add a parsing method to `SetEnviromentVariable` (now it only has meaning inside `node` and `executable`, the schema wasn't up to date).
We have a subtle issue with that, `SetEnviromentVariable` is an action that supports `if` and `unless` condition. But for `node` and `executable`, I was using `additional_env` argument of them, where I cannot use conditions. So exposing `SetEnviromentVariable` variable with the same name would be confusing.

The workarounds I see are the following:
- Document the limitation.
- Change `ExecuteProcess` and `Node` parsing methods. Instead of using `additional_env`, return a sandwich of `SetEnviromentVariable`, `Node/ExecuteProcess`, `UnsetEnvironmentVariable`.
- Use a different tag.

wjwwood: You could use `<set_env ...>` outside of execute process and/or node.

ivanpauno:
With https://github.com/ros2/launch/pull/272 and https://github.com/ros2/launch/pull/273, we will have exactly the functionality described in this document.

wjwwood (approved):
lgtm, thanks!

PR #208, opened 2019-02-06
Propose ROS 2 launch front-end design hints.

hidmic:
This is a proposal on how to go about implementing something like #207 in an extensible way. Approaches described in here are a combination of ideas I had and ideas that @wjwwood had and shared with me, thus the dual authorship. 

I hope that by the time this gets merged we either have chosen one of these approaches or came up with a another one to actually implement.

### DIFF
--- a/articles/160_roslaunch_frontend.md
+++ b/articles/160_roslaunch_frontend.md
@@ -343,7 +343,7 @@ Each launch entity that is to be statically described must provide a parsing pro
 
 *Python*
 ```python
-@launch.frontend.expose_action('some-action')  # Infers it's an Action
+@launch.frontend.expose_action('some-action')
 class SomeAction(launch.Action):
 
     @classmethod
@@ -358,7 +358,7 @@ class SomeAction(launch.Action):
             ]
         )
 
-@launch.frontend.expose_substitution('some-subst')  # Infers it's a Substitution
+@launch.frontend.expose_substitution('some-subst')
 class SomeSubstitution(launch.Substitution):
 
     @classmethod
@@ -384,7 +384,7 @@ class SomeAction : public launch::Action {
   }
 };
 
-LAUNCH_EXPOSE("some-action", SomeAction);  // Infers it's an Action
+LAUNCH_EXPOSE_ACTION("some-action", SomeAction);
 
 class SomeSubstitution : public launch::Substitution {
   static std::unique_ptr<SomeSubstitution>
@@ -393,7 +393,7 @@ class SomeSubstitution : public launch::Substitution {
   }
 };
 
-LAUNCH_EXPOSE("some-subst", SomeSubst);  // Infers it's a Substitution
+LAUNCH_EXPOSE_SUBSTITUTION("some-subst", SomeSubst);
 ```
 
 As can be seen above, procedures inspect the description through the given parsing entity, delegating further parsing to the parser recursively.
@@ -408,7 +408,7 @@ In the simplest case, the user may explicitly provide their own parsing procedur
 
 ##### Automatic Derivation
 
-If accurate type information is available (e.g.: type annotations in constructor), reflection mechanisms can aid derivation of a parsing procedure with no user intervention.
+If accurate type information is somehow available (e.g.: type annotations in constructor), reflection mechanisms can aid derivation of a parsing procedure with no user intervention.
 
 #### Advantages & Disadvantages
 
@@ -417,6 +417,9 @@ The abstraction layer allows.
 *-* Launch system implementations are aware of the parsing process.
 
 *+* The static description abstraction effectively decouples launch frontends and backends, allowing for completely independent development and full feature availability at zero cost.
+
+*-* No markup language specific sugars are possible.
+    REVISIT(hidmic): IMHO explicitly disallowing this is a good thing, it makes for more homogeneus descriptions and avoids proliferation of multiple representation of the same concepts (e.g. a list of strings).
 
 *+* The transfer function nature of the parsing procedure precludes the need for a rooted object type hierarchy in statically typed launch system implementations.
 
--- a/articles/160_roslaunch_frontend.md
+++ b/articles/160_roslaunch_frontend.md
@@ -26,28 +26,27 @@ This document describes parsing and integration approaches of different front en
 
 ## Proposed approaches
 
-In the following, different approaches to the solution are described.
+In the following, different approaches to parsing launch descriptions are described.
 This list is by no means exhaustive.
 
-It's worth to note some commonalities between them:
+It's worth noting some things they all have in common:
 
-- All of them attempt to solve the problem in a general and extensible way to help
-  the system scale with its community.
+- All of them attempt to solve the problem in a general and extensible way to help the system scale with its community.
 
-- All of them require a way to establish associations between entities, solved using
-  unique reference id (usually, a human-readable name but that's not a requisite).
+- All of them require a way to establish associations between entities, solved using unique reference id (usually, a human-readable name but that's not a requisite).
 
-- All of them need some form of instantiation and/or parsing procedure registry for
-  parsers to lookup.
+- All of them associate a markup language with a given substitution syntax.
+
+- All of them supply a general, interpolating substitution to deal with embedded substitutions e.g. "rooted/$(subst ...)".
+
+- All of them need some form of instantiation and/or parsing procedure registry for parsers to lookup.
   How that registry is populated and provided to the parser may vary.
-  For instance, in Python class decorators may populate a global dict or even its import
-  mechanism may be used if suitable, while in C++ convenience macros may expand into
-  demangled registration hooks that can later be looked up by a dynamic linker (assuming
-  launch entities libraries are shared libraries).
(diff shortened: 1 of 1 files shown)


articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,358 @@
...
+
+Static launch descriptions are an integral part to ROS 2 launch system, and the natural path to transition from predominant ROS 1 `roslaunch` XML description.
+This document describes parsing and integration approaches of different front ends i.e. different markup languages, with a focus on extensibility and scalability.
+
+## Proposed approaches
+
+In the following, different approaches to the solution are described.
+This list is by no means exhaustive.
+
+It's worth to note some commonalities between them:
```
wjwwood:
```suggestion
It's worth noting some things they all have in common:
```

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,358 @@
...
+*YAML*
+```yml
+launch:
+  version: x.1.0
+  actions:
+    - type: actions.ExecuteProcess
+      name: some-action
+      args:
+        cmd: /bin/ls
+  handlers:
```
wjwwood:
```suggestion
  event_handlers:
```

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,358 @@
...
+#### Parsing Hooks
+
+Some code samples in different programming languages are provided below:
+
+*Python*
+```python
+class SomeAction(LaunchDescriptionEntity):
+
+   @launch.parse_from('xml')
+   def parse(cls, element):
```
wjwwood: nitpick: I know this is a just a theoretical example, but it might be good (even in the examples) to separate the markup helpers (markup sugar or w/e you want to call them) from the actual action.

Something like:

```python
@launch_xml.handle_tag('process')
def process_tag_helper(xml_element):
    ...
    return launch.actions.ExecuteProcess(...)
```

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,358 @@
...
+*C++*
+```c++
+class SomeAction : public LaunchDescriptionEntity {
+public:
+  static std::unique_ptr<SomeAction> parse_from_xml(const XmlElement& e) {
+     ...
+  }
+};
+
+LAUNCH_XML_PARSING_HOOK("some-action", SomeEntity::parse_from_xml);
```
wjwwood: For this one it's be like:

```c++
std::unique_ptr<launch_cpp::actions::ExecuteProcess>
process_tag_helper(const XmlElement & xml_element)
{
  // ...
  return std::make_unique<launch_cpp::actions::ExecuteProcess>(/* ... */);
}

LAUNCH_CPP_XML_HANDLE_TAG('process', process_tag_helper);
```

wjwwood: Though having written both examples now, I wonder if this doesn't overlap with some proposal further down in the article...

wjwwood (commented):
Looks like a good start to me.

As for which implementation to use, the one I mostly closely identified with was FDM+, though perhaps this document is just not compelling enough. We might need to discuss off-line to hash out more detail.

Also, a big thing that is missing (but possibly orthogonal to these approaches) is substitutions and how they would be handled.

articles/160_roslaunch_frontend.md
hidmic: Same as above.

articles/160_roslaunch_frontend.md
hidmic: Yeah, I see where you're going and it makes sense. Changed.

articles/160_roslaunch_frontend.md
hidmic: Suggestion accepted (but I'm skipping the Github suggestion to avoid DCO issues).

articles/160_roslaunch_frontend.md
hidmic: Suggestion accepted (but I'm skipping the Github suggestion to avoid DCO issues).

hidmic:
> Also, a big thing that is missing (but possibly **orthogonal** to these approaches) is substitutions and how they would be handled.

And I think that's exactly why I skipped them in the first place. The document now covers a bit on how substitutions could be handled. PTAL! And thanks for reviewing it in the first place!

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,422 @@
...
+
+- All of them supply a general, interpolating substitution to deal with embedded substitutions e.g. "rooted/$(subst ...)".
+
+- All of them need some form of instantiation and/or parsing procedure registry for parsers to lookup.
+  How that registry is populated and provided to the parser may vary.
+  For instance, in Python class decorators may populate a global dict or even its import mechanism may be used if suitable, while in C++ convenience macros may expand into demangled registration hooks that can later be looked up by a dynamic linker (assuming launch entities libraries are shared libraries).
+
+### Forward Description Mapping (FDM)
+
+In FDM, the parser relies on a schema and well-known rules to map an static description (markup) to implementation specific instances (objects).
```
wjwwood:
```suggestion
In FDM, the parser relies on a schema and well-known rules to map a static description (markup) to implementation specific instances (objects).
```

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,422 @@
...
+        - type: actions.LogInfo
+          args:
+            message: "I'm done"
+```
+
+#### Advantages & Disadvantages
+
+*+* Straightforward to implement.
+
+*+* Launch implementations are completely unaware of the static description existence and its parsing process (to the extent that type agnostic instantiation mechanisms are available).
```
wjwwood:
```suggestion
*+* Launch implementations are completely unaware of the existence of the static description formats and their parsing processes (to the extent that type agnostic instantiation mechanisms are available).
```

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,422 @@
...
+LAUNCH_MARKUP_HANDLE_SUBST("env", env_subst_helper);
+```
+
+#### Advantanges & Disadvantages
+
+*+* Straightforward to implement.
+
+*-* Launch system implementations are aware of the parsing process, being completely involved with it if sugars are to be provided.
+
+*+* Allows leveraging the strenghts of each markup language.
```
jacobperron:
```suggestion
*+* Allows leveraging the strengths of each markup language.
```

articles/160_roslaunch_frontend.md
```diff
@@ -0,0 +1,422 @@
...
+
+*-* Opens the door to big differences in the representation of launch entities across different front end, and even within a given one by allowing the users to introduce multiple custom representations for the same concepts (e.g. a list of numbers).
```
jacobperron:
```suggestion
*-* Opens the door to big differences in the representation of launch entities across different front ends, and even within a given one by allowing the users to introduce multiple custom representations for the same concept (e.g. a list of numbers).
```

ivanpauno:
@hidmic I addressed the comments, and updated it to match better with the actual implementation (it was actually matching pretty well).

articles/160_roslaunch_frontend.md
```diff
@@ -412,9 +418,6 @@ The abstraction layer allows.
 
 *+* The static description abstraction effectively decouples launch frontends and backends, allowing for completely independent development and full feature availability at zero cost.
 
-*-* No markup language specific sugars are possible.
-    REVISIT(hidmic): IMHO explicitly disallowing this is a good thing, it makes for more homogeneus descriptions and avoids proliferation of multiple representation of the same concepts (e.g. a list of strings).
-
```
hidmic: @ivanpauno this still applies.

articles/160_roslaunch_frontend.md
```diff
@@ -336,31 +343,30 @@ Each launch entity that is to be statically described must provide a parsing pro
 
 *Python*
 ```python
-@launch.expose('some-action')  # Infers it's an Action
+@launch.frontend.expose_action('some-action')  # Infers it's an Action
```
hidmic: @ivanpauno I think we should remove `# Infers it's an Action`.

articles/160_roslaunch_frontend.md
```diff
@@ -336,31 +343,30 @@ Each launch entity that is to be statically described must provide a parsing pro
...
         return cls(
-            scoped=parser.parse(entity.scoped), entities=[
-                parser.parse(child) for child in entity.children
+            scoped=parser.parse_substitution(entity.get_attr(scoped)), entities=[
+                parser.parse_action(child) for child in entity.children
             ]
         )
 
-@launch.expose('some-subst')  # Infers it's a Substitution
+@launch.frontend.expose_substitution('some-subst')  # Infers it's a Substitution
```
hidmic: @ivanpauno I think we should remove `# Infers it's a Substitution`.

ivanpauno:
I think this PR is ready. We will deal with event_handlers in the next iteration, but I didn't delete reference to them as this is a design hints document.

wjwwood (approved):

jacobperron (approved):

PR #210, opened 2019-02-08
lock bundler version

dirk-thomas:
Fix failing CI on the default branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
 - 2.1
 sudo: false
 install:
-- gem install bundler --no-rdoc --no-ri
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Ruby 2.1 has ended support. I think an optimal fix would be to try updating the ruby version to something more current.

dirk-thomas:
> I think an optimal fix would be to try updating the ruby version to something more current.

Please feel free to do so in a follow up PR. This patch makes the CI pass with the same versions as before in order to continue merging other PRs.

wjwwood (approved):
It would be better to update to the latest version (or revisit GitHub's recommendations about this), but I guess this gets it going again.

dirk-thomas:
See follow up work in #211.

PR #211, opened 2019-02-08
Try bumping Ruby version instead of pinning bundler version.

nuclearsandwich:
This is attempted as an alternative to #210.

Looking at the node version that also may be a candidate for updating but I have no idea how well supported those node libraries are on later versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
 - 2.6
 sudo: false
 install:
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Assuming that this is doing what it's supposed to be doing. This _Just Worked_:tm:.

dirk-thomas (approved):
lgtm

PR #213, opened 2019-02-15
update mapping and information about deprecated types, add actions, minor tweaks

dirk-thomas:
Follow up of #190.

### DIFF
--- a/articles/143_legacy_interface_definition.md
+++ b/articles/143_legacy_interface_definition.md
@@ -224,7 +224,7 @@ Depending on the type the following values are valid:
 
 - `char`:
 
-  - an unsigned character with a numerical value in the following interval `[0, 255]`
+  - an unsigned integer value in the following interval `[0, 255]`
 
 - `float32` and `float64`:
 
@@ -268,7 +268,7 @@ Code generation uses the generated file.
 | -------- | ------------------ |
 | bool     | boolean            |
 | byte     | octet              |
-| char     | char               |
+| char     | uint8              |
 | float32  | float              |
 | float64  | double             |
 | int8     | int8               |
@@ -282,15 +282,14 @@ Code generation uses the generated file.
 | string   | string             |
 
 <div class="alert alert-info" markdown="1">
-  The mapping of <code>byte</code> uses a different type while still remaining an opaque 8-bit quantity.
+  The mapping of <code>byte</code> uses a different type than in ROS 1 while still remaining an opaque 8-bit quantity.
   [Definition in ROS 1](http://wiki.ros.org/msg#Field_Types): deprecated alias for <code>int8</code>.
   Definition in IDL (7.4.1.4.4.2.6): an opaque 8-bit quantity.
 </div>
 
 <div class="alert alert-info" markdown="1">
-  While the mapping of <code>char</code> maintains the same value range as defined in ROS 1 it uses a different type.
-  [Definition in ROS 1](http://wiki.ros.org/msg#Field_Types): deprecated alias for <code>uint8</code>.
-  Definition in IDL (7.2.6.2.1): an 8-bit quantity with a value between 0 and 255.
+  While the mapping of <code>char</code> is unintuitive it preserves compatibility with the [definition in ROS 1](http://wiki.ros.org/msg#Field_Types):
+  "deprecated alias for <code>uint8</code>".
 </div>
 
 | ROS type                | IDL type         |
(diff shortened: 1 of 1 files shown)


articles/143_legacy_interface_definition.md
```diff
@@ -40,10 +41,17 @@ Together with the name of the *package* a message can be uniquely identified.
...
+These pairs of data structures are called *services*.
 A service is identified by its *name* and the *package* it is in.
 Each service describes two messages, one for the request data structure, one for the reply data structure.
 
+### Actions
+
+For longer running request / reply style communication with feedback about the progress the exchanged data structures are related.
+These triplets of data structures are called *actions*.
+An action is identified by its *name* and the *package* it is in.
+Each service describes three messages, one for the goal data structure, one for the result data structure, and one for the feedback data structure.
```
hidmic: @dirk-thomas typo?

```suggestion
Each action describes three messages, one for the goal data structure, one for the result data structure, and one for the feedback data structure.
```

articles/143_legacy_interface_definition.md
```diff
@@ -92,11 +98,19 @@ A dynamic array can have between `0` and `N` elements of the specified type.
 
 #### Upper boundaries
 
+<div class="alert alert-info" markdown="1">
+  This feature has been added compared to the ROS 1 format.
```
hidmic: @dirk-thomas what do you mean by `has been added compared to`? That it has been added here, not like in ROS 1?

articles/143_legacy_interface_definition.md
```diff
@@ -205,11 +220,11 @@ Depending on the type the following values are valid:
 
 - `byte`:
 
-  - an unsigned integer value in the following interval `[0, 255]`
+  - an opaque 8-bit quantity with a numerical value in the following interval `[0, 255]`
 
 - `char`:
 
-  - an integer value in the following interval `[-128, 127]`
+  - an unsigned character with a numerical value in the following interval `[0, 255]`
```
hidmic: @dirk-thomas hmm, would it make more sense to either say the character is restricted to ASCII encoding or that we're dealing with an unsigned 8-bit quantity? Character and numerical on the same sentence sound like there's something left implicit about size and/or encoding.

mjcarroll (approved):
LGTM.

articles/143_legacy_interface_definition.md
mjcarroll: Can't believe I missed that!

articles/143_legacy_interface_definition.md
dirk-thomas: Yes, this is an addition in this specific compared to the ROS 1 format.

mjcarroll: These are all new features for ROS2, maybe it could be reworded as "New for ROS2" or "Not present in ROS1" to be more explicit?  I think it makes sense as-is, but may be difficult to parse?

articles/143_legacy_interface_definition.md
dirk-thomas: We don't prescribe any kind of semantic / encoding to these types. This only describes the data structure itself.

hidmic: Alright, then maybe stating 8-bit quantity would make more sense (like IDL does)? IMHO saying character and not specifying encoding nor platform may be confusing.

dirk-thomas: Imo the information "unsigned" is important here for the expectation what type that might be in each programming language (rather than signed).

dirk-thomas:
The goal of mapping ROS 1 `char` to IDL `char` was to have a way to express an unsigned character type in `.msg`. Looking back I don't think this change is worth doing. I reverted that part in eec42d6b3d0ed883d611beac043b2ccd8f373754 and `char` continues to be mapped to `uint8` (as it was in ROS 1).

For the case that there is no character type in `.msg`: well, that has always been the case and I don't think we need to resolve this. In IDL that will be easy to express...

articles/143_legacy_interface_definition.md
dirk-thomas: See https://github.com/ros2/design/pull/213#issuecomment-464169725 for reverting this part of the patch.

articles/143_legacy_interface_definition.md
dirk-thomas: See cb4739c83b64b13f92427a2d4e78d5262d12bf5c.

hidmic (approved):
LGTM!

PR #215, opened 2019-03-12
publish idl articles, add note to existing articles

dirk-thomas:
Follow up of ros2/rosidl#346 marking the IDL articles as published as well as adding notes to the existing articles.

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -4,7 +4,7 @@ title: IDL - Interface Definition and Language Mapping
 permalink: articles/idl_interface_definition.html
 abstract:
   This describes defining interfaces using a subset of the [Interface Definition Language](https://www.omg.org/spec/IDL/) (IDL).
-published: false
+published: true
 author: '[Dirk Thomas](https://github.com/dirk-thomas)'
 categories: Interfaces
 ---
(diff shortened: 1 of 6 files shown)


articles/142_idl.md
```diff
@@ -32,6 +32,14 @@ Further it describes how the interfaces are used to generate code in C, C++ and
...
 ROS 2 supports a subset of the [OMG IDL 4.2 specification](https://www.omg.org/spec/IDL/4.2).
 
+### Lexical Conventions
+
+#### 7.2.2 Comments
+
+#### 7.2.3 Identifiers
+
+#### 7.2.6 Literals
+
```
hidmic, line 46: @dirk-thomas are these placeholders for future additions?

hidmic (approved):
LGTM

articles/142_idl.md
dirk-thomas: These are only enumerating the sections of the IDL spec which are supported.

hidmic: I see. Would it make sense to state above that what follows is an enumeration of supported IDL spec sections? It wasn't obvious to me at first sight, the fact that these are the only "empty" sections didn't help inferring it either.

spiderkeys, line 46: Could a section be added here explicitly discussing the support or lack thereof for annotations? I'm not sure what the latest thinking is with respect to DDS's use of keyed fields/instances and whether or not RMW will ever support that concept, but of particular interest is whether or not annotations like @key fields might get preserved when generating vendor specific types. This would impact middleware behavior, even if ROS2 didn't explicitly provide support for the capabilities introduced, so it would be worth addressing either way.

dirk-thomas: I added a short text to each section instead: 826f0119a11b1420db7d1020fc5561b1c284f6b4.

dirk-thomas: I added some information about annotations in 94a37507b0710a19540fcbdccbc0c79b48aff913.

PR #216, opened 2019-03-13
Fix RVM issue in CI by switching Rubies

nuclearsandwich:
Although I suspect this is an issue with travis's networking on their VMs based on the curl output:

```
$ rvm use 2.6 --install --binary --fuzzy
curl: (22) The requested URL returned error: 403 Forbidden
Required ruby-2.6 is not installed - installing.
curl: (22) The requested URL returned error: 403 Forbidden
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm ruby-2.6 do rvm gemset create ' first, or append '--create'.
The command "rvm use 2.6 --install --binary --fuzzy" failed and exited with 2 during .
```

As of this morning, the result is a 404 not a 403. If I install rvm in a local trusty container, I can get the binary from... rubies.travis-ci.org... and it installs successfully.

```
# rvm use ruby-2.6 --install --fuzzy --binary
Required ruby-2.6.1 is not installed - installing.
Searching for binary rubies, this might take some time.
Found remote file https://rubies.travis-ci.org/ubuntu/14.04/x86_64/ruby-2.6.1.tar.bz2
```

I don't expect falling back to ruby 2.5 will fix the travis issue but this is a starting point to try and get it up again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
+dist: xenial
 language: ruby
 rvm:
-- 2.5
+- 2.6
 sudo: false
 install:
 - gem install bundler --no-doc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-- 2.5
+- 2.6.0
 sudo: false
 install:
 - gem install bundler --no-doc
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ install:
 - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install 10
 - npm install remark-cli remark-lint remark-lint-sentence-newline
 script:
-- bundler --version
 - bundle exec jekyll build
 - ./node_modules/remark-cli/cli.js --frail articles/
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
-dist: xenial
 language: ruby
 rvm:
-- 2.6
+- 2.5
 sudo: false
 install:
 - gem install bundler --no-doc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-- 2.6
+- 2.5
 sudo: false
 install:
 - gem install bundler --no-doc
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
Annoyingly, the above worked.

It found the Ruby from language-archives.travis-ci.com .....

```
20.16s$ rvm use 2.5 --install --binary --fuzzy
curl: (22) The requested URL returned error: 404 Not Found
Required ruby-2.5.1 is not installed - installing.
curl: (22) The requested URL returned error: 404 Not Found
Searching for binary rubies, this might take some time.
Found remote file https://language-archives.travis-ci.com/ruby/binaries/ubuntu/14.04/x86_64/ruby-2.5.1.tar.bz2
```

nuclearsandwich:
Well it irritates the hell out of me that falling back to 2.5 is working but it seems like a Travis networking issue I don't have the power to resolve and this gets CI turning over. We don't _really_ care about the 2.6 vs 2.5 here it was just pushing when I'd have to look at it again into The Future.

nuclearsandwich:
Definitely squash this to hide my experimentation.

nuclearsandwich:
Ah, found this reported on Travis's community site. https://travis-ci.community/t/travis-cannot-find-a-ruby-2-6-version-using-rvm/2641

nuclearsandwich:
Started tracking that topic on the Travis forum so we can revert this when the issue is resolved.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-- 2.6
+- 2.6.0
```
dirk-thomas, line 3: This change shouldn't be necessary.

dirk-thomas:
I guess the problem fixed itself - retriggered CI is now passing for #215.

.travis.yml
nuclearsandwich: This change switched the Ruby version from 2.6.1 implicit which was not being found to 2.6.0 explicit which was. It was intentional and necessary as an alternative to switching to 2.5.

nuclearsandwich:
No longer necessary. The last commit 2c8a3ac may be desired to fix other Bundler related issues if they come up in the future

nuclearsandwich:
> I guess the problem fixed itself - retriggered CI is now passing for #215.

https://travis-ci.community/t/travis-cannot-find-a-ruby-2-6-version-using-rvm/2641/3 with thanks to `@banzaiman`

PR #217, opened 2019-03-18
map sequence to Python array.array()

dirk-thomas:
This patch proposes to map (un)bounded sequences of numerics to [array.array](https://docs.python.org/3/library/array.html#array.array) instead of [numpy.ndarray](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.html).

An array provides the same interface as a Python list. It only uses memory closer to what the C representation needs (similar even though slightly higher than `numpy`) and also checks the value range of each item based on the typecode.

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -227,7 +227,8 @@ If a cell is blank then the default mapping is used.
 
 | IDL type                     |   | C type                            | C++ type       | Python type                                     |   |
 | ---------------------------- | - | --------------------------------- | -------------- | ----------------------------------------------- | - |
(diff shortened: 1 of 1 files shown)


hidmic (approved):
LGTM, but what about `char` and `wchar` types? As it stands, it looks like we'll end up with a `list` of them.

dirk-thomas:
> what about `char` and `wchar` types?

What are you suggesting for those types? As far as I can see there are no corresponding `dtypes` / `typecodes` for these primitives in `numpy` / `array`.

> As it stands, it looks like we'll end up with a `list` of them.

Correct.

hidmic:
Hmm, I initially noticed 'b', 'B' and 'u', but on a second look, it treats the first two as numbers and using 'u' would result on `sequence<char>` and `sequence<wchar>` being represented differently. Nevermind my comment.

PR #218, opened 2019-03-20
Add ROS 2 Threat Model article

thomas-moulard:
This adds an article describing a threat model for
ROS 2 robotic applications.

The objective of this work is to drive ROS 2 security
work and define, end-to-end, what it takes for a robotic
company to build a secure ROS 2 robotic application.

Signed-off-by: Thomas Moulard <tmoulard@amazon.com>


gbiggs:
Wow.

gavanderhoorn:
@thomas-moulard: nice article. <del>I notice however that there is just a single reference to related/existing work. Was there really no usable literature available, or was it all non-applicable? Grounding these sort of articles in the existing body of knowledge -- also (or preferably?) from outside the ROS community -- would significantly increase the value of them.</del> Edit: oops, I should scroll to the right.

Perhaps a *References* section would be a good idea?

olaldiko:
Great work @thomas-moulard! We have reviewed it and it looks very promising. I think that most of the attack vectors have been covered. Just a small concern, several risks for physical attacks to the OpenCR controller and the Raspberry Pi have been taken as acceptable for this architecture. Shouldn't we advise at least for physically protecting access to sensitive components? I would suggest a combination of shields or covers + usage of the ZimKey perimeter breach detection mechanisms. In our experience with commercial robots, this is the way to go.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+### Defining Robotic Systems Threats
+
+We will consider as a robotic system one or more general-purpose computer
+connected to one or more actuators or sensors. An actuator is defined as any
+device producing physical motion. A sensor is defined as any device capturing or
+recording a physical property.
+
+### Robot Application Actors, Assets, Business Goals and Entry Points
+
+This section defines actors, assets, business goals and entry points for this
```
jacobperron, line 79: This section does not explicitly define business goals. For clarity, consider organizing the entities as done in the threat analysis section.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+<table class="table">
+  <thead>
+    <th>Asset</th>
+    <th>Description</th>
+  </thead>
+
+  <tr><th colspan="2">Privacy</th></tr>
+  <tr>
+    <td>Sensor Data Privacy</td>
+    <td>Sensor data must not be accessed by attackers.</td>
```
jacobperron: More specifically, "unauthorized actors"?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+Privacy</th>
+    <th> </th>
+  </tr>
+
+  <tr><th colspan="29">Embedded / Software / Communication / Inter-Component
+Communication</th></tr>
+
+  <tr>
+    <td>An attacker spoofs a component identity.</td>
+    <td class="success">✓</td>
```
jacobperron, line 442: Maybe it's obvious, but it might be worth defining what "success", "danger", and "warning" mean for each category (STRIDE, DREAD, and impacted entities).

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+    <td class="danger">✘</td>
+    <td class="warning">▲</td>
+    <td class="success">✓</td>
+    <td>
+      <ul>
+        <li>Cloud data stores should encrypt data at rest</li>
+      </ul>
+    </td>
+    <td> </td>
+    <td> </td>
```
jacobperron: Remove extra `<td>` tag

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+
```
jacobperron: Recommend rephrasing: "The application considered in this section is tele-operation of a Turtlebot 3"

vmayoral, line 2042: Great diagram and great reading! I've learn tons from this article. Very good contribution. Blue lines in the diagram below are representing "DDS protocols" do you mean RTPS solely? If not, what else are you capturing here?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td>
+      <ul>
+        <li>Actuators should embed an identifier.</li>
+        <li>Command vector should be signed (ideally encryped) to prevent
```
vmayoral: Minor typo here
```suggestion
        <li>Command vector should be signed (ideally encrypted) to prevent
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+connected to one or more actuators or sensors. An actuator is defined as any
+device producing physical motion. A sensor is defined as any device capturing or
+recording a physical property.
+
+### Robot Application Actors, Assets, Business Goals and Entry Points
+
+This section defines actors, assets, business goals and entry points for this
+threat model.
+
+Actors are humans or external systems interacting with the robot. Considering
```
vmayoral: I had to go several times through this section. I'd suggest bolding the concept definition itself here for readability. 
```suggestion
**Actors** are humans or external systems interacting with the robot. Considering
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+which actors interact with the robot is helpful to determine how the system
+can be compromised. For instance, actors may be able to give commands to the
+robot which may be abused to attack the system.
+
+Assets represent any user, resource (e.g. disk space), or property (user physical
```
vmayoral: Same here
```suggestion
**Assets** represent any user, resource (e.g. disk space), or property (user physical
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+safety) of the system that should be defended against attackers.
+
+Entry points represent how the system is interacting with the world (communication
```
vmayoral: And here
```suggestion
**Entry points** represent how the system is interacting with the world (communication
```

articles/ros2_threat_model.md
vmayoral: +1, this was confusing.

articles/ros2_threat_model.md
olaldiko: Fix indentation

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+Herron, Tariq Yusuf, Junjie Yan, Tadayoshi Kohno, and Howard Jay Chizeck. “To
+Make a Robot Secure: An Experimental Analysis of Cyber Security Threats Against
+Teleoperated Surgical Robots.” ArXiv:1504.04339 [Cs], April 16, 2015.</a></td>
+  </tr>
+
+  <tr><th colspan="29">Embedded / Software / Communication / Short-Range
+Communication (e.g. Bluetooth)</th></tr>
+
+  <tr>
+  <td>Attacker executes arbitraty code using a short-range communication
```
olaldiko: An attacker

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+  <tr>
+    <td>Embedded Computer Physical Access</td>
+    <td>External (HDMI, USB...) and internal (PCI Express, SATA...) ports.</td>
+  </tr>
+</table>
+</div>
+
+### Robot Application Components and Trust Boundaries
+
+The system is divided into hardware (sensors, actuators), multiple components
```
vmayoral: Shouldn't you add the general purpose-computer of line https://github.com/ros2/design/pull/218/files#diff-60aa2b3022e7728b236f31c650141eceR69  to the hardware here? 
```suggestion
The system is divided into hardware (embedded general-purpose computer, sensors, actuators), multiple components
```

articles/ros2_threat_model.md
olaldiko: An attacker

olaldiko, line 2618: An attacker

olaldiko, line 3070: Add physical security measures, anti-tampering? (ZimKey perimeter breach detection)

olaldiko, line 3123: Add usage of TPM devices as well for the rest of the platforms?

olaldiko (changes requested):
Here go some typo fixes and suggestions!

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+        <li>Logs access should be monitored to enable later audits.</li>
+      </ul>
+    </td>
+    <td> </td>
+  </tr>
+
+  <tr><th colspan="29">Embedded / Hardware / Sensors</th></tr>
+
+  <tr>
+    <td>An attacker spoofs a robot sensor (by e.g. replacing the sensor itself
```
vmayoral, line 1420: True, I agree with this threat evaluation overall but I'm missing in the "Embedded / Hardware / <Sensor/Actuators>" section considerations of physical attacks that involve complete removal (not simple replacement) of the sensor. I still find shocking how several commercial and industrial robots don't provide any mechanism to secure hardware physically and simply present sensors totally exposed. I collaborated in the production of a framework (with a slightly different aim) that touched into these aspects https://arxiv.org/pdf/1806.04042.pdf.

Would you like me to make a few additional row suggestions? @olaldiko was also involved on this so maybe he'd like to jump in directly?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td>
+      <ul>
+        <li>If a joint command is exceeding the joint limits, a specific code
```
vmayoral, line 1608: I don't quite see how this could be realized @thomas-moulard. Do you imply that all trajectories should be monitored for an "out of bounds" potential action and in case such is found a particular routine should get executed? Which routine? Wouldn't it be simpler to just discard the trajectory completely?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+path for handling out-of-bounds command should be executed instead of
+triggering the E-Stop. Whenever safe, the command could be discarded and the
+error reported to the user for instance.</li>
+      </ul>
+    </td>
+    <td> </td>
+  </tr>
+
+  <tr><th colspan="29">Embedded / Hardware / Aucilliary Functions</th></tr>
```
vmayoral: Typo?
```suggestion
  <tr><th colspan="29">Embedded / Hardware / Auxilliary Functions</th></tr>
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+[Diagram Source](ros2_threat_model/RobotSystemThreatModel.json)
+(edited with [Threat Dragon][threat_dragon])
+
+### Threat Analysis and Modeling
+
+The table below lists all *generic* threats which may impact a
+robotic application.
+
+Threat categorization is based on the [STRIDE][wikipedia_stride]
+ (Spoofing / Tempering / Repudiation / Integrity / Denial of service
```
olaldiko: Tampering

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td>
+      <ul>
+        <li>Sensors should embed an identifier to detect hardware
+tempering.</li>
```
olaldiko: tampering

olaldiko (changes requested):

articles/ros2_threat_model.md
vmayoral, line 2628: How about adding another section to treat vulnerabilities in the _Communications Middleware_? Something like "Embedded / Software / Communication Middleware". There're several issues opened in ros2/ros2 describing how middleware implementations start demanding a considerable amount of resources as more publishers (or subscribers) are enabled into them. We ourselves while building robots found considerable limitations in several DDS implementations. This can lead to simple DoS attacks by simply accessing the local network of the robot. While we haven't yet tested it extensively (@LanderU, feel free to comment), we've been able to limit the implications to a certain extend by properly configuring the DDS layer.

articles/ros2_threat_model.md
olaldiko: @vmayoral, line 2717?

articles/ros2_threat_model.md
olaldiko: I think this would fall on "An attacker prevents a communication channel from being usable."

vmayoral, line 2099: @JaimeMartin and @BorjaOuterelo ping.

vmayoral: This is a nice thing to add to the micro-ROS project. @imuguruza please consider it.

articles/ros2_threat_model.md
olaldiko: This would be in line with adding physical security measures or tamper detections. In would combine them into spoofing/removal of the sensor, and add alerting/logging of physical intrusions and sensor communication losses as mitigation.

articles/ros2_threat_model.md
vmayoral: I think there's an indentation issue here.
```suggestion
  * SSH
```

vmayoral, line 2178: Are you open to consider more communication channels? E.g. serial ports such as SPI or UART should be available both from the RPi. Similarly, OpenCR exposes communication ports.

vmayoral: On a second reading, there're several more entry points worth mentioning here @thomas-moulard. It may be worth considering them. There're some simple aspects such as unused communication ports which can be simply disabled by configuring the embedded computer (and the underlying OS) appropriately. Would you be open for contributions that would affect (slightly to include those columns below) the threat model below? @XabierPB and @o-olalde, would you be interested on this?

vmayoral:
```suggestion
*TurtleBot Threat Diagram - An attacker deploys a malicious node on the robot*
```

vmayoral: I think the last space wasn't needed.

vmayoral: Or maybe this is intended to be a title? If so, a different format would be required

vmayoral:
That was very nice reading @thomas-moulard! Thanks a lot for the contribution. Very, very nice.

A few further comments:
- I'd support @gavanderhoorn above about a references section. It'll help many of us get the references across the document into a single place for further reading. I'm happy drafting a first Pull Request if needed.
- I find the first part of the article particularly well suited for the title of the document however the Threat Analysis that follows (the TurtleBot 3 one) is only of of many we could (and I hope we) produce. May I suggest to bring that second part of the article to yet another article? I can envision having a dedicated section in http://design.ros2.org to "**Security**" (at the same level as "**Overview**" or "**Middleware**"). @jacobperron, @nuclearsandwich, @dirk-thomas, @wjwwood is this actually feasible? It will certainly boost the contributions on security IMHO.

articles/ros2_threat_model.md
vmayoral: Maybe I wasn't clear enough, sorry, I don't quite understand how to implement the countermeasure proposed and I'm arguing about whether simply discarding the trajectory would be safer. Does that make more sense @olaldiko?

articles/ros2_threat_model.md
tfoote:
```suggestion
 (Spoofing / Tampering / Repudiation / Integrity / Denial of service
```

articles/ros2_threat_model.md
gbiggs: That has safety implications. Simply discarding the trajectory may not be safe in all conditions or for all systems.

Monitoring every single command input (e.g. every trajectory) for dangerous ones is a common technique in safety.

thomas-moulard:
Thanks a lot everyone for the feedback - I'll integrate all of those points and push an update to this PR ASAP.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3291 @@
...
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td>
+      <ul>
+        <li>Hardened kernel (prevent dynamic loading of kernel modules)</li>
+        <li>Ensure only trustable kernels are used (e.g. Secure Boot)</li>
+        <li>/boot should not be accessible by robot processes</li>
+      </ul>
```
cratliff, line 856: Outside of the Kernel modules the robot is likely running an implementation of NTP to modify the RTC on the device.  A successful attack on NTP enables a number of other attacks. From enabling the use of expired keys, distortion of time data when used with sensors or other components in the system, or hampering coordination with remote components with correct time data.

Mitigations would include using NTP with encryption or using multiple NTP pools to detect modified times.

olaldiko: Agree with that, having no hardware RTC on the Raspberry pi supposses a problem, as no valid time exists at boot. Depending on the implementation, it stores the latest known valid time or goes straight to 0 epoch. I would suggest the addition of a dedicated hardware RTC as a mitigation as well. Adding limitations to the maximum time difference to be taken as valid from the NTP server would also help.

articles/ros2_threat_model.md
thomas-moulard: We don't explicitly distinguish RTPS from XRCE in this diagram (both are in blue). The main reason is that we don't consider (at this point) protocol-level attack.
We should improve on that as XRCE does not support in-transit data encryption (see https://www.slideshare.net/GerardoPardo/overview-of-the-ddsxrce-specification slide 35)

articles/ros2_threat_model.md
thomas-moulard: I agree this is a real concern. @vmayoral yes, adding a few rows in the doc would be great.

articles/ros2_threat_model.md
thomas-moulard: Yes, @gbiggs is correct. Coming from a humanoid robotic background, stopping a walking robot can be more dangerous than allowing a slight violation of a joint max velocity or effort. The same idea can apply to any robot where "velocity = 0" is not a safe state (drones, planes, self-driving cars on a highway).

articles/ros2_threat_model.md
thomas-moulard: @vmayoral My understanding is that enabling SROS and enforcing all nodes to be secure would mitigate this. Maybe adding somewhere "the DDS network should be kept isolated as much as possible" would be good?

By the way, we've been running TSan on ROS 2 and we detected potential deadlocks when the graph changes. So what you're highlighting here is a real concern, due to the way DDS discovery works, one could imagine that spamming a network with new DDS participants could put ROS 2 nodes into a dead-locked state (note that we didn't attempt this yet so we don't know how hard it would be to reproduce this issue that TSan detected).

thomas-moulard: Yes, that'd be great. As with your other suggestion, it may be better to get this merged and iterate on the doc, but I don't have a very strong preference there.

thomas-moulard:
@gavanderhoorn @vmayoral I added a references section - open for ideas there if you feel it could be improved.

I think splitting the doc could be a good idea, but it could also probably be done after this first pass.

All feedback is not addressed yet and I'll take care of that tomorrow.

ruffsl:
I think the tables are a bit too wide to reasonably read with one screen:

image

Perhaps we could collapse the columns for: 
* Threat Category (STRIDE)
* Threat Risk Assessment (DREAD)
* Impacted Assets
* Impacted Entry Points
* Impacted Business Goals 

And use a fixed width encoding of flags and values, e.g:

 | Threat Category | Threat Risk Assessment | ... |
|--------------------------|--------------------------------|-----|
| S✓ T✘ R✘ I✓ D✓ E✓        | D3 R1 E1 A2 D3 = 10                 | ... |
| ...                      | ..                             | ... |

gbiggs:
> I think the tables are a bit too wide to reasonably read with one screen

You should try reviewing the document on an iPad.

articles/ros2_threat_model.md
vmayoral: All right, that makes sense, however does the same rationale apply for this mobile robot application? It feels like in this particular case we may want to elaborate? I still believe I’d rather have the TB3 stopping than falling/hitting something.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+
+<div class="alert alert-info" markdown="1">
+This section is intentionally independent from ROS as robotic systems
+share common threats and potential vulnerabilities. For instance, this section
+describes "robotic components" while the next section will mention "ROS 2 nodes".
+</div>
+
+### Defining Robotic Systems Threats
+
+We will consider as a robotic system one or more general-purpose computer
```
kyrofa:
```suggestion
We will consider as a robotic system one or more general-purpose computers
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+This section defines actors, assets, business goals and entry points for this
+threat model.
+
+**Actors** are humans or external systems interacting with the robot. Considering
+which actors interact with the robot is helpful to determine how the system
+can be compromised. For instance, actors may be able to give commands to the
+robot which may be abused to attack the system.
+
+**Assets** represent any user, resource (e.g. disk space), or property (user physical
+safety) of the system that should be defended against attackers.
```
kyrofa: I suggest rephrasing "user physical safety" as "physical safety of users". I also suggest including "e.g." to follow the pattern established by disk space a few words earlier.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+
+**Entry points** represent how the system is interacting with the world (communication
+channels, API, sensors, etc.).
+
+#### Robot Application Actors
+
+Actors are divided into multiple categories based on whether or not they are
+physically present next to the robot (could the robot harm them?), are they
+human or not and (human only) are they a "power user" or not. A power user
```
kyrofa: The "human only" makes this just a tad awkward: my eye wants that to be an example similar to "could the robot harm them" mentioned a few words previous. Might I suggest removing "(human only)" and simply adding the qualification when you define power user? "A power user is defined to be a human who is knowledgeable..."

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+    <td>A human tele-operating the robot or sending commands to it through a
+        client application (e.g. smartphone app) </td>
+  </tr>
+  <tr>
+    <td>Cloud Developer</td>
+    <td class="danger">N</td>
+    <td class="success">Y</td>
+    <td class="success">Y</td>
+    <td>A developer buildling a cloud service connected to the robot or an
+        analyst who has been granted access to robot data.</td>
```
kyrofa, line 146: There may be two actors lumped into one here, which would be fine except that I question whether or not one is considered a super user. For example, this might be exactly the same as the "Robot Developer" but someone who happens to be remote, in which case, definitely a super user. However, an analyst who has been granted access to the data but nothing more feels like another actor that isn't a super user (they can't affect the behavior of the robot), but also isn't a normal end-user. The set of assets covered by each role is very different. Should the data analyst be an actor of its own?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+
+**Entry points** represent how the system is interacting with the world (communication
+channels, API, sensors, etc.).
+
+#### Robot Application Actors
+
+Actors are divided into multiple categories based on whether or not they are
+physically present next to the robot (could the robot harm them?), are they
+human or not and (human only) are they a "power user" or not. A power user
+is defined as someone who is knowledgeable and execute tasks which are normally
```
kyrofa, line 98:
```suggestion
is defined as someone who is knowledgeable and can execute tasks which are normally
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+(usually processes) running on multiple computers (trusted or non-trusted
+components) and data stores (embedded or in the cloud).
+
+While the computers may run well-controlled, trusted software (trusted
+components), other off-the-shelf robotics components (non-trusted) nodes may be
+included in the application. Third-party components may be malicious (extract
+private data, install a root-kit, etc.) or their QA validation process may
+not be as extensive as in-house software. Third-party components releasing
+process create additional security threats (third-party component may be
+compromised during their distribution).
```
kyrofa, line 279: This last sentence was difficult to parse. I also would say that the release process of in-house components could lead to their compromise as well, no? Perhaps a simple rephrase to "The release process for each component may create additional security threats (the component may be compromised during its distribution)."

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+
+A trusted robotic component is defined as a node developed, built, tested and
+deployed by the robotic application owner or vetted partners. As the process is
+owned end-to-end by a single organization, we can assume that the node will
+respect its specifications and will not, for instance, try to extract and leak
+private information. While carefully controlled engineering processes can
+reduce the risk of malicious behavior (accidentally or voluntarily), it cannot
+completely eliminate it. Trusted nodes can still leak private data, etc.
+
+Trusted node should not trust non-trusted nodes. It is likely that more than one
```
kyrofa:
```suggestion
Trusted nodes should not trust non-trusted nodes. It is likely that more than one
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+non-trusted component is embedded in any given robotic application. It is
+important for non-trusted components to not trust each other as one malicious
+non-trusted node may try to compromise another non-trusted node.
+
+An example of trusted component could be an in-house (or carefully vetted) IMU
```
kyrofa:
```suggestion
An example of a trusted component could be an in-house (or carefully vetted) IMU
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+driver node. This component may communicate through unsafe channels with other
+driver nodes to reduce sensor data fusion latency. Trusting components is never
+ideal but it may be acceptable if the software is well-controlled.
+
+On the opposite, an non-trusted node can be a third-party object tracker.
+Deploying this node without adequate sandboxing could impact:
+
+* User privacy: the node is streaming back user video without their consent
+* User safety: the robot is following the object detected by the tracker and its
+  speed is proportional to the object distance. The malicious tracker estimate
```
kyrofa:
```suggestion
  speed is proportional to the object distance. The malicious tracker estimates
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+robotic application.
+
+Threat categorization is based on the [STRIDE][wikipedia_stride]
+ (Spoofing / Tampering / Repudiation / Integrity / Denial of service
+ / Elevation of privileges) model.
+Risk assessment relies on [DREAD][wikipedia_dread] (Damage / Reproducibility /
+ Exploitability / Affected users / Discoverability).
+
+In the following table, the "Threat Category (STRIDE)" columns indicates
+which categories a threat belong too. If the "Spoofing" column is marked with
```
kyrofa: Typo here, might just consider a rephrase.
```suggestion
the categories to which a threat belongs. If the "Spoofing" column is marked with
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+The table below lists all *generic* threats which may impact a
+robotic application.
+
+Threat categorization is based on the [STRIDE][wikipedia_stride]
+ (Spoofing / Tampering / Repudiation / Integrity / Denial of service
+ / Elevation of privileges) model.
+Risk assessment relies on [DREAD][wikipedia_dread] (Damage / Reproducibility /
+ Exploitability / Affected users / Discoverability).
+
+In the following table, the "Threat Category (STRIDE)" columns indicates
```
kyrofa: Typo here
```suggestion
In the following table, the "Threat Category (STRIDE)" columns indicate
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+which categories a threat belong too. If the "Spoofing" column is marked with
+a check sign (✓), it means that this threat can be used to spoof a component
+of the system. If it cannot be used to spoof a component, a cross sign will
+be present instead (✘).
+
+The "Threat Risk Assessment (DREAD)" columns contain a score indicating how
+easy or likely it is for a particular threat to be exploited. The allowed
+score values are 1 (not at risk), 2 (may be at risk) or 3 (at risk, need to
```
kyrofa:
```suggestion
score values are 1 (not at risk), 2 (may be at risk) or 3 (at risk, needs to
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+be mitigated).  For instance, in the damage column a 1 would mean
+"exploitation of the threat would cause minimum damages", 2 "exploitation of
+the treat would cause significant damages" and 3 "exploitation of the treat
```
kyrofa:
```suggestion
the threat would cause significant damages" and 3 "exploitation of the threat
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+would cause massive damages". The "total score" is computed by adding the
+score of each column. The higher the score is, the most critical is the
```
kyrofa:
```suggestion
score of each column. The higher the score, the more critical the
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+threat.
+
+Impacted assets, entry points and business goals columns indicate whether
+an asset, entry point or business goal is impacted by a given threat.
+A check sign (✓) means impacted, a cross sign (✘) means not impacted.
+A triangle (▲) means "impacted indirectly or under certain conditions".
+For instance, compromising the robot kernel may not be enough to steal
+user data but it makes the stealing data much easier.
```
kyrofa:
```suggestion
user data but it makes stealing data much easier.
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td>
+      <ul>
+        <li>Components should authenticate themselves.</li>
+        <li>Components should not be attributed similar identifiers.</li>
+        <li>Components identifiers should be chosen carefully.</li>
```
kyrofa:
```suggestion
        <li>Component identifiers should be chosen carefully.</li>
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td>
+      <ul>
+        <li>Components should only be allowed to access channels they
+require.</li>
+        <li>Internet-facing channels and robot-only channels should be
+isolated.</li>
+        <li>Components behaviors should be reliant to a lost of communication
```
kyrofa:
```suggestion
        <li>Components behaviors should be tolerant of a loss of communication
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+      </ul>
+    </td>
+    <td><a href="http://arxiv.org/abs/1504.04339">Bonaci, Tamara, Jeffrey
+Herron, Tariq Yusuf, Junjie Yan, Tadayoshi Kohno, and Howard Jay Chizeck. “To
+Make a Robot Secure: An Experimental Analysis of Cyber Security Threats Against
+Teleoperated Surgical Robots.” ArXiv:1504.04339 [Cs], April 16, 2015.</a></td>
+  </tr>
+
+  <tr>
+    <td>An attacker intercepts robot long-range communications (e.g. MitM)</td>
```
kyrofa, line 692: The difference between "hijack" and "intercept" is that the latter is only listening, then? Aren't both of these consider MitM?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+    <td class="warning">▲</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td class="danger">✘</td>
+    <td class="success">✓</td>
+    <td class="danger">✘</td>
+    <td>
+      <ul>
+        <li>Multiple long-range communication transport layers should be used
+when possible (e.g. cellular and WiFi)</li>
```
kyrofa, line 763: And the system should be designed to "fail closed." If the system isn't being commanded to do something or detects that it's lost communication, it might be a good time to halt.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+      </ul>
+    </td>
+    <td><a href="https://doi.org/10.3390/s18051643">Ahmad Yousef, Khalil, Anas
+AlMajali, Salah Ghalyon, Waleed Dweik, and Bassam Mohd. “Analyzing
+Cyber-Physical Threats on Robotic Platforms.” Sensors 18, no. 5 (May 21,
+2018): 1643.</a></td>
+  </tr>
+
+  <tr>
+    <td>An attacker access configuration values without authorization.</td>
```
kyrofa:
```suggestion
    <td>An attacker accesses configuration values without authorization.</td>
```

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td class="success">✓</td>
+    <td>
+      <ul>
+        <li>Robot filesystem must be encrypted. The key should be stored in a
+secure enclave (TPM).</li>
+        <li>Robot filesystem should be wiped out if the robot perimeter is
+breached.</li>
```
kyrofa: What does this actually mean? What is the robot perimeter?

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+  <td class="success">✓</td>
+  <td class="danger">✘</td>
+  <td class="warning">▲</td>
+  <td class="success">✓</td>
+  <td>
+    <ul>
+      <li>Remote users should be granted minimum privileges• Credentials on
+desktop machines should be stored securely (secure enclave, TPM, etc.)</li>
+      <li>User credentials should be revokable or expire automatically• User
+credentials should be tied to the user identity for audit purposes</li>
```
kyrofa: Might want to clean up the •'s here.

articles/ros2_threat_model.md
```diff
@@ -0,0 +1,3421 @@
...
+    </ul>
+  </td>
+  <td> </td>
+  </tr>
+
+  <tr><th colspan="29">Remote / Coud Integration</th></tr>
```
kyrofa:
```suggestion
  <tr><th colspan="29">Remote / Cloud Integration</th></tr>
```

articles/ros2_threat_model.md
thomas-moulard: I was thinking of perimeter-intrusion in this sense: https://community.zymbit.com/t/using-perimeter-detect/204

Rephrased to make it easier to understand.

tfoote:
With the first round of comments resolved and the structure laid out. This has been marked as a draft. Please open new PRs to discuss specific areas of the document so we can have deeper threaded discussions on this document.

articles/ros2_threat_model.md
ryanewel: This is addressed in this PR https://github.com/ros2/design/pull/227#pullrequestreview-223379001 .

articles/ros2_threat_model.md
thomas-moulard: @cratliff @olaldiko I opened https://github.com/ros2/design/pull/229 as a follow-up PR to address this feedback. I will close the discussion here - your feedback on this PR would be greatly appreciated!

articles/ros2_threat_model.md
ryanewel: Some more detail on the DDS protocol used is added in this PR https://github.com/ros2/design/pull/231. This also addresses an earlier comment on attacks for the OpenCR board.

ros-discourse:

https://discourse.ros.org/t/ros-2-threat-model/8385/1

PR #219, opened 2019-03-25
Fix Reversed Serialization Pipeline Images

Parallel-Hao:
The two images are used in the ROS 2.0 Message Research page, which was reversed based on the pipeline description. This pull request renames the two image names to fix it.


wjwwood:
Section in question: https://github.com/ros2/design/blob/gh-pages/articles/serialization.md#possible-message-storage-and-serialization-process

wjwwood (approved):
Makes sense to me, thanks @Parallel-Hao

dirk-thomas (approved):
Thanks for fixing this.

PR #220, opened 2019-03-25
Add section on uniquely identifier interface definitions

jacobperron:
Follow up from https://github.com/ros2/rmw_opensplice/pull/263#discussion_r267590563

This is a proposal for uniquely identifying ROS interfaces that does not make any assumptions about the subfolder(s) where the interface definitions are found.

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -40,7 +40,7 @@ Both line comments (`//`) as well as block comments (`/* ... */`) are being supp
 
 #### 7.2.3 Identifiers
 
-An identifier must start with an is an ASCII alphabetic characteran followed by any number of ASCII alphabetic, digit and underscore (`_`) characters.
+An identifier must start with an ASCII alphabetic character followed by any number of ASCII alphabetic, digit and underscore (`_`) characters.
 
 #### 7.2.6 Literals
 
@@ -129,7 +129,7 @@ the DDS-XTypes specification 1.2 defines it with 16-bit.
 | ---------------------- | ---------------------------------------------------------- |
 | sequence\<type\_spec>  | sequence of items of the specific type\_spec               |
 |                        | the sequence is unbounded and no maximum size is specified |
-| sequence<type_spec, N> | sequence of of up to N items of the specified type\_spec   |
+| sequence<type_spec, N> | sequence of up to N items of the specified type\_spec      |
 |                        | the sequence is bounded and contain between 0 and N items  |
 
 #### 7.4.1.4.4.3.2 Strings
@@ -158,6 +158,60 @@ An enumerated type consist of an ordered list of enumerators.
 
 A multidimensional, fixed-size array is defined by the type of each item and the explicit sizes for each dimension.
 For now only a one-dimensional, fixed-size array is supported though.
+
+### Interface Identification
+
+#### Uniform Resource Name (URN)
+
+Following 7.5.1 from the IDL 4.2 specification, every IDL definition can be uniquely identified with a qualified name of the form
+**<scoped\_name>::<identifier>**.
+In addition to the specification, the scope of an IDL definition is prefixed with the package name from which the defintions is derived.
+This is to guarantee uniqueness of definitions across packages (under the assumption that package names are unique).
+Therefore, each interface defintion can be resolved with a URN of the form: **<package\_name>/<scoped\_name>/<identifier>**.
+
+#### Uniform Resource Locator (URL)
+
+For source code to use an interface (e.g. include or import) there needs to be a way to reference the file containing the defintion.
+To reference the location of an IDL defintion we use a URL of the form: **rosidl:<package\_name>/<subfolder>/<file>**.
+Where *<subfolder>* is the relative path from the root of the package where generated defintions can be found.
+
+#### Example
+
+Imagine we have a package `foo_pkg` containing an IDL file `Bar.idl` with the following contents:
+
+```idl
+struct MsgA {
+  boolean bool_member;
+};
+
+module bar {
+  const int8 MY_CONSTANT = 42;
+
+  struct MsgB {
+    int8 int_member;
+  };
+
+  module baz {
+    struct MsgC {
+      string string_member;
+    };
+  };
+};
+```
+
+Also imagine the message generation pipeline places the generated implementation to a subfolder `interface`.
+
+Then, we can reference the interfaces with the following URNs:
+
+- `foo_pkg/MsgA`
+- `foo_pkg/bar/MY_CONSTANT`
+- `foo_pkg/bar/MsgB`
+- `foo_pkg/bar/baz/MsgC`
+
+And all can be located with the URL: `rosidl:foo_pkg/interface/Bar`.
+
+For example, the definitions could be included in C++ with `#include <foo_pkg/interface/Bar.hpp>`
+or imported in Python with `import foo_pkg.interface.Bar`.
 
 ### Annotations
 
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -202,17 +202,17 @@ module bar {
 
 Also imagine the message generation pipeline places the generated implementation to a subfolder `interface`.
 
-Then, we can reference the interfaces with the following URNs:
+Then, we can reference the interfaces with the following URIs:
 
 - `rosidl:foo_pkg/MsgA`
 - `rosidl:foo_pkg/bar/MY_CONSTANT`
 - `rosidl:foo_pkg/bar/MsgB`
 - `rosidl:foo_pkg/bar/baz/MsgC`
 
-And all can be located with the URL: `package://foo_pkg/interface/Bar`.
+And all can be located with the URL: `package://foo_pkg/interface/bar`.
 
-For example, the definitions could be included in C++ with `#include <foo_pkg/interface/Bar.hpp>`
-or imported in Python with `import foo_pkg.interface.Bar`.
+For example, the definitions could be included in C++ with `#include <foo_pkg/interface/bar.hpp>`
+or imported in Python with `import foo_pkg.interface.bar`.
 
 ### Annotations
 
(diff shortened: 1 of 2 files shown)


articles/143_legacy_interface_definition.md
```diff
@@ -38,22 +38,30 @@ Each field is described by a *type* and a *name*.
...
 These triplets of data structures are called *actions*.
-An action is identified by its *name* and the *package* it is in.
 Each action describes three messages, one for the goal data structure, one for the result data structure, and one for the feedback data structure.
 
+### Identifying data structures
+
+Every data structure can be uniquely identified with three pieces of information:
+
+1. **Package** - the name of the package containing the data structure definition.
+2. **Subfolders** - the list of subfolders within the package where the data structure defintion can be found.
```
gbiggs: I'm not keen on needing to keep the internal file structure of my package constant forever just to avoid breaking a name.

gbiggs:
Can we do better than relying on file system layout to uniquely identify a message? What about using package name and a hash generated from the comment-less version of the message definition?

articles/143_legacy_interface_definition.md
dirk-thomas: The information is not internal. The subfolder is part of the namespace of the data structure.

dirk-thomas:
> What about using package name and a hash generated from the comment-less version of the message definition?

That doesn't seem practical if the user has to specify a message type on the command line.

articles/143_legacy_interface_definition.md
jacobperron: I think this can be reworded and perhaps subfolder is not the best word to use. What about "interface namespace"?

dirk-thomas: Maybe just `subnamespaces`?

jacobperron: Updated to use "subnamespaces" 11564de

gbiggs:
> That doesn't seem practical if the user has to specify a message type on the command line.

True. I was thinking only in terms of automated tools, but the usability aspect for the user/programmer is probably more important.

articles/143_legacy_interface_definition.md
gbiggs: You will need to specify how a subnamespace is defined now, which comes back to the use of folders, i.e. package structure, again.

Perhaps we need a way to specify the message definition root in a package and all folders below that are treated as name spaces?

Either that, or we state outright that best practice is that messages should live in a subfolder of the package called `msg`, etc. so that people are likely to get it right from the start and not need to change it. I wouldn't be opposed to that.

jacobperron: I did some experimenting. IIUC, as far as typesupport is concerned, it doesn't matter where you place message definitions in a package. For example, `msg/Foo.msg` and `some/other/subfolder/Foo.msg` will generate the same files under the `msg` namespace. Likewise for services and actions (I guess the namespace is determined by the file suffix, `.msg`, `.srv`, and `.action`?). @dirk-thomas Is my understanding correct?

If it has any consequence, I did note that the original definition files (not the `.idl` files) appear to be installed in a directory named after the parent directory. In my example above, I see `subfolder/Foo.msg` in the packages `share` directory (note, it is missing the parent directories `some/other/`). Maybe this is a bug/limitation of the current IDL pipeline.

jacobperron:
After an offline discussion, I've updated this PR attempting to address the problem of identifying interfaces in general (pure IDL).

articles/143_legacy_interface_definition.md
```diff
@@ -39,21 +39,52 @@ Each field is described by a *type* and a *name*.
...
 Together with the name of the *package* a message can be uniquely identified.
+A message has the URN: package/msg/name`.
 
 ### Services
 
 For request / reply style communication the two exchanged data structures are related.
 These pairs of data structures are called *services*.
-A service is identified by its *name* and the *package* it is in.
 Each service describes two messages, one for the request data structure, one for the reply data structure.
+Together with the name of the *package* a
```
dirk-thomas: This seems to end in the middle of the sentence?

articles/143_legacy_interface_definition.md
```diff
@@ -39,21 +39,52 @@ Each field is described by a *type* and a *name*.
...
+
+- URN: `<package_name>/srv/<name>_Request`
+- URN: `<package_name>/srv/<name>_Response`
+
+#### Actions
+
+- URN: `<package_name>/action/<name>`
+- URL: `<package_name>/action/<name>`
+
+The underlying message definitions that make up an action are located in the same file (ie. have the same URL) and are in the `action` namespace:
```
dirk-thomas: Message and service definitions?

dirk-thomas (commented):
Considering a command line tool like `ros2 msg` I would expect it to take a URN as an argument. The problem I see is that it doesn't know what the correct URL is in order to import the module containing the type.

articles/143_legacy_interface_definition.md
jacobperron: Yeah, this is vestigial. Removed in 30a99ff

articles/143_legacy_interface_definition.md
jacobperron: b9a2826

jacobperron:
> Considering a command line tool like ros2 msg I would expect it to take a URN as an argument. The problem I see is that it doesn't know what the correct URL is in order to import the module containing the type.

Is it possible for typesupport to provide the URL? I.e. provide a mapping from a types URN to the URL.

dirk-thomas:
> Is it possible for typesupport to provide the URL? I.e. provide a mapping from a types URN to the URL.

Maybe that information could be put into the resource index. Currently each URN has a marker file in the resource index. Instead of using an empty file it could contain the URL.

jacobperron:
> Maybe that information could be put into the resource index. Currently each URN has a marker file in the resource index. Instead of using an empty file it could contain the URL.

That might work. But is it the case that there is a marker file for each URN? It looks like there is a single marker file for the package containing relative paths to the resource files (not necessarily URNs). For example, there's a file `install/test_msgs/share/ament_index/resource_index/rosidl_interfaces/test_msgs` containing:

<details><summary>content</summary>
<pre>
<code>
action/Fibonacci.action
action/Fibonacci.idl
action/NestedMessage.action
action/NestedMessage.idl
msg/BoundedArrayNested.idl
msg/BoundedArrayNested.msg
msg/BoundedArrayPrimitives.idl
msg/BoundedArrayPrimitives.msg
msg/BoundedArrayPrimitivesNested.idl
msg/BoundedArrayPrimitivesNested.msg
msg/Builtins.idl
msg/Builtins.msg
msg/DynamicArrayNested.idl
msg/DynamicArrayNested.msg
msg/DynamicArrayPrimitives.idl
msg/DynamicArrayPrimitives.msg
msg/DynamicArrayPrimitivesNested.idl
msg/DynamicArrayPrimitivesNested.msg
msg/DynamicArrayStaticArrayPrimitivesNested.idl
msg/DynamicArrayStaticArrayPrimitivesNested.msg
msg/Empty.idl
msg/Empty.msg
msg/Nested.idl
msg/Nested.msg
msg/Primitives.idl
msg/Primitives.msg
msg/StaticArrayNested.idl
msg/StaticArrayNested.msg
msg/StaticArrayPrimitives.idl
msg/StaticArrayPrimitives.msg
msg/StaticArrayPrimitivesNested.idl
msg/StaticArrayPrimitivesNested.msg
srv/Empty.idl
srv/Empty.srv
srv/Empty_Request.msg
srv/Empty_Response.msg
srv/Primitives.idl
srv/Primitives.srv
srv/Primitives_Request.msg
srv/Primitives_Response.msg
</code>
</pre>
</details>

articles/142_idl.md
```diff
@@ -159,6 +159,60 @@ An enumerated type consist of an ordered list of enumerators.
 A multidimensional, fixed-size array is defined by the type of each item and the explicit sizes for each dimension.
 For now only a one-dimensional, fixed-size array is supported though.
 
+### Interface Identification
+
+#### Uniform Resource Name (URN)
```
sloretz: This name could be confusing since it doesn't appear to follow [RFC 8141 - Uniform Resource names (URNs)](https://tools.ietf.org/html/rfc8141)

articles/142_idl.md
```diff
@@ -159,6 +159,60 @@ An enumerated type consist of an ordered list of enumerators.
...
+Following 7.5.1 from the IDL 4.2 specification, every IDL definition can be uniquely identified with a qualified name of the form
+**\<scoped\_name>::\<identifier>**.
+In addition to the specification, the scope of an IDL definition is prefixed with the package name from which the defintion is derived.
+This is to guarantee uniqueness of definitions across packages (under the assumption that package names are unique).
+Therefore, each interface defintion can be resolved with a URN of the form: **\<package\_name>/\<scoped\_name>/\<identifier>**.
+
+#### Uniform Resource Locator (URL)
+
+For source code to use an interface (e.g. include or import) there needs to be a way to reference the file containing the defintion.
+To reference the location of an IDL defintion we use a URL of the form: **rosidl:<package\_name>/<subfolder>/<file>**.
```
sloretz: Similar to the above comment, if using the name URL, it may make sense to follow [RFC 3986](https://tools.ietf.org/html/rfc3986). Then, `<package_name>` seems like an `authority` and `<subfolder>/<file>` seems like a `path` so this applies

```
The authority component is preceded by a double slash ("//") and is
   terminated by the next slash ("/"), question mark ("?"), or number
   sign ("#") character, or by the end of the URI.
```

meaning this should be `rosidl://<package_name>/<subfolder>/<file>`

sloretz: Instead of `rosidl`, ROS 1 `resource_retriever` used the scheme `package`. http://wiki.ros.org/resource_retriever/Tutorials/Retrieving%20Files , so that could be considered for compatibility.

articles/142_idl.md
jacobperron: Right, the syntax could be `urn:package_name:scoped/name/identifier` for a proper URN.
Or we could drop the `urn:` business and declare this as a more general URI, e.g. `rosidl:package_name/scoped/name/identifier`.

What do you think?

articles/142_idl.md
jacobperron: Updated 53ff8b2

articles/142_idl.md
jacobperron: Updated with the latter suggestion 53ff8b2

PR #221, opened 2019-03-26
char now maps to int in python

mikaelarguedas:
Submitting this documentation fix to match behavior observed using the nightly archive.
It may be that the code is not yet behaving like the design doc, in which case feel free to close this PR.

---
Additional information:

How to reproduce the issue:
```
>>> from test_msgs.msg import Primitives
>>> a = Primitives()
>>> a.char_value = 'a'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/ros/crystal/lib/python3.6/site-packages/test_msgs/msg/_primitives.py", line 186, in char_value
    "The 'char_value' field must of type 'int'"
AssertionError: The 'char_value' field must of type 'int'
```
The same code works on ROS 2 Crystal

The generated IDL file says that the field `char_values` is `uint8` and not `char`:
```
module test_msgs {
  module msg {
    struct Primitives {
      boolean bool_value;

      octet byte_value;

      uint8 char_value;

      float float32_value;
```

### DIFF
--- a/articles/142_idl.md
+++ b/articles/142_idl.md
@@ -191,7 +191,7 @@ The next table defines how IDL types are mapped to the following programming lan
 | float       | float            | float                          | float                |
 | double      | double           | double                         | float                |
 | long double | long double      | long double[<sup>2</sup>](#ld) | float                |
-| char        | unsigned char    | unsigned char                  | str with length 1    |
+| char        | unsigned char    | unsigned char                  | int                  |
 | wchar       | char16\_t        | char16\_t                      | str with length 1    |
 | boolean     | \_Bool           | bool                           | bool                 |
 | octet       | unsigned char    | std::byte[<sup>1</sup>](#byte) | bytes with length 1  |


articles/142_idl.md
```diff
@@ -191,7 +191,7 @@ The next table defines how IDL types are mapped to the following programming lan
 | float       | float            | float                          | float                |
 | double      | double           | double                         | float                |
 | long double | long double      | long double[<sup>2</sup>](#ld) | float                |
-| char        | unsigned char    | unsigned char                  | str with length 1    |
```
dirk-thomas, line 194: The mapping documented here is correct. A `char` in `IDL` is mapped to a `str` with length 1 in Python.

The mapping from the `char` type in `.msg` files is described in [143_legacy_interface_definition](https://github.com/ros2/design/blob/ae61bd9738c0896e97ff4768ffd7b898435f2ff0/articles/143_legacy_interface_definition.md) instead.

mikaelarguedas: thanks for clarifying, to summarize:
- the `char` type mapping for IDL is the same as the crystal one for msg
- `char` in msg for crystal will be `int` and not `str`
- the design document that needs to be updated is the one defining the python mapping for msg

Is that correct?

See https://github.com/ros2/design/pull/222 for a second attempt

dirk-thomas: In Crystal and before the `char` type in `.msg` was mapped to a single character (`str` with length 1 in Python). That mapping was different from how it is specified in ROS 1 and therefore as of Dashing the mapping as it was in ROS 1 has been restored which maps a `char` in `.msg` to an `uint8`. See #213.

mikaelarguedas: Thanks for the answer :+1: 

My question was mostly find a place to share the information with users (familiar or not with ROS1 mapping) that may face the same issue.
It seems that these documents are not the right place. Where do you think it would be appropriate? maybe on a Dashing migration page? or an answered question of ROS Answers?

dirk-thomas: The dashing release page would be the right place. It already contains a migration section where this information would fit in: https://index.ros.org/doc/ros2/Releases/Release-Dashing-Diademata/#changes-since-the-crystal-release

mikaelarguedas: :+1: thanks, submitting a PR there then

dirk-thomas: Already done: ros2/ros2_documentation#149

mikaelarguedas: :+1: 
closing https://github.com/ros2/ros2_documentation/pull/148 then

dirk-thomas: Sorry for the duplicate effort. Thanks for following up on this.

PR #222, opened 2019-03-26
legacy char now maps to int in python

mikaelarguedas:
redo of https://github.com/ros2/design/pull/221 but targeting the legacy design doc for `msg`.

### DIFF
--- a/articles/114_generated_interfaces_python.md
+++ b/articles/114_generated_interfaces_python.md
@@ -59,7 +59,7 @@ The Python module `<package_name>.msg` / `<package_name>.srv` exports all messag
 | -------- | ---------------------------- |
 | bool     | builtins.bool                |
 | byte     | builtins.bytes with length 1 |
-| char     | builtins.str with length 1   |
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This article starts with:

> With the transition to use ``IDL`` for specifying interfaces in ROS 2 Dashing this article has been superseded by ...

So I am not sure it makes sense to update it.

mikaelarguedas:
I understand your point.
What prompted this initially is that crystal code setting fields of type `char` in messages is now failing to run on master, and I was not able to find resources explaining that it was now mapping to an `int`.

Is there a place discussing the support (or discontinued support) of `msg` in Dashing and what the generated datastructures will be like for people using `msg` files?

dirk-thomas:
Please see https://github.com/ros2/design/pull/221#discussion_r269296677. The change basically restores compatibility to how it is being mapped in ROS 1.

mikaelarguedas:
Thanks for the answer, I replied in #221 . Closing this invalid PR then

PR #223, opened 2019-03-26
escape < to fix rendering

mikaelarguedas:
Reading this document on design.ros2.org. It appears several lines did not render properly. e.g.
`sequence<T>` renders as `sequence`


dirk-thomas:
Thanks for the patch.

PR #226, opened 2019-03-30
Update action goal state diagram

jacobperron:
* Change transition labels to match text in design doc.
* Renamed cancel_goal transition to request_cancel.
  This makes it more clear from the name that it is a *request* to cancel and not the actual canceling of a goal.

Related to changes happening for ros2/rcl#399.


articles/actions.md
```diff
@@ -196,7 +196,7 @@ State transitions triggered by the action client:
 
 - **send_goal** - A goal is sent to the action server.
 The state machine is only started if the action server *accepts* the goal.
-- **cancel_goal** - Request that the action server stop processing the goal.
+- **request_cancel** - Request that the action server stop processing the goal.
```
hidmic: @jacobperron should the section below on [`Cancel goal service`](https://github.com/ros2/design/pull/226/files?short_path=d9d51a6#diff-0890511cd0323c94976368b725d6ada2R234) be updated too?

hidmic (commented):
LGTM! Bummer we can't keep `cancel_goal`, it was more consistent with `send_goal`.

jacobperron:
> Bummer we can't keep cancel_goal, it was more consistent with send_goal

We *could* keep it, although, IMO, "request_cancel" is more clear.

articles/actions.md
jacobperron: Yeah, if we move forward with this renaming then that makes sense. We should probably also update the [diagram](http://design.ros2.org/articles/actions.html#middleware-implementation) above it too.

dirk-thomas:
I am not sure the "request" in the name is better. With any service your request is just that - a request - and the response could be "no, not going to do it". Therefore I think `cancel_goal` was the better name and doesn't need to be changed.

jacobperron:
> I am not sure the "request" in the name is better. With any service your request is just that - a request - and the response could be "no, not going to do it". Therefore I think cancel_goal was the better name and doesn't need to be changed.

We should distinguish the cancel request event from the event where the goal is actually canceled (see https://github.com/ros2/rcl/issues/399#issuecomment-478193156). I think the events "cancel_goal" and "cancel" are more ambiguous than "request_cancel" and "cancel". In any case, it's probably not helping that we're conflating the user request to cancel and the goal state transition.

dirk-thomas:
> We should distinguish the cancel request event from the event where the goal is actually canceled (see ros2/rcl#399 (comment)). I think the events "cancel_goal" and "cancel" are more ambiguous than "request_cancel" and "cancel". In any case, it's probably not helping that we're conflating the user request to cancel and the goal state transition.

From the wording "cancel" alone it is again not clear if that is just a request, it is about to happen or it is already done.

I would argue that the natural language already has a clean way to distinguish these two cases:

* `cancel_goal`: the verb expresses what you want to do with the subject - to cancel a specific goal
* `goal_cancelled` or just `cancelled` describes that has happens to a specific goal - that it has been cancelled

jacobperron:
I've reverted the change to the "cancel_goal" transition and renamed the "cancel" transition to "canceled" as per @dirk-thomas's suggestion. I've updated the state diagram accordingly.

lbegani:
> I've reverted the change to the "cancel_goal" transition and renamed the "cancel" transition to "canceled" as per @dirk-thomas's suggestion. I've updated the state diagram accordingly.

Typo: Should it not be `cancelled` instead of `canceled`

clalancette:
> Typo: Should it not be `cancelled` instead of `canceled`

It depends on which flavor of English you learned: https://www.grammarly.com/blog/canceled-vs-cancelled/

dirk-thomas (approved):

hidmic (approved):
LGTM too!

hidmic:
Though if I get really picky, I find the use of `canceled`, an adjective, along with `abort` or `succeed`, both verbs, a bit inconsistent. But I can't think of a better naming right now.

dirk-thomas:
> I find the use of canceled, an adjective, along with abort or succeed, both verbs, a bit inconsistent.

Could be changed to `succeeded` / `aborted`?

hidmic:
> Could be changed to succeeded / aborted?

I'm somewhat inclined towards verbs, but yeah, that's the alternative. Anyways, it isn't super important.

dirk-thomas:
> I'm somewhat inclined towards verbs

Using verbs to describe a status seems off to me.

hidmic:
> Using verbs to describe a status seems off to me.

Agreed, but IIUC it's not the status or state but the event that we're describing here.

jacobperron:
> Agreed, but IIUC it's not the status or state but the event that we're describing here.

Correct. These are the names of the transitions between states. I believe the word "canceled" can be used as an adjective and as a verb.

PR #231, opened 2019-04-09
adds mitigations for USB communications in TurtleBot3 model

ryanewel:
- Distinguishes between DDS-XRCE and DDS-RTPS communication channels
- Adds mitigation concerns coming from lack of default security for DDS-XRCE communication highlighted by these two comments: https://github.com/ros2/design/pull/218#issuecomment-474743786 https://github.com/ros2/design/pull/218#discussion_r268043153

Signed-off-by: Ryan Newell <ryanewel@amazon.com>


articles/ros2_threat_model.md
```diff
@@ -2492,6 +2495,7 @@ enable_discovery_protection to True to prevent malicious actors from
 enumerating and fingerprinting DDS participants.</li>
         <li>DDS Security Governance document must enable_liveliness_protection
 to True</li>
+        <li>XRCE agent to LIDAR should use secure UDP protocol like DTLS</li>
```
thomas-moulard: Is this actually do-able as of today? Or is this a suggestion for the spec?

articles/ros2_threat_model.md
```diff
@@ -2097,11 +2096,13 @@ computer.
...
-  on the [OpenCR 1.0 board][opencr_1_0], that publishes  IMU sensor data to ROS
+  on the [OpenCR 1.0 board][opencr_1_0], that publishes IMU sensor data to ROS
   topics, and controls the wheels of the TurtleBot based on the
-  [teleoperation][tb3_teleop] messages published as ROS topics.
+  [teleoperation][tb3_teleop] messages published as ROS topics. This channel uses
+  serial communication.
 * A Lidar driver process running on the Raspberry interfaces with the Lidar, and
-  uses a DDS-XRCE client to publish the sensor data to several ROS topics.
+  uses a DDS-XRCE client to publish the sensor data to several ROS topics. This
+  channel uses udp.
```
thomas-moulard: udp -> UDP

thomas-moulard: Also, I'm not clear on how it works, does this node actually use UDP over the wire or is just between the lidar driver and the XRCE agent (on localhost)?

articles/ros2_threat_model.md
```diff
@@ -2381,8 +2384,8 @@ through a network connection to e.g. stop the robot.</td>
       <ul>
         <li>Enable SROS / DDS Security Extension to authenticate and encrypt
         DDS communications. Message tampering is mitigated by DDS security as
-	message authenticity is verified by default (with preshared
-	HMACs / digital signatures)</li>
+  message authenticity is verified by default (with preshared
```
thomas-moulard: indent?

thomas-moulard (commented):
Mainly LGTM, but I'ld like to confirm two points:

1. Is TB3 actually using UDP over the wire?
2. Is XRCE supporting DTLS as of today?

articles/ros2_threat_model.md
ryanewel: This is a suggestion for the spec, as far as I can tell only there are only implementations for UDP, TCP, and serial transport layers under DDS-XRCE. I've tried to clarify that in the doc and given another suggestion for securing the UDP communication. Looking at the LIDAR source code for the turtlebot (https://github.com/ROBOTIS-GIT/turtlebot3/blob/ros2/turtlebot3_lidar/turtlebot3_lidar/main.cpp) it's just communicating with localhost so I'm not sure what general network security recommendations are applicable here.

articles/ros2_threat_model.md
ryanewel: added

articles/ros2_threat_model.md
ryanewel: You're right, there is a drive on the raspberry pi that interacts with the LIDAR sensor. The XRCE agent and client are both running on the raspberry pi. I'll try to clarify that

articles/ros2_threat_model.md
thomas-moulard: The main problem I see with that is that this doc needs to highlight actionable steps and we cannot modify the spec. What about recommending instead:
1/ never use XRCE for communicating over an insecure transport layer (ethernet unless one is using e.g. a VPN tunnel).
2/ If communication is only done locally, it should be done over the loopback interface.

articles/ros2_threat_model.md
```diff
@@ -2496,8 +2499,29 @@ to True</li>
...
+      <td class="success">✘</td>
+      <td class="success">✘</td>
+      <td class="danger">✓</td>
+      <td>TurtleBot LIDAR measurements are saved to a remote location controlled by
+  the attacker. </td>
+      <td> Limiting access to the UDP ports used by the XRCE agent and client to
+        a specific user running the agent and client processes
+      </td>
+      <td class="danger">A much better solution would be to use an encrypted
+      transport layer, however that would require significantly more effort to implement.
```
thomas-moulard: and overhead, VPN would be a solution, but you probably don't want a VPN client running on your micro-controller :)

thomas-moulard: Most likely, the story is: prefer XRCE for local communication only and serial-over-USB.

articles/ros2_threat_model.md
```diff
@@ -2496,8 +2499,29 @@ to True</li>
     </td>
     <td class="warning">Risk is mitigated if DDS-Security is configured
 appropriately.</td>
-    <td> </td>
-  </tr>
+    <td>As of writing, XRCE only has implementations for UDP, TCP, and serial
```
thomas-moulard: As mentioned in the other comment, that's not really actionable as we can't really modify the spec. If you start fighting against the spec, it just may not be the right tool for you.

articles/ros2_threat_model.md
```diff
@@ -2496,8 +2499,29 @@ to True</li>
...
+      <td class="danger">✓</td>
+      <td class="danger">✓</td>
+      <td class="danger">✓</td>
+      <td class="success">✘</td>
+      <td class="success">✘</td>
+      <td class="danger">✓</td>
+      <td>TurtleBot LIDAR measurements are saved to a remote location controlled by
+  the attacker. </td>
+      <td> Limiting access to the UDP ports used by the XRCE agent and client to
+        a specific user running the agent and client processes
```
thomas-moulard: How would you do that? cgroups? or set a port under 1024, run from root and fork?

articles/ros2_threat_model.md
ryanewel: The XRCE spec is setup (and recommends) using secure transport layers, but there doesn't appear to be an implementation for these yet (https://micro-xrce-dds.readthedocs.io/en/latest/client.html). But I agree it's not immediately actionable so I removed this recommendation.

thomas-moulard: Oh I see - thanks for looking into this. May be worth it to add exactly this comment in the doc too. This is super useful info to whoever tries to design a new robot system

articles/ros2_threat_model.md
ryanewel: For the Turtlebot specific threats I've removed mitigations that aren't immediately actionable. I think these recommendations are already covered in the more general ROS2 model with mitigations like "communicating over an encrypted channel"

articles/ros2_threat_model.md
ryanewel: Good point, I've removed this part.

articles/ros2_threat_model.md
ryanewel: I wasn't really sure on the details here. I've changed this to mention that the loopback interface should be sufficient

thomas-moulard (approved):
:ship:

PR #233, opened 2019-04-19
Design for new build tools for ROS 2 package cross-compilation

juanrh:
Design doc for task ros2/ros2#693

Signed-off-by: Juan Rodriguez Hortala <hortala@amazon.com>

To be reviewed before 2019-04-26

### DIFF
--- a/articles/cross_compilation_build_tools_design.md
+++ b/articles/cross_compilation_build_tools_design.md
@@ -47,7 +47,7 @@ We propose adding a set of commands to wrap the instructions on the
 cross-compilation tutorial.
 Those commands would use a new workspace
 directory for the target platform, determined by a convention based on
-on the target ECU, operating system, rmw implementation, and ROS distribution.
+the target ECU, operating system, rmw implementation, and ROS distribution.
 Let us call that directory the _cc-root_.
 Under that convention, the ECU
 implies the architecture, but generic ECUs like `generic_armhf` would also be
@@ -67,9 +67,9 @@ cc-setup-sysroot --arch generic_armhf --os ubuntu_bionic \
 ```
 
 - A new colcon mixin for each known platform, which adds options to the colcon
-  build task for using asysroot generated using `cc-setup-sysroot`, by using
+  build task for using a sysroot generated with `cc-setup-sysroot`, by using
   the same path conventions.
-  For example from a ROS 2 overlay workspace on a
+  For example, from a ROS 2 overlay workspace on a
   developer workstation, the following command would cross-compile the packages
   in the workspace up to a package `performance_test` for the platform
   specified.
@@ -94,30 +94,26 @@ colcon build --mixin cc-generic_armhf-ubuntu_bionic-fastrtps-crystal \
 
 Under the hood, the `cc-setup-sysroot` command would do the following:
 
-- The command downloads a base ROS 2 Docker image for the target platform, and
-  builds a workspace dependent sysroot image by
-  running `COPY` to get the contents of the workspace into the container, and
-  running `rosdep` on the copy.
-  That ensures no workspace system dependency is
-  missing.
-  Note we can run `rosdep` in Docker for platform like ARM-HF thanks
-  to [QEMU](https://www.qemu.org/).
-  That image's file system is then exported by
-  launching a container running `docker container export` on it, to a
-  `sysroot` subdirectory of the cc-root, so it can be used as a sysroot during
-  the build.
-  - Those ROS 2 base images are variants of
-    [sysroot/Dockerfile_ubuntu_arm](https://github.com/ros2/cross_compile/blob/master/sysroot/Dockerfile_ubuntu_arm)
-  from ros2/cross_compile.
-  - OSRF would publish those base ROS 2 images, with ROS 2 pre built from
-    source.
-    OSRF would also publish other base ROS 2 images that install
-    `ros-$ROS_DISTRO-desktop` with Apt, instead of installing the system
-    dependencies for ROS 2 with `rosdep install`.
-- The colcon mixin then sets the variable `SYSROOT` to the sysroot path,
-  and other variables required to configure the toolchain file [generic_linux.cmake](https://github.com/ros2/cross_compile/blob/master/cmake-toolchains/generic_linux.cmake)
-  from ros2/cross_compile, and configures `DCMAKE_TOOLCHAIN_FILE` in
-  `cmake-args` pointing to that toolchain file.
+1. Downloads a base ROS 2 Docker image for the target platform.
+1. Builds a workspace dependent sysroot image by using a Dockerfile that starts
+   from that base image and:
+   1. Runs `COPY` to get the contents of the workspace into the container.
+   1. Uses `rosdep` to ensure no workspace system dependency is missing.
+      Note we can run `rosdep` in Docker for platforms like ARM-HF thanks
+      to [QEMU](https://www.qemu.org/).
+1. Launches a container for that image and exports its file system into a
+   `sysroot` subdirectory of the cc-root.
+1. During a build, the colcon mixin sets the variable `SYSROOT` to the sysroot path,
+   as well as any other variables required to configure the toolchain file [generic_linux.cmake](https://github.com/ros2/cross_compile/blob/master/cmake-toolchains/generic_linux.cmake).
+   It also sets `DCMAKE_TOOLCHAIN_FILE` through `cmake-args`, so it points to that toolchain file.
+
+The ROS 2 base images are variants of [sysroot/Dockerfile_ubuntu_arm](https://github.com/ros2/cross_compile/blob/master/sysroot/Dockerfile_ubuntu_arm)
+from ros2/cross_compile.
+OSRF would publish base ROS 2 images with ROS 2 prebuilt, that should be used
+for cros compiling workspaces with user packages.
+OSRF would also publish other base ROS 2 images with the system setup and
+basic tools for building ROS 2 preinstalled, that should be used for building ROS
+2 from source.
 
 The `cc-setup-sysroot` command will also support the following optional arguments:
 
@@ -166,26 +162,32 @@ we would need to:
   from the pre-built binary if available for that platform.
 - Adapt the toolchain file for the new platform, if required.
   For that the toolchains in [ruslo/polly](https://github.com/ruslo/polly) might be useful.
-- Add a new `cross-compile*` colcon mixin to for the new platform, add perform
+- Add a new `cross-compile*` colcon mixin to for the new platform and perform
   any modifications needed to use a suitable toolchain file.
 
-We could then test cross-compile support for the new platform by building the
-base ROS 2 Docker image, and a custom sysroot image for some reference package
-(e.g. [ApexAI/performance_test](https://github.com/ApexAI/performance_test/)),
-and then launching a Docker container for the that sysroot image, running
-`colcon test` using the cross-compiled binaries.
-The container would be launched with a [bind mount](https://docs.docker.com/storage/bind-mounts/)
-of the workspace so they have access to the compiled binaries.
+We can then test cross-compile support for the new platform by:
+
+1. Building the base ROS 2 Docker image.
+1. Using `cc-setup-sysroot` to build a custom sysroot image for some reference package,
+   e.g. [ApexAI/performance_test](https://github.com/ApexAI/performance_test/), and export its
+   file system into a sysroot.
+1. Cross-compiling that package using the sysroot.
+1. Launching a Docker container for that sysroot image and running
+   `colcon test` using the cross-compiled binaries.
+   The container would be launched with a [bind mount](https://docs.docker.com/storage/bind-mounts/)
+   of the workspace so they have access to the compiled binaries.
 
 In order to simplify the development of base ROS 2 Docker images for new
 architecture-OS combinations, on a future iteration of this initiative we might add
-a new command `build-sysroot-base-image` that will 1) copy the required
-QEMU files and ROS 2 source into a temporary file; 2) build a Docker image from
-a specified Dockerfile, with access to the those resources; 3) optionally
-publish the image into a Docker registry, using a suitable naming convention
-that corresponds to the platform identifier.
-Dockerfiles compatible with this command would be keep on the `ros2/cross_compile` repository,
-so they can be used in CI/CD pipeline if needed.
+a new command `build-sysroot-base-image` that would:
+
+1. Copy the required QEMU files and ROS 2 source files into a temporary location.
+1. Build a Docker image from a specified Dockerfile, with access to the those resources.
+1. Optionally publish the image into a Docker registry, using a suitable naming convention
+   that corresponds to the platform identifier.
+
+Dockerfiles compatible with that command would be kept on the `ros2/cross_compile` repository,
+so they can be used in CI/CD pipelines if needed.
 
 ### Advantages of the approach
 
@@ -210,7 +212,7 @@ build setup, running the build commands inside a Docker container, although
 that would imply some performance degradation due to the virtualization layer
 that Docker employs on those platforms.
 
-We should determine a process for building an publishing the base ROS 2 Docker
+We should determine a process for building and publishing the base ROS 2 Docker
 images for each platform.
 Ideally a nightly job would publish an image built
 from source for the tip of master, but that might be not feasible depending on
--- a/articles/cross_compilation_build_tools_design.md
+++ b/articles/cross_compilation_build_tools_design.md
@@ -2,13 +2,13 @@
 layout: default
 title: Cross-compiling ROS 2 packages
 abstract:
-  This article is a design proposal for extending the build tools for ROS 2 to simplify 
-  cross-compiling ROS 2 packages for platforms currently not supported by ROS 2, or 
-  building on a host with different platforms than the target platform. That later 
-  use case is particularly useful for speeding up build times by building the code on
-  a powerful development workstation, and then deploying the binaries to a constrained 
-  resources hardware that uses the target platform, thus making the development cycle
-  much faster.
+  This article is a design proposal for extending the build tools for ROS 2 to
+  simplify cross-compiling ROS 2 packages for platforms currently not supported
+  by ROS 2, or building on a host with different platforms than the target
+  platform. That later use case is particularly useful for speeding up build
+  times by building the code on a powerful development workstation, and then
+  deploying the binaries to a constrained resources hardware that uses the
+  target platform, thus making the development cycle much faster.
 author: >
   [Thomas Moulard](https://github.com/thomas-moulard),
   [Juan Rodriguez Hortala](https://github.com/juanrh)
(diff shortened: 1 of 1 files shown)


articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
+---
+layout: default
+title: Cross-compiling ROS 2 packages
+abstract:
+  This article is a design proposal for extending the build tools for ROS 2 to
+  simplify cross-compiling ROS 2 packages for platforms currently not supported
+  by ROS 2, or building on a host with different platforms than the target
+  platform. That later use case is particularly useful for speeding up build
```
esteve: "latter" is the proper spelling here.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Proposed Approach
+
+We propose continuing with the general approach of the cross-compilation
+tutorial, based on building a sysroot for the target platform using qemu and
+Docker, and then using `cmake-toolchains` with a toolchain file pointing
+`CMAKE_SYSROOT` to the sysroot location to compile with C and C++
+cross-compilers.
+
+On top of that we propose using a **colcon plugin `cc-build`** to wrap the
```
esteve: Instead of adding a new verb, wouldn't it more appropriate to add this logic to the CMake extension for colcon? `cc-build` seems to assume that all packages that are to be crosscompiled use CMake.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Background
+
+The [cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation)
```
esteve, line 26: Just a minor note, both https://github.com/esteve/ros2_java and https://github.com/esteve/ros2_objc predate the instructions that ARM wrote up. https://discourse.ros.org/t/crosscompiling-ros2-for-raspberry-pi-or-pretty-much-any-other-arm-based-board/2995 used the same technique as ARM's instructions (Docker + QEmu), but for armhf as well.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+contains instructions for cross-compiling ROS 2 to unsupported platforms, using
+some configuration files for CMake and Docker that is setting maintained in the
+[ros2/cross_compile repository](https://github.com/ros2/cross_compile/).
+However, that tutorial doesn't provide instructions to build other ROS 2
+packages, but just for building a whole ROS 2 distribution. Ideally, we should
+be able to cross-compile a ROS package for ROS 2 by launching a single
+`colcon build` command. Finally, cross-compiling for other platforms besides
+ARM64 is not supported in the [ros2/cross_compile repository](https://github.com/ros2/cross_compile/).
+
```
esteve, line 37: Worth mentioning is https://github.com/ruslo/polly, which contains a bunch of toolchains for crosscompiling CMake projects for a variety of platforms.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+### Advantages of the approach
+
+The proposed design leads to a simple development experience, where cross
+compilation is triggered by a one-liner command.
+The design is extensible and new platforms can be supported easily.
+
+For simple packages with additional dependencies, there is no need to build the
+sysroot locally with Docker, and a readily available base Docker image can be
+used. For packages with custom dependencies, the base sysroot can be extended
+using `rosdep install`.
```
esteve: I'd clarify that `rosdep install` here needs to be run inside QEmu.

esteve:
@filiperinaldi @lmayencourt @sgermanserrano thought you might be interested in this as well.

articles/cross_compilation_build_tools_design.md
thomas-moulard: The tricky part is the sysroot management which is based on Docker. I think it's too brutal to directly pull a dependency on Docker in all cases when one is using `colcon`. Being in a plugin is a convenient way to get users to "opt-in".

articles/cross_compilation_build_tools_design.md
thomas-moulard: Sounds awesome, thanks!

articles/cross_compilation_build_tools_design.md
thomas-moulard: (through `binfmt_misc`)

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+
+- The [development tools described in the cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation/#install-development-tools)
+  (`qemu-user-static`, `g++-aarch64-linux-gnu`, ...) should be installed in the
+  development workstation used for the build. The `cc-build` plugin will
+  declare those system dependencies using a `stdeb.cfg` file like is done on
+  [other colcon plugins](https://github.com/colcon/colcon-core/blob/master/stdeb.cfg).
+- In order to build the sysroot, the plugin downloads a base ROS 2 Docker image
+  for the target platform, and builds a workspace dependent sysroot image by
+  doing a `COPY` to copy the contents of the workspace into the container, and
+  running `rosdep` on the copy. That ensures no workspace system dependency is
```
thomas-moulard: I'm wondering whether we could execute a "well-known script" like `./initialize.sh` or something to avoid making the assumption that `rosdep` will always be the only required step there.

This would be helpful to test the plugin too as we could run non-ROS 2 images (e.g. base OS image only) to make sure we can cross-compile a simple "hello world" binary.

thomas-moulard (approved):
SGTM, one comment on whether we could remove the assumption that `rosdep update` is the only thing we will ever need to do to "update" the sysroot image.

articles/cross_compilation_build_tools_design.md
nuclearsandwich: Posing @esteve's remark as a suggestion.

```suggestion
  platform. That latter use case is particularly useful for speeding up build
```

dirk-thomas:
Reading the current proposal I see a few issues with the approach:

* The initial design goals of `colcon` explicitly mentions that dealing with external dependencies (anything outside of the workspace) is explicitly out of scope for the tool. The idea behind this is basically that a tool should have one purpose, focus on that topic and do it well. To me it sounds like a different tool (like `rosdep`) should do that job.

* While I understand the desire to come up with a single-line-invocation the new verb seems to do two completely separate things: building a sysroot image and building the workspace. I think all the wording around when the sysroot is being build, when it needs to be rebuild and options like `--force-sysroot-build` are indicators that it might be better to separate these two parts.

* Also I am not sure that `colcon` needs a way to use Docker. Instead of coupling `colcon` with knowledge about Docker why no just invoke `colcon` within Docker - `colcon` doesn't need to know about this imo.

thomas-moulard:
> Reading the current proposal I see a few issues with the approach:
> 
> * The initial design goals of `colcon` explicitly mentions that dealing with external dependencies (anything outside of the workspace) is explicitly out of scope for the tool. The idea behind this is basically that a tool should have one purpose, focus on that topic and do it well. To me it sounds like a different tool (like `rosdep`) should do that job.

That seems fair to me. What about having something like `rosdep_docker` building Docker sysroot images for us?

We can start by putting this script alongside the colcon plugin and fork it into its own repo if that stuff keeps growing. There is some benefit IMHO into piggy-backing the repo to reduce the maintenance burden (e.g. release management and such).

Otherwise, woud you be OK with us teaching `rosdep` how to build Docker images? That'd be another way to do it. We don't need to pull a dependency to Docker for that so it would be a very localized change (probably built on top of `rosdep` `-s` flag?)

> * While I understand the desire to come up with a single-line-invocation the new verb seems to do two completely separate things: building a sysroot image and building the workspace. I think all the wording around when the sysroot is being build, when it needs to be rebuild and options like `--force-sysroot-build` are indicators that it might be better to separate these two parts.

That's fair. I guess `rosdep` is not updating the locally installed packages so it could be the same for the sysroot image. If that's becoming a usability problem, we can always revisit this decision down the road so I'm not worried about this.

> * Also I am not sure that `colcon` needs a way to use Docker. Instead of coupling `colcon` with knowledge about Docker why no just invoke `colcon` within Docker - `colcon` doesn't need to know about this imo.

This part is the tricky part: the Docker image we're talking about there not a native image. It could e.g. an **ARM** Docker image. Meaning that anything inside is running through `qemu` with a 10x penalty (relying on `binfmt_misc` hooks).

That's what we're proposing is not equivalent to just run `colcon` in Docker. If we do that, then development becomes a pain because all of colcon code (Python, etc.) is qemu based. The approach described in this PR is only paying the cost of qemu for the initial build of the workspace as well as every time the image gets updated. Obviously, I don't have number to show you but the hope is that having rosdep + apt-get being slow is not too much of a problem compared to having colcon + gcc/g++ running through qemu.

articles/cross_compilation_build_tools_design.md
gbiggs: If it's added to the CMake extension, then wouldn't that cause it to exclude packages that only use Python? I know that they don't need cross-compiling, but it would be convenient if `colcon` could build an entire application in the crosscompile target file system, not just the parts that are done with CMake, so that it is really easy for the user to just copy the install directory to the target device.

gbiggs:
> Otherwise, woud you be OK with us teaching rosdep how to build Docker images?

Wouldn't that have the same problem of adding a new, different function to a tool? `rosdep` is for installing dependencies, so I think that your proposal should be reversed somewhat, and that you should provide a script that wraps `rosdep` and produces a Docker image.

thomas-moulard:
> > Otherwise, woud you be OK with us teaching rosdep how to build Docker images?
> 
> Wouldn't that have the same problem of adding a new, different function to a tool? `rosdep` is for installing dependencies, so I think that your proposal should be reversed somewhat, and that you should provide a script that wraps `rosdep` and produces a Docker image.

Sure that SGTM.

gavanderhoorn:
<del>Introducing a tool (or a verb) to build sysroots sounds like what things like OpenEmbedded do. From the design doc it seems like going that route is not something @juanrh and @thomas-moulard have in mind. Is that correct?</del>

<del>Otherwise the work that LG does/has done with superflore (https://github.com/ros-infrastructure/superflore/pull/168) might be relevant?</del>

Edit: blah, I'd actually read the doc, and it already states this:

> However, that tutorial doesn't provide instructions to build other ROS 2
packages, but just for building a whole ROS 2 distribution. Ideally, we should
be able to cross-compile a ROS package for ROS 2 [..]

That's indeed not a typical use-case for OpenEmbedded et al (but it can be done afaik).

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
+---
+layout: default
+title: Cross-compiling ROS 2 packages
+abstract:
+  This article is a design proposal for extending the build tools for ROS 2 to
+  simplify cross-compiling ROS 2 packages for platforms currently not supported
```
dejanpan, line 6: I would suggest that we define what a platform is, e.g. in the automotive world a platform is a tupple of {ECU:OS:Framework}. So for instance an example of a platform would be: {Renesas R-Car H3:QNX:ROS2 C}.
Note that Renesas R-Car H3 implicitly implies that this is an `aarch64` architecture for computer processor.

Since ROS 2 depends greatly on an underlying middle-ware it would be good to even include that into the tupple, so {ECU:OS:Middle-ware:Framework}.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+allows us to run the single command below from a ROS 2 overlay workspace on a
+developer workstation, for cross-compiling the packages in the workspace up to
+a package `performance_test` for the ARM-HF platform.
+
+```bash
+colcon cc-build --arch armhf --os ubuntu_bionic \
+  --packages-up-to performance_test # other `colcon build` arguments
+```
+
+That would should result on a new workspace directory `armhf-ubuntu_bionic`
```
dejanpan: remove `should`

dejanpan: on => in

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+with subdirectories `build`, `install`, and `log` directories, with the
+ross-compiled binaries. All options for `colcon build` will be supported by
```
dejanpan: ross => cross

I actually propose to proof read this PR for spelling mistakes and other grammar.

anup-pem:
The design looks good and can be a first step in getting cross compile ROS 2 packages and testing on the target platform. 

 We follow the following steps at Apex.AI to cross compile and test. Open for suggestions/comments to make it better and contribute it back.

1. Develop on a x86_64 ubuntu linux system

2. Cross compile on X86_64 for aarch64. This PR should make it one step process

3. SCP the build and install folder on to the target aarch64 hardware and do unit testing.

4. Submit to CI

We follow the above steps for aarch64/linux, x86_64/qnx and aarch64/qnx.

As part of step-3, we will start using the qemu/docker image for aarch64/Linux.

We follow the above steps for application testing, performance testing.

articles/cross_compilation_build_tools_design.md
juanrh: I'll add that precedents to the document, thanks for bringing them to our attention

articles/cross_compilation_build_tools_design.md
juanrh: I'll add explicit reference to the QEmu usage there

articles/cross_compilation_build_tools_design.md
gbiggs: Perhaps we can have a generic part and a platform-specific set of samples? Or a small set of supported platforms, complying with the relevant REP?

gbiggs:
I'm surprised it hasn't been pointed out yet (has the policy changed?) but the format guidelines say one sentence per line and don't word wrap.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,248 @@
...
+Original Author: {{ page.author }}
+
+## Background
+
+The [cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation)
+contains instructions for cross-compiling ROS 2 to unsupported platforms, using
+some configuration files for CMake and Docker that is setting maintained in the
+[ros2/cross_compile repository](https://github.com/ros2/cross_compile/).
+However, that tutorial doesn't provide instructions to build other ROS 2
+packages, but just for building a whole ROS 2 distribution. Ideally, we should
```
lmayencourt: The "Cross-compiling against a pre-built ROS2" section aims to provide an example on how to CC a single ros2 package without building the full distribution.

articles/cross_compilation_build_tools_design.md
lmayencourt: armhf is also supported with the [ros2/cross_compile repository](https://github.com/ros2/cross_compile/). The architecture is an argument of the "sysroot dockerfile"

> Use an argument to specify the architecture of the target-file system.
This can be specified with the `--build-arg` argument on the docker
build command:
  *for armhf: docker build --build-arg ARM_ARCH=arm32v7 ...
  *for aarch64: docker build --build-arg ARM_ARCH=arm64v8 ...

articles/cross_compilation_build_tools_design.md
esteve: In addition to CMake and Python, `colcon` supports Gradle, Bazel, Cargo and probably more, hence why I believe this functionality should be part of each build plugin, because it's hard to cover them all in a single plugin. I'm thinking something like `colcon build --cmake-sysroot ... --python-sysroot ...` would work. Perhaps we could enforce that any new build tool that gets integrated into `colcon` has support for cross compiling.

dirk-thomas: > Perhaps we could enforce that any new build tool that gets integrated into `colcon` has support for cross compiling.

This goes against one of the [stated software design goals](https://colcon.readthedocs.io/en/released/developer/design.html#software-design) of `colcon`:

    > Each functionality added should follow the principle “you don’t pay for what you don’t use”.

So if a plugin proving support for X doesn't need sysroot support I shouldn't need to implement it.

dirk-thomas:
> I'm surprised it hasn't been pointed out yet (has the policy changed?) but the format guidelines say one sentence per line and don't word wrap.

I didn't focus on stylistic feedback yet since before that the conceptional problems / questions need to be addressed. But yes, the [style guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#markdown-restructured-text-docblocks) contains the following:

    > Each sentence must start on a new line.

dirk-thomas (changes requested):
Just to make my feedback visible that imo this draft requires conceptional changes as commented above.

articles/cross_compilation_build_tools_design.md
juanrh: [Dockerfile_ubuntu_arm](https://github.com/ros2/cross_compile/blob/master/sysroot/Dockerfile_ubuntu_arm) indeed supports that argument, but [entry_point.sh](https://github.com/ros2/cross_compile/blob/master/entry_point.sh#L28) assumes ARM64 when it sets `export TARGET_TRIPLE=aarch64-linux-gnu`. However I agree we could be more precise and state that the tutorial does support ARMHF, but the automation described in the "Automated Cross-compilation" section does not fully support it. I'll be more precise about that, as this proposal doesn't try to change the basic approach to cross compilation, but to pack it into a development tool that is easy to use

articles/cross_compilation_build_tools_design.md
juanrh: Fixed

articles/cross_compilation_build_tools_design.md
juanrh: I have made a pass to fix spelling and grammar mistake, please let me know if you find something else

articles/cross_compilation_build_tools_design.md
juanrh: That makes sense, but I'm not sure where to put it. Some options that come to my mind: 

- Make it dynamic: and optional argument for the plugin to provide and `./initialize.sh` to run at the end of the workspace image. If not provided we use a default `./initialize.sh` that runs rosdep. 
- Make it static: for each (os, arch) combination (or whichever definition of platform we use) the plugin uses a different Dockerfile,  extending a base Dockerfile that extends the base image for ROS 2 specified and does the COPY of the workspace contents. This means the plugin has to be modified to add or modify the Dockerfile specific for a platform. 

Does some of these option sound good to you?

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,251 @@
...
+That would should result on a new workspace directory `armhf-ubuntu_bionic`
+with subdirectories `build`, `install`, and `log` directories, with the
+ross-compiled binaries. All options for `colcon build` will be supported by
+forwarding them to the `BuildVerb` class. That would be the only command
+required for cross-compiling the packages in a workspace. Under the hood, the
+plugin would do the following:
+
+- The [development tools described in the cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation/#install-development-tools)
+  (`qemu-user-static`, `g++-aarch64-linux-gnu`, ...) should be installed in the
+  development workstation used for the build. The `cc-build` plugin will
```
tfoote: I would like to suggest a different approach that will not require the installation of the cross compile toolchain and development tool on the users development machine. Instead we can create a docker image which contains the toolchain and all the cross compilation tools. This allows us to ensure that the build is much more reproducible by allowing us to validate the versions of the development tools in the image with the toolchain. And the only content coming from the users environment is the workspace.

We can still generate a sysroot that can be mounted in for testing and providing dependencies. But the sysroot can be  mounted or COPY'd into the built container using multistage docker builds. 

This will break the need for the plugin to have dependencies on the tools. This will be especially important as we start to try to support new architectures. Requiring all the tools to be installed locally would become cumbersome.

Similarly it allows us to QA the tools on a specific platform and can run reliably even if the user is not running on one of those platforms. Say the user is running on an older distro without the necessary recent enough version of the gcc armhf, if the tools are up to date in the bionic image it can build fine on their machine.

articles/cross_compilation_build_tools_design.md
tfoote: Related to this see below where I suggest having the platform support builds embedded into docker images instead of on the host. So that you don't need to have the cross compilation resources on your local machine, just a docker dependency.

articles/cross_compilation_build_tools_design.md
tfoote: I would suggest holding off on this until it's necessary for a use case instead of speculatively extending the scope.

articles/cross_compilation_build_tools_design.md
dirk-thomas: While doing the build within Docker sounds like a good idea I think the (to be developed?) tools should be not limited to that use case. Imo there should be "tools" which can do they job (in whatever environment) and then there can be a specific "meta tool" which uses the tools in a certain way - e.g. within a Docker container. Developing something only for a very use case will likely make the result not really reusable and be only useful to a small subset of users.

articles/cross_compilation_build_tools_design.md
juanrh: I'll remove that statement, and make it clear that this proposal only contributes build tools to make it easier to use the existing instructions

articles/cross_compilation_build_tools_design.md
juanrh: Another proposal would be splitting this into two commands, as suggested in this thread several times:

- A new command `setup-sysroot` would use Docker to generate the sysroot
- A new colcon mixin that adds options for using the sysroot to the colcon build task. This command requires having the cross compilation toolchain installed.
- On an second iteration we might:
  - Maintain a docker image for cross compilation that has installed both the cross compilation toolchain, and the aforementioned colcon mixin, based on the images at the ros2/cross_compile repository.
  - Implement a new `cross-build` command (definitive name TBD) that will ensure the sysroot is created using `setup-sysroot`, and then launch the cross compilation into the docker container with the cross compilation toolchain.

How does that sound?

dirk-thomas: > A new command `setup-sysroot` would use Docker to generate the sysroot

What exactly is this *verb* supposed to be doing? (Depending on the answer) why should this be a `colcon` verb?

> A new colcon mixin that adds options for using the sysroot to the colcon build task. This command requires having the cross compilation toolchain installed.

Sounds good to me.

juanrh: > What exactly is this verb supposed to be doing? (Depending on the answer) why should this be a colcon verb?

The idea is keeping a directory in the workspace for that target platform, for example  `armhf-ubuntu_bionic`, determined by convention based on arguments for architecture and OS. Let us call that directory the cc-root. The command `setup-sysroot` would create a `sysroot` subdirectory of the cc-root (e.g.  `armhf-ubuntu_bionic/sysroot`) with the sysroot for the target platform. The colcon mixin for cross-compilation would then use the same convention as `setup-sysroot` to determine the cc-root, and use the sysroot stored there. We'd have mixins that provide shortcuts for cross compiling known architecture-os combinations. 

Here I'm not proposing having `setup-sysroot` as a new colcon verb, but as a new separate command. I'm also open to use a different name instead of "setup-sysroot".

dirk-thomas: > Here I'm not proposing having setup-sysroot as a new colcon verb, but as a new separate command.

Oh, outside of `colcon`? Please feel free to choose whatever you like - I don't have an opinion on that.

articles/cross_compilation_build_tools_design.md
juanrh: I agree with the proposal to add the ROS distro and rmw implementation. For the ECU I propose starting with a `generic_armhf`, and then adding some specific ECUs as @gbiggs suggests. So we for example we could have `generic_armhf-ubuntu_bionic-fastrtps-crystal` as an example  platform identifier.

articles/cross_compilation_build_tools_design.md
juanrh: I have updated the design to replace the colcon plugin by the new command and the colcon mixin, please let me know what you think

dirk-thomas: Sounds reasonable to me. With this update I might not be the best person to provide more feedback since the `colcon` part of this boils down to a mixin. Maybe it would be good to update the title of the ticket to reflect the current state?

I am not sure I understand how the user gets the CMake toolchain file which would be referenced from the mixin. But that might be a too detailed question for now.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+
+- _How can we run the resulting binaries?_ Packaging the resulting binaries is
+  outside of the scope of this document.
+  The cross-compiled binaries will be
+  available in the workspace in the cc-root (e.g.
+  `generic_armhf-ubuntu_bionic-fastrtps-crystal`).
+  We make the assumption that
+  the user will copy the whole workspace to the remote device (including `src`)
+  and run `rosdep` on the target platform, to install the dependencies before
+  running the binaries.
```
dirk-thomas, line 153: Just a note: this will also install all build dependencies which would't be necessary.

juanrh:
Do we have any other comments on the new design with a separate command to setup the sysroot, and a colcon mixin for running the build?

tfoote:
@juanrh When you add/update commits ready for review please comment soon after so we know it's ready for rereview otherwise we don't get a notification and won't know it's been updated. I'll take a look at this shortly.

articles/cross_compilation_build_tools_design.md
filiperinaldi: The Dockerfiles and scripts under ros2/cross_compile are just exampless to go along with the tutorial explaining cross-compilation and what it takes to achieve it (independent of the arch and target platform). The idea of integrating this support into Colcon is great. I assume the proposal here is to have a list of supported platforms (e.g. armhf) instead of just deriving the arch/plat from the new Colcon plugins, right?

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,161 @@
...
+ross-compiled binaries. All options for `colcon build` will be supported by forwarding them to the `BuildVerb` class. That would be the only command required for cross-compiling the packages in a workspace. Under the hood, the plugin would do the following:
+
+- The [development tools described in the cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation/#install-development-tools)
+ (`qemu-user-static`, `g++-aarch64-linux-gnu`, ..) should be installed in the development workstation used for the build. The `cc-build` plugin will
+ declare those system dependencies using a `stdeb.cfg` file like is done on [other colcon plugins](https://github.com/colcon/colcon-core/blob/master/stdeb.cfg).
+- In order to build the sysroot, the plugin downloads a base ROS 2 Docker image for the target platform, and builds a workspace dependent sysroot image by doing a `COPY` to copy the contents of the workspace into the container, and running `rosdep` on the copy. That ensures no workspace system dependency is missing. Note we we can run `rosdep`
```
filiperinaldi: Double we in "Note we we can run"

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,161 @@
...
+in Docker for the ARM-HF platform thanks to [qemu](https://www.qemu.org/). That image's file system is then exported by launching a container a running `docker container export` on it, to a path following a sensible convention, like a workspace-local `armhf-ubuntu_bionic/sysroot` directory, so it can be used as a sysroot during the build.
```
filiperinaldi: Extra "a" in "...launching a container a running..."

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,161 @@
...
+- `--force-sysroot-build`: Boolean flag to force exporting the sysroot even if it already available in the workspace. This is useful for example when new system
+dependencies are added to a package in the workspace. This leads to rebuilding the dependent sysroot image after the `COPY` statements that copies the workspace into
+the image, including running `rosdep install` from scratch. This does not rebuild the base ROS 2 Docker image.
+- `--sysroot-path`: Specifies a local path for an existing sysroot. This can be useful for example if different workspaces share the same set of system dependencies.
+- `--use-base-image-sysroot`: Boolean flag, when true the base ROS 2 Docker image is used to build the sysroot, so the dependencies declared in packages in the current
+workspace are ignored. This is useful to speed up builds for packages without additional dependencies.
+
+Support for other platforms can be added by creating corresponding ROS 2 Docker images, and variants of the toolchain file if required.
+We could include support for ARM-HF and ARM64 in the first release of the plugin.
```
filiperinaldi: You may also to restrict the valid OS (when using --os) to Ubuntu to get started.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,161 @@
...
+`--sysroot-base-image` from the `--arch` and `--os` arguments, for the architecture-OS combinations that have an image published for them on
+[osrf/ros2 on Docker Hub](https://hub.docker.com/r/osrf/ros2/).
+- `--sysroot-image`: Specifies a Docker image for the workspace sysroot. The build will export the sysroot from that image, instead of creating a sysroot image from
+the workspace. No check is performed to ensure the image has all the dependencies required by the packages in the workspace.
+- `--force-sysroot-build`: Boolean flag to force exporting the sysroot even if it already available in the workspace. This is useful for example when new system
+dependencies are added to a package in the workspace. This leads to rebuilding the dependent sysroot image after the `COPY` statements that copies the workspace into
+the image, including running `rosdep install` from scratch. This does not rebuild the base ROS 2 Docker image.
+- `--sysroot-path`: Specifies a local path for an existing sysroot. This can be useful for example if different workspaces share the same set of system dependencies.
```
filiperinaldi: --sysroot-path is also very useful to target products handling their file system outside Docker. In other words, the whole built-in support for preparing a sysroot being proposed under this design is a great way for a developer to focus on the "arch" support, but in practice, companies and individuals might be using other options to create their file system: Buildroot, Yocto, other pre-built distros, etc.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,161 @@
...
+That would allows us to run the single command below from a ROS 2 overlay workspace on a developer workstation, for cross-compiling
+the packages in the workspace up to a package `performance_test` for the ARM-HF platform.
+
+```bash
+colcon cc-build --arch armhf --os ubuntu_bionic \
+  --packages-up-to performance_test # other `colcon build` arguments
+```
+
+That would should result on a new workspace directory `armhf-ubuntu_bionic` with subdirectories `build`, `install`, and `log` directories, with the 
+ross-compiled binaries. All options for `colcon build` will be supported by forwarding them to the `BuildVerb` class. That would be the only command required for cross-compiling the packages in a workspace. Under the hood, the plugin would do the following:
```
filiperinaldi: ross-compiled -> cross-compiled

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,161 @@
...
+(e.g. [ApexAI/performance_test](https://github.com/ApexAI/performance_test/)), and then launching a Docker container for the that sysroot image, running
+`colcon test` using the cross-compiled binaries. The container would be launched with a [bind mount](https://docs.docker.com/storage/bind-mounts/) of the
+workspace so they have access to the compiled binaries.
+
+In order to simplify the development of base ROS 2 Docker images for new architecture-OS combinations, on a second iteration of this plugin we might add a
+new verb `colcon cc-build-sysroot-base-image` that will 1) copy the required qemu files and ROS 2 source into a temporary file; 2) build a Docker image from a
+specified Dockerfile, with access to the those resources; 3) optionally publish the image into a Docker registry, using a suitable naming convention like
+`${docker_repo_uri}:${os}_${arch}_${prebuilt?}_${rosdistro}`. Dockerfiles compatible with this command, for each architecture-OS pair would be keep on the
+`ros2/cross_compile` repository, so they can be used in CI/CI pipeline if needed.
```
filiperinaldi: CI/CI pipeline -> CI/C**D** pipeline ?

articles/cross_compilation_build_tools_design.md
juanrh: Thanks for letting me know about the typo, it's fixed in a later commit.

Regarding the list of supported platforms, after some feedback in this PR the idea is that instead of a colcon plugin we'd add a command for setting up the sysroot, and a colcon mixin for adding the required cross compilation flags to `colcon build`. For example we might build ROS 2 for armhf as follows

```bash
# Setup the sysroot for the platform (generic_armhf, ubuntu_bionic, ROS crystal, RMW FastRTPS)
# this leaves the sysroot at generic_armhf-ubuntu_bionic-fastrtps-crystal/sysroot
cc-setup-sysroot --arch generic_armhf --os ubuntu_bionic --distro crystal --rmw fastrtps \
   --sysroot-base-image ros2:armhf_bionic_crystal_base-crystal
# this creates some symlinks like /lib/arm-linux-gnueabihf -> $(pwd)/generic_armhf-ubuntu_bionic-fastrtps-crystal/sysroot/lib/arm-linux-gnueabihf in line with the "Hack to find Poco" from https://github.com/ros2/cross_compile/blob/master/entry_point.sh#L38
bash generic_armhf-ubuntu_bionic-fastrtps-crystal/cc_system_setup.bash

# this setups some env vars for cross compilation
source generic_armhf-ubuntu_bionic-fastrtps-crystal/cc_build_setup.bash

# Run colcon build using the mixin for this platform
colcon build --mixin cc-generic_armhf-ubuntu_bionic-fastrtps-crystal
```

The image `ros2:armhf_bionic_crystal_base-crystal` above refers to a Docker image with just the basics depedencies to build ROS2. To cross compile a workspace with additional packages that are not part of ROS 2, we use a different argument for `--sysroot-base-image` that corresponds to an image with ROS 2 pre built for that architecture. Then `cc_build_setup.bash` would source `generic_armhf-ubuntu_bionic-fastrtps-crystal/sysroot/opt/ros/crystal/setup.bash` as part of the setup.

This means we'd have a different colcon mixin for each platform.

articles/cross_compilation_build_tools_design.md
juanrh: I'll fix it, thanks

articles/cross_compilation_build_tools_design.md
juanrh: Thanks for the suggestion, other comments in this PR had similar concerns. After incorporating that feedback that idea now is that the platform is determined by a tuple ` ('arch', 'os', 'distro', 'rmw')`, for example `(generic_armhf, ubuntu_bionic, ROS crystal, RMW FastRTPS)`, as seen in my previous comment. That would allow extending this in the future with support for more specific ECUs than `generic_armhf`, as suggested by @dejanpan above

articles/cross_compilation_build_tools_design.md
dirk-thomas: > This means we'd have a different colcon mixin for each platform.

If the duplication between them becomes an issue we could consider supporting one mixin to extend another mixin - just a thought.

articles/cross_compilation_build_tools_design.md
juanrh: With the new design with a separate `cc-setup-sysroot` command we could add a `--sysroot-path` optional argument to it, so  `cc-setup-sysroot`  would just generate the `cc_system_setup.bash` and `cc_build_setup.bash` when that argument is present. That should work provided the sysroot contains standard Linux paths line `/lib` and `/usr/lib/`

articles/cross_compilation_build_tools_design.md
juanrh: I'll fix it, thanks

articles/cross_compilation_build_tools_design.md
juanrh: I'll fix that, thanks

gbiggs: I don't know about you but I feel the need to do CI on my CI all the time. Who watches the watchers and all that.

juanrh:
Any other comments on this? Thanks

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+We propose continuing with the general approach of the cross-compilation
+tutorial, based on building a sysroot for the target platform using QEMU and
+Docker, and then using [`cmake-toolchains`](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html)
+with a toolchain file pointing `CMAKE_SYSROOT` to the sysroot location, to
+compile with C and C++ cross-compilers.
+
+We propose adding a set of commands to wrap the instructions on the
+cross-compilation tutorial.
+Those commands would use a new workspace
+directory for the target platform, determined by a convention based on
```
jacobperron, line 49: typo: "on on"

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+  This command would take arguments to specify the target platform,
+  for example:
+
+```bash
+cc-setup-sysroot --arch generic_armhf --os ubuntu_bionic \
+                 --rmw fastrtps --distro crystal
+```
+
+- A new colcon mixin for each known platform, which adds options to the colcon
+  build task for using asysroot generated using `cc-setup-sysroot`, by using
```
jacobperron:
```suggestion
  build task for using a sysroot generated with `cc-setup-sysroot`, by using
```

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+  the same path conventions.
+  For example from a ROS 2 overlay workspace on a
```
jacobperron:
```suggestion
  For example, from a ROS 2 overlay workspace on a
```

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+- On an second iteration of this initiative we would:
+  - Maintain a docker image for cross compilation, that has both the cross
+  compilation toolchain and the aforementioned colcon mixins installed.
+  - Implement a new `cc-build` command that would ensure the sysroot is created
+  using `cc-setup-sysroot`, and then launch the cross compilation into a docker
+  container using the corresponding Docker image.
+
+Under the hood, the `cc-setup-sysroot` command would do the following:
+
+- The command downloads a base ROS 2 Docker image for the target platform, and
```
jacobperron: Consider breaking this long point into a numbered list for clarity. E.g.

> 1. Downloads a base ROS 2 Docker image for the target platform.
> 2. Builds a workspace dependent sysroot image by running `COPY` to get the contents of the workspace into the container.
> 3. Run `rosdep` to ensure no workspace system dependency is missing.
> 4. etc...

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+  builds a workspace dependent sysroot image by
+  running `COPY` to get the contents of the workspace into the container, and
+  running `rosdep` on the copy.
+  That ensures no workspace system dependency is
+  missing.
+  Note we can run `rosdep` in Docker for platform like ARM-HF thanks
```
jacobperron:
```suggestion
  Note we can run `rosdep` in Docker for platforms like ARM-HF thanks
```

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+
+Regarding maintenance, in order to support cross-compilation for a new platform
+we would need to:
+
+- Prepare and publish a base ROS 2 Docker image for the new platform, both
+  fetching the dependencies using the ROS 2 source using `rosdep install`, and
+  from the pre-built binary if available for that platform.
+- Adapt the toolchain file for the new platform, if required.
+  For that the toolchains in [ruslo/polly](https://github.com/ruslo/polly) might be useful.
+- Add a new `cross-compile*` colcon mixin to for the new platform, add perform
```
jacobperron:
```suggestion
- Add a new `cross-compile*` colcon mixin to for the new platform and perform
```

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+  any modifications needed to use a suitable toolchain file.
+
+We could then test cross-compile support for the new platform by building the
```
jacobperron: Consider making this a numbered list as well. E.g.

> We can then test cross-compile support for the new platform by:
> 1. Building the base ROS 2 Docker image.
> 2. Building a custom sysroot image for a reference package.
> 3. etc...

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+base ROS 2 Docker image, and a custom sysroot image for some reference package
+(e.g. [ApexAI/performance_test](https://github.com/ApexAI/performance_test/)),
+and then launching a Docker container for the that sysroot image, running
+`colcon test` using the cross-compiled binaries.
+The container would be launched with a [bind mount](https://docs.docker.com/storage/bind-mounts/)
+of the workspace so they have access to the compiled binaries.
+
+In order to simplify the development of base ROS 2 Docker images for new
+architecture-OS combinations, on a future iteration of this initiative we might add
+a new command `build-sysroot-base-image` that will 1) copy the required
```
jacobperron: Consider making this a proper markdown list.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,266 @@
...
+
+This setup is focused on cross-compiling for Linux based platforms.
+For example we couldn't use this to cross-compile for Windows, because we cannot install
+the Visual Studio compiler using Apt.
+On the other hand, a Windows or Mac development workstation could use this
+build setup, running the build commands inside a Docker container, although
+that would imply some performance degradation due to the virtualization layer
+that Docker employs on those platforms.
+
+We should determine a process for building an publishing the base ROS 2 Docker
```
jacobperron:
```suggestion
We should determine a process for building and publishing the base ROS 2 Docker
```

jacobperron (commented):
Cosmetic comments.

juanrh:
Thanks for your comments @jacobperron, I have addressed them in a new commit

juanrh:
Any other comments on this design?

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+
+- A new `cc-setup-sysroot` command would use Docker to generate the sysroot
+  for the target platform, and store it in a `sysroot` subdirectory of the
+  cc-root.
+  This command would take arguments to specify the target platform,
+  for example:
+
+```bash
+cc-setup-sysroot --arch generic_armhf --os ubuntu_bionic \
+                 --rmw fastrtps --distro crystal
```
nuclearsandwich, line 66:
```suggestion
                 --rmw fastrtps --rosdistro crystal
```

--rosdistro, abbreviated `-r` would be more consistent with other tools in the ROS community and less likely to be misinterpreted as a Linux ("distro") distribution.

nuclearsandwich: An rmw can consist of more than the underlying DDS mechanism. For example there are static and dynamic typesupport variants of rmw_fastrtps. On some level only the underlying middleware is significant for setting up the sysroot as it defines what's available for rmw implementation packages to be built on top of it, but I do want to make sure that we're not equating **one** middleware provider with **one** rmw implementation.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+1. Launches a container for that image and exports its file system into a
+   `sysroot` subdirectory of the cc-root.
+1. During a build, the colcon mixin sets the variable `SYSROOT` to the sysroot path,
+   as well as any other variables required to configure the toolchain file [generic_linux.cmake](https://github.com/ros2/cross_compile/blob/master/cmake-toolchains/generic_linux.cmake).
+   It also sets `DCMAKE_TOOLCHAIN_FILE` through `cmake-args`, so it points to that toolchain file.
+
+The ROS 2 base images are variants of [sysroot/Dockerfile_ubuntu_arm](https://github.com/ros2/cross_compile/blob/master/sysroot/Dockerfile_ubuntu_arm)
+from ros2/cross_compile.
+OSRF would publish base ROS 2 images with ROS 2 prebuilt, that should be used
+for cros compiling workspaces with user packages.
```
nuclearsandwich, line 113:
```suggestion
for cross compiling workspaces with user packages.
```
Just fixing spelling.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+
+Regarding maintenance, in order to support cross-compilation for a new platform
+we would need to:
+
+- Prepare and publish a base ROS 2 Docker image for the new platform, both
+  fetching the dependencies using the ROS 2 source using `rosdep install`, and
+  from the pre-built binary if available for that platform.
+- Adapt the toolchain file for the new platform, if required.
+  For that the toolchains in [ruslo/polly](https://github.com/ruslo/polly) might be useful.
+- Add a new `cross-compile*` colcon mixin to for the new platform and perform
```
nuclearsandwich, line 165: There's a subject missing.

"Add a new `cross-compile*` colcon mixin to ______ for the new platform..."

nuclearsandwich (commented):
Found some clerical / cosmetic things and a couple of comments about the platform identifier and argument names.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+
+Original Author: {{ page.author }}
+
+## Background
+
+The [cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation)
+and the [steve/ros2_raspbian_tools repository](https://github.com/esteve/ros2_raspbian_tools)
+contain instructions for cross-compiling ROS 2 to unsupported platforms like
+ARM64 and ARM-HF, using some configuration files for CMake and Docker.
+Other projects like [steve/ros2_raspbian_tools](https://github.com/esteve/ros2_raspbian_tools),
```
pokitoz, line 30: Looks like
`[steve/ros2_raspbian_tools](https://github.com/esteve/ros2_raspbian_tools)` is already mentioned above. Did you mean https://github.com/esteve/ros2_objc  (for iOS) ?

zmichaels11:
PR moved to https://github.com/ros2/design/pull/249

juanrh:
moved to https://github.com/ros2/design/pull/249

rosbuild:
This pull request has been mentioned on **ROS**. There might be relevant details there:

https://discourse.ros.org/t/announcing-first-release-of-ros-cross-compiler-tool/12601/6

PR #234, opened 2019-05-06
Expanded first use of CDR acronym

nickolai-voyage:
Acronyms should always be expanded upon first use


jacobperron (approved):
LGTM, thanks!

PR #236, opened 2019-05-22
use consistent naming of ROS versions

dirk-thomas:
The guideline for referring to specific major version of ROS should be:
* a space between "ROS" and the major version "1" or "2"
* no minor version like "2.0" or "1.x"

This patch changes various variations to follow that guideline.

* Before: `ROS2`, `ROS 2.0`, `ROS 1.x`
* After: `ROS 1`, `ROS 2`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# ROS 2.0 design
+# ROS 2 design
 
 This repository is a [Jekyll](http://jekyllrb.com/) website hosted on [Github Pages](http://pages.github.com/) at http://design.ros2.org/.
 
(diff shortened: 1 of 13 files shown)


tfoote (approved):

dirk-thomas:
@davetcoleman Maybe this could be applied to the naming of the upcoming MoveIt! alpha, too - naming it "MoveIt! 2" rather than "MoveIt! 2.0".

vmayoral:
@dirk-thomas, we decided to drop the exclamation mark and leave it as `MoveIt`. I personally would support dropping the `.x` but let's hear what @davetcoleman thinks on this.

@lalzolak and @sergiovara, you may want to consider https://github.com/ros2/design/pull/236#issue-281382474 for future communications.

PR #239, opened 2019-06-13
New intra-process communication design

alsora:
This PR adds a description of the current intra-process communication mechanism, together with a design proposal for improving its performances and supporting more QoS.
The new design is supported by experimental evaluations.

An initial discussion about this design can be found [here](https://github.com/irobot-ros/ros2-performance/pull/1).

The implementation of this design can be found [here](https://github.com/alsora/rclcpp/tree/alsora/new_ipc_proposal).

@dgoel @raghaprasad


gbiggs:
Please format with one sentence per line and no line breaks within a sentence so that changes made in response to comments are easier to track.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,437 @@
...
+There are three possible data-types that can be stored in the buffer:
+
+- `MessageT`
+- `shared_ptr<const MessageT>`
+- `unique_ptr<MessageT>`
+
+The choice of the buffer data-type is controlled through an additional field in the `SubscriptionOptions`. The default value for this option is denominated `CallbackDefault`, which corresponds to selecting the type between `shared_ptr<constMessageT>`
+and `unique_ptr<MessageT>` that better fits with its callback type. This is deduced looking at the output of `AnySubscriptionCallback::use_take_shared_method()`.
+
+If the history QoS is set to `keep all`, the buffers are dynamically allocated. On the other hand, if the history QoS is set to `keep last`, the buffers have a size equal to the depth of the history and they act as ring buffers (overwriting the
```
gbiggs: This is different from how these QoS settings work in DDS. Not having the same semantics for QoS settings between inter-process and intra-process would make it hard to reuse a node.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,437 @@
...
+   its callback, and a weak pointer for the `Subscription` object. An `uint64_t sub_id` unique within the `rclcpp::Context` is assigned to the `Subscription`. The `IntraProcessManager` contains a `std::map<uint64_t, SubscriptionInfo>` object where it
+   is possible to retrieve the `SubscriptionInfo` of a specific `Subscription` given its id. There is also an additional structure `std::map<uint64_t, std::pair<std::set<uint64_t>, std::set<uint64_t>>>`. The key of the map is the unique id of a
+   `Publisher`. The value of the map is a pair of sets of ids. These sets contain the ids of the `Subscription`s that can communicate with the `Publisher`. We have two different sets because we want to differentiate the `Subscription`s depending on
+   whether they request ownership of the received messages or not.
+6. The `SubscriptionIntraProcessWaitable` object is added to the list of Waitable interfaces of the node through `node_interfaces::NodeWaitablesInterface::add_waitable(...)`.
+
+### Publishing only intra-process
+
+#### Publishing unique_ptr
```
gbiggs, line 200: Sometimes there's something to be said for [using UML diagrams](http://plantuml.com/).

gbiggs:
This is a tangential comment, but I wonder if we could achieve the same zero-copies-when-same-process result by reducing the number of copies requires for going into and out of the `rmw` layer to zero and using a DDS implementation that also supports zero copies (ignoring that there may not be any and that the standard API may not support this, both of which are solvable issues). One of the reasons for using DDS is to push all the communication issues down into an expert-vendor-supplied library, after all.

gbiggs:
The current prototype implementation and the design focus on `rclcpp`. What about `rclc`? How are we going to ensure that `rclpy` has exactly the same capabilities and semantics?

raghaprasad:
> The current prototype implementation and the design focus on `rclcpp`. What about `rclc`? How are we going to ensure that `rclpy` has exactly the same capabilities and semantics?

I am not sure about the status of `rclc` in terms of feature parity, seems like the [ros2 supported](https://index.ros.org/doc/ros2/Concepts/ROS-2-Client-Libraries/#supported-client-libraries) clients are `rclcpp` & `rclpy`. 

While I agree about maintaining feature parity between the 2 supported clients, this design proposal is an improvement to the existing intra-process support, which currently is [supported](https://index.ros.org//doc/ros2/Tutorials/Intra-Process-Communication) only in `rclcpp`. 

If we continue to have the intra process manager live in `rclcpp`, we should have a separate discussion about what it means to do the same on `rclpy` (and if it is even required).

raghaprasad:
> This is a tangential comment, but I wonder if we could achieve the same zero-copies-when-same-process result by reducing the number of copies requires for going into and out of the `rmw` layer to zero and using a DDS implementation that also supports zero copies (ignoring that there may not be any and that the standard API may not support this, both of which are solvable issues). One of the reasons for using DDS is to push all the communication issues down into an expert-vendor-supplied library, after all.

How about moving the `intra_process_management` into an rmw ?
This rmw could handle only `intra_process` communication and delegate `inter-process` communication to a any of the chosen DDS rmw implementations. 

Support for zero copies is an important objective, but its not the only one. It has been observed that creating DDS participants is pretty resource heavy in terms of net memory required (atleast for FastRTPS & OpenSplice) and the discovery process is CPU intensive (due to multicast). 
This new rmw could drastically simplify the discovery process and most certainly reduce the memory footprint by needing only one participant per process to support `inter_process` communication. 

I don't want to derail this PR as this is totally tangential topic and needs more thought and planning. Should we move this discussion to a discourse post ?

articles/intraprocess_communication.md
dgoel: We meant to say "buffers are dynamically adjusted in size _up to some max limit_ specified as part of KEEP_ALL QoS". Then the semantics are the same, right?

alsora:
The aim of this PR is to improve the current intra-process communication, that is only supported in `rclcpp` at the moment.

> I don't want to derail this PR as this is totally tangential topic and needs more thought and planning. Should we move this discussion to a discourse post ?

As @raghaprasad says, I think that moving the intra-process manager to a different place from `rclcpp` 
will involve many considerations not necessarily related with how the intra-process communication should actually take place, which is the focus of this PR.

However, I think that once it has been finalized and implemented there shouldn't be any major blockers avoiding to move it to a COMMON `rmw` layer.
This can be used to have common behavior between C++ and Python applications.

While working to this PR I also explored the idea of implementing the intra-process manager logic at the RMW layer (I have a very dummy proof of concept for RMW DPS).
One of the issues I had was that published messages have to go through the `rclc` layer, thus it's not possible to use smart pointers and message templates. 
Apart from this, the performances were comparable with the `rclcpp` implementation (slightly worst).

articles/intraprocess_communication.md
alsora: Sorry, but I don't get what do you mean

raghaprasad: @alsora I am guessing the comment is recommending UML based diagrams to describe visually some of the sections of this PR. For. Eg, Creating a publisher,  Creating a subscription,  Publishing only intra-process etc can have [sequence diagrams](http://plantuml.com/sequence-diagram) in addition to the description.

gbiggs:
> One of the issues I had was that published messages have to go through the rclc layer, thus it's not possible to use smart pointers and message templates.

But it is possible to do the `rmw` and `rcl` APIs and implementations such that they manage their raw pointers properly and provide a `smart_ptr` interface-compatible object in `rclcpp`. I'm not saying it would be easy, but this is how the STL is designed to be used and it would be the most powerful solution.

I feel that having inter-process only available in `rclcpp` is a further continuation of not putting things in `rclc` where they can be easily wrapped by other languages, which makes it harder to get back to what we were told at the start of ROS 2 development is a goal: to have a universal C client library provide all the functionality and other languages, including C++, just wrap it. I understand this policy was sidelined by a need to get features out faster but if we just keep doing things in C++ we will make it increasingly difficult to get back to that approach, and in the long run end up with a bunch of mostly-compatible client libraries again.

dgoel:
> I feel that having inter-process only available in `rclcpp` is a further continuation of not putting things in `rclc` where they can be easily wrapped by other languages, which makes it harder to get back to what we were told at the start of ROS 2 development is a goal: to have a universal C client library provide all the functionality and other languages, including C++, just wrap it. I understand this policy was sidelined by a need to get features out faster but if we just keep doing things in C++ we will make it increasingly difficult to get back to that approach, and in the long run end up with a bunch of mostly-compatible client libraries again.

ROS2 architecture [1] depicts very explicitly that intra-process communication (along with type masquerading optimizations) belongs in the language-specific client library. Also, intra-process communication is merely an optimization, and not a "feature" per say. All the features like services, parameters, etc. are rightly implemented in `rcl` layer already.

[1] http://docs.ros2.org/dashing/developer_overview.html#internal-ros-interfaces

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+[reference](https://github.com/ros2/design/pull/212).
+
+### Dependent on the RMW
+
+The current implementation of intra-process communication has to send meta-messages from the `Publisher` to the `Subscription`s.
+This is done using the `rmw_publish` function, the implementation of which depends on the chosen middleware.
+This results in the performance of a ROS2 application with intra-process communication enabled being heavily dependent on the chosen RMW implementation.
+
+Given the fact that these meta-messages have only to be received from entities within the same process, there is space for optimizing how they are transmitted by each RMW.
+However, at the moment none of the supported RMW is actively tackling this issue.
```
ivanpauno, line 74: I think that connext have a shared transport mechanism (I don't know if we are using it). And I think opensplice too.
FastRTPS doesn't, but has it in its long term roadmap.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+However, comparing the publication/reception of an intra and an inter-process message, the former requires several additional operations: it has to store the message in the ring buffer, monitor the number of `Subscription`s, and extract the message.
+
+The result is that from the latency and CPU utilization point of view, it is convenient to use intra-process communication only when the message size is at least 5KB.
+
+
+### Problems when both inter and intra-process communication are needed
+
+Currently, ROS2 does not provide any API for making nodes or `Publisher` and `Subscription` to ignore each other.
+This feature woudl be useful when both inter and intra-process communication are needed.
```
ivanpauno: nit: `woudl` -> `would`

articles/intraprocess_communication.md
ivanpauno: Here a link with some discussion about it https://discourse.ros.org/t/is-there-any-dds-implementation-using-shared-memory-inside-system/7609/20 (which you have probably already seen).

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+The choice of having independent buffers for each `Subscription` leads to the following advantages:
+ - It is easy to support different QoS for each `Subscription`, while, at the same time, simplifying the implementation.
+ - Multiple `Subscription`s can extract messages from their own buffer in parallel without blocking each other, thus providing an higher throughput.
+
+The only drawback is that the system is not reusing as much resources as possible, compared to sharing buffers between entities.
+However, from a practical point of view, the memory overhead caused by the proposed implementation with respect to the current one, will always be only a tiny delta compared to the overall memory usage of the application.
+
+There are three possible data-types that can be stored in the buffer:
+
+- `MessageT`
```
ivanpauno, line 137: When is this case useful? I'm just curious.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+- `shared_ptr<const MessageT>`
+- `unique_ptr<MessageT>`
+
+The choice of the buffer data-type is controlled through an additional field in the `SubscriptionOptions`.
+The default value for this option is denominated `CallbackDefault`, which corresponds to selecting the type between `shared_ptr<constMessageT>` and `unique_ptr<MessageT>` that better fits with its callback type.
+This is deduced looking at the output of `AnySubscriptionCallback::use_take_shared_method()`.
+
+If the history QoS is set to `keep all`, the buffers are dynamically adjusted in size up to the maximum resource limits specified by the underlying middleware.
```
ivanpauno, line 145: I don't now exactly how we can ensure that, because we could have "duplicated" resources between the interprocess communication and the intraprocess.
It's also a problem in the current implementation.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+The data-type stored in the `Publisher` buffer is always `shared_ptr<const MessageT>`.
+
+A new class derived from `rclcpp::Waitable` is defined, denominated `SubscriptionIntraProcessWaitable`.
+An object of this type is created by each `Subscription` with intra-process communication enabled and it is used to notify the `Subscription` that a new message has been pushed into its ring buffer and that it needs to be processed.
+
+The `IntraProcessManager` class stores information about each `Publisher` and each `Subscription`, together with pointers to these structures.
+This allows to know which entities can communicate with each other and to have access to methods for pushing data into the buffers.
+
+The decision whether to publish inter-process, intra-process or both is made every time the `Publisher::publish()` method is called.
+For example, if the `NodeOptions::use_intra_process_comms_` is enabled and all the known `Subscription`s are in the same process, then the message is only published intra-process.
```
ivanpauno, line 163: Currently, we weren't allowing to set intraprocess comm with transient local qos. If we start to do so, we should still publish interprocess message in this case. We could add a way of explicitly avoiding interprocess comm.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+If the `Publisher` durability is set to `transient_local` an additional buffer on the `Publisher` side is used to store the sent intra-process messages.
+
+Late-joiner `Subscription`s will have to extract messages from this buffer once they are added to the `IntraProcessManager`.
+In this case the `IntraProcessManager` has to check if the recently created `Subscription` is a late-joiner, and, if it is, it has to retrieve messages from the `Transient Local` `Publisher`s.
+
+1. Call `IntraProcessManager::find_matching_publishers(SubscriptionInfo sub_info)` that returns a list of stored `PublisherInfo` that have a QoS compatible for sending messages to this new `Subscription`.
+These will be all `Transient Local` `Publisher`s, so they have a ring buffer.
+2. Copy messages from all the ring buffers found into the ring buffer of the new `Subscription`.
+**TODO:** are there any constraints on the order in which old messages have to be retrieved? (i.e. 1 publisher at the time; all the firsts of each publisher, then all the seconds ...).
```
ivanpauno, line 300: mmm, will check too.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+3. If at least 1 message was present, trigger the `rcl_guard_condition_t` member of the `SubscriptionIntraProcessWaitable` associated with the new `Subscription`.
+
+However, this is not enough as it does not allow to handle the scenario in which a `transient local` `Publisher` has only intra-process `Subscription`s when it is created, but, eventually, a `transient local` `Subscription` in a different process joins.
+Initially, published messages are not passed to the middleware, since all the `Subscription`s are in the same process.
+This means that the middleware is not able to store old messages for eventual late-joiners.
+
+The solution to this issue consists in always publishing both intra and inter-process when a `Publisher` has `transient local` durability.
+For this reason, when `transient local` is enabled, the `do_intra_process_publish(...)` function will always process a shared pointer.
+This allows us to add the logic for storing the published messages into the buffers only in one of the two `do_intra_process_publish(...)` cases and also it allows to use buffers that have only to store shared pointers.
```
ivanpauno, line 309: I think this is basically what I proposed in a previous comment, so it sounds good.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+
+The possibility of setting the data-type stored in each buffer becomes helpful when dealing with more particular scenarios.
+
+Considering a scenario with N `Subscription`s all taking a unique pointer.
+If the `Subscription`s don't actually take the message (e.g. they are busy and the message is being overwritten due to QoS settings) the default buffer type (`unique_ptr` since the callbacks require ownership) would result in the copy taking place anyway.
+By setting the buffer type to `shared_ptr`, no copies are needed when the `Publisher` pushes messages into the buffers.
+Eventually, the `Subscription`s will copy the data only when they are ready to process it.
+
+On the other hand, if the published data are very small, it can be advantageous to do not use C++ smart pointers, but to directly store the data into the buffers.
```
ivanpauno, line 373: Does this improve performance? I don't see much benefit of this vs using an unique ptr. If memory allocation has to be avoided, an allocator can be passed.
I think it's fine to having this option too, but I'm just curious if there is another reason for this.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+These results show that if there is at least one node in a different process, with the current implementation it is better to keep intra-process communication disabled.
+The proposed implementation does not require the ROS2 middleware when publishing intra-process.
+This allows to easily remove the connections between nodes in the same process when it is required to publish also inter process, potentially resulting in a very small overhead with respect to the only intra-process case.
+
+
+## Open Issues
+
+There are some open issues that are not addressed neither on the current implementation nor on the proposed one.
+
+ - The proposal does not take into account the problem of having a queue with twice the size when both inter and intra-process communication are used.
```
ivanpauno, line 496: I think that just documenting it would be enough.

ivanpauno: And in some cases, the duplication is just a buffer of "pointers".

ivanpauno (commented):
The proposal looks quite good to me.
I left some minimal comments. I will do a second read before jumping to the PR.

articles/intraprocess_communication.md
joxoby: I don't think it's needed.

joxoby: Originally, it was thought for increased performance in the case of small messages. But I'm not sure if this would actually be the case.

articles/intraprocess_communication.md
alsora: Yes, it is more a problem from a "conceptual" point of view rather than for the performances. 
In any case enforcing this would be quite complex as it requires constant communication between `rclcpp` and `rmw` layers.

articles/intraprocess_communication.md
alsora: Yes, I read that discussion.
However, with the risk of going off-topic, I would like to say that I have done multiple experiments and the performances of OpenSplice and Connext (evaluation version for both) were always much much worst than Fast-RTPS in terms of latency (that does not implement shared memory). 
I performed tests both using a single process as well as different processes in the same machine. Are you aware if is it necessary to set some particular configuration in order to enable shared memory transportation?

ivanpauno: I think it's enabled by default in connext, but I'm really not sure. Will try to check.
In opensplice, you need to [start it in an special way](https://istkb.adlinktech.com/article/configuring-shared-memory/), but I've never done that. That's only available in the commercial binaries.

ivanpauno: Maybe, commenting out [this](https://github.com/ros2/rmw_connext/blob/45596c37d7aca76b1806751b96863b0378f55af5/rmw_connext_shared_cpp/src/shared_functions.cpp#L246-L260) improves performance ...
That's forcing using both loopback and shared memory.

articles/intraprocess_communication.md
ivanpauno: Yeah, I wouldn't enforce that. Just document the extra memory requirements.

articles/intraprocess_communication.md
alsora: That file does not exist anymore. Now these settings have been moved [here](https://github.com/ros2/rmw_connext/blob/master/rmw_connext_shared_cpp/src/node.cpp#L54-L100).
I will try this out.

ivanpauno: > That file does not exist anymore. Now these settings have been moved [here](https://github.com/ros2/rmw_connext/blob/master/rmw_connext_shared_cpp/src/node.cpp#L54-L100).
> I will try this out.

Ups, sorry for the wrong permalink.

ivanpauno:
Some thoughts about moving intraprocess communication to rmw:

----

> ROS2 architecture [1] depicts very explicitly that intra-process communication (along with type masquerading optimizations) belongs in the language-specific client library. Also, intra-process communication is merely an optimization, and not a "feature" per say. All the features like services, parameters, etc. are rightly implemented in `rcl` layer already.
> 
> [1] http://docs.ros2.org/dashing/developer_overview.html#internal-ros-interfaces

That's mostly "historical". I think that if we have a good rationale for implementing it at rmw level, we should.

----

> But it is possible to do the rmw and rcl APIs and implementations such that they manage their raw pointers properly and provide a smart_ptr interface-compatible object in rclcpp. I'm not saying it would be easy, but this is how the STL is designed to be used and it would be the most powerful solution.

I think something like that could be possible, but quite complicated.

----

I would like to see something mimicking connext [Zero Copy Transfer Over Shared Memory](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html) semantics (by default connext use shared memory, but it doesn't use zero copy transfer, which have an specific semantics). Basically, instead of creating a unique pointer and then publishing it:
```cpp
auto msg = std::make_unique<MSG_TYPE>();
/* Fill the message here */
publisher->publish(std::move(msg))
```
You ask to the publisher a piece of memory, fill it, and then publish:
```cpp
auto msg = publisher->new_message();
/* Fill the message here */
publisher->publish(std::move(msg)); // I'm using move semantics because the message will be undefined after calling publish. But how we wrap the msg for this is an implementation detail.
```

For dds vendors that have implemented zero copy transport, this could just wrap it.
For others, we could have a default implementation that's used in those cases. That implementation could not use shared memory that allows INTERprocess zero copy transport, but just use a preallocated buffer in each publisher that allows INTRAprocess zero copy transport. This implementation is a good start for later doing something like this (if we want to do it).

I also think this idea will look idiomatic in other languages (for example, in python), and performance should be quite similar.

----

> As @raghaprasad says, I think that moving the intra-process manager to a different place from rclcpp
will involve many considerations not necessarily related with how the intra-process communication should actually take place, which is the focus of this PR.

I agree, I think this implementation will improve performance compared with the current intraprocess implementation. We could later decide if moving intraprocess comm to the `rmw` layer is a good idea or not.

emersonknapp:
@ivanpauno I agree with the above discussion that we should evaluate this PR for what it is (an improvement to the existing scenario of Intra-Process existing only in rclcpp) - but I'm also very interested in making this feature more globally accessible. I was sold that one of ROS2's goals was to _stop_ reimplementing features for each language, and to make language clients as thin as possible, therefore functionality like this should live in `rcl` or lower. 

My question is, where can we start and maintain a focused discussion for this topic today, rather than just saying "we can decide later"? That way we can take this discussion off of this PR, but still know that we have actually begun to focus on the bigger-picture architectural discussion somewhere. Perhaps a new `ros2/design` PR would be most appropriate? (e.g. "Design for intra-process comms available to all ros2 language clients"?)

ivanpauno:
> @ivanpauno I agree with the above discussion that we should evaluate this PR for what it is (an improvement to the existing scenario of Intra-Process existing only in rclcpp) - but I'm also very interested in making this feature more globally accessible. I was sold that one of ROS2's goals was to _stop_ reimplementing features for each language, and to make language clients as thin as possible, therefore functionality like this should live in `rcl` or lower.

I fully agree about making the feature globally available. Actually, it's currently available in rclcpp only for publish/subscribe communication, but for example, not for services.

> My question is, where can we start and maintain a focused discussion for this topic today, rather than just saying "we can decide later"? That way we can take this discussion off of this PR, but still know that we have actually begun to focus on the bigger-picture architectural discussion somewhere. Perhaps a new `ros2/design` PR would be most appropriate? (e.g. "Design for intra-process comms available to all ros2 language clients"?)

Maybe a PR/issue here, and post the link in discourse too.
If you already have an idea of the design, I think it's better to directly open a PR. If we just want to continue the discussion, I would rather open an issue.

As the topic is complex, if we want to move ahead with it fast, I think that it would be good to organize a meeting for discussing the topic. I would be interested in participate.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
+---
+layout: default
+title: Intraprocess Communications in ROS 2
```
wjwwood: We should be consistent, `intra-process`/`inter-process` seems like the consensus in this document and elsewhere. We should use it everywhere, including here.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+permalink: articles/intraprocess_communications.html
+abstract: Description of the current intra-process communication mechanism in ROS2 and of its drawbacks. Design proposal for an improved implementation. Experimental results.
```
wjwwood: Please use `ROS 2` everywhere.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Introduction
+
+The subscriptions and publications mechanisms in ROS2 fall in two categories:
+
+* intra-process: messages are sent from a publication to subscription via in-process memory.
```
wjwwood: publication -> publisher

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+Several shortcomings of the current implementation are listed below.
+
+### Incomplete Quality of Service support
+
+The current implementation can't be used when the QoS durability value is set to `Transient Local`.
+
+The current implementation does not enforce the depth of the QoS history in a correct way.
+The reason is that there is a single ring buffer per `Publisher` and its size is equal to the depth of the `Publisher`'s history.
+A `Publisher` stores a message in the ring buffer and then it sends a meta-message to allow a `Subscription` to retrieve it.
+The `Subscription` correctly stores meta-messages up to the number indicated by its depth of the history, but, depending on the frequency at which messages are published and callbacks are triggered, it may happen that a meta-message processed from the `Subscription` does not correspond anymore to a valid message in the ring buffer, because it has been already overwritten.
```
wjwwood, line 54: Maybe it should be clarified here that the danger here is not that you'll receive incorrect data, but that you may not receive data via intra-process that you would via inter-process due to it getting overwritten prematurely.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+Moreover, even if the use of meta-messages allows to deleagate the enforcement of other QoS settings to the RMW layer, every time a message is added to the ring buffer the `IntraProcessManager` has to compute how many `Subscription`s will need it.
+This potentially breaks the advantage of having the meta-messages.
+For example, the `IntraProcessManager` has to take into account that potentially all the known `Subscription`s will take the message, regardless of their reliability QoS.
+If a `Publisher` or a `Subscription` are best-effort, they may not receive the meta-message thus preventing the `IntraProcessManager` from releasing the memory in the buffer.
```
wjwwood, line 60: Why is this an issue? It will simply overwrite the "dangling" message when needed.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+Publishing a meta-message has the same overhead as that of publishing a small inter-process message.
+
+However, comparing the publication/reception of an intra and an inter-process message, the former requires several additional operations: it has to store the message in the ring buffer, monitor the number of `Subscription`s, and extract the message.
+
+The result is that from the latency and CPU utilization point of view, it is convenient to use intra-process communication only when the message size is at least 5KB.
+
+
+### Problems when both inter and intra-process communication are needed
+
+Currently, ROS2 does not provide any API for making nodes or `Publisher` and `Subscription` to ignore each other.
```
wjwwood: There is this... https://github.com/ros2/rmw/blob/2250b3eee645d90f9e9d6c96d71ce3aada9944f3/rmw/include/rmw/rmw.h#L517

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+As before, the messages would be discarded immediately after being received, but they would still affect the performances.
+The overhead caused by the additional publication of meta-messages can be potentially reduced by appending to the intra-process topic names a process specific identifier.
+
+## Proposed implementation
+
+### Overview
+
+The new proposal for intra-process communication addresses all the issues previously mentioned.
+It has been designed with performance in mind, so it avoids any communication through the middleware between nodes in the same process.
+Moreover, it supports all the ROS2 Quality of Service configurations.
```
wjwwood: Including the new ones? like deadline and such?

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+Consider a simple scenario, consisting of `Publisher`s and `Subscription`s all in the same process and with the durability QoS set to `volatile`.
+The proposed implementation creates one buffer per `Subscription`.
+When a message is published to a topic, its `Publisher` pushes the message into the buffer of each of the `Subscription`s related to that topic and raises a notification, waking up the executor.
+The executor can then pop the message from the buffer and trigger the callback of the `Subscription`.
+
+![Proposed IPC Block Diagram](../img/intraprocess_communication/new_ipc.png)
+
+The choice of having independent buffers for each `Subscription` leads to the following advantages:
+ - It is easy to support different QoS for each `Subscription`, while, at the same time, simplifying the implementation.
```
wjwwood, line 129:
```suggestion

  - It is easy to support different QoS for each `Subscription`, while, at the same time, simplifying the implementation.
```

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+- `MessageT`
+- `shared_ptr<const MessageT>`
+- `unique_ptr<MessageT>`
+
+The choice of the buffer data-type is controlled through an additional field in the `SubscriptionOptions`.
+The default value for this option is denominated `CallbackDefault`, which corresponds to selecting the type between `shared_ptr<constMessageT>` and `unique_ptr<MessageT>` that better fits with its callback type.
+This is deduced looking at the output of `AnySubscriptionCallback::use_take_shared_method()`.
+
+If the history QoS is set to `keep all`, the buffers are dynamically adjusted in size up to the maximum resource limits specified by the underlying middleware.
```
wjwwood, line 145: What happens in keep all + reliable when the subscription's buffers are full? This is where not having a publisher queue breaks down. In this scenario in DDS the publisher queue would then start to fill up, and then finally the publish call would block when that queue is full.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+On the other hand, if the history QoS is set to `keep last`, the buffers have a size equal to the depth of the history and they act as ring buffers (overwriting the oldest data when trying to push while its full).
+
+Buffers are not only used in `Subscription`s but also in each `Publisher` with a durability QoS of type `transient local`.
+The data-type stored in the `Publisher` buffer is always `shared_ptr<const MessageT>`.
```
wjwwood, line 154: I'm confused, where is the publisher buffer in this diagram:

https://github.com/ros2/design/blob/5cb9d22fc0b321bc7a3ebd31f66f3bf845a327eb/img/intraprocess_communication/new_ipc.png

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+A new class derived from `rclcpp::Waitable` is defined, denominated `SubscriptionIntraProcessWaitable`.
```
wjwwood: I think you mean to use `designated` here? `denominated` usually refers to numbering or categorization in my experience.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+An object of this type is created by each `Subscription` with intra-process communication enabled and it is used to notify the `Subscription` that a new message has been pushed into its ring buffer and that it needs to be processed.
+
+The `IntraProcessManager` class stores information about each `Publisher` and each `Subscription`, together with pointers to these structures.
+This allows to know which entities can communicate with each other and to have access to methods for pushing data into the buffers.
```
wjwwood: This allows who to know? The "system", "us", or something else?

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+It is added to the same callback group used for the standard inter-process communication of that topic.
+
+### Publishing only intra-process
+
+#### Publishing unique_ptr
+
+1. User calls `Publisher::publish(std::unique_ptr<MessageT> msg)`.
+2. `Publisher::publish(std::unique_ptr<MessageT> msg)` calls `IntraProcessManager::do_intra_process_publish(uint64_t pub_id, std::unique_ptr<MessageT> msg)`.
+3. `IntraProcessManager::do_intra_process_publish(...)` uses the `uint64_t pub_id` to call `IntraProcessManager::get_subscription_ids_for_pub(uint64_t pub_id)`.
+This returns the ids correspondin to `Subscription`s that have a QoS compatible for receiving the message.
```
wjwwood:
```suggestion
This returns the ids corresponding to `Subscription`s that have a QoS compatible for receiving the message.
```

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+These ids are divided into two sublists, according to the data-type that is stored in the buffer of each `Susbscription`: requesting ownership (`unique_ptr<MessageT>`) or accepting shared (`shared_ptr<MessageT>`, but also `MessageT` since it will copy data in any case).
+4. The message is "added" to the ring buffer of all the items in the lists.
+The `rcl_guard_condition_t` member of `SubscriptionIntraProcessWaitable` of each `Subscription` is triggered (this wakes up `rclcpp::spin`).
+
+The way in which the `std::unique_ptr<MessageT>` message is "added" to a buffer, depends on the type of the buffer.
+
+ - `BufferT = unique_ptr<MessageT>` The buffer receives a copy of `MessageT` and has ownership on it; for the last buffer, a copy is not necessary as ownership can be transferred.
+ - `BufferT = shared_ptr<const MessageT>` Every buffer receives a shared pointer of the same `MessageT`; no copies are required.
+ - `BufferT = MessageT` A copy of the message is added to every buffer.
```
wjwwood, line 214: When would this be preferable to a `shared_ptr<const MessageT>`? You could simply make the copy when ready to deliver the data rather than copying ahead of time. This way if the message is never taken, it does not need to be copied ever.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+![Sequence UML diagram](../img/intraprocess_communication/intra_process_only.png)
+
+#### Publishing other message types
+
+The `Publisher::publish(...)` method is overloaded to support different message types.
```
wjwwood:
```suggestion
The `Publisher::publish(...)` method is overloaded to support different message types:
```

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+As previously described, whenever messages are added to the ring buffer of a `Subscription`, a condition variable specific to the `Subscription` is triggered.
+This condition variable has been added to the `Node` waitset so it is being monitored by the `rclcpp::spin`.
+
+Remember that the `SubscriptionIntraProcessWaitable` object has access to the ring buffer and to the callback function pointer of its related `Subscription`.
+
+1. The guard condition linked with the `SubscriptionIntraProcessWaitable` object awakes `rclcpp::spin`.
+2. The `SubscriptionIntraProcessWaitable::is_ready()` condition is checked.
+This has to ensure that the ring buffer is not empty.
+3. The `SubscriptionIntraProcessWaitable::execute()` function is triggered.
```
wjwwood, line 240: It seems like the `is_read()` is redundant and race-like. Can't was just always run execute and let if fail gracefully if nothing is there?

Especially once you start to consider things like `lifespan` which might cause the message to be dropped (or become "un-take-able") between calling `is_ready()` and `execute()` (which internally I guess calls a `take`/`get`/`pop` kind of method on the buffer which is thread-safe.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+Note that in this step, if the type of the buffer is a smart pointer one, no message copies occurr, as ownership has been already taken into account when pushing a message into the queue.
+
+
+### Publishing intra and inter-process
+
+1. User calls `Publisher::publish(std::unique_ptr<MessageT> msg)`.
+2. The message is moved into a shared pointer `std::shared_ptr<MessageT> shared_msg = std::move(msg)`.
+3. `Publisher::publish(std::unique_ptr<MessageT> msg)` calls `IntraProcessManager::do_intra_process_publish(uint64_t pub_id, std::shared_ptr<MessageT> shared_msg)`.
+
+The following steps are identical to steps 3, 4 and 5 applied when publishing only intra-processs.
```
wjwwood:
```suggestion
The following steps are identical to steps 3, 4, and 5 applied when publishing only intra-process.
```

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+4. `IntraProcessManager::do_intra_process_publish(...)` uses the `uint64_t pub_id` to call `IntraProcessManager::get_subscription_ids_for_pub(uint64_t pub_id)`.
+Then it calls `IntraProcessManager::find_matching_subscriptions(PublisherInfo pub_info)`.
+This returns the ids correspondin to `Subscription`s that have a QoS compatible for receiving the message.
```
wjwwood:
```suggestion
This returns the ids corresponding to `Subscription`s that have a QoS compatible for receiving the message.
```

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+After the intra-process publication, the inter-process one takes place.
+
+6. `Publisher::publish(std::unique_ptr<MessageT> msg)` calls `Publisher::do_inter_process_publish(const MessageT & inter_process_msg)`, where `MessageT inter_process_msg = *shared_msg`.
+
+The difference from the previous case is that here a `std::shared_ptr<const MessageT>` is being "added" to the buffers.
+Note that this `std::shared_ptr` has been just created from a `std::unique_ptr` and it is only used by the `IntraProcessManager` and by the RMW, while the user application has no access to it.
+
+ - `BufferT = unique_ptr<MessageT>` The buffer receives a copy of `MessageT` and has ownership on it.
+ - `BufferT = shared_ptr<const MessageT>` Every buffer receives a shared pointer of the same `MessageT`, so no copies are required.
+ - `BufferT = MessageT` A copy of the message is added to every buffer.
```
wjwwood, line 271: This seems out of place here and the exact same exists above, is this supposed to be here?

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+The difference with publishing a unique_ptr is that here it is not possible to move the ownership of the message to one of the `Subscription`, potentially saving a copy.
```
wjwwood: I agree you cannot save a copy, but you can at least give the original to one of the subscriptions, which I think is worth preserving for cases where the address is important or just even for demonstrating the feature is working in tutorials or tests by comparing pointers.

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+
+![Sequence UML diagram](../img/intraprocess_communication/intra_inter_process.png)
+
+### QoS features
+
+The proposed implementation can handle all the different QoS.
+
+ - If the history is set to `keep_last`, then the depth of the history corresponds to the size of the ring buffer.
+ On the other hand, if the history is set to `keep_all`, the buffer becomes a standard FIFO queue with an unbounded size.
+ - The reliability is only checked by the `IntraProcessManager` in order to understand if a `Publisher` and a `Subscription` are compatible.
```
wjwwood, line 284: I mentioned this above, but this may also affect how the behavior of how a publisher should block when subscriptions are full.

articles/intraprocess_communication.md
wjwwood: I don't know if there is an order because I think it may be asynchronous when DDS does this, it "requests" the back filled messages and then they arrive later. See this conversation starting here:

https://github.com/ros2/geometry2/pull/160#pullrequestreview-285066035

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+This means that the middleware is not able to store old messages for eventual late-joiners.
+
+The solution to this issue consists in always publishing both intra and inter-process when a `Publisher` has `transient local` durability.
+For this reason, when `transient local` is enabled, the `do_intra_process_publish(...)` function will always process a shared pointer.
+This allows us to add the logic for storing the published messages into the buffers only in one of the two `do_intra_process_publish(...)` cases and also it allows to use buffers that have only to store shared pointers.
+
+
+### Number of message copies
+
+
```
wjwwood:
```suggestion
```

articles/intraprocess_communication.md
```diff
@@ -0,0 +1,494 @@
...
+| ----------------------- | ----------------------- | ----------------------- |
+| shared_ptr\<MsgT\> @1   | unique_ptr\<MsgT\>      |                     @2  |
+| shared_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> unique_ptr\<MsgT\>     |     @2 <br> @3 |
+| shared_ptr\<MsgT\> @1   | shared_ptr\<MsgT\>      |                     @1  |
+| shared_ptr\<MsgT\> @1   | shared_ptr\<MsgT\> <br> shared_ptr\<MsgT\>     |     @1 <br> @1 |
+| shared_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> shared_ptr\<MsgT\>     |     @2 <br> @1 |
+| shared_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> shared_ptr\<MsgT\>   <br> shared_ptr\<MsgT\>   |  @2 <br> @1 <br> @1|
+| shared_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> unique_ptr\<MsgT\> <br> shared_ptr\<MsgT\>   <br> shared_ptr\<MsgT\>   |  @2 <br> @3 <br> @1 <br> @1|
+
+
```
wjwwood:
```suggestion
```

articles/intraprocess_communication.md
wjwwood: I think this only makes sense if you also avoid the necessity that all intra-process starts as a `unique_ptr`. Once you've created or were given a unique pointer, you can either just deliver that (in the case of a single owning subscription) and/or make a single copy into a shared pointer until you're ready to deliver each (in the case of more subscriptions) and only when they're taken make another copy.

wjwwood (commented):
I've just finished going over this again. Some of my comments maybe don't apply after reading further in the document, but I tried to remove those that I could.

I'm going through the implementation pr simultaneously and will continue that in the next few days.

articles/intraprocess_communication.md
alsora: https://wikidiff.com/denominate/designate

articles/intraprocess_communication.md
alsora: Should I also change the name of the file?

articles/intraprocess_communication.md
alsora: I'll add something like
```
This results in the loss of the message and it is also a difference in behavior between intra and inter-process communication, since, with the latter, the message would have been received.
```

articles/intraprocess_communication.md
alsora: Yes, you are right, but a problem is present if the message will not be overwritten.
This may happen in cases of a publisher which publishes at a slow rate combined with a long history (i.e. its queue will never be completely full) or in case of a keep all policy.

articles/intraprocess_communication.md
alsora: A couple of lines below I specify that 
```
Note that these messages will be discarded, but they will still cause an overhead.
```

And this is exactly obtained using the `ignore_local_publication` flag.  The overhead comes from the middleware layer

Do you think that I should describe this a little bit more?

articles/intraprocess_communication.md
alsora: Nope.
Honestly I barely know the specification of these new QoS.

I can update the sentence specifying which exactly QoS are supported.
Do you know how the new ones are implemented? Do you think it could be problematic to support them?

articles/intraprocess_communication.md
alsora: The first thing that comes to my mind is that we can block as soon as one of the subscription buffers is full.
I'm wondering if this may cause problems if both reliable and best-effort subscriptions are present

articles/intraprocess_communication.md
alsora: The diagram shows the "default" scenario, i.e. all entities are volatile.

articles/intraprocess_communication.md
alsora: The original idea behind that was to use it for very small data (i.e. base types) especially in case of high number of subscriptions. In this way you avoid the overhead of synchronizing the shared pointers and of having many threads accessing the same memory address.

I did some quick tests, but I haven't seen noticeable improvements. 
There are obviously alternatives, such as using unique pointers that are already supported.

articles/intraprocess_communication.md
alsora: This looks more an implementation-related discussion.

I can remove point 2 from the design specification.

articles/intraprocess_communication.md
alsora: It's **almost** the same. The first point is different as explained in the following line.
```
The difference with publishing a unique_ptr is that here it is not possible to move the ownership of the message to one of the `Subscription`, potentially saving a copy.
```

However, reading it again after 2 months confused also me. It's probably better to only highlight the difference with the unique pointer.

articles/intraprocess_communication.md
alsora: I don't think you can.
Here it's describing the situation in which you are publishing both inter and intra process.
The IntraProcess manager does not have owenership of the message.

I'm pretty sure that this applies also to the current implementation.
The only way for preserving the address is to publish only intra-process a unique pointer.

alsora:
Github does not allow me to reply to the last comment https://github.com/ros2/design/pull/239#discussion_r326365860 ...
However, I agree with you. When I tested this I also added new APIs for publish and callbacks that do not use pointers.

articles/intraprocess_communication.md
ivanpauno: I also thought about this and I think that lifetime related QoS are the only ones that are much harder to implement in this way than sending a meta-message.

See for example:
- https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/DEADLINE_QosPolicy.htm
- https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/LIFESPAN_QoS_Policy.htm

alsora: I'm thinking that is probably worth restricting intra-process communication to a simple, scenario. This also will allow to better optimize it.

No matter which intra-process implementation we will end up having: as long as it is in the `rclcpp` layer it will require a not negligible amount of work to implement and support quality of service and all this will be at the expenses of optimization which should be the main purpose for enabling this feature.

IMO we should only care about the default DDS settings and eventually whether in such a configuration the behaviors of inter and intra-process are coincident.

If an user wants to specify QoS or DDS features, he should go through the RMW and rely on the optimization implemented in the underlying middleware.

Some DDS vendors already have (CycloneDDS) or are implementing (FastRTPS) shared memory communication.
The intra-process communication at the `rclcpp` layer would also act as a benchmark for the performances of all the DDS, allowing to easily evaluate the overhead introduced by a specific middleware.

Intra-process communication would be a feature that has to be manually enabled by the user (i.e. it will not be a default).

ivanpauno: Yes, I completely agree that we shouldn't focus on supporting all the QoS policies.
Doing that at `rclcpp` level would be hard.

At some point in the future, we should just wrap an intra-process (or inter-process) API provided by the middle-ware itself (there is a proposal about that, see https://github.com/ros2/design/pull/256).

articles/intraprocess_communication.md
wjwwood: Up to you. I think as a file name, you could consider collapsing the `-` into nothing (and therefore leaving it as is), but I don't feel strongly about that.

articles/intraprocess_communication.md
wjwwood: It's not necessarily true that when using `ignore_local_publication` that there is overhead in the middleware. Some DDS implementations will do publisher (datawriter) side filtering in this case, so no unused data is received.

Either way, I think the sentence being referenced here is factually inaccurate, because there is an API that deals with this in part, regardless of whether or not the way it is implemented is efficient or not (though depending on the middleware it may actually be quite efficient).

articles/intraprocess_communication.md
wjwwood: > No matter which intra-process implementation we will end up having: as long as it is in the rclcpp layer it will require a not negligible amount of work to implement and support quality of service and all this will be at the expenses of optimization which should be the main purpose for enabling this feature.

I've said this before, but it would be non-negligible work in both cases, but using the meta message would be significantly simpler for many cases, including these new QoS (in my opinion).

That being said, I think the benefits of the new system are worth it even if we cannot support as many QoS settings as easily.

> Intra-process communication would be a feature that has to be manually enabled by the user (i.e. it will not be a default).

That's not clear to me right now. I think there is a good case for intra-process being on by default.

> At some point in the future, we should just wrap an intra-process (or inter-process) API provided by the middle-ware itself (there is a proposal about that, see #256).

I'm not sure this ever 100% replace intra-process in rclcpp, so until that happens, it's not a great reason to ignore this issue.

wjwwood: > I can update the sentence specifying which exactly QoS are supported.

Please do update it, if you can, so we can get this merged. In general, I think it would be good to avoid superlatives or all inclusive statements like this. Instead, enumerating what you support is more likely to age well as things change around the document.

articles/intraprocess_communication.md
wjwwood: Blocking when one of the subscription buffers is full does not emulate the behavior of the middleware, as the middleware will also have a publisher buffer which must fill up before blocking.

articles/intraprocess_communication.md
wjwwood: Hmm, that's not clear from the diagram, maybe there's some way to clarify that?

articles/intraprocess_communication.md
wjwwood: It still doesn't sound right... how about just "which is named ..."?

articles/intraprocess_communication.md
wjwwood: Sounds good.

articles/intraprocess_communication.md
wjwwood: I don't think that's true, a copy is made for inter-process and intra-process owns the message. That's the entire reason the pointer addresses remain the same in this tutorial:

https://index.ros.org/doc/ros2/Tutorials/Intra-Process-Communication/#the-cyclic-pipeline-demo

clalancette:
@alsora Now that https://github.com/ros2/rclcpp/pull/778 is merged, is there anything we need to update here?

alsora:
No I think this is fine. 
On the rclcpp side a follow up PR will be necessary to re introduce the use of intra process manager impl class, but it should not affect the design

clalancette:
> No I think this is fine.
> On the rclcpp side a follow up PR will be necessary to re introduce the use of intra process manager impl class, but it should not affect the design

All right, thanks.  @wjwwood @ivanpauno You two had the earlier comments and still some unresolved conversations.  Are you satisfied with this article now, or do we still need to resolve the open conversations?

articles/intraprocess_communication.md
ivanpauno: This has to be updated after https://github.com/ros2/rclcpp/pull/778/commits/3013e02cc0f4a97083510a29a99f3d99304d6d87

alsora:
Should  we merge this?

ivanpauno (approved):

ivanpauno:
Merged. If there are any further comments, they can be addressed in a follow-up.

ros-discourse:

https://discourse.ros.org/t/intra-process-communications-for-all-clients/10416/1

PR #241, opened 2019-07-04
Extend ROS resource addressing design document

hidmic:
This pull request generalizes the existing topic and service name addressing design document to cover the current goals in terms of addressing topics, services, actions and parameters in ROS 2.


hidmic:
As a follow-up, the design document on static name remapping will need an update as well.

articles/140_resource_names.md
```diff
@@ -1,10 +1,10 @@
 ---
 layout: default
-title: Topic and Service name mapping to DDS
-permalink: articles/topic_and_service_names.html
+title: Resource Names
+permalink: articles/resource_names.html
```
wjwwood, line 4: I'm hesitant to rename the permalink, maybe we can setup a redirect? I just don't want to break existing links if we can help it.

articles/140_resource_names.md
```diff
@@ -19,13 +19,33 @@ published: true
...
 Original Author: {{ page.author }}
 
-## Context
+## Background
 
-Before proposing constraints for ROS 2 topic and service names and a mapping to underlying DDS topics, this article will first summarize the existing guidelines and restrictions for both ROS 1 and DDS.
+### Concepts
 
-### ROS 1 Names
+#### Content-oriented resources
```
wjwwood, line 27: I think it would be good to motivate our use of this terminology (and potentially change it if something else fits better). There is some stuff on wikipedia, but it's mostly focused on specific technologies like actually implementing internet networks using named resources or are tightly coupled to web technologies:

- https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed
- https://en.wikipedia.org/wiki/Content_centric_networking
- https://en.wikipedia.org/wiki/Information-centric_networking
- https://en.wikipedia.org/wiki/Named_data_networking
- https://en.wikipedia.org/wiki/Resource-oriented_architecture

I had trouble finding an article which summarized the basic difference in addressing between having a host and not. The URI page talks about host and path, and that host is optional:

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Definition

I don't think we need to spend much time here, but a few citations might lend credibility to the distinction we're making here.

articles/140_resource_names.md
```diff
@@ -19,13 +19,33 @@ published: true
...
-In ROS 1, topic and service name guidelines are all covered under the umbrella of "ROS Names" and have these restrictions:
+Content-oriented resources are not bound to any particular node in a given system.
+ROS topics, which may be freely published or subscribed to by multiples ROS nodes, fall under this classification.
+
+Note that in ROS 1, every resource i.e. topics, services, actions and parameters, is a content-oriented resource.
+
+#### Host-oriented resources
+
+Host-oriented resources are provided by a node in a given system and thus are bound to it.
+Multiple resources with the same name and type may by provided by different nodes.
```
wjwwood:
```suggestion
Multiple resources with the same name and type may be provided by different nodes.
```

articles/140_resource_names.md
```diff
@@ -19,13 +19,33 @@ published: true
...
+Therefore, the provider node name is necessary to unambiguously address a host-oriented resource.
+In the preceding, it is assumed that node name uniqueness has been ensured by means not prescribed in this document.
+
+ROS services, actions and parameters fall under this classification.
```
wjwwood:
```suggestion
ROS services, actions, and parameters fall under this classification.
```

Subjective, but we use the Oxford comma in other documents, so I recommend it here for consistency.

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
-- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), or forward slashes (`/`)
-- may use balanced curly braces (`{}`) for substitutions
-- may start with a tilde (`~`), the private namespace substitution character
-- must not start with a numeric character (`[0-9]`)
-- must not end with a forward slash (`/`)
-- must not contain any number of repeated forward slashes (`/`)
-- must not contain any number of repeated underscores (`_`)
-- must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`
-- must have balanced curly braces (`{}`) when used, i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`
+#### Provider Names
```
wjwwood: This goes along with the other comment I made, I think we should consider adopting the URI terminology where it makes sense. Would `host` make more sense than `provider` in this case? I'm not 100% sure, but I'd like to know your opinion.

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
 
-The content of substitutions, i.e. the string in side of balanced curly braces (`{}`), follow very similar rules to names.
-The content of substitutions:
+For host-oriented resources i.e. services, actions and parameters, the provider name refers to the node that is exposing said resources.
```
wjwwood:
```suggestion
For host-oriented resources, i.e. services, actions, and parameters, the provider name refers to the node that is exposing said resources.
```

articles/140_resource_names.md
```diff
@@ -19,13 +19,33 @@ published: true
...
+### Concepts
 
-### ROS 1 Names
+#### Content-oriented resources
 
-In ROS 1, topic and service name guidelines are all covered under the umbrella of "ROS Names" and have these restrictions:
+Content-oriented resources are not bound to any particular node in a given system.
+ROS topics, which may be freely published or subscribed to by multiples ROS nodes, fall under this classification.
+
+Note that in ROS 1, every resource i.e. topics, services, actions and parameters, is a content-oriented resource.
```
wjwwood:
```suggestion
Note that in ROS 1, every resource, i.e. topics, services, actions, and parameters, is a content-oriented resource.
```

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
-The content of substitutions:
+For host-oriented resources i.e. services, actions and parameters, the provider name refers to the node that is exposing said resources.
 
-- must not be empty
-- may contain alphanumeric characters (`[0-9|a-z|A-Z]`) and underscores (`_`)
-- must not start with a numeric character (`[0-9]`)
+##### Definition
 
-### Fully Qualified Names
+A provider name is made up of one or more tokens, delimited by periods (`.`) i.e. using URL hostname style syntax.
```
wjwwood: The structure of this is covered by FQDN's I think (https://en.wikipedia.org/wiki/Fully_qualified_domain_name) if you want another citation.

wjwwood: We use the "fully qualified [node] name" phrase in places already.

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
-For example, as outlined in further detail in the sections that follow, names may be relative (e.g. `foo` versus the absolute `/foo`), they may contain the private namespace substitution character (`~`), or arbitrary substitutions which are within the curly braces (`{}`) syntax.
-With context, each of these features can be expanded to some simple string to form the fully qualified name.
-Fully qualified names have these additional restrictions:
+Each token:
 
-- must start with a forward slash (`/`), i.e. they must be absolute
-- must not contain tilde (`~`) or curly braces (`{}`)
+- must not be empty.
+- may use alphanumeric characters (`[0-9|a-z|A-Z]`) and/or underscores (`_`) only.
+- must not start with numeric characters (`[0-9]`).
```
wjwwood, line 95: Are they also not allowed to have `__` in the name? I think that might be a restriction for node names and/or namespaces right now.

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
-Note the triple forward slash (`/`), which is a similar style to the `file://` scheme.
-A relative name `foo/bar` could would be represented (as a topic) with `rostopic://foo/bar`.
+The last token in a provider name is its base name and any preceding tokens make up its namespace.
+For example, a `foo.bar.baz` provider name is made up of `foo`, `bar` and `baz` tokens, and refers to a node in the `foo.bar` namespace whose base name is `baz`.
 
-### ROS 2 Name Examples
+#### Resource Names
 
-For example, these are valid names:
+All resources, whether content-oriented or host-oriented, have a name.
```
wjwwood, line 110: I like name myself, but again looking to converge on "standard" terminology, I believe this would be referred to as the `path` in URI parlance. Perhaps just an aside the first time we mention it (here)? Like `have a name (the 'path' of the URI)`? I don't know what would be more appropriate, I just wanted to bring it up.

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
-| `~`        | `foo/bar`  | `~/foo` | `{foo}_bar`         | `foo/{ping}/bar`     |
-| `foo/_bar` | `foo_/bar` | `foo_`  | `rosservice:///foo` | `rostopic://foo/bar` |
+##### Definition
 
-But these are not valid names:
+A resource name is made up of one or more tokens, delimited by forward slashes (`/`) i.e. using URL path style syntax.
 
-| `123abc`    | `123`  | `foo bar`  | ` `        | `foo//bar` |
-| `/~`        | `~foo` | `foo~`     | `foo~/bar` | `foo/~bar` |
-| `foo/~/bar` | `foo/` | `foo__bar` |            |            |
```
wjwwood, line 113: Hmm, this (`foo__bar`) was in the "not valid name" list, but I didn't see it in the bullet points any where. I know it was part of a previous alternative we tried (described at the bottom of the document), but I wonder if the code checks for this still...

wjwwood: Looks like not, so forget my previous comment: https://github.com/ros2/rcl/blob/21ff57b14b6f406d948f6e9ea97db75a1b2004ef/rcl/src/rcl/validate_topic_name.c#L97-L99

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
+Content-oriented resource names may be absolute or relative.
 
-### Private Namespace Substitution Character
+An absolute name begins with a forward slash (`/`) and it's not affected by the addressing node's namespace.
+It can be regarded as a _global_ name, or a name at the _root_ namespace.
+
+A relative name does not begin with a forward slash (`/`) and it is resolved _relative_ to the addressing node's namespace.
+Periods (`.`) in the addressing node's namespace are replaced by forward slashes (`/`) and a leading forward slash (`/`) is added during resolution.
+
+For example, if a `ping.pong.ball` node addresses (e.g. subscribes to) a `foo/bar` topic, then the topic name will be resolved to `/ping/pong/foo/bar`.
```
wjwwood:
```suggestion
For example, if a node with the address `ping.pong.ball` publishes or subscribes to a topic `foo/bar`, then the topic name will be resolved to `/ping/pong/foo/bar`.
```

articles/140_resource_names.md
```diff
@@ -38,152 +58,157 @@ A valid name has the following characteristics:
...
+##### Substitutions
 
-### Substitutions
+The bracket syntax (`{substitution_name}`) may be used in resource names to substitute useful contextual information into the them.
+The set of substitution keys (names) are not set in this document, but some reasonable examples might be `{node}`, expanding to the addressing node's name, or `{ns}`, expanding to the addressing node's namespace.
 
-The bracket syntax (`{substitution_name}`) may be used in non-fully qualified names to substitute useful contextual information into the name.
-The set of substitution keys (names) are not set in this document, but some reasonable examples might be: `{node}` expands to the current node's name or `{ns}` expands to the current node's namespace.
+Substitutions must result in a valid name and comply with the name constraints.
+An example of an invalid substitution would be `{sub}/foo` and replace `{sub}` with a numeric value, which thus leads to a (invalid) resource name starting with a numeric character.
```
wjwwood:
```suggestion
An example of an invalid substitution would be `{sub}/foo` and replace `{sub}` with a numeric value, which thus leads to an *invalid* resource name starting with a numeric character.
```

articles/140_resource_names.md
```diff
@@ -192,81 +217,153 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
 This is implicitly enforced by the rules above that say substitution names may only contain alphanumerics and underscores (`_`).
 For example, given the name `{% raw %}/foo/{{bar}_baz}{% endraw %}` would result in an error because `{` and `}` are not allowed in a substitution names and the substitution name `{bar}_baz` does contain them.
 
-### Hidden Topic or Service Names
+#### Fully Qualified Resource Names (FQRNs)
+
+Regular resource names resolution rules allow for convenient syntax, which in some cases requires additional context to expand to a fully qualified name that can be mapped to e.g. a DDS topic name.
```
wjwwood:
```suggestion
Regular resource name resolution rules allow for convenient syntax, which in some cases requires additional context to expand to a fully qualified name that can be mapped to another type of name, e.g. a DDS topic name.
```

articles/140_resource_names.md
```diff
@@ -192,81 +217,153 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
+- must start with a forward slash (`/`), i.e. they must be absolute.
+- must not contain tilde (`~`) nor balanced curly braces (`{}`).
+
+### Resource Qualification
 
-Any topic or service name that contains any tokens (either namespaces or a topic or service name) that start with an underscore (`_`) will be considered hidden and tools may not show them unless explicitly asked.
+#### Hidden nodes
 
-## Mapping of ROS 2 Topic and Service Names to DDS Concepts
+If a (resource provider) node name contains any tokens that start with an underscore (`_`), the node it refers to will be considered hidden and tools may not show them unless explicitly asked to.
```
wjwwood:
```suggestion
If a node name (a.k.a. resource provider) contains any tokens that start with an underscore (`_`), the node it refers to will be considered hidden and tools may not show them unless explicitly asked to do so.
```

articles/140_resource_names.md
```diff
@@ -192,81 +217,153 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
 
-The ROS topic and service name constraints allow more types of characters than the DDS topic names because ROS additionally allows the forward slash (`/`), the tilde (`~`), and the balanced curly braces (`{}`).
-These must be substituted or otherwise removed during the process of converting the topic or service name to DDS concepts.
-Since ROS 2 topic and service names are expanded to fully qualified names, any balanced bracket (`{}`) substitutions and tildes (`~`) will have been expanded.
-Additionally any URL related syntax, e.g. the `rostopic://` prefix, will be removed once parsed.
-Previously forward slashes (`/`) were disallowed in DDS topic names, now the restriction has been lifted (see [issue](https://issues.omg.org/issues/lists/dds-rtf5#issue-42236) on omg.org) and therefore the ROS topic names are first prefixed with ROS Specific Namespace prefix (described below) are then mapped directly into DDS topic names.
+#### Hidden resources
+
+If a resource name contains any tokens that start with an underscore (`_`), the resource it refers to will be considered hidden and tools may not show them unless explicitly asked to.
```
wjwwood:
```suggestion
If a resource name contains any tokens that start with an underscore (`_`), the resource it refers to will be considered hidden and tools may not show them unless explicitly asked to do so.
```

articles/140_resource_names.md
```diff
@@ -192,81 +217,153 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
+
+For convenience, here is a summary of all the rules that apply to URLs in ROS.
+
+Provider names in a URL:
+
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`) and periods (`.`).
+- must not start with a numeric character (`[0-9]`).
+- must not end nor start with a period (`.`).
+- must not contain any number of repeated periods (`.`).
+- must not contain any number of repeated underscores (`_`).
```
wjwwood, line 274: So this is back :D. We should figure out what we're doing here and be consistent. Unless I missed it this wasn't a constraint mentioned above. I think we may have wanted to keep it to allow us to have a "safe" delimiter to use in the future, but it looks like the code doesn't check this, at least for topics.

articles/140_resource_names.md
```diff
@@ -192,81 +217,153 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`) and underscores (`_`).
+- must not start with a numeric character (`[0-9]`).
+
+## Mapping Resource Names to DDS Concepts
+
+### Naming Constraints 
+
+#### Allowed characters
+
+ROS URL syntax allows for more characters than e.g. DDS topic names allow: the forward slash (`/`), the tilde (`~`), the balanced curly braces (`{}`), the period (`.`) and the semicolon (`:`).
```
wjwwood:
```suggestion
The ROS URL syntax allows for more characters than are allowed in DDS topic names.
For example, the forward slash (`/`), the tilde (`~`), the balanced curly braces (`{}`), the period (`.`) and the semicolon (`:`) are all allowed in the ROS URL syntax, but are not allowed in DDS topic names.
```

and actually we are using `/` in the DDS topic names, we just have a understanding with the DDS vendors that in the future they will update the standard to allow at least `/` in addition to the current constraints. And for now, all the implementations allow us to use them.

articles/140_resource_names.md
```diff
@@ -192,81 +217,153 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
+These must be substituted or otherwise removed during the process of converting a ROS URL to e.g. a DDS topic name.
```
wjwwood:
```suggestion
These must be substituted or otherwise removed during the process of converting a ROS URL to another kind of name, e.g. a DDS topic name.
```

articles/140_resource_names.md
wjwwood: I see your not below now, maybe we should mention that earlier, like in the DDS topic name restrictions overview near the beginning of the article.

articles/140_resource_names.md
hidmic: Sure, we can! See redirect in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: > I think it would be good to motivate our use of this terminology

I agree, and I think that CCN (Content Centric Networking) gets pretty close. A related [Linux Foundation project](https://wiki.fd.io/view/Cicn) adds Publish/Subscribe Networking as a potential alias for the same concept. See 7680627ee8f2fa8675fd58e72f814585c6721f2a.

> The URI page talks about host and path, and that host is optional:

For URIs, a host is always defined, either explicitly or implicitly through a sane default e.g. `localhost` for `file://` URIs, see [RFC 3986 Section 3.2.2](https://tools.ietf.org/html/rfc3986#section-3.2.2).

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Sounds good. Added in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: So, the rationale here is that since nodes provide services instead of hosting data it made more sense to use the former terminology, but the distinction is subtle. I'm fine with `host` too. See 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: That's right. Added missing constraint in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: I agree with establishing a relationship between resource name and path, but I wouldn't switch to URI parlance. Talking about topic paths coming from a ROS 1 background is somewhat awkward. See 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: I cleaned up the examples table and removed the `foo__bar` entry in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Made this constraint explicit everywhere else in 7680627ee8f2fa8675fd58e72f814585c6721f2a. Do you think we shouldn't?

articles/140_resource_names.md
hidmic: Corrected phrasing, moved the note and took away forward slashes as a forbidden character in DDS topic names in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
hidmic: Corrected in 7680627ee8f2fa8675fd58e72f814585c6721f2a.

articles/140_resource_names.md
wjwwood: Yeah, I don't feel strongly about it, host is fine and I would be ok with provider as well so long as there is a rationale and it's not arbitrary.

articles/140_resource_names.md
wjwwood: Hmm, not sure. I think we only _needed_ it for the alternative we're no longer using. I don't think the code enforces it at the moment. Maybe it should be removed. @sloretz or anyone on the @ros2/team do you have an opinion here?

wjwwood (approved):

articles/140_resource_names.md
```diff
@@ -1,10 +1,11 @@
...
 layout: default
-title: Topic and Service name mapping to DDS
-permalink: articles/topic_and_service_names.html
+title: Resource Names
+permalink: articles/resource_names.html
+redirect_from: articles/topic_and_service_names.html
 abstract:
-  This article describes the proposed mapping between ROS topic and service names to DDS topic and service names.
-author: '[William Woodall](https://github.com/wjwwood)'
+  This article describes an specification proposal for resource addressing in ROS, as well as its implementation in a DDS setting.
```
jacobperron:
```suggestion
  This article describes a specification proposal for resource addressing in ROS, as well as its implementation in a DDS setting.
```

articles/140_resource_names.md
```diff
@@ -192,81 +219,155 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
+- must not be empty.
+- may contain alphanumeric characters (`[0-9|a-z|A-Z]`), underscores (`_`), and forward slashes (`/`).
+- may use balanced curly braces (`{}`) in the resource name for substitutions.
+- may start with a tilde (`~`), the private namespace substitution character.
+- must not start with a numeric character (`[0-9]`).
+- must not end with a forward slash (`/`).
+- must not contain any number of repeated forward slashes (`/`).
+- must not contain any number of repeated underscores (`_`).
+- must separate a tilde (`~`) from the rest of the name with a forward slash (`/`), i.e. `~/foo` not `~foo`.
+- must have balanced curly braces (`{}`) when used, i.e. `{sub}/foo` but not `{sub/foo` nor `/foo}`.
```
jacobperron: Is this point redundant with L280?

articles/140_resource_names.md
```diff
@@ -192,81 +219,155 @@ Substitutions are also not allowed to be nested, i.e. substitutions may not cont
...
+
+The length of a DDS topic name must not exceed 256 characters. Therefore the length of a ROS topic name, including the namespace hierarchy, the base name of the topic and any ros specific prefixes must not exceed 256 characters since this is mapped directly as DDS topic.
+
+##### Considerations for RTI Connext
+
+While testing our implementation with Connext, we encountered some additional limitations when it comes to the topic length.
+That is, for example the length of a service name has tighter limits than the length of a topic.
+In the RTI Connext implementation, service names are suffixed with the GUID value of the DDS participant for the service response topic.
+Additionally, a content filtered topic (max length 256 characters) is created which is mapped from the suffixed service name.
+Therefore when linking against `rmw_connext_c` or `rmw_connext_cpp`, service names cannot be longer than 185 characters including the namespace hierarchy and any ros specific prefixes.
```
jacobperron:
```suggestion
Therefore when linking against `rmw_connext_c` or `rmw_connext_cpp`, service names cannot be longer than 185 characters including the namespace hierarchy and any ROS specific prefixes.
```

articles/140_resource_names.md
jacobperron: I don't have a strong preference. But it seems okay to remove the constraint if there's no reason to have it. If we really need it in the future, we could start using repeated `.` as a delimiter since it's already disallowed.

articles/140_resource_names.md
hidmic: Corrected in c753fe9dd2f9bdfe03f92245cab4c80b932a70ed.

articles/140_resource_names.md
hidmic: Good catch. Removed in c753fe9dd2f9bdfe03f92245cab4c80b932a70ed.

articles/140_resource_names.md
hidmic: Corrected in c753fe9dd2f9bdfe03f92245cab4c80b932a70ed.

jacobperron (approved):

mjcarroll:
@hidmic What is the status on this?  Can it be merged or does it need to move forward to the Foxy board?

hidmic:
It has yet to be implemented, so if anything it has to go to the Foxy board. It was brought up during our brainstorming for Foxy, but AFAIK it has a rather low priority.

hidmic:
https://github.com/ros2/ros2/issues/881 to track progress on this topic.

PR #243, opened 2019-07-15
Add article about how ROS 2 integrates with DDS-Security

kyrofa:
This is not a new design proposal. Rather, it is a description of the way ROS 2 _currently_ integrates with DDS-Security (i.e. SROS 2). No new ideas are contained here; this is meant to fill a hole in existing design documentation and provide a good basis for future design of SROS 2, providing necessary context both for the designer as well as the reviewer.

### DIFF
--- a/articles/ros2_dds_security.md
+++ b/articles/ros2_dds_security.md
@@ -29,7 +29,8 @@ Original Author: {{ page.author }}
 
 # DDS-Security overview
 
-The [DDS-Security specification][dds_security] expands upon the [DDS specification][dds], adding security enhancements by defining a Service Plugin Interface (SPI) architecture, a set of builtin implementations of the SPIs, and the security model enforced by the SPIs. Specifically, there are five SPIs defined:
+The [DDS-Security specification][dds_security] expands upon the [DDS specification][dds], adding security enhancements by defining a Service Plugin Interface (SPI) architecture, a set of builtin implementations of the SPIs, and the security model enforced by the SPIs.
+Specifically, there are five SPIs defined:
 
 - **Authentication**: Verify the identity of a given domain participant.
 - **Access control**: Enforce restrictions on the DDS-related operations that can be performed by an authenticated domain participant.
--- a/articles/ros2_dds_security.md
+++ b/articles/ros2_dds_security.md
@@ -2,8 +2,12 @@
 layout: default
 title: ROS 2 DDS-Security integration
 permalink: articles/ros2_dds_security.html
-abstract:
-  Robotics is full of experimentation: evaluating different hardware and software, pushing ahead with what works, and culling what doesn't. ROS was designed to be flexible to enable this experimentation; to allow existing components to easily be combined with new ones or swapped with others. In ROS 1, this flexibility was valued above all else, at the cost of security. By virtue of being designed on top of DDS, ROS 2 is able to retain that flexibility while obtaining the ability to be secured by properly utilizing the DDS-Security specification. This article describes how ROS 2 integrates with DDS-Security.
+abstract: >
+  Robotics is full of experimentation: evaluating different hardware and software, pushing ahead with what works, and culling what doesn't.
+  ROS was designed to be flexible to enable this experimentation; to allow existing components to easily be combined with new ones or swapped with others.
+  In ROS 1, this flexibility was valued above all else, at the cost of security.
+  By virtue of being designed on top of DDS, ROS 2 is able to retain that flexibility while obtaining the ability to be secured by properly utilizing the DDS-Security specification.
+  This article describes how ROS 2 integrates with DDS-Security.
 author:  >
   [Kyle Fazzari](https://github.com/kyrofa)
 
@@ -33,43 +37,55 @@ The [DDS-Security specification][dds_security] expands upon the [DDS specificati
 - **Logging**: Provides the ability to audit DDS-Security-related events
 - **Data tagging**: Provides the ability to add tags to data samples
 
-ROS 2's security features currently utilize only the first three of these, as not all DDS implementations support the **Logging** or **Data Tagging** plugins. Let's delve a little further into those first three plugins.
+ROS 2's security features currently utilize only the first three of these, as not all DDS implementations support the **Logging** or **Data Tagging** plugins.
+Let's delve a little further into those first three plugins.
 
 
 ## Authentication
 
 The **Authentication** plugin (see section 8.3 of the [DDS-Security spec][dds_security]) is central to the entire SPI architecture, as it provides the concept of a confirmed identity without which further enforcement would be impossible (e.g. it would be awfully hard to make sure a given ROS node could only access specific topics if it was impossible to securely determine which node it was).
 
-The SPI architecture allows for a number of potential authentication schemes, but ROS 2 uses the builtin authentication plugin (called "DDS:Auth:PKI-DH", see section 9.3 of the [DDS-Security spec][dds_security]), which uses the proven Public Key Infrastructure (PKI). It requires a public and private key per domain participant, as well as an x.509 certificate that binds the participant's public key to a specific name. Each x.509 certificate must be signed by (or have a signature chain to) a specific Certificate Authority (CA) that the plugin is configured to trust.
+The SPI architecture allows for a number of potential authentication schemes, but ROS 2 uses the builtin authentication plugin (called "DDS:Auth:PKI-DH", see section 9.3 of the [DDS-Security spec][dds_security]), which uses the proven Public Key Infrastructure (PKI).
+It requires a public and private key per domain participant, as well as an x.509 certificate that binds the participant's public key to a specific name.
+Each x.509 certificate must be signed by (or have a signature chain to) a specific Certificate Authority (CA) that the plugin is configured to trust.
 
 
 ## Access control
 
-The **Access control** plugin (see section 8.4 of the [DDS-Security spec][dds_security]) deals with defining and enforcing restrictions on the DDS-related capabilities of a given domain participant. For example, it allows a user to restrict a particular participant to a specific DDS domain, or only allow the participant to read from or write to specific DDS topics, etc.
+The **Access control** plugin (see section 8.4 of the [DDS-Security spec][dds_security]) deals with defining and enforcing restrictions on the DDS-related capabilities of a given domain participant.
+For example, it allows a user to restrict a particular participant to a specific DDS domain, or only allow the participant to read from or write to specific DDS topics, etc.
 
-Again the SPI architecture allows for some flexibility in how the plugins accomplish this task, but ROS 2 uses the builtin access control plugin (called "DDS:Access:Permission", see section 9.4 of the [DDS-Security spec][dds_security]), which again uses PKI. It requires two files per domain participant:
+Again the SPI architecture allows for some flexibility in how the plugins accomplish this task, but ROS 2 uses the builtin access control plugin (called "DDS:Access:Permission", see section 9.4 of the [DDS-Security spec][dds_security]), which again uses PKI.
+It requires two files per domain participant:
 
 - **Governance** file: A signed XML document that specifies how the domain should be secured
 - **Permissions** file: A signed XML document that contains the permissions of the domain participant, bound to the name of the participant as defined by the authentication plugin (which is done via an x.509 cert, as we just discussed).
 
-Both of these files must be signed by a CA which the plugin is configured to trust. This may be the same CA that the **Authentication** plugin trusts, but that isn't required.
+Both of these files must be signed by a CA which the plugin is configured to trust.
+This may be the same CA that the **Authentication** plugin trusts, but that isn't required.
 
 
 ## Cryptographic
 
-The **Cryptographic** plugin (see section 8.5 of the [DDS-Security spec][dds_security]) is where all the cryptography-related operations are handled: encryption, decryption, signing, hashing, etc. Both the **Authentication** and **Access control** plugins utilize the capabilities of the **Cryptographic** plugin in order to verify signatures, etc. This is also where the functionality to encrypt DDS topic communication resides.
+The **Cryptographic** plugin (see section 8.5 of the [DDS-Security spec][dds_security]) is where all the cryptography-related operations are handled: encryption, decryption, signing, hashing, etc.
+Both the **Authentication** and **Access control** plugins utilize the capabilities of the **Cryptographic** plugin in order to verify signatures, etc.
+This is also where the functionality to encrypt DDS topic communication resides.
 
 ROS 2 uses the builtin cryptographic plugin (called "DDS:Crypto:AES-GCM-GMAC", see section 9.5 of the [DDS-Security spec][dds_security]), which provides authenticated encryption using Advanced Encryption Standard (AES) in Galois Counter Mode (AES-GCM).
 
 
 # DDS-Security integration with ROS 2: SROS2
 
-Now that we have established some shared understanding of how security is supported in DDS, let's discuss how that support is exposed in ROS 2. By default, none of the security features of DDS are enabled in ROS 2. The set of features and tools in ROS 2 that are used to enable them are collectively named "Secure ROS 2" (SROS 2).
+Now that we have established some shared understanding of how security is supported in DDS, let's discuss how that support is exposed in ROS 2.
+By default, none of the security features of DDS are enabled in ROS 2.
+The set of features and tools in ROS 2 that are used to enable them are collectively named "Secure ROS 2" (SROS 2).
 
 
 ## Features in the ROS client library (RCL)
 
-Most of the user-facing runtime support for SROS 2 is contained within the [ROS Client Library](https://github.com/ros2/rcl). Once its requirements are satisfied it takes care of configuring the middleware support for each supported DDS implementation. RCL includes the following features for SROS 2:
+Most of the user-facing runtime support for SROS 2 is contained within the [ROS Client Library](https://github.com/ros2/rcl).
+Once its requirements are satisfied it takes care of configuring the middleware support for each supported DDS implementation.
+RCL includes the following features for SROS 2:
 
 - Support for security files for each domain participant
 - Support for both permissive and strict enforcement of security
@@ -80,7 +96,9 @@ Let's discuss each of these in turn.
 
 ### Security files for each domain participant
 
-As stated earlier, the DDS-Security plugins require a set of security files (e.g. keys, governance and permissions files, etc.) per domain participant. Domain participants map to a specific instance of a node in ROS 2, so each node requires a set of these files. RCL supports this in two different ways:
+As stated earlier, the DDS-Security plugins require a set of security files (e.g. keys, governance and permissions files, etc.) per domain participant.
+Domain participants map to a specific instance of a node in ROS 2, so each node requires a set of these files.
+RCL supports this in two different ways:
 
 - Manual specification
 - Directory tree of all security files
@@ -90,7 +108,8 @@ Let's delve further into these.
 
 #### Manual specification
 
-RCL supports specifying the path to the directory containing the set of security files for the exact node instance that needs to be launched. The set of files expected within that directory are:
+RCL supports specifying the path to the directory containing the set of security files for the exact node instance that needs to be launched.
+The set of files expected within that directory are:
 
 - **identity_ca.cert.pem**: The x.509 certificate of the CA trusted by the **Authentication** plugin (the "Identity" CA)
 - **cert.pem**: The x.509 certificate of this node instance (signed by the Identity CA)
@@ -104,7 +123,8 @@ This can be specified by setting the `$ROS_SECURITY_NODE_DIRECTORY` environment 
 
 #### Directory tree of all security files
 
-RCL supports finding security files in one directory that is the root of a directory structure corresponding to the fully-qualified names of every node instance (i.e. namespace + node name). For example, for the `/front/camera` node, the directory structure would look like:
+RCL supports finding security files in one directory that is the root of a directory structure corresponding to the fully-qualified names of every node instance (i.e. namespace + node name).
+For example, for the `/front/camera` node, the directory structure would look like:
 
     <root>
     └── front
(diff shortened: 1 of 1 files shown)


articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+Original Author: {{ page.author }}
+
+
+# DDS-Security overview
+
+The [DDS-Security specification][dds_security] expands upon the [DDS specification][dds], adding security enhancements by defining a Service Plugin Interface (SPI) architecture, a set of builtin implementations of the SPIs, and the security model enforced by the SPIs. Specifically, there are five SPIs defined:
+
+- **Authentication**: Verify the identity of a given domain participant
+- **Access control**: Enforce restrictions on the DDS-related operations that can be performed by an authenticated domain participant
+- **Cryptographic**: Takes care of all encryption, signing, and hashing operations required
```
mikaelarguedas: `Takes` -> `Take` as previous bullets don't use the conjugated form, same for following bullet points

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+- **Logging**: Provides the ability to audit DDS-Security-related events
+- **Data tagging**: Provides the ability to add tags to data samples
+
+ROS 2's security features currently utilize only the first three of these, as not all DDS implementations support the **Logging** or **Data Tagging** plugins.
```
mikaelarguedas: Nit: only the first three ~~of these~~

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+Let's delve a little further into those first three plugins.
+
+
+## Authentication
+
+The **Authentication** plugin (see section 8.3 of the [DDS-Security spec][dds_security]) is central to the entire SPI architecture, as it provides the concept of a confirmed identity without which further enforcement would be impossible (e.g. it would be awfully hard to make sure a given ROS node could only access specific topics if it was impossible to securely determine which node it was).
+
+The SPI architecture allows for a number of potential authentication schemes, but ROS 2 uses the builtin authentication plugin (called "DDS:Auth:PKI-DH", see section 9.3 of the [DDS-Security spec][dds_security]), which uses the proven Public Key Infrastructure (PKI).
```
mikaelarguedas, line 50: Nit: allows for a number of potential authentication schemes, ~~but~~ ROS 2 uses the builtin authentication plugin

I would remove the `but` here as it is just defining what the default behavior is, same below for the access control plugin

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+## Access control
+
+The **Access control** plugin (see section 8.4 of the [DDS-Security spec][dds_security]) deals with defining and enforcing restrictions on the DDS-related capabilities of a given domain participant.
+For example, it allows a user to restrict a particular participant to a specific DDS domain, or only allow the participant to read from or write to specific DDS topics, etc.
+
+Again the SPI architecture allows for some flexibility in how the plugins accomplish this task, but ROS 2 uses the builtin access control plugin (called "DDS:Access:Permission", see section 9.4 of the [DDS-Security spec][dds_security]), which again uses PKI.
+It requires two files per domain participant:
+
+- **Governance** file: A signed XML document that specifies how the domain should be secured
+- **Permissions** file: A signed XML document that contains the permissions of the domain participant, bound to the name of the participant as defined by the authentication plugin (which is done via an x.509 cert, as we just discussed).
```
mikaelarguedas: Nit: some bullets end with a dot, others don't

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Access control
+
+The **Access control** plugin (see section 8.4 of the [DDS-Security spec][dds_security]) deals with defining and enforcing restrictions on the DDS-related capabilities of a given domain participant.
+For example, it allows a user to restrict a particular participant to a specific DDS domain, or only allow the participant to read from or write to specific DDS topics, etc.
+
+Again the SPI architecture allows for some flexibility in how the plugins accomplish this task, but ROS 2 uses the builtin access control plugin (called "DDS:Access:Permission", see section 9.4 of the [DDS-Security spec][dds_security]), which again uses PKI.
+It requires two files per domain participant:
+
+- **Governance** file: A signed XML document that specifies how the domain should be secured
```
mikaelarguedas: `that specifies` -> `specifying`, same below: `that contains` -> `containing`
(seems more natural to me but I'm not a native speaker so feel free to discard if it's not the case)

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+## Cryptographic
+
+The **Cryptographic** plugin (see section 8.5 of the [DDS-Security spec][dds_security]) is where all the cryptography-related operations are handled: encryption, decryption, signing, hashing, etc.
+Both the **Authentication** and **Access control** plugins utilize the capabilities of the **Cryptographic** plugin in order to verify signatures, etc.
+This is also where the functionality to encrypt DDS topic communication resides.
+
+ROS 2 uses the builtin cryptographic plugin (called "DDS:Crypto:AES-GCM-GMAC", see section 9.5 of the [DDS-Security spec][dds_security]), which provides authenticated encryption using Advanced Encryption Standard (AES) in Galois Counter Mode (AES-GCM).
+
+
+# DDS-Security integration with ROS 2: SROS2
```
mikaelarguedas: Nit: `SROS2` -> `SROS 2` to be consistent with the rest of the document

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+
+Now that we have established some shared understanding of how security is supported in DDS, let's discuss how that support is exposed in ROS 2.
+By default, none of the security features of DDS are enabled in ROS 2.
+The set of features and tools in ROS 2 that are used to enable them are collectively named "Secure ROS 2" (SROS 2).
+
+
+## Features in the ROS client library (RCL)
+
+Most of the user-facing runtime support for SROS 2 is contained within the [ROS Client Library](https://github.com/ros2/rcl).
+Once its requirements are satisfied it takes care of configuring the middleware support for each supported DDS implementation.
```
mikaelarguedas, line 104: `configuring the middleware support` -> `configuring the DDS-Security plugins`

mikaelarguedas: Not sure how you feel about this but this may benefit from including a table of DDS Implementation and plugins supported (ala https://github.com/ros2/sros2/issues/110#issuecomment-489644205).

articles/ros2_dds_security.md
mikaelarguedas: Maybe worth mentioning that **Logging** and **Data Tagging** are optional plugins not required to claim DDS-Security compliance. This is one of the reasons the support is expected to be less homogeneous among vendors regarding these plugins (could be captured in the table suggested in the other comment below)

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+Let's discuss each of these in turn.
+
+
+### Security files for each domain participant
+
+As stated earlier, the DDS-Security plugins require a set of security files (e.g. keys, governance and permissions files, etc.) per domain participant.
+Domain participants map to a specific instance of a node in ROS 2, so each node requires a set of these files.
+RCL supports this in two different ways:
+
+- Manual specification
```
mikaelarguedas: While this is defined more in details below but from reading this it's unclear that RCL accepts and operates only on folders and never on individual files.
Not sure how to address it, maybe make the section names more verbose ?

- `Manual specification of the node security files directory`
- `Directory tree of all nodes' security files directories`

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+The set of files expected within that directory are:
+
+- **identity_ca.cert.pem**: The x.509 certificate of the CA trusted by the **Authentication** plugin (the "Identity" CA)
+- **cert.pem**: The x.509 certificate of this node instance (signed by the Identity CA)
+- **key.pem**: The private key of this node instance
+- **permissions_ca.cert.pem**: The x.509 certificate of the CA trusted by the **Access control** plugin (the "Permissions" CA)
+- **governance.p7s**: The XML document that specifies to the **Access control** plugin how the domain should be secured  (signed by the Permissions CA)
+- **permissions.p7s**: The XML document that specifies the permissions of this particular node instance to the - **Access control** plugin (also signed by the Permissions CA)
+
+This can be specified by setting the `$ROS_SECURITY_NODE_DIRECTORY` environment variable to point to the directory containing the security files.
```
mikaelarguedas: IIRC this takes precedence over `ROS_SECURITY_ROOT_DIRECTORY` defined in next section, maybe swapping the 2 paragraph and specifying it here would make it more obvious to the reader

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+This can be specified by setting the `$ROS_SECURITY_ROOT_DIRECTORY` environment variable to point to the root of the directory tree.
+The set of files expected within each node instance directory are the same as outlined in the "Manual specification" section, above.
+
+
+##### Support security files lookup methods
+
+If using the directory tree approach to organize security files, RCL supports two different methods for looking up a given node instance's security files in the tree:
+
+- **Exact**: Only load security files from a directory exactly matching the fully-qualified name of the node instance.
+For example, given a node named "baz" within the "/foo/bar/" namespace, only load security files from `<root>/foo/bar/baz/`.
```
mikaelarguedas: using the same example node name `baz_123` in here and in the next section would make the difference between the options clearer

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+In order to enable SROS 2, set the `$ROS_SECURITY_ENABLE` environment variable to "true" (case-sensitive).
+To disable, set to any other value.
+
+
+## Features in the SROS 2 CLI
+
+Configuring a ROS 2 system to be secure in RCL involves a lot of new technology (PKI, DDS governance and permissions files and their syntax, etc.).
+If a user is comfortable with these technologies, the above information should be all that's necessary to properly lock things down.
+However, the [SROS 2 CLI](https://github.com/ros2/sros2) includes a number of commands to help those who don't want to set it all up themselves.
+This section is not meant to be a tutorial, but a reference of capabilities that are available.
```
mikaelarguedas: Not sure if it belongs in a design doc but this could link to the SROS 2 tutorial for people interested

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+Nodes with the security features enabled will not communicate with nodes that don't, but what should RCL do if one tries to launch a node that has no discernable keys/permissions/etc.? It has two options:
+
+- **Permissive mode**: Try to find security files, and if they can't be found, launch the node without enabling any security features.
+This is the default behavior.
+- **Strict mode**: Try to find security files, and if they can't be found, fail to run the node.
+
+The type of mode desired can be specified by setting the `$ROS_SECURITY_STRATEGY` environment variable to "Enforce" (case-sensitive) for strict mode, and anything else for permissive mode.
+
+
+### Support for a master "on/off" switch for all SROS 2 features
```
mikaelarguedas, line 188: Maybe this should be moved up as the first user-facing configuration option. Setting this to off will ignore all the configuration options described above

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+    $ ros2 security create_key <keystore path> <fully-qualified node instance name>
+
+This creates a new identity for the node, generating a keypair and signing its x.509 certificate using the CA created in `create_keystore`.
+It also creates an initial permissions document (that allows everything by default) and signs it with the same CA.
+It then copies in the governance file created in `create_keystore` as well as the x.509 certificate for the CA to act as both the identity CA as well as the permissions CA.
+
+
+    $ ros2 security generate_policy <policy file>
+
+This takes the current state of the ROS graph (i.e. all the nodes currently running, along with their topic publications/subscriptions, services, etc.) and turns it into an [SROS 2 policy file](/articles/sros2_policy_files.html).
```
mikaelarguedas: Maybe add a TODO here for the in progress/unpublished policy file design document, as this link will be broken for the time being?

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,220 @@
...
+
+This takes an SROS 2 policy file and uses it to generate (and sign) the DDS permissions document for the given node instance.
+
+
+    $ ros2 security generate_artifacts \
+        -k <keystore path> \
+        -n <space-separated list of fully-qualified node instance names> \
+        -p <space-separated list of policy files>
+
+This is essentially a combination of `create_key` and `create_permission`, taking a set of fully-qualified node instance names as well as a set of policy files and generating the entire set of keys and permissions necessary for them to run securely.
```
mikaelarguedas: This is also a combination of `create_keystore` if the one provided in cmd line or via environment variable doesn't exist or is not valid

mikaelarguedas (commented):
Thanks @kyrofa for the new iteration, this looks great!

I added a couple comments inline.

Not sure how you prefer to track the TODOs not addressed in this first PR (policy file design doc, not implemented features to be designed etc)?

articles/ros2_dds_security.md
kyrofa: I disagree. What I'm trying to say here is that DDS-Security doesn't mandate how authentication is actually supposed to work, and ROS 2 happens to use a specific implementation that works THIS way. It doesn't have to, it could have used anything, and we might change it at some point. I could just as happily change this to "While the SPI architecture allows for a number of potential authentication schemes, ROS 2 uses [...]" but the "while" seems to be serving the same purpose, there. Do you like that better?

articles/ros2_dds_security.md
kyrofa: I'm getting a bit pedantic, I'm sorry, but I'd say technically the actual configuring of the DDS-Security plugins is done in the rmw layer, e.g. https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_shared_cpp/src/rmw_node.cpp#L281, wouldn't you agree? RCL just passes the info along once it sorts out where everything is. As a result I think the current phrasing is more accurate than what you suggest.

Regarding the table, I think it's very helpful for understanding the current status, but I'm not sure it's particularly relevant to a document that doesn't actually discuss individual DDS vendors at all other than mentioning why all plugins aren't currently utilized by SROS 2. If you'd still like to see it here it would probably make the most sense above where we discuss which plugins are used. Thoughts?

articles/ros2_dds_security.md
kyrofa: Ah! I didn't actually know that, thank you.

kyrofa: Updated.

articles/ros2_dds_security.md
kyrofa: Oh right, I meant to add a comment to the PR here before anyone noticed that, and then promptly forgot :stuck_out_tongue: . Are you wanting that TODO to actually go in the document, then?

articles/ros2_dds_security.md
kyrofa: I considered that, and I'm still kinda on the fence. In one sense it makes sense to show the simplest thing first, but the reason I put it here is so folks have context and understand what exactly is being disabled by the time they get to it (and then I can say "all the above stuff will be ignored" essentially, instead of being more abstract). I don't feel strongly about it, but that's my reasoning. If you'd like it moved, let me know.

articles/ros2_dds_security.md
kyrofa: I think it's worth doing, but only once the tutorial is somewhere that [_isn't_ a github repo](https://github.com/ros2/sros2/blob/master/SROS2_Linux.md), otherwise it's way too easy for that link to break.

articles/ros2_dds_security.md
kyrofa: Excellent idea, thank you-- done.

articles/ros2_dds_security.md
kyrofa: Right, thank you, I intended to say something about that and never got around to it. Done.

articles/ros2_dds_security.md
kyrofa: Those section names seem a little too long, so I reworded the paragraph leading up to it in 1b7a13b5729fdf21a9efe0506f4c515483003466 instead. What do you think?

articles/ros2_dds_security.md
kyrofa: Argh, I tried so hard on that one. Done!

articles/ros2_dds_security.md
kyrofa: No strong feelings, I think it reads fine either way-- updated.

articles/ros2_dds_security.md
kyrofa: :roll_eyes: that's one of my OWN nits, how embarrassing. Fixed!

articles/ros2_dds_security.md
kyrofa: An excellent point, thank you. I reworded this in e1100ac4e08e7d1b26de46ea37b2d651b9e2e8c1 and referred to the section of the spec that seems to lay that out, let me know if I got it wrong.

articles/ros2_dds_security.md
kyrofa: +1 for consistency, thank you-- done.

kyrofa:
Thanks for the review, @mikaelarguedas, I think it's ready for another pass and some more discussion.

> Not sure how you prefer to track the TODOs not addressed in this first PR (policy file design doc, not implemented features to be designed etc)?

We should be proposing the policy file design doc in the next few days; we can wait for it, remove that link from here and add it once it's available, or add a blurb in this doc about how it's (design-wise) TBD... I don't have strong feelings, do you have a preference?

Regarding iterating on this to include design for not-yet-implemented features, I say we land this PR and then propose new ones as necessary. That way folks have a foundation of what they know exists, and can review new designs with the proper context.

articles/ros2_dds_security.md
```diff
@@ -0,0 +1,224 @@
...
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+
+# DDS-Security overview
+
+The [DDS-Security specification][dds_security] expands upon the [DDS specification][dds], adding security enhancements by defining a Service Plugin Interface (SPI) architecture, a set of builtin implementations of the SPIs, and the security model enforced by the SPIs. Specifically, there are five SPIs defined:
```
jacobperron: One sentence per line
```suggestion
The [DDS-Security specification][dds_security] expands upon the [DDS specification][dds], adding security enhancements by defining a Service Plugin Interface (SPI) architecture, a set of builtin implementations of the SPIs, and the security model enforced by the SPIs.
Specifically, there are five SPIs defined:
```

articles/ros2_dds_security.md
jacobperron: I like it as-is.

articles/ros2_dds_security.md
jacobperron: We could make this section a little more abstract; similar to the [CLI section for Actions](http://design.ros2.org/articles/actions.html#introspection-tools). Rather than adding command signatures (that are subject to change), we could list the features that we want from a CLI tool with the descriptions.
Just a suggestion, I don't feel strongly either way.

jacobperron (commented):
LGTM :+1:

articles/ros2_dds_security.md
kyrofa: @jacobperron you may have a good point. I reworked this section to similar to the Actions doc, basically a list of requirements to be met. What do you think? Too abstract?

jacobperron: LGTM :+1:

articles/ros2_dds_security.md
mikaelarguedas: Thanks for the detailed answer.
I don't feel strongly about the wording and lgtm as is :+1:

---

As this is reflecting a restrictive choice we could consider providing a rationale regarding the choice of using builtin plugins (for Auth and Access Control); though the rationale would be limited to the fact that this is the only approach described in details in the spec and that the mandatory support for its implementation and interoperability between vendors made it the most sensible option.
+1 for getting this merged with or without a rationale on the plugins choice.

articles/ros2_dds_security.md
mikaelarguedas: > but I'd say technically the actual configuring of the DDS-Security plugins is done in the rmw layer, e.g. https://github.com/ros2/rmw_fastrtps/blob/master/rmw_fastrtps_shared_cpp/src/rmw_node.cpp#L281, wouldn't you agree? RCL just passes the info along once it sorts out where everything is.

That's right

> As a result I think the current phrasing is more accurate than what you suggest.

Fair enough :+1:

> Regarding the table, I think it's very helpful for understanding the current status, but I'm not sure it's particularly relevant to a document that doesn't actually discuss individual DDS vendors at all other than mentioning why all plugins aren't currently utilized by SROS 2

Agreed, it's still unclear where such documentation would live that's why I suggested to add it to this "current state of DDS-Security integration document" but I agree that this level of implementation state detail does not belong in a design doc so let's leave it out :+1:

articles/ros2_dds_security.md
mikaelarguedas: Perfect! Much better than what I suggested

articles/ros2_dds_security.md
mikaelarguedas: > the reason I put it here is so folks have context and understand what exactly is being disabled by the time they get to it (and then I can say "all the above stuff will be ignored" essentially, instead of being more abstract)

Fine by me :+1:

mikaelarguedas:
> We should be proposing the policy file design doc in the next few days; we can wait for it, remove that link from here and add it once it's available, or add a blurb in this doc about how it's (design-wise) TBD... I don't have strong feelings, do you have a preference?

I'd have a slight preference for the first option as it's only a matter of days, adding "todo: link to upcoming access control policy design doc" instead of the link (similar to https://github.com/ros2/design/blob/gh-pages/articles/114_generated_interfaces_python.md#magic-methods) and replace the todo with the link in the PR adding the policy design doc sounds good to me.

> Regarding iterating on this to include design for not-yet-implemented features, I say we land this PR and then propose new ones as necessary. That way folks have a foundation of what they know exists, and can review new designs with the proper context.

:+1:

>  I think it's ready for another pass and some more discussion.

Thanks for the quick update, this looks great :+1:

kyrofa:
> I'd have a slight preference for the first option as it's only a matter of days [...]

I don't think I split out the three options very well, which makes it hard for you to specify which one you actually prefer, sorry about that :stuck_out_tongue: . However, given the following it sounds like you prefer the second option:

> adding "todo: link to upcoming access control policy design doc" instead of the link (similar to https://github.com/ros2/design/blob/gh-pages/articles/114_generated_interfaces_python.md#magic-methods) and replace the todo with the link in the PR adding the policy design doc sounds good to me.

So I've gone ahead and done that. If you do actually prefer the first option (i.e. waiting for the policy proposal) then we can easily add it back.

articles/ros2_dds_security.md
kyrofa: > we could consider providing a rationale regarding the choice of using builtin plugins

Oh yeah, great idea! Done.

kyrofa:
I believe all comments have been addressed here, please let me know if I'm mistaken.

jacobperron (approved):

PR #245, opened 2019-07-22
Add ROS 2 command line arguments design document

hidmic:
Follow-up PR after the discussions that took place in #242.

### DIFF
--- a/articles/150_ros_command_line.md
+++ b/articles/150_ros_command_line.md
@@ -33,7 +33,9 @@ Flags, in contrast with other custom syntax alternatives, are:
 
 Unfortunately, since these flags coexist with user-defined ones, additional guarding and extraction devices must be put in place -- one of the reasons why these were avoided entirely in ROS 1 command lines.
 
-## Namespacing
+## Features
+
+### Namespacing
 
 To prevent ROS specific command line flags from colliding with user-defined ones, the former are scoped using the `--ros-args` flag and a trailing double dash token (`--`):
 
@@ -56,16 +58,16 @@ ros2 run some_package some_node --ros-args <ros-specific-arg-0> -- [<user-define
 
 This way, multiple sources, potentially unaware of each other, can append flags to the command line with no regard for previous sets.
 
-## Capabilities
+### Capabilities
 
-### Summary
+#### Summary
 
 As a quick summary of ROS command line capabilities:
 
 - For name remapping, use either `--remap from:=to` or `-r from:=to`.
 - For parameter assignment, use either `--param name:=value` or `-p name:=value`.
 
-### Name remapping rules
+#### Name remapping rules
 
 Remapping rules may be introduced using the `--remap`/`-r` option.
 This option takes a single `from:=to` remapping rule.
@@ -82,22 +84,36 @@ or its shorter equivalent:
 ros2 run some_package some_node --ros-args -r foo:=bar
 ```
 
-### Parameter assignment statements
+#### Parameter assignment statements
 
 Parameter assignment may be achieved using the `--param`/`-p` option.
-This option takes a single `name:=value` assignment statement, where `value` is in YAML format.
+This option takes a single `name:=value` assignment statement, where `value` is in [YAML format](https://yaml.org/spec/) and thus YAML type inference rules apply.
 
-As an example, to assign an integer value of `2` to an `int_param` upon running `some_node`, one may execute:
+As an example, to assign a string value `test` to a parameter `string_param` upon running `some_node`, one may execute:
 
 ```sh
-ros2 run some_package some_node --ros-args --param int_param:=2
+ros2 run some_package some_node --ros-args --param string_param:=test
 ```
 
 or its shorter equivalent:
 
 ```sh
-ros2 run some_package some_node --ros-args -p int_param:=2
+ros2 run some_package some_node --ros-args -p string_param:=test
 ```
+
+## Implementation
+
+### Extraction
+
+Command line argument extraction happens within `rcl`.
+When an instance of the `--ros-args` flag is found in `argv`, until either a double dash token (`--`) is found or the end of the argument array is reached, all arguments that follow are taken as ROS specific arguments to be parsed as such.
+Remaining arguments can still be accessed by the user via `rcl` API.
+
+### Parsing
+
+At the time of writing, most ROS specific arguments target and are thus parsed by `rcl`.
+This is the case for name remapping rules or parameter assignments flags, to name a few.
+However, to support ROS specific arguments that target upper ROS layers e.g. a ROS client library like `rclcpp`, arguments unknown to `rcl` are left unparsed but accessible by these layers, which in turn can continue parsing or eventually warn the user if unknown arguments remain.
 
 ## Alternative designs
 
--- a/articles/150_ros_command_line.md
+++ b/articles/150_ros_command_line.md
@@ -22,30 +22,39 @@ Original Author: {{ page.author }}
 ## Overview
 
 As it was the case in ROS 1, ROS 2 nodes allow configuration via command line arguments to a certain degree.
-However, a larger set of configuration options, an ambiguity in remapping rules and parameter assignment syntax (as a result of the leading underscore name convention for hidden resources), a one-to-many relationship between executables and nodes, among others, add up complexity to an otherwise simple interface.
+In ROS 2, this interface had to become more complex to cope with a larger set of configuration options, an ambiguity in remapping rules and parameter assignment syntax (as a result of the leading underscore name convention for hidden resources), a one-to-many relationship between executables and nodes, to name a few.
 
-Increasingly precise addressing mechanisms, as well as leading double underscores (`__`) in positional arguments, solve part of the problem.
-In ROS 2, however, command line options are introduced too as these are:
+Because of this, increasingly precise addressing mechanisms as well as leading double underscores (`__`) in some positional arguments, both natural extensions of existing ROS 1 command line features, are combined with ROS 2 specific command line flags.
+Flags, in contrast with other custom syntax alternatives, are:
 
+- Widely known and used.
 - Explicit and less error prone.
 - Suitable for tab completion.
 
-Unfortunately, additional devices must be put in place as these coexist with user-defined options -- part of the rationale behind ROS 1 command line design.
+Unfortunately, since these flags coexist with user-defined ones, additional guarding and extraction devices must be put in place -- one of the reasons why these were avoided entirely in ROS 1 command lines.
 
 ## Namespacing
 
-To prevent ROS specific command line options from interacting with user-defined ones, the former are scoped using the `--ros-args` option and a trailing double dash token (`--`):
+To prevent ROS specific command line flags from colliding with user-defined ones, the former are scoped using the `--ros-args` flag and a trailing double dash token (`--`):
 
 ```sh
-ros2 run some_package my_node [<user-defined-pre-arg0>...] \
+ros2 run some_package some_node [<user-defined-arg0>...] \
   --ros-args <ros-specific-arg-0> [<ros-specific-arg-1>...] -- <user-defined-arg-1> [<user-defined-arg-2>...]
 ```
 
 If no user defined arguments are provided after ROS specific arguments are, the double dash token (`--`) may be elided:
 
 ```sh
-ros2 run some_package my_node [<user-defined-arg-0>...] --ros-args <ros-specific-arg-0> [<ros-specific-arg-1>...]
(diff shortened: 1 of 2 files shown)


hidmic:
I think we can (and probably should) augment this document to encompass all ROS 2 nodes command line features. Also, I wonder now that we're introducing options, if we should move some features away from the double leading underscore syntax e.g. `__log_level:=...` to `--log-level ...`. It's a cosmetic change though.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Overview
+
+As it was the case in ROS 1, ROS 2 nodes allow configuration via command line arguments to a certain degree.
+However, a larger set of configuration options, an ambiguity in remapping rules and parameter assignment syntax (as a result of the leading underscore name convention for hidden resources), a one-to-many relationship between executables and nodes, among others, add up complexity to an otherwise simple interface.
```
wjwwood: I'd reword this to have the conclusion then followed by the reasons (just because it's hard to see how the enumerated items are related which reading).

Something like this (might need more polish, but just an example of flipping the order of the sentence):

> However, in ROS 2 the complexity of these command line arguments have increased due to: a larger set of configuration options, an ambiguity in remapping rules and parameter assignment syntax (as a result of the leading underscore name convention for hidden resources), a one-to-many relationship between executables and nodes, etc.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+
+Increasingly precise addressing mechanisms, as well as leading double underscores (`__`) in positional arguments, solve part of the problem.
+In ROS 2, however, command line options are introduced too as these are:
```
wjwwood: For me "command line options" includes the "positional" `from:=to` style arguments. Something like `--opt` or `-o` is what I'd call "flags". I'm not sure that's the right wording, but as it's written it's a confusing. Maybe something like:

> In ROS 2, however, additional command line options have been introduced in order to be:

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Explicit and less error prone.
+- Suitable for tab completion.
+
+Unfortunately, additional devices must be put in place as these coexist with user-defined options -- part of the rationale behind ROS 1 command line design.
+
+## Namespacing
+
+To prevent ROS specific command line options from interacting with user-defined ones, the former are scoped using the `--ros-args` option and a trailing double dash token (`--`):
```
wjwwood: Personally, I would use `interfering` or `colliding` rather than `interacting`.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+```sh
+ros2 run some_package my_node [<user-defined-pre-arg0>...] \
+  --ros-args <ros-specific-arg-0> [<ros-specific-arg-1>...] -- <user-defined-arg-1> [<user-defined-arg-2>...]
+```
+
+If no user defined arguments are provided after ROS specific arguments are, the double dash token (`--`) may be elided:
+
+```sh
+ros2 run some_package my_node [<user-defined-arg-0>...] --ros-args <ros-specific-arg-0> [<ros-specific-arg-1>...]
+```
```
wjwwood: One more thing to clarify, can you have more than one instance of `--ros-args ... --`? I would say we should so that multiple sources of ROS options can just append them to the command line arguments without having to worry if there was a previous instance or not.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+
+Other, alternative designs were under discussion.
+
+### Additional operators
+
+Stop using the same `:=` operator for parameter assignments and name remapping rules and introduce additional operators e.g. `:=` for parameter assignment and `~=` for name remapping. This keeps the command line verbosity at a minimum, but is error prone.
+
+### Full name addressing
+
+Rely on full name addressing to disambiguate operator significance e.g. `rosparam://this:=that` would mean to assign a `that` string value to parameter `this` while `rostopic://this:=that` would mean to remap topic `this` to `that`. This signficantly increases the command line verbosity.
```
wjwwood: Wouldn't it need to be `rosremap://` or something? Right now remaps apply to topics and services.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Prefixed option names
+
+Remove the need for double dash tokens (`--`), conventionally used to signify the end of CLI options for a command, by adding the `--ros-` prefix to all ROS specific command line options e.g. `--ros-remap`, `--ros-param`, etc. This somewhat increases the command line verbosity.
```
wjwwood: I would also say that this one makes the extraction of the ROS arguments slightly more difficult because you need to know about all ros options ahead of time, whereas the currently proposal will always be removable with just two or three rules to consider.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,107 @@
...
+ros2 run some_package some_node --ros-args -p int_param:=2
+```
+
+## Alternative designs
+
+Other, alternative designs were under discussion.
+
+### Additional operators
+
+Stop using the same `:=` operator for parameter assignments and name remapping rules and introduce additional operators e.g. `:=` for parameter assignment and `~=` for name remapping. This keeps the command line verbosity at a minimum, but is error prone.
```
jacobperron: nit: one sentence per line

It might be worth clarifying that this option avoids the need for options at all (ie. no need for `--ros-args`).

articles/150_ros_command_line.md
hidmic: Yeah, it's poorly written. I almost rewrote the entire section in beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
hidmic: I like `colliding`. See beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
hidmic: That's a very good point. See beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
hidmic: To force a remap for all matching interfaces regardless of their type? Sure, yeah. See beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
hidmic: True that. See beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
hidmic: Done in beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
hidmic: Reworded in beaf7bec3dfc45db907f8c89e30801dcfdc3eba1.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,121 @@
...
+```sh
+ros2 run some_package some_node --ros-args -r foo:=bar
+```
+
+### Parameter assignment statements
+
+Parameter assignment may be achieved using the `--param`/`-p` option.
+This option takes a single `name:=value` assignment statement, where `value` is in YAML format.
+
+As an example, to assign an integer value of `2` to an `int_param` upon running `some_node`, one may execute:
```
dirk-thomas: I am not sure if `integer` is a good example here. That also opens the question what makes this a parameter of type since there needs to be some magic to convert the type solely based on its value which has problems on its own.

Therefore I would suggest to simply use a string value here.

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,121 @@
...
+title: ROS Command Line
+permalink: articles/ros_command_line.html
+abstract:
+  This article describes ROS 2 nodes command line capabilities and its syntax.
+author: '[Michel Hidalgo](https://github.com/hidmic)'
+published: true
+---
+
+- This will become a table of contents (this text will be scraped).
+{:toc}
```
dirk-thomas: Maybe add a short section about the implementation. Basically where the extraction from `argc` / `argv` happens and that all remaining arguments are still available for the user code to process.

articles/150_ros_command_line.md
hidmic: Hmm, I agree with using a string value to simplify the example but I also think that parameter value syntax cannot be left out of a command line specification. Here, just like when [parameter files are parsed in `rcl`](https://github.com/ros2/rclcpp/blob/8c48dbede7de2bc7e5199f340b6bfd33ffd6a784/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L137) and when [`launch` substitutions occur in `Node` parameters](https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/utilities/evaluate_parameters.py#L70), we delegate on YAML format rules for type inference based on values. Whether that's a good call or not is likely worth a separate discussion.

articles/150_ros_command_line.md
hidmic: That makes sense. It also raises an interesting question. Assuming extraction happens in `rcl`, should we support arguments specific to upper layers e.g. `rclcpp` or `rclpy`? Those would have to be available unparsed, separate from user defined arguments.

wjwwood: We could, but they could live in between the `--ros-args` and `--`, and `rcl` could just ignore them and leave them in a "left over args" output. Then `rclcpp` and/or `rclpy` can use those arguments or make warnings/errors because there are unknown parameters left over.

wjwwood (approved):

hidmic:
@dirk-thomas PTAL

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,137 @@
+---
+layout: default
+title: ROS Command Line
+permalink: articles/ros_command_line.html
```
dirk-thomas: Change to: `ros_command_line_arguments`

articles/150_ros_command_line.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Features
+
+### Namespacing
+
+To prevent ROS specific command line flags from colliding with user-defined ones, the former are scoped using the `--ros-args` flag and a trailing double dash token (`--`):
+
+```sh
+ros2 run some_package some_node [<user-defined-arg0>...] \
+  --ros-args <ros-specific-arg-0> [<ros-specific-arg-1>...] -- <user-defined-arg-1> [<user-defined-arg-2>...]
```
dirk-thomas: Does `--ros-args` need to be `nargs='+'`? Can't it be defined as `nargs='*'` allowing no ROS specific arguments to follow?

articles/150_ros_command_line.md
hidmic: Changed in 4c1b29833c98640308a183a28d9a9b35e8ed1ca5.

articles/150_ros_command_line.md
hidmic: Hmm, what use case do you have in mind?

dirk-thomas: Just common logic to build these arguments. If at least one is required the logic needs to double check / special case that when generating the command line.

It is the same for any other argument taking multiple parameters in command line tools, e.g. `colcol build --cmake-args`, `ros2 security generate_artifacts --node-names`.

wjwwood: I don't understand what you mean.

dirk-thomas: `ros2 run some_package some_node --ros-args` should be a valid invocation. The current design requires at least one argument (`<ros-specific-arg-0>`) to be present.

wjwwood: I understand now, yeah, I think it should be fine to not actually have any ros arguments follow `--ros-args`, so `--ros-args --` would also be ok.

dirk-thomas (approved):

hidmic:
@dirk-thomas @wjwwood PTAL at https://github.com/ros2/design/pull/245/commits/c079dc73cc07f5f2e91b2406fc2a5afccb66fd0f so we can call it done.

dirk-thomas (approved):
lgtm

hidmic:
Alright, going in!

wjwwood (approved):
late lgtm too :)

PR #249, opened 2019-08-07
Design for new tools for ROS 2 package cross-compilation

zmichaels11:
Old PR: https://github.com/ros2/design/pull/233

### DIFF
--- a/articles/cross_compilation_build_tools_design.md
+++ b/articles/cross_compilation_build_tools_design.md
@@ -2,10 +2,8 @@
 layout: default
 title: Cross-compiling ROS 2 packages
 abstract:
-  This article is a design proposal for developing a ROS 2 tool that sets up and
-  manages sysroot environments for cross-compilation with the objective of being simple
-  and extensible. Extending support for new cross compilation configurations using colcon 
-  mixins is also proposed.
+  This article is a design proposal for developing a ROS 2 tool that sets up and manages sysroot environments for cross-compilation with the objective of being simple and extensible.
+  Extending support for new cross compilation configurations using colcon mixins is also proposed.
 author: >
   [Thomas Moulard](https://github.com/thomas-moulard),
   [Juan Rodriguez Hortala](https://github.com/juanrh),
@@ -44,7 +42,7 @@ These commands would use a new workspace directory for the target platform, dete
 - ROS distribution
 
 Let us call this workspace directory the _cc-root_.
-Under this convention, the platform implies the architecture, but generic platforms like `generic_armhf` would also be available. 
+Under this convention, the platform implies the architecture, but generic platforms like `generic_armhf` would also be available.
 An example _platform identifier_ would be `generic_armhf-ubuntu_bionic-fastrtps-crystal` or `turtlebot_armhf-ubuntu_bionic-fastrtps-dashing`.
 To implement the proposal, we would add the following commands:
 
@@ -85,10 +83,10 @@ OSRF would also publish other base ROS 2 images with the system setup and basic 
 
 The `create-cc-sysroot` command will also support the following optional arguments:
 
-- `--sysroot-base-image`: Specifies a Docker image that will be used as the base image for the workspace dependent sysroot image. 
+- `--sysroot-base-image`: Specifies a Docker image that will be used as the base image for the workspace dependent sysroot image.
   This should be a string that is a valid argument for `docker image pull` and that can be used in the `FROM` Dockerfile statement.
   If it is not specified, the command will deduce the value of `--sysroot-base-image` from the target platform and pull from a published image on [Docker's Official ROS Images](https://hub.docker.com/_/ros), the [arm64v8 Dockerhub repo](https://hub.docker.com/u/arm64v8) or from the [arm32v7 Dockerhub repo](https://hub.docker.com/u/arm32v7).
-- `--sysroot-image`: Specifies a Docker image for the workspace sysroot. 
+- `--sysroot-image`: Specifies a Docker image for the workspace sysroot.
   The build will export the sysroot from that image, instead of creating a sysroot image from the workspace.
   No check is performed to ensure the image has all the dependencies required by the packages in the workspace.
 - `--use-base-image-sysroot`: Boolean flag, when true the base ROS 2 Docker image is used to build the sysroot, so the dependencies declared in packages in the current workspace are ignored.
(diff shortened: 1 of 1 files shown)


piraka9011:
@dirk-thomas @pokitoz @nuclearsandwich @filiperinaldi Could you PTAL at this PR and recent updates?

I've updated the article based on your comments in #233. We've also opened a PR in the [`ros2/cross_compile`](https://github.com/ros2/cross_compile/pull/20) repo with the requested changes in design doc. 

Let us know if there's any updates to make. 
@prajakta-gokhale and I will address any issues moving forward.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
+---
+layout: default
+title: Cross-compiling ROS 2 packages
+abstract:
+  This article is a design proposal for extending the build tools for ROS 2 to
```
dirk-thomas: Just a nitpick about naming / phrasing:

In a over simplification the article describes:
* to use mixins to describe cross compilation configurations as well as
* a tool to setup / manage sysroot environments

Neither of these is something I would call "extend" the existing "build tool" (which is `colcon`). Therefore I think the abstract should be rephrased to give the reader a better idea what the article describes. The title of the PR (which will become the commit message when merged) should be updated accordingly. As a comparison the title of the article describes the scope better better imo.

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Background
+
+The [cross-compilation tutorial](https://index.ros.org/doc/ros2/Tutorials/Cross-compilation)
+and the [steve/ros2_raspbian_tools repository](https://github.com/esteve/ros2_raspbian_tools)
+contain instructions for cross-compiling ROS 2 to unsupported platforms like
```
pokitoz: nitpick about phrasing: arm64 and armhf are architectures and not platforms.

I would rephrase to:
>  [...] instructions for cross-compiling ROS 2 to unsupported architecture like arm64 and armhf

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+  That would create `build`, `install`, and `log`
+  subdirectories of the cc-root `generic_armhf-ubuntu_bionic-fastrtps-crystal`,
+  with the cross-compiled binaries.
+
+```bash
+colcon build --mixin cc-generic_armhf-ubuntu_bionic-fastrtps-crystal \
+  --packages-up-to performance_test # any other other `colcon build` arguments
+```
+
+- On an second iteration of this initiative we would:
```
pokitoz: > a

instead of

> an

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,268 @@
...
+  container using the corresponding Docker image.
+
+Under the hood, the `cc-setup-sysroot` command would do the following:
+
+1. Downloads a base ROS 2 Docker image for the target platform.
+1. Builds a workspace dependent sysroot image by using a Dockerfile that starts
+   from that base image and:
+   1. Runs `COPY` to get the contents of the workspace into the container.
+   1. Uses `rosdep` to ensure no workspace system dependency is missing.
+      Note we can run `rosdep` in Docker for platforms like ARM-HF thanks
```
pokitoz: I would update

> in Docker for platforms like ARM-HF

to 

> in Docker for architecture like ARM-HF

(architecture versus platform)

piraka9011:
@dirk-thomas @pokitoz I've updated the design doc with the requested changes along with some updates based on what we've done so far in the [ros2/cross_compile PR](https://github.com/ros2/cross_compile/pull/20).

dirk-thomas:
Looks good to me - but I don't have much insight into the content so others should "approve" this.

Nitpick: each sentence should start on a new line (following the style guide).

articles/cross_compilation_build_tools_design.md
```diff
@@ -0,0 +1,170 @@
+---
+layout: default
+title: Cross-compiling ROS 2 packages
+abstract:
+  This article is a design proposal for developing a ROS 2 tool that sets up and
+  manages sysroot environments for cross-compilation with the objective of being simple
+  and extensible. Extending support for new cross compilation configurations using colcon
```
dirk-thomas: Another case of one-sentence-per-line.

It also seems that several lines have trailing whitespaces which should be removed.

piraka9011: Are we using any linters for this? I went through the contributing guidelines and the linting checks all pass.

Updating in the next commit.

dirk-thomas: Yes, they are supposed to catch this kind of stuff: https://github.com/ros2/design/blob/3f33a66aa5736c2557130181a2af72b0c37ca4f2/.travis.yml#L10

piraka9011:
If there aren't any other updates/suggestions could we get this merged/closed?

dirk-thomas:
Maybe someone should review / approve the content? Not sure who would be best for this. I was only looking at the content from a format / terminology point of view.

piraka9011:
Pinging @pokitoz @alsora and @filiperinaldi to see if they have any suggestions.

drastx:
A lot of these arm and mips houses that make sbc development or reference boards use openwrt, buildroot or yocto. There are others, but those are the three major ones.

So a lot of the times, developers would have sysroot already built by openwrt or buildroot so being able to use that existing standard mechanism instead of creating specialized docker images would be helpful.

Doing it with docker, you're really hoping that the qemu image matches in terms of libc, special flags, and processor nuances for particular hardware but in reality, I think that will be difficult. The qemu arm support is a bit generalist in my opinion and can be used for platforms that are _well_ supported. But with any new platforms, you're basically limiting your portability because _you_ have to create a specialized docker image yourself. I doubt that any of the SBC houses will invest time doing that.

I used to maintain the tool-chain(compiler,linker,libc, etc...) for a major US-based arm/mips chip manufacturer and personally, I've never seen anyone using docker or qemu for cross-compilation. The approach seems really up the stack for very well supported platforms but not a right solution for hardware bring-up.

So, in my opinion, if you supported standard sysroot options like openwrt and buildroot etc then you are addressing the embedded market better.

In summary, it's ok if you want to use docker/qemu, but that shouldn't be the only supported sysroot. You should be able to point the build to say an openwrt or buildroot sysroot image. My personal feeling is that companies/projects that are doing new hardware support/enablement are more likely to use the openwrt or buildroot, or yocto. And then on the other hand docker approach is going to be used mostly by the audience who's just trying to figure out what this arm thing is... dipping their toes in for like a school project or alike.

tfoote:
It seems like we've resolved most of the initial issues. There are some additional suggestions/extension points suggested by @drastx above. 

I'd suggest that if no one objects we merge this as is in draft state and people can start working on it. And @drastx if you can make any specific suggestions to extend the proposal to add the other sysroots that would be valuable and can be reviewed on top of the current draft.

thomas-moulard:
@tfoote We have updated the PR, let us know if it's OK and we'll merge it.

thomas-moulard:
@tfoote Merging as it meets the bar you set.

PR #254, opened 2019-08-28
Add launch XML substitution for a packages share directory

jacobperron:
Rename find-pkg to find-pkg-prefix.
Add find-pkg-share substitution for the share directory.

Implementation is in https://github.com/ros2/launch_ros/pull/57

### DIFF
--- a/articles/151_roslaunch_xml.md
+++ b/articles/151_roslaunch_xml.md
@@ -65,7 +65,7 @@ The included launch file description is not necessarily written in this format n
 
 ```xml
 <include file="/opt/my_launch_file.py"/>
-<include file="$(find-pkg my_pkg)/launch/some_launch_file.xml"/>
+<include file="$(find-pkg-share my_pkg)/launch/some_launch_file.xml"/>
 <include file="/opt/my_other_launch_file.xml">
   <arg name="some_argument" value="dummy_value"/>
 </include>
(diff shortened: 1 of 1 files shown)


jacobperron:
Alternatively, we could re-use the name `find-pkg` to refer to the share directory, which I think is the more common case. See https://github.com/ros2/launch_ros/pull/57#issuecomment-525920400

Thoughts?

dirk-thomas:
> we could re-use the name `find-pkg` to refer to the share directory, which I think is the more common case.

Common or not my argument is that it is opaque to the user what it maps to. The user must read docs to understand its behavior which I think is not good naming.

wjwwood:
I think they'll have to read the documentation (or infer its behavior from an example) anyways. I always prefer more explicit names, so I'm ok with `find-pkg-share`, but I know other people don't want to type that all the time and for the vast majority of the time when they say "find-pkg" they mean the share directory. So I'd also be ok with `find-pkg` as an alias to `find-pkg-share`.

ivanpauno:
I prefer the explicit names `find_pkg_prefix` and `find_pkg_share`, as it is now.

articles/151_roslaunch_xml.md
```diff
@@ -239,11 +239,16 @@ All substitutions are enclosed by `$(...)`.
...
-`$(find-pkg <pkg-name>)`
+`$(find-pkg-prefix <pkg-name>)`
 : Substituted by the install prefix path of the given package.
   Forward and backwards slashes will be resolved to the local filesystem convention.
   Substitution will fail if the package cannot be found.
 
+`$(find-pkg-share <pkg-name>)`
+: Substituted by the share directory path of the given package.
+  Forward and backwards slashes will be resolved to the local filesystem convention.
+  Substitution will fail if the package cannot be found.
```
wjwwood, line 251: Might want to mention that the "share directory" includes the package's name, e.g. `<prefix>/share/<pkg>`.

wjwwood (approved):

jacobperron:
Thanks for the feedback everyone! I'll proceed with explicit names (`find-pkg-prefix` and `find-pkg-share`) and we can always add the alias `find-pkg` in the future, if the community wants it.

articles/151_roslaunch_xml.md
jacobperron: 0fc46f1

Issue #251, opened 2019-08-23
Intra-Process Communications for all language clients


emersonknapp:
## Description
This issue is a call for a design of zero-copy intra-process communications available to all ROS2 language clients.

The current implementation of this feature exists in `rclcpp` and therefore is not usable for Python (or less-supported languages C, Java, Rust, etc.)

## Acceptance Criteria
To close this issue, we want a design document that proposes the architecture for
* Intra-process optimized communications
* zero copy
* Available to all ROS2 language clients (therefore exists at or below the level of `rcl`)

As a follow up, will attempt to collect existing thoughts from https://github.com/ros2/design/pull/239 and add as comments below.

## Note
I do not consider myself personally an expert on this, however I'm very interested in collaborating towards a top down view of what this part of the ROS2 core should look like, and figuring out how the community can pull together towards a solution

ivanpauno:
I think that we can take some ideas from Connext "zero copy transfer over shared memory":
- https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html
- https://community.rti.com/static/documentation/connext-dds/current/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/SendingLDZeroCopy.htm

That's actually interprocess communication over shared memory, but something similar can be replicated using a buffer instead of a piece of shared memory.

The basic idea is that you have to ask to the publisher for a new message, instead of allocating an `unique_ptr`:
```cpp
msg = publisher->get_new_message();
if (msg != nullptr) {
  msg->data = "asd";
  publisher->publish(msg);
}
```

Currently, message lifetime can be extended to be longer than the scope of the callback (in cpp). That would not be possible if we go ahead with something like this (or at least, it will be really hard to implement that feature). 

----

The implementation could live in `rcl` or `rmw`, I'm not sure what would be better.

allenh1:
@ivanpauno I don't think `publisher->get_new_message()` ever return `nullptr`. I'd prefer a more asynchronous way to fetch a message, or potentially blocking on that call instead. I'm not very fond of the blocking call idea, but maybe an asynchronous trigger could be set up?

Maybe it could be set up so that we can `std::invoke` a callback in the `publish()` function? This isn't great though, since this would need to be done in `rcl`, which means it would be wasting cycles checking if there are `std::bind`ed callbacks on non-shared memory platforms.

[...]

PR #256, opened 2019-09-17
Design Document for Zero-Copy via Loaned Messages

mjcarroll:
Design document for using loaned messages to provide support for zero-copy middleware implementations.

This is from notes that @Karsten1987 and @wjwwood composed, reformatted and expanded.

### DIFF
--- a/articles/zero_copy.md
+++ b/articles/zero_copy.md
@@ -26,15 +26,15 @@ This document outlines desired changes in the middleware and client libraries re
 
 The motivation for message loaning is to increase performance and determinism in ROS2.
 
-As additional motivation, there are specific middle implementations that allow for zero-copy via shared memory mechanisms.
+As additional motivation, there are specific middleware implementations that allow for zero-copy via shared memory mechanisms.
 These enhancements would allow ROS2 to take advantage of the shared memory mechanisms exposed by these implementations.
-An example of zero-copy transfer is [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html)
+An example of zero-copy transfer is [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html).
 
 ### Publisher Use Cases
 
 There are two primary kinds of use cases when publishing data which are relevant in this article:
 
-1. The user creates and owns an instance of a message which they which to publish and then reuse after publishing.
+1. The user creates and owns an instance of a message which they wish to publish and then reuse after publishing.
 2. The user borrows a message instance from the middleware, copies data into the message, and returns the ownership of the message during publish.
 
 Currently, only the first case is possible with the `rclcpp` API.
@@ -45,7 +45,7 @@ In order to support the second use case, we need a way for the user to get at le
 In the second case, the memory that is used for the loaned message should be optionally provided by the middleware.
 For example, the middleware may use this opportunity to use a preallocated pool of message instances, or it may return instances of messages allocated in shared memory.
 However, if the middleware does not have special memory handling or preallocations, it may refuse to loan memory to the client library, and in that case, a user provided allocator should be used.
-This allows the user to have control over the allocation of the message when the middleware would otherwise use the standard allocator (new/malloc)
+This allows the user to have control over the allocation of the message when the middleware might otherwise use the standard allocator (new/malloc).
 
 #### Additional Publisher Use Case
 
@@ -101,7 +101,7 @@ These requirements are driven by idiosyncrasies of various middleware implementa
 
 #### RTI Connext DDS
 
-* The Connext API (more generally the DDS API) requires that you use a sequence of messages when taking or reading.
+* The Connext API (more generally the DDS API) requires that the user to use a sequence of messages when taking or reading.
   * This means the ROS API needs to do the same, otherwise the middleware would be giving a "loan" to a message in a sequence, but it would also need to keep the sequence immutable.
 * Needs to keep sample and sample info together, therefore `rclcpp` loaned message sequence needs to own both somehow.
 
--- a/articles/zero_copy.md
+++ b/articles/zero_copy.md
@@ -38,13 +38,13 @@ There are two primary kinds of use cases when publishing data which are relevant
 2. The user borrows a message instance from the middleware, copies data into the message, and returns the ownership of the message during publish.
 
 Currently, only the first case is possible with the `rclcpp` API.
-After calling publish, the user still owns the message and may reuse it immedeately.
+After calling publish, the user still owns the message and may reuse it immediately.
 
 In order to support the second use case, we need a way for the user to get at least a single message from the middleware, which they may then populate, then return when publishing.
 
 In the second case, the memory that is used for the loaned message should be optionally provided by the middleware.
 For example, the middleware may use this opportunity to use a preallocated pool of message instances, or it may return instances of messages allocated in shared memory.
-However, if the middleware does not have special memory handling or preallocations, it may refuse to loan memory to the client library, and in that case, a user provided allocator should be used.
+However, if the middleware does not have special memory handling or pre-allocations, it may refuse to loan memory to the client library, and in that case, a user provided allocator should be used.
 This allows the user to have control over the allocation of the message when the middleware might otherwise use the standard allocator (new/malloc).
 
 #### Additional Publisher Use Case
@@ -62,19 +62,19 @@ There are two ways the users may take message instances from a subscription when
 1. Taking direction from the `Subscription` after polling it for data availability or waiting via a wait set.
 2. Using an `Executor`, which takes the data from the user and delivers it via a user-defined callback.
 
-Note: It is assumed that the user will be abe to take multiple messages at a time if they are available.
+Note: It is assumed that the user will be able to take multiple messages at a time if they are available.
 
 In the first case, the user could choose to either:
 
 * Manage the memory for the message instance themselves, providing a reference to it, into which the middleware should fill the data.
 * Take one or more loaned messages from the middleware and return the loans later.
 
-In the second case, the user is delegating the memory management ot he client library via the `Executor`.The `Executor` may or may not borrow data from the middleweare, but the user callback does not care, so this can be considered an implementation detail.
+In the second case, the user is delegating the memory management of the client library via the `Executor`.The `Executor` may or may not borrow data from the middleware, but the user callback does not care, so this can be considered an implementation detail.
 The user should be able to influence what the `Executor` does, and in the case that memory needs to be allocated, the user should be able to provide an allocator or memory management strategy which would influence the `Executor`'s behavior.
 
 ## Requirements
 
-Based on the the use cases above, the general requirements are as follows:
+Based on the use cases above, the general requirements are as follows:
 
 * Users must be able to avoid all memory operations and copies in at least one configuration.
 
--- a/articles/zero_copy.md
+++ b/articles/zero_copy.md
@@ -142,8 +142,8 @@ Extend publisher API for loaned messages:
 ```
 rmw_ret_t
 rmw_publish_loaned_message(
-  const rmw_publisher_t * publisher
-  const void * ros_message
+  const rmw_publisher_t * publisher,
+  const void * ros_message,
   rmw_publisher_allocation_t * allocation
 );
 ```
@@ -154,7 +154,7 @@ Because a subscription does not necessarily allocate new memory for the loaned m
 ```
 rmw_ret_t
 rmw_take_loaned_message(
-  const rmw_subscription_t * subscription
+  const rmw_subscription_t * subscription,
   void ** loaned_message,
   bool * taken,
   rmw_subscription_allocation_t * allocation
--- a/articles/zero_copy.md
+++ b/articles/zero_copy.md
@@ -124,14 +124,14 @@ Introduce APIs for creating/destroying loaned messages, as well as structure for
 
 ```
 void *
-rmw_allocate_loaned_message(
+rmw_borrow_loaned_message(
   const rmw_publisher_t * publisher,
   const rosidl_message_type_support_t * type_support,
-  size_t message_size
+  void ** ros_message
 );
 
 rmw_ret_t
-rmw_deallocate_loaned_message(
+rmw_return_loaned_message(
   const rmw_publisher_t * publisher,
   void * loaned_message
 );
@@ -141,18 +141,30 @@ Extend publisher API for loaned messages:
 
 ```
 rmw_ret_t
-rmw_publish(
+rmw_publish_loaned_message(
   const rmw_publisher_t * publisher
   const void * ros_message
-  rmw_publisher_allocation_t * allocation,
-  bool is_loaned
+  rmw_publisher_allocation_t * allocation
 );
 ```
 
-Extend subscription API for taking loaned message:
+Extend subscription API for taking loaned message.
+Because a subscription does not necessarily allocate new memory for the loaned message during a take, the message only needs to be released rather than returned.
 
 ```
-// TODO
+rmw_ret_t
+rmw_take_loaned_message(
+  const rmw_subscription_t * subscription
+  void ** loaned_message,
+  book * taken,
+  rmw_subscription_allocation_t * allocation
+);
+
+rmw_ret_t
+rmw_release_loaned_message(
+  const rmw_subscription_t * subscription,
+  void * loaned_message
+);
 ```
 
 ### RCL API
@@ -161,14 +173,14 @@ Introduce APIs for creating/destroying loaned messages, as well as structure for
 
 ```
 void *
-rcl_allocate_loaned_message(
+rcl_borrow_loaned_message(
   const rcl_publisher_t * publisher,
   const rosidl_message_type_support_t * type_support,
-  size_t message_size
+  void ** ros_message
 );
 
 rcl_ret_t
-rcl_deallocate_loaned_message(
+rcl_return_loaned_message(
   const rcl_publisher_t * publisher,
   rcl_loaned_message_t * loaned_message
 );
@@ -178,11 +190,10 @@ Extend publisher API for loaned messages:
 
 ```
 rcl_ret_t
-rcl_publish(
+rcl_publish_loaned_message(
   const rcl_publisher_t * publisher,
   const void * ros_message,
-  rmw_publisher_allocation_t * allocation,
-  bool is_loaned
+  rmw_publisher_allocation_t * allocation
 );
 
 bool
@@ -192,7 +203,22 @@ rcl_publisher_can_loan_messages(const rcl_publisher_t * publisher);
 Extend subscription API for taking loaned message:
 
 ```
-// TODO
+rcl_ret_t
+rcl_take_loaned_message(
+  const rcl_subscription_t * subscription,
+  void ** loaned_message,
+  rmw_message_info_t * message_info,
+  rmw_subscription_allocation_t * allocation
+);
+
+rcl_ret_t
+rcl_release_loaned_message(
+  const rcl_subscription_t * subscription,
+  void * loaned_message
+);
+
+bool
+rcl_subscription_can_loan_messages(const rcl_subscription_t * subscription);
 ```
 
 ### RCLCPP LoanedMessage
@@ -246,4 +272,14 @@ rclcpp::Publisher::can_loan_messages()
 ```
 
 ### RCLCPP Subscription
+
+```
+/// Test if the middleware supports loaning messages
+bool
+rclcpp::Subscription::can_loan_messages()
+
+/// Subscription path for middlewares supporting loaned messages.
+void
+rclcpp::Subscription::handle_loaned_message(void * loaned_message, const rmw_message_info_t & message_info)
+```
 
--- a/articles/zero_copy.md
+++ b/articles/zero_copy.md
@@ -1,0 +1,78 @@
+---
+layout: default
+title: Zero Copy via Loaned Messages
+permalink: articles/zero_copy.html
+abstract:
+author: '[Karsten Knese](https://github.com/karsten1987) [William Woodall](https://github.com/wjwwood) [Michael Carroll](https://github.com/mjcarroll)'
+published: false
+---
+
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+## Overview
+
+There is a need to eliminate unnecessary copies throughout the ROS2 stack to maximize performance and determinism.
+In order to eliminate copies, the user must have more advanced control over memory management in the client library and middleware.
+One method of eliminating copies is via message loaning, that is the middleware can loan messages that are populated by the end user.
+This document outlines desired changes in the middleware and client libraries required to support message loaning.
+
+## Motivation
+
+The motivation for message loaning is to increase performance and determinism in ROS2.
+
+As additional motivation, there are specific middle implementations that allow for zero-copy via shared memory mechanisms.
+These enhancements would allow ROS2 to take advantage of the shared memory mechanisms exposed by these implementations.
+An example of zero-copy transfer is [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html)
+
+### Publication Use Cases
+
+There are two primary kinds of use cases when publishing data which are relevant in this article:
+
+1. The user creates and owns an instance of a message which they which to publish and then reuse after publishing.
+2. The user borrows a message instance from the middleware, copies data into the message, and returns the ownership of the message during publish.
+
+Currently, only the first case is possible with the `rclcpp` API.
+After calling publish, the user still owns the message and may reuse it immedeately.
+
+In order to support the second use case, we need a way for the user to get at least a single message from the middleware, which they may then populate, then return when publishing.
+
+In the second case, the memory that is used for the loaned message should be optionally provided by the middleware.
+For example, the middleware may use this opportunity to use a preallocated pool of message instances, or it may return instances of messages allocated in shared memory.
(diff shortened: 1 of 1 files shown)


articles/zero_copy.md
```diff
@@ -0,0 +1,251 @@
...
+There is a need to eliminate unnecessary copies throughout the ROS2 stack to maximize performance and determinism.
+In order to eliminate copies, the user must have more advanced control over memory management in the client library and middleware.
+One method of eliminating copies is via message loaning, that is the middleware can loan messages that are populated by the end user.
+This document outlines desired changes in the middleware and client libraries required to support message loaning.
+
+## Motivation
+
+The motivation for message loaning is to increase performance and determinism in ROS2.
+
+As additional motivation, there are specific middle implementations that allow for zero-copy via shared memory mechanisms.
```
clalancette:
```suggestion
As additional motivation, there are specific middleware implementations that allow for zero-copy via shared memory mechanisms.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,251 @@
...
+These enhancements would allow ROS2 to take advantage of the shared memory mechanisms exposed by these implementations.
+An example of zero-copy transfer is [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html)
```
clalancette:
```suggestion
An example of zero-copy transfer is [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html).
```

articles/zero_copy.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Publisher Use Cases
+
+There are two primary kinds of use cases when publishing data which are relevant in this article:
+
+1. The user creates and owns an instance of a message which they which to publish and then reuse after publishing.
```
clalancette:
```suggestion
1. The user creates and owns an instance of a message which they wish to publish and then reuse after publishing.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,251 @@
...
+
+Currently, only the first case is possible with the `rclcpp` API.
+After calling publish, the user still owns the message and may reuse it immedeately.
+
+In order to support the second use case, we need a way for the user to get at least a single message from the middleware, which they may then populate, then return when publishing.
+
+In the second case, the memory that is used for the loaned message should be optionally provided by the middleware.
+For example, the middleware may use this opportunity to use a preallocated pool of message instances, or it may return instances of messages allocated in shared memory.
+However, if the middleware does not have special memory handling or preallocations, it may refuse to loan memory to the client library, and in that case, a user provided allocator should be used.
+This allows the user to have control over the allocation of the message when the middleware would otherwise use the standard allocator (new/malloc)
```
clalancette:
```suggestion
This allows the user to have control over the allocation of the message when the middleware might otherwise use the standard allocator (new/malloc).
```

articles/zero_copy.md
```diff
@@ -0,0 +1,251 @@
...
+* The loaned message or sequence must be returned by the user.
+* When taking a loan, the middleware should not do anything "special", that is that the user must be able to influence the allocation.
+
+### Special Requirements
+
+These requirements are driven by idiosyncrasies of various middleware implementations and some of their special operating modes, e.g. zero copy:
+
+#### RTI Connext DDS
+
+* The Connext API (more generally the DDS API) requires that you use a sequence of messages when taking or reading.
```
clalancette:
```suggestion
* The Connext API (more generally the DDS API) requires that the user to use a sequence of messages when taking or reading.
```

clalancette (commented):
I just added some typo fixes.  I haven't reviewed the whole thing yet, particularly the proposed API.

Karsten1987:
fyi @michael-poehnl

articles/zero_copy.md
```diff
@@ -0,0 +1,251 @@
...
+  rmw_loaned_message_t * loaned_message
+);
+```
+
+Extend publisher API for loaned messages:
+
+```
+// This would not need knowledge of the rmw_publisher_t, as the message allocation
+// is tied to the publisher.
+rmw_ret_t rmw_publish(
```
Karsten1987: I think we've discussed changing the existing `rmw_publish` function solely by passing a boolean argument `is_loaned` to it, as the rest of the function arguments can stay the same, i.e. the `void *` to a ros message instance.

mjcarroll: Will update accordingly, and change the allocate/deallocate to work on `void *`

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Subscription Requirements
+
+The following requirements should hold whether the user is polling or using a wait set.
+
+* The user must be able to have the middleware fill data into messages allocated in the user's stack or heap.
+* The user must be able to get a loaned message from the middleware when calling take.
+* The user must be able to get a sequence of loaned messages from the middleware when calling take.
+* The loaned message or sequence must be returned by the user.
+* When taking a loan, the middleware should not do anything "special", that is that the user must be able to influence the allocation.
```
Karsten1987: Should that be when _not_ taking a loan? To my understanding the middleware should be completely in charge of allocations when taking a loan. However, if the middleware isn’t capable of zero copy, the user should have means to pass in custom allocators and thus control the allocation.

mjcarroll: I was a bit unclear on this as I was going through the previous notes.

Is the idea that users should _always_ take a loan, regardless of if the middleware supports it?  That would put the burden on the middleware for managing that memory.  Or should it be the case that there is no allocation when loans aren't supported?

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Subscription Requirements
+
+The following requirements should hold whether the user is polling or using a wait set.
+
+* The user must be able to have the middleware fill data into messages allocated in the user's stack or heap.
+* The user must be able to get a loaned message from the middleware when calling take.
+* The user must be able to get a sequence of loaned messages from the middleware when calling take.
+* The loaned message or sequence must be returned by the user.
```
mjcarroll, line 95: Note from @Karsten1987:

> We should specify how we can make sure that the user doesn't hold too many of these loaned messages when not taking them via the executor/callback but rather via polling the waitset.

> AUTOSAR specifies this as not possible to gain access to the sequence without specifying an according callback where the sequence eventually gets passed in via std::unique_ptr.

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+* The user must be able to get a loaned message, use it, and return it during publication.
+* When taking a loan, the middleware should not do anything "special", that is that the user must be able to influence the allocation.
+
+### Subscription Requirements
+
+The following requirements should hold whether the user is polling or using a wait set.
+
+* The user must be able to have the middleware fill data into messages allocated in the user's stack or heap.
+* The user must be able to get a loaned message from the middleware when calling take.
+* The user must be able to get a sequence of loaned messages from the middleware when calling take.
```
mjcarroll, line 94: Note from @Karsten1987 :

> There should be a parameter for specifying how many items such a sequence can hold. That would help to use fixed size containers in its implementation. This could be defaulted to the subscription history depth though.

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+  // Loan message from the middleware, if loaning is not available,
+  // allocate memory using Alloc
+  LoanedMessage(
+    const rclcpp::Publisher<MsgT, Alloc> * pub);
+
+private:
+  const rclcpp::Publisher<MsgT, Alloc> * pub_;
+
+  // Holds details of the message loan
+  rcl_loaned_message_t * loaned_msg_;
```
Karsten1987, line 251: I think we come around having this a `void *` pointing to memory appropriately allocated by `rmw_allocate_loaned_message()`

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+// Publish a loaned message, returning the loan
+void
+rclcpp::Publisher::publish(std::unique_ptr<rclcpp::LoanedMessage<MsgT, Alloc>> loaned_msg);
+
+// Test if the middleware supports loaning messages
+bool
+rclcpp::Publisher::can_loan_messages()
+```
+
+### RCLCPP Subscription
```
ivanpauno, line 276: I guess, that no modification in RCLCPP subscription would be needed.
Maybe we would need to add a new subscription signature more convenient for loaned messages, but not more than that. Probably, similar changes would be needed in rcl/rmw.

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+
```
ivanpauno, line 277: We should also add to the document an RCLPY API

ivanpauno (commented):
I think that the idea of loaned messages is the best solution for zero-copy intraprocess communication.
Currently, most of the middleware vendors we're supporting doesn't have a loaned message API.
So, it may be a good idea to also push development to add this in some of them.

Fixes https://github.com/ros2/design/issues/251.
That issue have some comments in line with this proposal.
Also, there is a similar previous discussion in https://github.com/ros2/design/pull/239.

Karsten1987:
I think we talked about this, but I think we didn't come to an conclusion.

When we talk about taking sequences for messages which can be loaned, this implies that there has to be some sort of adjustment/enhancement on for a `rmw_take_loaned_message_sequence` function.
Right now I envision something like:
```c++
rmw_ret_t
rmw_take_loaned_message_sequence(rmw_message_sequence_t * message_sequence, size_t n);
```
Where `n` is the amount of messages being fetched and stored in the message sequence.

If I bring this all the way up to a possible implementation in the `rclcpp::Executor::execute_subscription` function, this would like as such:

```c++
void
Executor::execute_subscription(
  rclcpp::SubscriptionBase::SharedPtr subscription)
{
  rmw_message_info_t message_info;
  message_info.from_intra_process = false;

  // handle serialized messages
  if (subscription->is_serialized()) {
    auto serialized_msg = subscription->create_serialized_message();
    auto ret = rcl_take_serialized_message(
      subscription->get_subscription_handle().get(),
      serialized_msg.get(), &message_info, nullptr);
    if (RCL_RET_OK == ret) {
      auto void_serialized_msg = std::static_pointer_cast<void>(serialized_msg);
      subscription->handle_message(void_serialized_msg, message_info);
    } else if (RCL_RET_SUBSCRIPTION_TAKE_FAILED != ret) {
      RCUTILS_LOG_ERROR_NAMED(
        "rclcpp",
        "take_serialized failed for subscription on topic '%s': %s",
        subscription->get_topic_name(), rcl_get_error_string().str);
      rcl_reset_error();
    }
    subscription->return_serialized_message(serialized_msg);
    return;
  }

  // if a middleware can provide loaned messages, we always take them as such
  if (subscription->can_loan_messages()) {
    size_t n = 1;  // how many messages are we fetch at a time
    // Could be part of the subscription API in case of polling
    // Maybe take a parameter n for reserving space in the loaned message sequence
    auto LoanedMessageSequence lms = subscription->create_loaned_sequence(n);
    auto ret = rcl_take_loaned_message_sequence(lms.get_sequence_handle(), n);
    if (RCL_RET_OK == ret) {
      subscription->handle_message(lms.msg_at(0), lms.msg_info_at(0));
    } else if (RCL_RET_TAKE_FAILED != ret) {
      RCUTILS_LOG_ERROR_NAMED(
        "rclcpp",
        "could not take loaned message sequence on topic '%s': %s",
        subscription->get_topic_name(), rcl_get_error_string().str);
      rcl_reset_error();
    }
    subscription->return_loaned_sequence(lms);
    return;
  }

  // if the middleware can't loan message, we create our own instances and take them
  std::shared_ptr<void> message = subscription->create_message();
  auto ret = rcl_take(
    subscription->get_subscription_handle().get(),
    message.get(), &message_info, nullptr);
  if (RCL_RET_OK == ret) {
    subscription->handle_message(message, message_info);
  } else if (RCL_RET_SUBSCRIPTION_TAKE_FAILED != ret) {
    RCUTILS_LOG_ERROR_NAMED(
      "rclcpp",
      "could not deserialize serialized message on topic '%s': %s",
      subscription->get_topic_name(), rcl_get_error_string().str);
    rcl_reset_error();
  }
  subscription->return_message(message);
}
```

Looking at the scribble above, you can see an inconsistency between serialized/non-loaned messages and loaned messages. The first two types are always fetch one-by-one, whereas loaned messages are always fetched in a sequence - main reason for this being a strong requirements for RTI Connext Micro in order to enable zero-copy.

The question which arises now is whether any `rmw_take_*` should be modified in a way that it always takes `n` instances or if we restrict that to loaned messages exclusively. 
If we did change `rmw_take` to always take multiple, then we have to push this consequently to `rmw_serialized_messages` as well as ROS messages allocated by the `rclcpp::MessageMemoryStrategy` and implement C-based container structures for it. 

My thinking here is that in a context of an executor, we probably would get around implementing the message sequences for serialized and non-loaned messages as we handle one callback at a time.
But thinking this further and having the ability to use the API outside of the scope of an executor (by polling the waitset or having different implementations of an executor), I personally believe it would be the right thing to do and extend the sequencing also to these types. As the `rmw_take_multiple` functions can be implemented next to the existing `rmw_take`, we can still use the latter inside the executor.

Opinions?

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+Currently, only the first case is possible with the `rclcpp` API.
+After calling publish, the user still owns the message and may reuse it immedeately.
+
+In order to support the second use case, we need a way for the user to get at least a single message from the middleware, which they may then populate, then return when publishing.
+
+In the second case, the memory that is used for the loaned message should be optionally provided by the middleware.
+For example, the middleware may use this opportunity to use a preallocated pool of message instances, or it may return instances of messages allocated in shared memory.
+However, if the middleware does not have special memory handling or preallocations, it may refuse to loan memory to the client library, and in that case, a user provided allocator should be used.
+This allows the user to have control over the allocation of the message when the middleware might otherwise use the standard allocator (new/malloc).
```
allenh1, line 48: > it may refuse to loan memory to the client library, and in that case, a user provided allocator should be used.

this makes a lot of sense.

> This allows the user to have control over the allocation of the message when the middleware might otherwise use the standard allocator (new/malloc)

This seems to contradict that prior statement, though. Maybe I'm not reading this quite right?

I think this should read something more along the lines of "If the middleware does not have specific protocols for handling preallocations of messages, it should use the `rcutils` allocator."

I could be way off base in my interpretation, though.

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+);
+```
+
+Extend publisher API for loaned messages:
+
+```
+rmw_ret_t
+rmw_publish(
+  const rmw_publisher_t * publisher
+  const void * ros_message
```
allenh1: Missing some ','s on these lines.

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+// TODO
+```
+
+### RCL API
+
+Introduce APIs for creating/destroying loaned messages, as well as structure for management:
+
+```
+void *
+rcl_allocate_loaned_message(
```
allenh1: Has the message already been loaned? If it is planned to be loaned, maybe this should be called `rcl_allocate_loaner_message`?

Karsten1987:
connected to https://github.com/ros2/ros2/issues/785

articles/zero_copy.md
```diff
@@ -0,0 +1,249 @@
...
+### RMW API
+
+Introduce APIs for creating/destroying loaned messages, as well as structure for management:
+
+```
+void *
+rmw_allocate_loaned_message(
+  const rmw_publisher_t * publisher,
+  const rosidl_message_type_support_t * type_support,
+  size_t message_size
```
dirk-thomas: While some messages are PODs many messages containing sequences are not. In the pending PR in `rclcpp` e.g. the value `sizeof(MessageT)` is passed for this argument. How does that handle the memory allocation of the sequence fields since it doesn't seem to be the case that the RMW interface is aware of how many items should be allocated for the message?

Karsten1987: When creating a publisher, it has to be distinguished whether the publisher can loan messages of the given message type or not. If not, the call to `rmw_allocate_loaned_message` shall fail.

dirk-thomas: Your response doesn't rely answer my question. I understand that if loaning isn't supported the calling code can detect that and use a different approach if desired. So the question remains: is this design draft and the implementation limited to POD message? If yes, this needs to be mentioned in this article imo. Also from my point of few that would be a severe limitation of the proposal and would need a very good rational why non-POD aren't / can't be supported (since they are the majority of message types).

Karsten1987: non-POD must be supported. The only thing not being guaranteed with non-POD is that a zero-copy mechanism can be supported. That is, for heap allocated data structures, copies and eventual serializations have to be done. 
But I see your point that this has to be noted more explicitly in this design doc.

dirk-thomas: My concern is that I am not convinced that the currently proposed API is capable of supporting non-POD messages. And if that is the case I would like to see an API which is able to support that or a rational why that is not possible / feasible.

articles/zero_copy.md
deeplearningrobotics: @mjcarroll: How does the AUTOSAR approach help? I still can just keep the `unique_ptr` in my container forever.

articles/zero_copy.md
dirk-thomas: > My concern is that I am not convinced that the currently proposed API is capable of supporting non-POD messages. And if that is the case I would like to see an API which is able to support that or a rational why that is not possible / feasible.

@Karsten1987 Can you please comment on my last reply when you have a chance.

dirk-thomas: The API in question has been removed in the mean time. The new signature of `rmw_borrow_loaned_message` is not limited like this so this looks fine to me.

articles/zero_copy.md
```diff
@@ -0,0 +1,356 @@
...
+  void * loaned_message
+);
+
+bool
+rcl_subscription_can_loan_messages(const rcl_subscription_t * subscription);
+```
+
+### RCLCPP LoanedMessage
+
+In order to support loaned messages in `rclcpp`, we tntroduce the concept of a `LoanedMessage`.
```
alsora:
```suggestion
In order to support loaned messages in `rclcpp`, we introduce the concept of a `LoanedMessage`.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,356 @@
...
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+## Overview
+
+There is a need to eliminate unnecessary copies throughout the ROS2 stack to maximize performance and determinism.
```
jacobperron: I think we prefer "ROS 2" over "ROS2" (as it's written all over the [documentation](https://index.ros.org/doc/ros2/)). The same change can be made throughout this document.
```suggestion
There is a need to eliminate unnecessary copies throughout the ROS 2 stack to maximize performance and determinism.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,356 @@
...
+This use case comes up when all or part of the data being published is located in memory that the middleware cannot allocate from, e.g. a hardware buffer via memory mapped I/O or something similar, and the user wants to still have zero copy and asynchronous publishing.
+In this case, the user needs to call publish, then keep the message instance immutable until the middleware lets the user know that it is done with the loaned data.
+
+This is a narrow use case and will require additional interfaces to support, therefore it will be out of scope for this document, but it is mentioned for completion.
+
+### Subscription Use Cases
+
+There are two ways the users may take message instances from a subscription when data is available:
+
+1. Taking direction from the `Subscription` after polling it for data availability or waiting via a wait set.
```
jacobperron: "directly"?
```suggestion
1. Taking directly from the `Subscription` after polling it for data availability or waiting via a wait set.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,356 @@
...
+
+Extend subscription API for taking loaned message.
+Because a subscription does not necessarily allocate new memory for the loaned message during a take, the message only needs to be released rather than returned.
+
+```
+rmw_ret_t
+rmw_take_loaned_message(
+  const rmw_subscription_t * subscription
+  void ** loaned_message,
+  book * taken,
```
jacobperron:
```suggestion
  bool * taken,
```

articles/zero_copy.md
jacobperron: I agree it could be more clear.

> "If the middleware does not have specific protocols for handling preallocations of messages, it should use the rcutils allocator."

I think this is almost right, but I don't think the user is limited to the `rcutils` allocator. They could provide their own.

articles/zero_copy.md
jacobperron: I think @Karsten1987 is correct, based on what I read in the use-case section above,

> the user is delegating the memory management of the client library via the Executor. The Executor may or may not borrow data from the middleware, but the user callback does not care, so this can be considered an implementation detail. The user should be able to influence what the Executor does, and **in the case that memory needs to be allocated**, the user should be able to provide an allocator or memory management strategy which would influence the Executor's behavior.

I've made bold the key phrase "in the case that memory needs to be allocated". My interpretation is the the client library will try to take a loan if possible, and if it is not possible then use any allocator/memory management strategy provided by the user.

articles/zero_copy.md
allenh1: > I don't think the user is limited to the rcutils allocator. They could provide their own.

Yeah, I like that better. :+1:

articles/zero_copy.md
```diff
@@ -0,0 +1,357 @@
...
+## Overview
+
+There is a need to eliminate unnecessary copies throughout the ROS 2 stack to maximize performance and determinism.
+In order to eliminate copies, the user must have more advanced control over memory management in the client library and middleware.
+One method of eliminating copies is via message loaning, that is the middleware can loan messages that are populated by the end user.
+This document outlines desired changes in the middleware and client libraries required to support message loaning.
+
+## Motivation
+
+The motivation for message loaning is to increase performance and determinism in ROS2.
```
jacobperron:
```suggestion
The motivation for message loaning is to increase performance and determinism in ROS 2.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,357 @@
...
+
+As additional motivation, there are specific middleware implementations that allow for zero-copy via shared memory mechanisms.
+These enhancements would allow ROS2 to take advantage of the shared memory mechanisms exposed by these implementations.
```
jacobperron:
```suggestion
These enhancements would allow ROS 2 to take advantage of the shared memory mechanisms exposed by these implementations.
```

articles/zero_copy.md
```diff
@@ -0,0 +1,357 @@
...
+
+### Subscription Requirements
+
+The following requirements should hold whether the user is polling or using a wait set.
+
+* The user must be able to have the middleware fill data into messages allocated in the user's stack or heap.
+* The user must be able to get a loaned message from the middleware when calling take.
+* The user must be able to get a sequence of loaned messages from the middleware when calling take.
+* The loaned message or sequence must be returned by the user.
+  * In general, users should return loans as soon as feasibly possible, as the underlying mechanism has finite resources to loan messages from. Holding loans too long may cause messages to dropped or publications to stall.
```
jacobperron: One sentence per line.

jacobperron (commented):
A few more minor nitpicks. Besides that there are still some unresolved comments that would be good to address before merging this.

jacobperron:
> The question which arises now is whether any rmw_take_* should be modified in a way that it always takes n instances or if we restrict that to loaned messages exclusively.

It makes sense to me to also extend the "take N" functionality to non-loaned messages, but we probably don't need to include this detail/discussion in this document since it is not exactly related to loaned messages. Probably worth discussing as part of the implementation ticket or a new issue.

mjcarroll:
>  Probably worth discussing as part of the implementation ticket or a new issue.

Agreed.

articles/zero_copy.md
mjcarroll: I don't think that AUTOSAR helps, this was from our initial brainstorming notes.  The important point was that if users hold on to loaned messages, at some point the underlying mechanism is going to stop working.

articles/zero_copy.md
mjcarroll: So does this need additional clarification?  I believe that what is there summarizes the correct behavior?

articles/zero_copy.md
mjcarroll: Python API is out of the scope for the moment.

articles/zero_copy.md
jacobperron: Re-reading what's in the document, it makes sense to me. I don't have a solid suggestion for changing it.

jacobperron (approved):
LGTM

articles/zero_copy.md
```diff
@@ -0,0 +1,358 @@
...
+One method of eliminating copies is via message loaning, that is the middleware can loan messages that are populated by the end user.
+This document outlines desired changes in the middleware and client libraries required to support message loaning.
+
+## Motivation
+
+The motivation for message loaning is to increase performance and determinism in ROS 2.
+
+As additional motivation, there are specific middleware implementations that allow for zero-copy via shared memory mechanisms.
+These enhancements would allow ROS 2 to take advantage of the shared memory mechanisms exposed by these implementations.
+An example of zero-copy transfer is [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html).
```
Karsten1987, line 31:
```suggestion
Examples of zero-copy transfer are [RTI Connext DDS Micro](https://community.rti.com/static/documentation/connext-micro/3.0.0/doc/html/usersmanual/zerocopy.html) and [Eclipse iceoryx](https://github.com/eclipse/iceoryx).
```

Karsten1987 (approved):

ros-discourse:

https://discourse.ros.org/t/ros-2-galactic-default-middleware-announced/18064/1

PR #264, opened 2019-11-05
Fix dead link to OMG IDL

teutobod:


hidmic (commented):
Hmm, it looks like the exact replacement for that link should be https://www.omg.org/spec/IDL/3.5. @dirk-thomas which IDL version do we currently support / rely on?

dirk-thomas:
Since the article describes an outdated state (see line 17) I am not sure it should be updated at all.

The current article references IDL 4.2 (see http://design.ros2.org/articles/idl_interface_definition.html).

hidmic:
> Since the article describes and outdated state (see line 17) 

Good point, though we could at least restore the link to https://www.omg.org/spec/IDL/3.5

hidmic:
@teutobod I took the liberty to push https://github.com/ros2/design/pull/264/commits/b539102d68796315cedac57b77a587a8678a19fb. @dirk-thomas do you agree? If so, let's merge this.

dirk-thomas:
Thank you both for the update.

PR #266, opened 2019-11-25
Design document for Node Interface Definition Language (IDL)

artivis:
Design document for the 'Node Interface Definition Language (IDL)',

> a simple and standardized manner to export the complete interface (action/message/parameter/service) of node(s) in a package.

Initially discussed within the Security WG, this proposal is _not_ security-related only. 
Security features and tooling building on top of this will be proposed in a subsequent document.

FYI @ruffsl @jacobperron @thomas-moulard @kyrofa


articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+- **server**: Whether or not the node provides an action server for the action.
+Valid values are "true" or "false". Defaults to "false".
+- **client**: Whether or not the node provides a client for the action.
+Valid values are "true" or "false". Defaults to "false".
+
+#### `parameter`
+
+Define the interface for a given parameter.
+
+Attributes:
```
artivis: ROS 2 introduced [read only parameters](https://github.com/ros2/rclcpp/pull/495). Should the parameters attributes includes 'ro/rw' ?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+### Other motivations
+
+Outside of security, there are several fascinating possibilities unlocked by having such an interface.
+For example, consider how this could impact [ROS 2 launch] [launch_ros]. It would be able to statically (i.e. before running anything) determine if parameter names or remappings are incorrect, among other similar sanity checks.
+Another example of the usefulness of having a static interface is the ability to create graphical tools for putting a ROS system together.
+Yet another example would be an additional feature in `ros2 pkg create` that would allow a developer to hand it an IDL and have it generate scaffolding for a node with that interface.
+
+These examples are only a subset of use-cases made possible by such an interface.
+It's clear that this is useful well beyond security.
+
```
artivis:
```suggestion

## Challenges to overcome

This proposal has a number of potential upsides, but it also has some downsides worthy of discussion. 

### This is only really useful if it gains significant adoption in upstream packages

This is a valid criticism, particularly from a security standpoint. 
However, its usefulness hopefully outweighs the work required to implement it upstream, and it's certainly something that can be contributed by community members given that the interface would be reviewed by the experts in the package.

### Declared and actual interface can get out of sync

This is certainly a concern: an out-of-date interface is debatably less useful than having no interface at all.
One potential solution for this is to more tightly couple the declared and actual interface by creating a library that consumes the declared interface and creates the corresponding ROS entities.
That has its own challenges, of course (e.g. how does it deal with wildcards?), and would need to be properly designed.
This is an area of future investigation beyond the scope of this article.

```

artivis: Not sure if those concerns belong to a design doc...
We therefore keep them as `suggestion` here until we figure that out.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+- Services (server or client)
+- Topics (publisher or subscriber)
+
+The information contained within that interface is obviously very valuable on different levels and from different perspectives.
+While it is usually readily available to a developer looking at the code, it cannot reliably be automatically extracted.
+It therefore calls for the creation of a standardized way to explicitly define and export this information.
+
+This article defines a high-level abstraction allowing upstream packages to specify the communication requirements of the nodes in the package, such that the final user, be it a developer or a static analysis tool, can benefit from it.
+The Interface Definition Language (IDL) specified in the next section is meant to be distributed alongside its associated package.
+This is true for the package source and also for the generated release. Whether the interface is declared or not is up to the package author and should not prevent the correct execution of any system pre-existing the IDL.
```
jacobperron: I'm not sure what this sentence is trying to communicate; seems like it can be removed.
```suggestion
Whether the interface is declared or not is up to the package author and should not prevent the correct execution of any system pre-existing the IDL.
```

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+Thanks to [DDS-Security][dds_security] and [SROS 2][sros2_design], security is at the heart of ROS 2.
+DDS enforces access control using a signed permissions document for each domain participant (see the [DDS-Security spec][dds_security], section 9.4).
+In SROS 2, that permissions document is generated from a ROS-specific [XML][xml_wiki] policy file that may include the permissions for one or many nodes.
+
+Currently policy files can be created in one of two ways:
+- Written by hand.
+- A snapshot of the live ROS 2 graph can be taken and written into a policy that covers its current state via `ros2 security generate_policy`.
+
+While the first option is obviously very tedious and error-prone, the second only partially alleviates the burden due to the fact that it cannot fully cover the dynamic nature of a ROS 2 graph and all of its interactions.
+More problematic than these issues, though, is that both options put the onus of security squarely on the shoulders of end users. This introduces two problems:
```
jacobperron: nit: one sentence per line
```suggestion
More problematic than these issues, though, is that both options put the onus of security squarely on the shoulders of end users.
This introduces two problems:
```

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+
+While developers will be able to define the set of rules securing their own ROS 2 nodes, the nodes developed in-house are often outnumbered by upstream components when it comes to the entire node graph, and the developers are typically not experts in every component being used.
+Without that expertise, the entire node graph cannot be properly locked down.
+Consider a complex and popular upstream component, perhaps parts of the navigation stack. **Every end user** of this component must duplicate the effort of attempting to properly lock it down.
```
jacobperron:
```suggestion
Consider a complex and popular upstream component, perhaps parts of the navigation stack.
**Every end user** of this component must duplicate the effort of attempting to properly lock it down.
```

articles/ros_node_idl.md
jacobperron: Also, I only read about one problem below. Is there a second motivation?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+
+If ROS 2 provided a way for upstream package authors to specify the interface required by the nodes in their package, and if the tools to generate security policies from that interface existed, neither of these problems would exist.
+
+### Other motivations
+
+Outside of security, there are several fascinating possibilities unlocked by having such an interface.
+For example, consider how this could impact [ROS 2 launch] [launch_ros]. It would be able to statically (i.e. before running anything) determine if parameter names or remappings are incorrect, among other similar sanity checks.
```
jacobperron:
```suggestion
For example, consider how this could impact [ROS 2 launch][launch_ros].
It would be able to statically (i.e. before running anything) determine if parameter names or remappings are incorrect, among other similar sanity checks.
```

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+These examples are only a subset of use-cases made possible by such an interface.
+It's clear that this is useful well beyond security.
+
+## Package interface
+
+How do upstream packages specify their interface requirements?
+Through a high-level description of all the actions, parameters, services and topics, provided or required, by each node within the package.
+
+The package interface is defined in a separate [XML][xml_wiki] file and exported from the `package.xml` using [REP 149's export mechanism][rep149_export], thereby avoiding pollution of the package manifest.
+The interface may cover only a subset of nodes in a package, as long as the nodes that _are_ covered are done so completely.
```
jacobperron: In the "background" section, it says the opposite; that a partial IDL is allowed. Although it would be nice, it sounds difficult to enforce that a nodes interface is covered completely by the IDL (e.g. changes may cause them to become out-of-sync). Maybe it's better to write something like,

> Though it is not enforced, it is preferred that a nodes interface is covered completely by the description.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+
+{% include_relative ros2_node_idl/interface_declaration.xml %}
+
+Once an IDL file is written, it is exported from the package manifest:
+
+{% include_relative ros2_node_idl/package.xml %}
+
+### Schema for `package.xml`'s export tag
+
+#### `interface`
```
jacobperron: To add a little more specificity, I would call this `node_interface`.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+
+This is how the package exports its defined IDL for other tools to consume.
```
jacobperron: I might also be worth mentioning that there could be one or more instances of this tag.
I could imagine a package author wanting to describe their nodes in separate files.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+
+Considering the high level abstraction of the IDL, the `interface` tag is not meant for a specific package but rather declares intrinsic properties for anyone to process it.
+However, the keyword is likely solidly descriptive enough to either be accepted as falling under [REP 149][rep149_export] or else to motivate an amendment of the REP.
+
+### IDL Schema
+
+#### `interface`
+
+This tag contains the package interface, made up of a collection of node interfaces.
+Root tag of the IDL file. There must be only one <interface> tag per IDL file.
```
jacobperron: Suggested rephrasing:

> The root tag made up of a collection of node interfaces.
> There must be only one <interface> tag per IDL file.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,182 @@
...
+
+Attributes:
+- **version**: version of schema in use, allowing for future revisions.
+
+#### `node`
+
+Encapsulate a sequence of ROS entity interfaces.
```
jacobperron:
```suggestion
Encapsulates a sequence of ROS entity interfaces.
```

articles/ros2_node_idl/interface_declaration.xml
```diff
@@ -0,0 +1,15 @@
+```xml
+<interface version="1">
+  <node name="node_1" >
+    <parameter name="verbose" type="bool" />
+    <topic name="chatter" type="std_msgs/String" publisher="true" />
+  </node>
+
+  <node name="node_2" >
+    <parameter name="rate" type="int" />
+    <topic name="/foo/bar" type="std_msgs/String" subscribe="true" />
```
jacobperron:
```suggestion
    <topic name="/foo/bar" type="std_msgs/msg/String" subscription="true" />
```

articles/ros2_node_idl/interface_declaration.xml
```diff
@@ -0,0 +1,15 @@
+```xml
+<interface version="1">
+  <node name="node_1" >
+    <parameter name="verbose" type="bool" />
+    <topic name="chatter" type="std_msgs/String" publisher="true" />
```
jacobperron:
```suggestion
    <topic name="chatter" type="std_msgs/msg/String" publisher="true" />
```

articles/ros_node_idl.md
jacobperron: I think it's worth adding potential issues and how we plan/could address them.

I'm not convinced that adoption is a challenge. I'd argue a node IDL is still useful to individuals, or organizations using it internally.

Regarding sync issues, we could at the very least have a tool that launches each node and does a best-effort sanity check against the IDL (maybe this is what you are describing, I couldn't tell). It could report any discrepancies between the declared and actual interfaces. I bet this check could be added as a unit test in a lot of cases. Beyond that, I think we just have to rely on authors and community reports of inconsistencies.

jacobperron (commented):
Thanks for working on this, I think it's a good initiative and could be quite valuable.

In addition to my comments below, it would also be nice to add an "alternatives" section where we document other options considered (if any) related to the node IDL.

articles/ros_node_idl.md
artivis: The "background" section and this part are not in contradiction, it is a matter of re-phrasing I guess.
For short, the IDL may be _partial_ at the package level, e.g. only 1 node out of 2 has its IDL exported. However the IDL must be _complete_ at the node level, if it exists.
This does raise some difficulties , such as out-of-sync as you pointed out.

articles/ros_node_idl.md
artivis: Other tags descriptions are still infinitive. Shall they all be fixed or discard this?

articles/ros_node_idl.md
artivis: It essentially tries to say that the idl xml file should be distributed along with the source and the released package (deb etc). I'll rephrase that.

articles/ros_node_idl.md
artivis: Ok. I'll keep this as suggestion until updated with plan to address them.

The adoption challenge mentioned here is maybe a little too security (future plan/tools) centric and its tone could be voiced down a little.

As for your suggestion on the sync issue it will be added as potential plan to address it.

artivis:
@jacobperron Thanks for this initial review. I'll address it asap.

artivis:
> In addition to my comments below, it would also be nice to add an "alternatives" section where we document other options considered (if any) related to the node IDL.

I'm not sure of what alternatives we are talking about, alternative ideas to the concept presented here (explicitly exported node api) or alternative implementation design of the IDL or something else ?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,185 @@
...
+Every ROS 2 node has an associated interface that describes how it communicates with other nodes, as well as how it is to be configured.
+This interface is defined in code, and consists of:
+
+- Actions (server or client)
+- Parameters
+- Services (server or client)
+- Topics (publisher or subscriber)
+
+The information contained within that interface is obviously very valuable on different levels and from different perspectives.
+While it is usually readily available to a developer looking at the code, it cannot reliably be automatically extracted.
```
dirk-thomas: I assume you are implicitly assuming non-runtime here which would be good to mention. At runtime it should absolutely be possible to introspect these information from a node.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,185 @@
...
+- **server**: Whether or not the node provides a server for the service.
+Valid values are "true" or "false". Defaults to "false".
+- **client**: Whether or not the node provides a client for the service.
+Valid values are "true" or "false". Defaults to "false".
+
+#### `topic`
+
+Define the interface for a given topic.
+
+Attributes:
```
dirk-thomas: Since the QoS settings are deciding of two endpoints on the same topic name are being matched wouldn't it make sense to also specify those? Otherwise any "connectivity analysis" on these metadata might be incorrect.

articles/ros_node_idl.md
jacobperron: It's good as it is.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,185 @@
...
+
+### IDL Schema
+
+#### `interface`
+
+This tag contains the package interface, made up of a collection of node interfaces.
+Root tag of the IDL file. There must be only one <interface> tag per IDL file.
+
+Attributes:
+- **version**: version of schema in use, allowing for future revisions.
```
dirk-thomas: Why should the `interface` tag itself be versioned and not the `node` tag?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### `node`
```
dirk-thomas: Why should multiple independent nodes be described in the same file? I would think a one node-per-file mapping would be much more natural since they can evolve independently.

dirk-thomas (commented):
Using IDL for this is imo a really confusing choice. ROS 2 already used the OMG IDL spec for defining message definitions. Therefore I strongly suggest a different abbreviation / terminology for this.

articles/ros_node_idl.md
jacobperron: IMO, it's okay to not include the read only status. We can already query this constraint on parameters (and other constraints) at runtime. Unless there's a strong case for static analysis, I feel like describing it in the node IDL just adds another point where information can go out-of-sync.

articles/ros_node_idl.md
jacobperron: I agree that it's more natural to have one node per file. I think both option should be possible.

artivis:
> Using IDL for this is imo a really confusing choice. ROS 2 already used the OMG IDL spec for defining message definitions. Therefore I strongly suggest a different abbreviation / terminology for this.

We were expecting this particular concern to be raised as we share it too. However we kept this terminology  - 'Node IDL' - as it best describes the proposal. We're open to suggestions tho.

articles/ros_node_idl.md
artivis: In our mind this mechanism is very similar to that of plugins and therefore both options are possible. I'll make this explicit in the text.

articles/ros_node_idl.md
artivis: The `interface` version correspond to the version of the IDL schema in use, in case it evolves in the future. What should the `node` version represents? The same thing at a finer grain? Is that necessary?

articles/ros_node_idl.md
artivis: Fair point, going to update that.

articles/ros_node_idl.md
dirk-thomas: I misunderstood the meaning of the version. I was somehow thinking about evolving versions of a nodes interfaces.

Yes, identifying the XML schema in the `interface` tag sounds good - no need for that in the node tag.

articles/ros_node_idl.md
artivis: No we are not implicitly assuming non-runtime.
This is somewhat related to a point made in the `Security motivation` section were we consider the `ros2 security generate_policy` utility that only captures 
> snapshot of the live ROS 2 graph

Such snapshot(s) may very well miss some dynamic aspects of the overall graph.

On an other hand we considered code introspection tools as far from ideal - e.g. for cpp.

articles/ros_node_idl.md
dirk-thomas: It might be good if the parameter description could be separated. I could see it being read independently by a node which doesn't have a "Node IDL" but wants to read the declared parameters from a file. That would imply that the parameter would need several additional arguments though (all which are possible to pass using the parameter declaration API: ranges, read-only, ...).

dirk-thomas (commented):
> We were expecting this particular concern to be raised as we share it too. However we kept this terminology - 'Node IDL' - as it best describes the proposal. We're open to suggestions tho.

Also the "definition language" part of IDL doesn't make sense since the information is described in XML and not like message interfaces in an actual "definition language" (specific syntax).

Maybe "Node API definition / descriptor"? (We can figure that out later though to not derail the discussion with a bike shed topic like this.)

artivis:
> Also the "definition language" part of IDL doesn't make sense since the information is described in XML and not like message interfaces in an actual "definition language" (specific syntax).

We did tripped a little over the `IDL` acronym. Depending on the source the `D` letter stands for different things and we started this document with `Description` in mind rather than `Definition`.

> Maybe "Node API definition / descriptor"? (We can figure that out later though to not derail the discussion with a bike shed topic like this.)

Agreed.

articles/ros_node_idl.md
dirk-thomas: > No we are not implicitly assuming non-runtime.

I would argue you do. All topics / service which are created during startup can perfectly be introspected at runtime. (Or do you see a case where it is not?) It would be technically possible to 1) start a (well written) component (which hopefully uses a lifecycle) and 2) query its "API" when it has finished initialization and using that information 3) generate the XML file representing the nodes interface.

Even with the "Node IDL" there is no limitation that a node can't change its interface later at runtime (add new topic, rename parameter, remove services). This kind of dynamic behavior can obviously not be captured by any static description.

articles/ros_node_idl.md
artivis: @jacobperron Not that one of the point of this proposal is to access the API info in a 'pre-runtime' fashion - e.g. at launch-parsing time. So querying for ro/rw and other constraints at runtime is kinda too late already.

@dirk-thomas I agree that it might be interesting, but it does feel a little out of scope here. Do you mean something a la ROS 1 [rosparam_handler](https://github.com/cbandera/rosparam_handler/blob/develop/doc/HowToWriteYourFirstParamsFile.md)?

articles/ros_node_idl.md
artivis: The "Node IDL" indeed does not enforce anything and could possibly go out-of-sync. It is one of the 'challenge' discussed in a comment/suggestion above.

The rational precisely covers the kind of dynamic behavior you are describing, assuming that the node author knows best the whole API and writes the (up to date) "Node IDL". Depending on the application, we foresee the development of some tools to catch the other dynamic behavior such as launch-time remapping.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,187 @@
...
+- Parameters
+- Services (server or client)
+- Topics (publisher or subscriber)
+
+The information contained within that interface is obviously very valuable on different levels and from different perspectives.
+While it is usually readily available to a developer looking at the code, it cannot reliably be automatically extracted.
+It therefore calls for the creation of a standardized way to explicitly define and export this information.
+
+This article defines a high-level abstraction allowing upstream packages to specify the communication requirements of the nodes in the package, such that the final user, be it a developer or a static analysis tool, can benefit from it.
+The Interface Definition Language (IDL) specified in the next section is meant to be distributed alongside its associated package, be it in the source code or a generated release packaging format (e.g. debian).
```
gbiggs: I suggest you find a catchy name. "IDL" is a known term in computer science, ROS 1 defines an IDL, and DDS uses "the" IDL (the one from OMG).

"NoDL"?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,187 @@
...
+Whether the interface is declared or not is up to the package author and should not prevent the correct execution of any system pre-existing the IDL.
+Similarly, the declared interface may be only partial and allow for the full use of pre-existing systems and the use of dependent systems on the parts covered by the partial interface.
```
gbiggs: What will be the impact of not having a complete description of node interfaces? Will you provide a way for a user to know if parts of their system are missing node descriptions? If not, will tools still be useful if the user does not know if they are impacting the entire system? Even if you do, will tools still be useful if applied to just part of the system?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,187 @@
...
+While the first option is obviously very tedious and error-prone, the second only partially alleviates the burden due to the fact that it cannot fully cover the dynamic nature of a ROS 2 graph and all of its interactions.
+More problematic than these issues, though, is that both options put the onus of security squarely on the shoulders of end users.
+This introduces two problems:
+
+While developers will be able to define the set of rules securing their own ROS 2 nodes, the nodes developed in-house are often outnumbered by upstream components when it comes to the entire node graph, and the developers are typically not experts in every component being used.
+Without that expertise, the entire node graph cannot be properly locked down.
+Consider a complex and popular upstream component, perhaps parts of the navigation stack.
+**Every end user** of this component must duplicate the effort of attempting to properly lock it down.
+
+If ROS 2 provided a way for upstream package authors to specify the interface required by the nodes in their package, and if the tools to generate security policies from that interface existed, neither of these problems would exist.
```
gbiggs: Why is this problem not solved by upstream package authors providing template security policies?

articles/ros_node_idl.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Other motivations
+
+Outside of security, there are several fascinating possibilities unlocked by having such an interface.
+For example, consider how this could impact [ROS 2 launch][launch_ros].
+It would be able to statically (i.e. before running anything) determine if parameter names or remappings are incorrect, among other similar sanity checks.
+Another example of the usefulness of having a static interface is the ability to create graphical tools for putting a ROS system together.
+Yet another example would be an additional feature in `ros2 pkg create` that would allow a developer to hand it an IDL and have it generate scaffolding for a node with that interface.
+
+These examples are only a subset of use-cases made possible by such an interface.
```
gbiggs: It would be better if each of these use cases were written up in detail, at least as detailed as the security one. Otherwise you might miss details relevant to making the node IDL correct for those use cases.

articles/ros_node_idl.md
gbiggs: Have a verb under `ros2 nodl` such as `verify`. Make it really quick and people could potentially run it regularly at runtime, e.g. once a minute, as a sanity check.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,187 @@
...
+It's clear that this is useful well beyond security.
+
+## Package interface
+
+How do upstream packages specify their interface requirements?
+Through a high-level description of all the actions, parameters, services and topics, provided or required, by each node within the package.
+
+The package interface is defined in a separate [XML][xml_wiki] file and exported from the `package.xml` using [REP 149's export mechanism][rep149_export], thereby avoiding pollution of the package manifest.
```
gbiggs: Why does it have to be tied to the package? Or the node, for that matter? I think it could be useful to distribute node interface descriptions separately and have several different nodes from different packages that support the same interface.

In computer science, this is one of the primary reasons for using interfaces: it allows substituteability. Particularly if you are going after static analysis tools and system constrution tools, such capability would be close to essential.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,187 @@
...
+
+{% include_relative ros2_node_idl/package.xml %}
+
+Note that several IDL files can be exported, allowing for writing an IDL file per node.
+
+### Schema for `package.xml`'s export tag
+
+#### `interface`
+
+This is how the package exports its defined IDL for other tools to consume.
```
gbiggs: I was not aware that ROS 2 had such a tight coupling between packages and nodes.

articles/ros_node_idl.md
gbiggs: I think that allowing multiple nodes in a file is something you would come to regret.

I also think that if you are allowing mutliple nodes in a file, then you are describing a *Package* IDL, not a *Node* IDL. Although this comment applies generally to the tone throughout this design document.

gbiggs (changes requested):

articles/ros_node_idl.md
gavanderhoorn: So I see the potential for confusion, but what is proposed does seem like it could be called an IDL. For nodes. So *Node IDL* seems appropriate.

articles/ros_node_idl.md
artivis: I rephrased this. Can you please checked if it is fine now and eventually resolve this?

articles/ros_node_idl.md
artivis: I rephrased this. Can you please checked if it is fine now and eventually resolve this?

articles/ros_node_idl.md
artivis: @gbiggs Do you have an example to illustrate this? As long as one can parse the info on a per-node basis I don't see any problem.

articles/ros_node_idl.md
artivis: I believe the terminology is appropriate, however I do agree that it could be confusing. As discussed in another thread with @dirk-thomas we leave the potential renaming to later on.

ps - I like 'NoDL', other acronymes include 'NAPI', 'NAPID' (Node API Description).

articles/ros_node_idl.md
artivis: For clarification, just in case: _partial_ here applies at the package level, not all node are required to have their interface described. However, if the interface of a given node is described, then it must be described _fully_.

The take here is that not every node in a package are meant to be part of a live graph, some may be meant for visualization, debugging and what not and therefore exporting their interface may not be useful.

> What will be the impact of not having a complete description of node interfaces?

It is assumed that a tool *cannot* be used for a node with no exported interface, nor for a node partially described (the later may be subject to technical difficulties).

> Will you provide a way for a user to know if parts of their system are missing node descriptions?

This is a bit of a chicken & egg problem. The node IDL proposal addresses the issue that a system cannot be fully described with automatic tools since such tool may not capture the entire dynamic aspect of the graph. There could be tools such as the `ros nodl verify` you mentioned but we should not sell it as an exhaustive one. Maybe `ros nodl try_verify` would then be more appropriate.
Related to this [discussion](https://github.com/ros2/design/pull/266#discussion_r350378953).

> will tools still be useful if the user does not know if they are impacting the entire system? Even if you do, will tools still be useful if applied to just part of the system?

From our perspective - automating security - the entire live system should be described.
But that may not be true for all tools, thus it would depend on the tool in use, assuming the aforementioned specifications on '_partial_'.

articles/ros_node_idl.md
artivis: I'd say that's a multi factor issue. I'll cite two at least that are related to one another,
- Adoption, security is likely among the last of people's concerns.
- Difficulty, it's not that simple to write policies xml doc(s).

This Node IDL proposal together with a following security-specific proposal aim at tackling both at the same time with a relatively simple solution (for the pkg developer). 
Decoupling the Node IDL from security seemed like an opportunity to enable other possibilities and tools.

articles/ros_node_idl.md
artivis: Our primary concern is security, that's how this proposal born. These other use cases are only speculation at this point. We could develop them a little but likely not as much as the sec one.

articles/ros_node_idl.md
gavanderhoorn: > Decoupling the Node IDL from security seemed like an opportunity to enable other possibilities and tools.

Certainly.

I've posted a link to this PR in the [ROS Driver Common Interface Standards](https://discourse.ros.org/t/ros-driver-common-interface-standards/11435/37) discussion on ROS Discourse.

articles/ros_node_idl.md
ruffsl: With XML, one could always xinclude from external paths, such that you could break your "package" IDL into separate node "IDL" files, if one wanted to.

articles/ros_node_idl.md
jacobperron: This is a nice idea. Then there's the question: how to communicate that your node implements a particular NoDL?

jacobperron: We could have common packages that just contain NoDL files, e.g. `sensor_nodls`, `nav_nodls`, etc.

articles/ros_node_idl.md
jacobperron: lgtm

articles/ros_node_idl.md
jacobperron: Okay, makes sense. Similar to https://github.com/ros2/design/pull/266#discussion_r350549359, I suggest that we re-focus the tone of this article to be about the node description and less so about the package integration.

articles/ros_node_idl.md
jacobperron: > From our perspective - automating security - the entire live system should be described.
But that may not be true for all tools, thus it would depend on the tool in use, assuming the aforementioned specifications on 'partial'.

Makes me think of possible use-cases for partial NoDLs. For example, I write a LIDAR driver node that implements a common scan_nodl, but I also have an extra publisher. For security purposes, I want to describe all topics of my node, but for a system construction tool I only care that it implements scan_nodl. As the node author, it'd be nice to include scan_nodl into my custom NoDL and somehow communicate that my node implements scan_nodl.

On the other hand, maybe composable NoDLs are too fragile. If the upstream scan_nodl changes, my node may break my system unexpectedly at runtime. It might be easier to maintain a non-composed NoDL for my node, compared to relying on external descriptions.

articles/ros_node_idl.md
artivis: Yeah how one would tied node and NoDL back together if they are distributed separately. Also I believe that people are somewhat familiar with the plugin export scheme, so a look alike scheme would ease adoption.
Finally separating NoDL files in a separate pkg(s) just increase the chances that the files go out-of-sync.

I do see and appreciate @gbiggs' point on having some kind of template NoDL files tho.

articles/ros_node_idl.md
artivis: Well this isn't really an example of a partial NoDL file but rather a case for template/inheritable/composable one. The final NoDL of your driver would still export the scan_nodl + extra publisher and thus be complete.

I do see the potential for composition for NoDL, but it requires some more thinking - how to distribute template NoDL files ? How to prevent breaking from upstream ? Etc.

articles/ros_node_idl.md
gbiggs: I'm sure your primary concern is security, but if you are promoting this as something for use cases outside of security then I feel that those use cases should also be properly described and taken into account. Perhaps gather some feedback specifically on those use cases at Discourse?

articles/ros_node_idl.md
gbiggs: @artivis I think you are looking at this too much from the point of view of describing existing nodes, and not enough from the point of view of describing types. Describing types is ultimately what you are doing, and I think the correct approach is to find a way to state that a node implements a particular NoDL interface rather than saying that every node is a completely unique, unsubstitutable type that needs its own NoDL file.

In some ways this is a philosophical question about where the substitution level is in ROS. There are probably those who say it is at the level of topic, and those who say it is at the level of node. There are definite benefits to doing the later, but it does require more careful engineering and/or tools to support and ROS users may not have an appetite for that. I would argue, though, that if they are at the level of caring about proper security they are probably at the level of caring about good engineering practices and so wouldn't mind some extra complexity if it gives useful benefits, like powerful static analysis tools.

However I think the most important is that the reality is that people treat many nodes in this way already, and the state publisher node discussion on Discourse is a currently-being-discussed example of this.

articles/ros_node_idl.md
gbiggs: > Decoupling the Node IDL from security seemed like an opportunity to enable other possibilities and tools.

I absolutely agree, but I think this needs to be done properly to ensure the full potential is realisable. I think you are currently too focused on the package concept and I'm not sure that is correct. I could be convinced otherwise, of course.

articles/ros_node_idl.md
gbiggs: I didn't mean to imply the terminology is incorrect, it is perfectly fine. But it's an over-used term and a catchy name avoids that. :)

articles/ros_node_idl.md
gbiggs: Welcome to the wonderful world of interface design! ;) There's oodles of theory behind all this stuff.

articles/ros_node_idl.md
kyrofa: Yeah, no strong feelings on the name. It is indeed an IDL, but we can name it whatever we like.

articles/ros_node_idl.md
kyrofa: Alright, updated suggestion with what was discussed here and pushed it.

articles/ros_node_idl.md
kyrofa: In this context, I think that describing nodes versus types is a distinction without a difference. In the end, node A has interface X, regardless of where X comes from. As @ruffsl points out in [another thread](https://github.com/ros2/design/pull/266/files#r350898103), we can xinclude files from other packages if we want to declare X in one place and have multiple packages/nodes use it. I don't see a solid use-case for designing much more than that at this time.

articles/ros_node_idl.md
kyrofa: Can you elaborate on this? I tend to lean toward the simplest system that will work instead of immediately trying to enable everything a developer would possibly want to do. Having multiple interface files is a complexity that I'm not sure is currently warranted, particularly given the ability to xinclude. We can of course add it, I just want to learn more about what's in your head here.

jacobperron: I was thinking that we could define nodes interfaces in one or more files. This comment thread is more relevant: https://github.com/ros2/design/pull/266#discussion_r350382380.
In that case, we might expect there to be one or more occurrences of this tag in the package.xml.

articles/ros_node_idl.md
kyrofa: This outer tag represents the interface for the entire package, which is _made up_ of a set of node interfaces. We waffled between `interface` (i.e. "the package interface") and `nodes`, but having a `version` on the `interface` felt nicer than having a `version` on `nodes`.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,205 @@
...
+It's true that, if not all of the packages in one's system have adopted this, its gains are incomplete.
+However, it's still useful even if only a subset of the packages adopt it (e.g. one's own packages), which means even without signifant upstream adoption it will still be useful to individuals or organizations.
+Upstream packages that haven't adopted this simply won't benefit from it.
+Also, its usefulness hopefully outweighs the work required to implement it upstream, and it's certainly something that can be contributed by community members given that the interface would be reviewed by the experts in the package.
+
+### Declared and actual interface can get out of sync
+
+This is certainly a concern: an out-of-date interface is debatably less useful than having no interface at all.
+There are a number of possibilities that will help with this issue.
+One possibility is to more tightly couple the declared and actual interface by creating a library that consumes the declared interface and creates the corresponding ROS entities.
```
mjcarroll: I discussed this with some people, but I don't recall who was there, perhaps @ruffsl and @thomas-moulard.  We could create an extension to the `Node` concept (much like `LifecycleNode`) that consumes the NoDL file as part of the constructor (or early in execution) and would not allow topics/services/parameters/actions outside of that policy to be created either _at all_ or past a certain point in execution (perhaps coupled with the `configure` state of `LifeCycleNode`.

kyrofa: That would be an excellent way to ensure these stayed in sync. It's essentially what I was thinking when I mentioned that this could live within RCL (the next sentence, I think). The only thing I dislike about the idea is that it still requires the developers to duplicate the information: once in the NoDL and again in code. It would really be ideal if they only had to write it once, but that's a harder problem to solve.

articles/ros_node_idl.md
artivis: I added the QoS settings for actions, services and topics (da63aa62fabb216e04cd3bec6d35f54c33559e7f). I'm not familiar with the later two so please let me know your take on this.
I also went ahead assuming we would like those to defaults to the same values they do in code, hopefully I'm correctly linking to the defaults.

artivis: Forgot to mention, I don't know what type should the `qos` attribute be. I'm not sure of to translate it to a xml tag attribute here...
If anyone has a suggestion?

kyrofa: Given that they are objects in code, it may not be suited to being an attribute at all, but an object nested under the respective tags.

artivis: You mean something similar to `<rosparam>` in `roslaunch`? 
E.g.
```xml
<topic name="/foo/bar" type="std_msgs/msg/String" subscription="true">
  <qos>
    policy_history: best_effort
    policy_reliabiliity: reliable
    ...
  </qos>
</topic>
```

kyrofa: That's the right idea, yeah, although I'd make those QoS attributes, maybe something like:

```
<topic name="/foo/bar" type="std_msgs/msg/String" subscription="true">
  <qos policy_history="keep_last" policy_reliabiliity="reliable" \>
</topic>
```

> I also went ahead assuming we would like those to defaults to the same values they do in code

I think this makes sense. Given that the rationale for the defaults is "keep behavior like ROS 1" I doubt they will change and get out of sync.

articles/ros_node_idl.md
gbiggs: I think to solve that you would have to take the same approach as messages, and resort to generating base classes for node implementations to inherit from. At this point you are starting to get well into the realm of model-based tool chains and all the benefits and risks that entails. One big risk is the restrictions it places on developers in how they write and manage their code, which can become a barrier to adoption due to the increased learning required. Tools that have done this in the past have seen limited adoption while ROS has been happily adopted by people all over the place.

It's difficult to say which is the correct way to go because there are many facets to the decision.

articles/ros_node_idl.md
gbiggs: I'm not asking for designing too much, but I think that we need to make sure what we design now does not close off future potential. Having a way to define node interfaces, parameters a node can take, etc. outside of the source code opens up huge possibilities for better engineering practices for ROS users, but it needs careful thought for what exactly is defined and how to ensure that we don't make it impossible to do, for example, easy substitution of equivalent nodes, or nodes with optional extensions to a defined interface, or verifying the compatibility of two nodes at launch time or in some pre-launch system check.

gbiggs:
There is something I haven't seen made explicit yet, but which I think we should. This is about where the definitions are gathered and what they mean, semantically.

The current proposal says that a package can define the interfaces of the nodes it includes. Accepting that this can be done using `xinclude`s to include multiple XML files that define individual nodes, this is what I think is semantically being created by the current proposal:

- A package is a namespace, defined by a directory and a package-level IDL file
- Within the package namespace are one or more object types, defined in the package-level IDL file or with individual files collected into that (i.e. the IDL descriptions of nodes)
- Within the package namespace are instances of those types, defined by source code that implements them (i.e. the node implementations)
- Within the package namespace are one or more configuration set types, defined in the package-level IDL file or with individual files collected into that (i.e. the IDL descriptions of parameters)
- Within the package namespace are uses of the configuration set types (i.e. the actual declaration and use of parameters by nodes)
- A node from one package could declare that it implements a node interface defined in a Node IDL in another package.

I may have missed some parts by I think it gets my point across.

At the source level, the "package namespace" concept does not really exist in current ROS. However it *does* exist at the tooling level, in tools such as `ros2 run` and `launch`.

What I want to be clear on is if we are all happy making the "package namespace" concept much more formal than it currently is and extend it throughout the run-time ROS system. Having run-time checks of "package interfaces" will do that because it makes the package a run-time entity.

Personally I am fine with that, I think it's a natural extension of where things are currently and there are benefits to doing so for identifying specific node implementations, even in cases where two nodes from two packages implement the exact same interface declared in the exact same Node IDL file.

articles/ros_node_idl.md
artivis: Updated `qos` so that it is a nested object in `action`, `service` and `topic`.
Updated the example to highlight `qos`. The example is not fully finished.

peterpolidoro:
Would the Node IDL make it possible to create some sort of rosbridge between nodes of different ROS 2 distributions?

It could be very liberating in large heterogeneous systems to allow nodes running on different distributions to communicate with each other. Right now if one node is stuck on an old distribution, for one reason or another like being proprietary code, it cannot participate in the system unless all other nodes are held or regressed back to that old distribution. Even if it was less efficient, a bridge could allow most nodes to upgrade at will while still letting old nodes participate in the system.

kyrofa:
That's an interesting idea, @peterpolidoro. Off-hand I'd say it could help with it inasmuch as the node in question might change what it calls a given topic across distros, but the IDL doesn't dive into message contents, so it would be fairly limited. Specifically, if the message used by the given topic changed its structure, the IDL can't really help you there.

peterpolidoro:
> but the IDL doesn't dive into message contents

Perhaps the message contents are outside the scope of the Node IDL, but that information could be very valuable in a variety of situations. It could allow you to verify, for safety or consistency, that a message is defined as you expect. It might also be used to create some sort of plug-and-play interface that allows you to first ask a node to give you its IDL and use that to generate the messages necessary to communicate with that node so you do not need to access the original package that defined the node messages. It might be nice at times to be able to get everything you need to communicate with a node locally from the node itself, instead of relying on a connection to the outside world and finding the proper interface version for that node.

articles/ros_node_idl.md
```diff
@@ -0,0 +1,249 @@
...
+- **publisher**: Whether or not the node publishes on the topic.
+Valid values are "true" or "false". Defaults to "false".
+- **subscription**: Whether or not the node subscribes to the topic.
+Valid values are "true" or "false". Defaults to "false".
+
+Sub-tag:
+- **qos**: The Quality of Service setting for the topic - see the [qos tag](####qos).
+Defaults to [rmw_qos_profile_default][rmw_qos_profile_default_link]
+
+#### `qos`
```
Arnatious: rcl defines a number of qos presets (https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/qos.py#L343 contains some  of them in rclpy).

We should offer the ability to refer to these by name. Adding an attribute for qos to the topic, action, and service tag would allow this.

For instance, the user could specify

```xml
<topic qos="sensor_data">
  ...
</topic>
```
where `sensor_data` corresponds to `qos_profile_sensor_data`. The idl parser can prepend `qos_profile_` when looking up the qos struct for validation purposes.

articles/ros2_node_idl/interface_declaration.xml
```diff
@@ -0,0 +1,25 @@
...
+        reliability=""
+        durability=""
+        deadline=""
+        lifespan=""
+        liveliness=""
+        liveliness_lease_duration=""
+        avoid_ros_namespace_conventions="false" />
+    </topic>
+    <service name="/example_service" type="std_srvs/srv/Empty" client="true" />
+    <action name="/example_action", type="example_interfaces/action/Fibonacci" client="true" />
```
Arnatious:
```xml
<action name="/example_action", type="example_interfaces/action/Fibonacci" client="true" />
                           ---^--- Invalid token
```

articles/ros_node_idl.md
artivis: Following the discussion [here](https://github.com/ros2/design/pull/266#discussion_r350380468) we opted for a nested object approach to define `qos` (see e.g. [here](https://github.com/ros2/design/pull/266/files#diff-ddd4bb0e2616a5d1c2e7667c294d309eR11)) rather than an attribute as you seem to be suggesting. I agree that it would be handy to be able to re-use those pre-defined profiles somehow. Maybe there could be another `qos` attribute such as follow,
```xml
<qos profile="sensor_data">
```
but then we would have to handle cases such as defining a `profile` and some other options, e.g.,
```xml
<qos profile="sensor_data" history=5>
```
What would we do in such case? Override the `history` param of the `sensor_data` `profile`, raise an exception, something else?

ros-discourse:

https://discourse.ros.org/t/guidelines-on-how-to-architect-ros-based-systems/12641/17

articles/ros_node_idl.md
git-afsantos: > I do see the potential for composition for NoDL, but it requires some more thinking - how to distribute template NoDL files ? How to prevent breaking from upstream ? Etc.

A possibility would be to specify that a node implements a *version* of a template (as if pointing to a git commit). If the upstream version changes, you are not obligated to implement newer versions. Of course, this might be too much clutter already.

git-afsantos: > Makes me think of possible use-cases for partial NoDLs.

I have a use case for partial descriptions in static analysis, but I am kind of in a rush right now. I will write an elaborate post later on.

articles/ros_node_idl.md
git-afsantos: As I mentioned in another comment, I believe I have a use case for this in static analysis. I am still reading all of this for the first time though. I will write a more detailed comment once I go through all of it.

Arnatious:
The proposal has been updated to match the implementation in https://github.com/ubuntu-robotics/nodl .

Changes of note:
- Added "executable" as an attribute at the node level, specifying the executable containing the node (needed for launch integration)
- Export mechanism specified as using the ament index plugin system
- Added details of how to export NoDL files from a package using CMakeLists.txt or setup.py
- Removed QoS related functionality, can be proposed later as an extension/future revision to NoDL
- Added mention of a formal ".xsd" schema that is provided in the NoDL package
- "Node IDL" changed to "NoDL" or "Node Definition Language" throughout

articles/ros2_nodl.md
```diff
@@ -0,0 +1,241 @@
...
+Define the interface for a given action.
+
+Attributes:
+- **name**: The name of the action.
+- **type**: The type of the action.
+Valid values are any ROS action types.
+- **server**: Whether or not the node provides an action server for the action.
+Valid values are "true" or "false". Defaults to "false".
+- **client**: Whether or not the node provides a client for the action.
+Valid values are "true" or "false". Defaults to "false".
```
Arnatious: Currently, the two bools for server/client (or publisher/subscription in the topic case) have an *OR* requirement imposed on them. A node should never have both be false. 

This adds a bit of logic that can't be expressed in the `.xsd` and requires a check in the parsing library and additional error handling (https://github.com/ubuntu-robotics/nodl/blob/f143f51174fb2b5954651534af89bdd165c82bc5/nodl_python/nodl/_parsing/_v1/_parsing.py#L29).

Proposal:
Create a required enum typed "role" tag for Action, Service, and Topic.

Enums would be `ServerClientRole: server, client, both` and `PubSubRole: publisher, subscription, both`.

This would allow the schema to enforce this rule and give Action, Service, and Topic an identical interface (name, type, role).

Arnatious (changes requested):

articles/ros2_nodl.md
kyrofa: That seems okay to me, does anyone else have any thoughts?

ros-discourse:

ros-discourse:

https://discourse.ros.org/t/matching-the-dynamic-nature-of-ros-to-the-static-nature-of-dds-security/15653/1

ros-discourse:

https://discourse.ros.org/t/announcing-a-ros-package-generator/16317/3

clalancette:
@artivis What's the status of this design?  I see there has been lots of discussion, though no approvals as of yet.  Is there more work that needs to be done here?

artivis:
> @artivis What's the status of this design? I see there has been lots of discussion, though no approvals as of yet. Is there more work that needs to be done here?

Let me forward this to @Arnatious who has picked up this up.

Arnatious:
> 
> 
> @artivis What's the status of this design? I see there has been lots of discussion, though no approvals as of yet. Is there more work that needs to be done here?

There's been no comment since the last revision of the nodl schema, which added the `role` attribute. I'm taking that as a good sign.

We have a reference implementation in https://github.com/ubuntu-robotics/nodl, and are nearly done with a simple `launch_ros` integration pending some changes in SROS2. I think this proposal is ready to move ahead.

clalancette:
> We have a reference implementation in https://github.com/ubuntu-robotics/nodl, and are nearly done with a simple `launch_ros` integration pending some changes in SROS2. I think this proposal is ready to move ahead.

All right, then we need some approvals.  Once we have that, I'll go ahead and merge.

budrus:
@Arnatious Have you thought of using yaml instead of xml? I just read some PRs and saw that for QoS settings first xml was discussed in #296 and then they decided to go the yaml way like for parameters in #300. Is it because of DDS or security interaction?

kyrofa:
> Have you thought of using yaml instead of xml?

We actually prefer YAML, but DDS and SROS2 use XML, and the package.xml (which the nodl would be alongside) is also XML. Using YAML just didn't seem consistent.

articles/ros2_nodl.md
```diff
@@ -0,0 +1,235 @@
...
+- Services (server or client)
+- Topics (publisher or subscriber)
+
+The information contained within that interface is obviously very valuable on different levels and from different perspectives.
+While it is usually readily available to a developer looking at the code, it cannot reliably be automatically extracted.
+It therefore calls for the creation of a standardized way to explicitly define and export this information.
+
+This article defines a high-level abstraction allowing upstream packages to specify the communication requirements of the nodes in the package, such that the final user, be it a developer or a static analysis tool, can benefit from it.
+The Node Definition Language (NoDL) specified in the next section is meant to be distributed alongside its associated package, be it in the source code or a generated release packaging format (e.g. debian).
+Whether the interface is declared or not is up to the package author and should not prevent the correct execution of any system pre-existing the NoDL.
```
bergercookie, line 43: Would `predating` be more appropriate than `pre-existing`?

ros-discourse:

https://discourse.ros.org/t/eprosima-presents-visual-ros/17632/3

articles/ros2_nodl.md
git-afsantos: I would say either *'correct execution of any pre-existing system'* or *'correct execution of any system predating the NoDL'*. The latter is probably better to portray the intent.

ros-discourse:

https://discourse.ros.org/t/eprosima-presents-visual-ros/17632/4

articles/ros2_nodl.md
```diff
@@ -0,0 +1,235 @@
...
+            ['package.xml', package_name + '.nodl.xml']),
+    ],
+```
+
+### NoDL Schema
+
+An `.xsd` xml schema is provided alongside the NoDL implementation.
+This can be used to programmatically validate the NoDL's `.xml` document.
+There are some semantics that cannot be expressed in this schema, so the the `.xsd` is not authoritative.
+Rather, it is a heuristic, and the [NoDL reference implementation][nodl-reference] can reject a document that does not conform to other requirements.
```
dirk-thomas, line 168: I just noticed that the link in this sentence doesn't render correctly.

aprotyas:
Hi @Arnatious, over the last few weeks - with the help of this design doc and the [`ubuntu-robotics/nodl` API](https://github.com/ubuntu-robotics/nodl) - some preliminary work has been done in writing a [`nodl_to_policy`](https://github.com/aprotyas/nodl_to_policy) tool, that converts NoDL description files to a corresponding [ROS 2 access control policy file](http://design.ros2.org/articles/ros2_access_control_policies.html). In doing so, we've come across some ambiguities with the NoDL description that I wanted to raise for discussion here:

- The NoDL description seems to be lacking the concept of namespace. Every `<profile>` tag in an access control policy needs a node namespace, and insofar, the only option for namespace is to assume that all nodes are in the default "`/`" namespace, which is an invalid assumption for many reasons. Having said that, are there/should there be plans to incorporate the namespace concept into NoDL? Possibly as an additional (optional) attribute of the `<node>` tag?
- [Security enclaves](http://design.ros2.org/articles/ros2_security_enclaves.html) are seen as collections of profiles, and a valid access control policy requires their fully qualified paths. Since the NoDL description does not specify enclave paths, an alternative is to qualify an enclave path using the namespace and node name, for example: "`/[ns]/[node]`". Having said that, should namespacing get formalized as a concept in the NoDL description, enclave paths are probably a non-issue. Else, we should consider adding the concept of a node's enclave path to the NoDL description too.
- Thoughts on including default parameter services (`describe_parameters`, `get_parameters`, etc.), default topics (`rosout`, `parameter_events`), and default actions in a NoDL description? I'm assuming these are implicit with every node, and hence considered redundant from the description's POV. However, when converting from a NoDL description to an access control policy, certain assumptions have to be made about the presence/absence of said default services/topics/actions. This is not ideal from a security POV, even if this simply grants "empty" permissions. A fallout from this discussion is that a node's "type" (is it a lifecycle node?) is currently not expressed in the NoDL description. The "type" of node changes the set of "default" permission tags to consider for it.
- Should there be a field to specify a node's "type"? I'm thinking of a regular node versus a lifecycle node. An alternate could be to capture the "extra" services/topics of lifecycle nodes with the regular service/topic tags, but that might go against omitting said implicit permission types (previous point).

FYI @iche033 @clalancette @wjwwood

danthony06:
I was wondering if this is still under active development and will get merged in?

artivis:
Hi @danthony06,
We do plan on picking up this work soon. Note that altho this PR is note yet merged, NoDL-related packages are already part of [REP 2005 - ROS 2 Common Packages](https://ros.org/reps/rep-2005.html).

vmayoral:
+1 to this. Merging it would give it more visibility and we need to
encourage participation and contributors.

rkent:
I'm a little late to this discussion, but I am looking at this from the perspective of generating package documentation with a tool like rosdoc2. When we document interfaces in Python or C++, classes and parameters typically have a short description. It would not be difficult to extend rosdoc2 to read this NoDL file and extract information for documentation purposes - but nodes and parameters would need descriptions. Could you add a description attribute (type string) to the **node** and **parameter** elements for use by documentation utilities?

[update] I see that rosindex has some [recommendations](https://github.com/ros-infrastructure/rosindex/blob/ros2/contribute/metadata.md#nodes) for node documentation in the package.xml file. Coming from a documentation perspective, their examples show a lot of text for nodes descriptions.

What is the relationship of this PR to the rosindex proposal? I don't think you really want to ask package maintainers to document their packages in two places. If this PR becomes the definitive source, then you'll need the same info (that is a description attribute or element) that rosindex wanted.

fujitatomoya (commented):
+1 for this. this is 4 years old design, but description and motivation looks good. this is useful for ROS 2 end users and security enclave configuration.

emersonknapp:
To add an update here: the ROSGraph Working Group is working on this concept, and has taken on maintainership of the existing NoDL repositories and name to push the concept forward. We're working on a next phase design now.
See https://github.com/ros-tooling/nodl for the new home

A note that our main focus right now is on boilerplate code generation, node/graph validation, and documentation generation - not on SROS2 directly. However, that slightly different focus doesn't change the core idea at all and we also are keeping in mind not to hinder that use case.

I don't want to lose the context here, but I'm also not sure that adding to `ros2/design` is a prerequisite to making progress. We'll be iterating on designs and prototype implementations over at the source repo. I wonder if we should close this PR?

Also note https://github.com/ros2/rcl_interfaces/pull/188 which is strongly linked to NoDL schema, for communicating node definitions within a running ROS system.

rkent:
I'm very late to this party, and I did not carefully review the documents, but let me still give some comments from my very different perspective.

I really have two perspectives:
1. The NoDL spec, like many other ROS specs, does not consider the need to generate automatic documentation from the results.

This NoDL effort as I understand it is an attempt to add nodes, parameters, and interfaces with a spec similar to package.xml  Great, but don't forget the description fields for everything!

The main issue here is that every item that ultimately will be documented needs a description field. In the existing ROS2 system, repos have packages, packages have nodes, and nodes have parameters and interfaces. Also packages have dependencies. Within ros2, only packages have description fields. So packages can be listed and searched based on descriptions, but repos cannot. Also, system dependencies do not have descriptions, which makes listing them much more difficult.

So documentation generation is forced to try to recreate descriptions from external sources. In the case of dependencies, that is from debian and pip sources. That is currently implemented (albeit incompletely). In the case of repos, the underlying github repo often have descriptions which could be web scraped to find descriptions. That is proposed but not implemented.

The fundamental problem though is that the original specs for distribution.yaml do not include description fields for repos nor system dependencies. That forces later documention efforts to recreate them from less reliable remote sources.

I see the same thing happening here in NoDL. Ultimately, the specs being generated will want to be used in indexes and node-level documentation. But without descriptions for nodes and parameters, the resulting documentation is not very useful.

TLDR: take account of the documentation needs, and include text descriptions for everything that is being generated.

2. The ship has sailed for this to be a standard part of ROS2 package code, so realistically this is going to mostly be implemented using AI.

A couple of observations from that. First, because it will be AI generated, you should be able to add it without having to modify existing code (like package.xml) That means using standard filenames or directories instead that could be merged with an existing repo. Second, the spec should be ai-friendly. I don't know enough to really know what this means yet (though I am learning), but one thing I've done is add constraints on the length of description fields. A description that will appear in a single line is a table is very different from a summary that could be a full paragraph in documentation.

PR #267, opened 2020-01-16
Update launch frontend <let> attribute

jacobperron:
The actually implementation uses the attribute `name` instead of `var`.


ivanpauno (approved):
Thanks for the fix!

PR #268, opened 2020-02-06
avoid the term legacy for .msg/.srv/.action file format

dirk-thomas:
While we use IDL in the internal pipeline and in all message generators `.msg`/.`srv`/.`action` files are unlikely to be deprecated but will continue to be used if the additional IDL features are not needed.

### DIFF
--- a/articles/110_interface_definition.md
+++ b/articles/110_interface_definition.md
@@ -14,7 +14,7 @@ categories: Interfaces
 # {{ page.title }}
 
 <div class="alert alert-warning" markdown="1">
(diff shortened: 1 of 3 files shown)


jacobperron (approved):
LGTM

mxgrey:
Just a thought: What about calling them "simplified interface definitions"? Or maybe some other descriptive term to differentiate from IDL?

jacobperron:
I've always differentiated them as "ROS interfaces" verses IDL interfaces.

mxgrey:
I'm a little worried that terminology might give the appearance that IDL isn't supported by ROS2, but I don't know how legitimate my concern is.

PR #269, opened 2020-02-06
Publish article on loaned messages

Karsten1987:
@mjcarroll @wjwwood Are you okay with publishing this article or was it held back on purpose?


wjwwood:
I'll leave that to @mjcarroll, I would publish it so long as it's ready to be consumed.

mjcarroll (approved):

PR #278, opened 2020-04-16
Fix invalid launch XSD

hidmic:
Precisely what the title says. Addresses https://github.com/ros2/launch/issues/396.


articles/specs/launch.0.1.1.xsd
```diff
@@ -8,80 +8,82 @@
     </xs:documentation>
   </xs:annotation>
 
-  <xs:element name="launch" maxOccurs="1">
```
ivanpauno, line 11: Why should this be removed?

articles/specs/launch.0.1.1.xsd
```diff
@@ -475,7 +477,7 @@
           </xs:documentation>
         </xs:annotation>
       </xs:attribute>
-      <xs:assert test="(@name and not(@from)) or (not(@name) and @from)"/>
+      <!-- <xs:assert test="(@name and not(@from)) or (not(@name) and @from)"/> -->
```
ivanpauno: I guess this isn't supported

ivanpauno (approved):
LGTM!

Unfortunately we aren't using this schema to validate anything, so it's hard to tell if it's correct or not.

articles/specs/launch.0.1.1.xsd
```diff
@@ -8,80 +8,82 @@
...
-      <xs:simpleType>
-        <xs:restriction base="xs:string">
-          <xs:pattern value="[0-9]+\.[0-9]+\.[0-9]+"/>
-        </xs:restriction>
-      </xs:simpleType>
-    </xs:attribute>
+        <xs:element name="arg">
+          <xs:annotation>
+            <xs:documentation xml:lang="en">
+              Declares launch file-level arguments.
```
clalancette: I guess this was already there, but could you explain a little more what is meant by `file-level` arguments?  It's not clear to me.

articles/specs/launch.0.1.1.xsd
clalancette: Why are we commenting this out?

articles/specs/launch.0.1.1.xsd
```diff
@@ -524,8 +526,7 @@
         <xs:element ref="param"/>
         <xs:element ref="remap"/>
       </xs:choice>
-    </xs:complexType>
-    <xs:attribute name="package" type="xs:string" use="required">
+          <xs:attribute name="package" type="xs:string" use="required">
```
clalancette: The indentation seems odd here.

articles/specs/launch.0.1.1.xsd
```diff
@@ -608,5 +609,6 @@
           </xs:documentation>
         </xs:annotation>
       </xs:attribute>
+    </xs:complexType>
```
clalancette, line 611: It's not really clear to me that this is the right thing to do here.  I think the "complexType" here is indeed the ref to "env", "param", and "remap", and it should end there.  The rest of the attributes seem to be attributes on "node", so I think this should be moved back where it was, and the indentation should be fixed.  Or maybe I'm misunderstanding?

articles/specs/launch.0.1.1.xsd
```diff
@@ -172,6 +151,29 @@
           </xs:complexType>
         </xs:element>
       </xs:sequence>
+      <xs:attribute name="file" type="xs:string" use="required">
```
clalancette, line 154: Out of curiosity, does moving this around have any affect on what is allowed by the schema?

clalancette (commented):
A bunch of mostly questions.

articles/specs/launch.0.1.1.xsd
hidmic: Because it is not valid XSD. It is already the case as it stands that only one root `launch` element is allowed.

articles/specs/launch.0.1.1.xsd
hidmic: Oops. This shouldn't be commented out, I only did so for `xmllint` to accept this schema. There're Python modules with full XSD 1.1 support we can use if we ever attempt to validate given launch files prior to execution.

hidmic: Shouldn't be. See https://github.com/ros2/design/pull/278#discussion_r409649671.

articles/specs/launch.0.1.1.xsd
hidmic: Attributes apply to the type of the element, not the element itself (see https://www.w3schools.com/xml/el_attribute.asp). It was invalid XSD.

articles/specs/launch.0.1.1.xsd
hidmic: Same as in https://github.com/ros2/design/pull/278#discussion_r409651844.

articles/specs/launch.0.1.1.xsd
hidmic: > Declares a launch file argument.

conveys the same amount of information more clearly. Changed in  25db0215cb72dc84d9828dd25895593512f57582.

articles/specs/launch.0.1.1.xsd
hidmic: It is. Fixed in 25db0215cb72dc84d9828dd25895593512f57582.

hidmic:
@clalancette PTAL !

ivanpauno:
@hidmic I'm not sure why, but checks aren't passing.

hidmic:
> @hidmic I'm not sure why, but checks aren't passing.

Did not update all places.

clalancette (approved):

hidmic:
Alright, going in!

PR #282, opened 2020-05-01
ContentFilteredTopic Design Proposal.

fujitatomoya:
ROS 2 ContentFilteredTopic Feature Support.

- adding content-filtered-topic interfaces for Humble API freeze

* [x]  [ros2/rmw](https://github.com/ros2/rmw/pull/302)
* [x]  [ros2/rmw_implementation](https://github.com/ros2/rmw_implementation/pull/181)
* [x]  [ros2/rmw_connextdds](https://github.com/ros2/rmw_connextdds/pull/68) (CFT supported)
* [x]  [ros2/rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/pull/513) (CFT supported)
* [x]  [ros2/rmw_cyclonedds](https://github.com/ros2/rmw_cyclonedds/pull/289) (not supported)
* [x]  [ros2/rcl](https://github.com/ros2/rcl/pull/894)
* [x]  [ros2/rclcpp](https://github.com/ros2/rclcpp/pull/1561)

- follow-ups for Humble Release

* [x] [Tutorial](https://github.com/ros2/ros2_documentation/pull/2441)
* [x] [Update Humble Release Note](https://github.com/ros2/ros2_documentation/pull/2396)
* [x] [demo_nodes_cpp](https://github.com/ros2/demos/pull/557)
* [x] [examples](https://github.com/ros2/examples/pull/341)

- Fallback Filtering after Humble Release

* [ ]  [Fallback filtering subscription](https://github.com/ros2/rcl/issues/982)

- `/parameter_events` & action `feedback` topics.

* [ ]  [ros2/rcl](https://github.com/ros2/rcl/pull/901)
* [ ]  [ros2/rclcpp](https://github.com/ros2/rclcpp/pull/1592)

- rclpy support

* [ ]  [ros2/rclpy](https://github.com/ros2/rclpy)

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>


fujitatomoya:
@wjwwood @ivanpauno 
CC: @eboasson

would you give us some feedback what/how you feel about this feature extension?
pointing you, because i usually have communication with you these layers, such as rmw/rcl/rclxxx.

also anyone, any idea and comments will do good.

ros-discourse:

https://discourse.ros.org/t/content-filtering-topic/13889/1

ros-discourse:

https://discourse.ros.org/t/ros-2-real-time-working-group-online-meeting-16-apr-29-2020-meeting-minutes/13999/4

fujitatomoya:
@wjwwood @dirk-thomas 

friendly ping, but not urgent.

dirk-thomas:
Someone will take a look at this but likely only after the Foxy release is out.

articles/content_filtering.md
```diff
@@ -0,0 +1,222 @@
...
+<img src="../img/content_filter/parameter_event_sequence_with_filter.png">
+
+### Interfaces
+
+#### rcl
+
+- add API `is_cft_supported` to allow frontend(rclcpp/rclpy) can know CFT is supported or not.
+  e.g) `rcl_subscription_is_cft_supported` to access `is_cft_supported`.
+- add `rcl_create_cft_subscription` to call `rmw_create_cft_subscription`.
+- add `rcl_cft_set/get_expression_parameters` to call `rmw_cft_set/get_expression_parameters`.
```
ivanpauno: Having a `rcl_is_cft_supported` function sounds like a good idea, but I don't get why we need all the others.

A cft expression in the subscription options sounds like a better idea than a special function to create a cft_subscription.
The `cft` expression can be a custom object, and we can add a proper API to configure it.

articles/content_filtering.md
```diff
@@ -0,0 +1,222 @@
...
+| rmw_fastrtps | false |
+| rmw_cyclonedds | false |
+
+- new `DDSContentFilteredTopic` object needs to be managed in implementation.
+- parent topic information should be cached so that we can create CFT based on that parent topic.
+- `rmw_create_cft_subscription` interface.
+  - input: rmw_subscription_t, rmw_node_t, parent topic information.
+  - output: rmw_subscription_t (with new data reader)
+- `rmw_cft_set/get_expression_parameters` interface.
+  - input: `DDSContentFilteredTopic`, (string sequence) expression parameters.
```
ivanpauno: types in `rmw` shouldn't be a DDS type.

articles/content_filtering.md
```diff
@@ -0,0 +1,222 @@
...
+| rmw_connext | true |
+| rmw_fastrtps | false |
+| rmw_cyclonedds | false |
+
+- new `DDSContentFilteredTopic` object needs to be managed in implementation.
+- parent topic information should be cached so that we can create CFT based on that parent topic.
+- `rmw_create_cft_subscription` interface.
+  - input: rmw_subscription_t, rmw_node_t, parent topic information.
+  - output: rmw_subscription_t (with new data reader)
+- `rmw_cft_set/get_expression_parameters` interface.
```
ivanpauno: similar comments to `rcl` API

ivanpauno (commented):
I'm not going to say that content filters aren't cool and have a lot of good use cases, but if we add support to them non-DDS based rmw implementations will become really difficult.

Adding an optional feature sounds fine to me, but you're proposing making use of cft in some built-in use cases (parameter events and actions).

How will parameter events work in the case the rmw implementation doesn't support content filtering?
i.e. How will filtering happen in that case?

If we add this, I'm not sure if we should support the content filtering expressions supported in the DDS standard, or a custom designed one for ROS2 (with a way to convert between the two).

P.S.: parameter events and multi-client actions are a great example of where using cft is a performance win.

ivanpauno:
@fujitatomoya I see you have been working on draft PRs, but bare in mind that this has high chances of being rejected (specially, because how DDS specific is this feature).
I would really wait for the design approval before having a full implementation.

Sorry for the delay in reviewing.

fujitatomoya:
@ivanpauno 

really appreciate for checking on this and advice that is helpful.

> I would really wait for the design approval before having a full implementation.

i was thinking that too, i would discuss more how we could make approach on this before implementation.

fujitatomoya:
> Adding an optional feature sounds fine to me, but you're proposing making use of cft in some built-in use cases (parameter events and actions).

It does not have to be only for built-in use cases, but user classes also. that is what we want to discuss.

> P.S.: parameter events and multi-client actions are a great example of where using cft is a performance win.

I believe so. Something we'd like to know from the others is `is there any approach that we can solve the problems defined in this draft?`. I mean, if there is and it is better, we will do that to support the requirement instead of cft.

> How will parameter events work in the case the rmw implementation doesn't support content filtering?
> i.e. How will filtering happen in that case?

i guess that we have options,

1. filtering will be done in subscription side. (not rmw implementation, but rclcpp/rclpy instead)
2. return NOT_SUPPORT if rmw implementation does not support cft.

thinking about good user experience for `it just works out-of-the-box`, option-1 can be an actual choice. we can support this as well.

> If we add this, I'm not sure if we should support the content filtering expressions supported in the DDS standard, or a custom designed one for ROS2 (with a way to convert between the two).

as far as i see, DDS standard expressions are really flexible so it would be nice to keep the standard.

articles/content_filtering.md
fujitatomoya: user application is likely to change filtering expression and parameter expressions at runtime, so i think that we need to provide methods to do that dynamically.

articles/content_filtering.md
fujitatomoya: correct, that should be `rmw_subscriotion`, that will be fixed.

gbiggs:
> return NOT_SUPPORT if rmw implementation does not support cft.

This is tempting, but it would have the side-effect of preventing a non-DDS-based RMW being used for a node that decides it wants content filtering. I feel that whether or not to content-filter a topic should either be decidable by the system integrator outside of the node's implementation, or by the node implementer and done in the node's implementation (not in the RMW layer) if we want to remain completely portable.

I think that if we want content filtering, we would have to provide a generic interface for doing it that can internally switch between using RMW-provided content filtering when available or doing it at a higher layer when not available in the RMW layer. Without that generic support also being provided I think adding access to it in the RMW layer is risky.

fujitatomoya:
@gbiggs 

thanks for sharing you thoughts! i do agree on that.

articles/content_filtering.md
ivanpauno: > user application is likely to change filtering expression and parameter expressions at runtime, so i think that we need to provide methods to do that dynamically.

Do you mean changing the expression in an already created subscription?
Or dynamically creating an expression?

In the first case, having a special object for the `cft` subscription makes sense.
In the second, only having a object for the expression makes sense.

ivanpauno:
> 1. filtering will be done in subscription side. (not rmw implementation, but rclcpp/rclpy instead)
> 2. return NOT_SUPPORT if rmw implementation does not support cft.

Yeah, in the case content filtered topics are added, I think that having a default implementation of subscription side filtering for when the `rmw` implementation don't support the feature is a good idea.

We will need full support of dds cft expressions somewhere.

> but rclcpp/rclpy instead

It could be implemented in `rcl` instead.

fujitatomoya:
> It could be implemented in rcl instead.

i was thinking about that, it will be better.

articles/content_filtering.md
fujitatomoya: I am not sure if i follow you precisely,

> Do you mean changing the expression in an already created subscription?

I say yes, this is needed especially for action topics. filtering expression and expression parameters for goal id need to be adjusted at runtime.

ivanpauno: Oh yeah, we need to be able to modify the filtering expression of an already created cft subscription (or at least, the expression parameters).

ros-discourse:

https://discourse.ros.org/t/rmw-proposal-content-filtered-topic-suport/16113/1

ivanpauno:
@wjwwood friendly ping

fujitatomoya:
we are doing some test about efficiency of ContentFitleredTopic based on RTI Connext with scenarios.

- Confirm filtering happens either on writer or reader side. (Network transmission if that happens or not.)

note,

- DDS specification does not specify ContentFilteredTopic implementation, so it is up to implementation.
- To support different DDS implementation, reader must be responsible to filter the message at the end.
- filtering depends on how many actual messages need to be delivered to reader. that said, if # of receiver is more than 80%, it would be better to do the filtering process in the reader side. and vice-versa.

articles/content_filtering.md
```diff
@@ -0,0 +1,282 @@
...
+
+<img src="../img/content_filter/parameter_event_sequence_with_filter.png">
+
+### Interfaces
+
+#### rcl
+
+- add API `is_cft_supported` to allow frontend(rclcpp/rclpy) can know CFT is supported or not.
+  e.g) `rcl_subscription_is_cft_supported` to access `is_cft_supported`.
+- add `rcl_create_cft_subscription` to call `rmw_create_cft_subscription`.
```
iuhilnehc-ynos: @fujitatomoya 

Based on your design, I have a question about the function `rmw_create_cft_subscription`.
If using this function, a subscriber can dynamically use `rmw_create_cft_subscription` to set the [topic_reader_](https://github.com/ros2/rmw_connext/blob/25721877d133a6ea09c1f3a864ed892d288fa918/rmw_connext_cpp/src/rmw_subscription.cpp#L224-L225) inside it at any time, which means we need to add an extra lock to protect the `subscriber_info[topic_reader_, read_condition_]` in `rmw_create_cft_subscription` and some other locations like [rmw_connext-wait](https://github.com/ros2/rmw_connext/blob/25721877d133a6ea09c1f3a864ed892d288fa918/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/wait.hpp#L224-L230).

I was thinking that using `rmw_create_subscription` can create normal topic or content filtered topic based on rmw_subscription_options_t that adding `const char * filter_expression` and `const rcutils_string_array_t * expression_parameters` if CFT is supported. Is it necessary to separate this operation into two actions?

fujitatomoya: I think that is fair and reasonable just to extend the interface to support cft subscription. I will update the design as you suggested.

ros-discourse:

https://discourse.ros.org/t/cyclonedds-roadmap-multi-network-async-content-filtering-ql1-et-al/17561/5

iuhilnehc-ynos:
To add content-filtered-topic interfaces and an empty stub implementation for rmw_connext,rmw_fastrtps,rmw_cyclonedds

* [ ]  [ros2/rmw](https://github.com/ros2/rmw/pull/302)
* [ ]  [ros2/rmw_implementation](https://github.com/ros2/rmw_implementation/pull/181)
* [ ]  [ros2/rmw_connext](https://github.com/ros2/rmw_connext/pull/490)
* [ ]  [ros2/rmw_fastrtps](https://github.com/ros2/rmw_fastrtps/pull/513)
* [ ]  [ros2/rmw_cyclonedds](https://github.com/ros2/rmw_cyclonedds/pull/289)
* [ ]  [ros2/rcl](https://github.com/ros2/rcl/pull/894)
* [ ]  [ros2/rclcpp](https://github.com/ros2/rclcpp/pull/1561)

fujitatomoya:
@wjwwood @ivanpauno @clalancette 
CC: @iuhilnehc-ynos 

friendly ping, ContentFiteredTopic 1st step is out here: https://github.com/ros2/design/pull/282#issuecomment-785522547

i know you are busy and there are other stuff to review in the queue. but we really appreciate if you have some time to take a look and give us feedback 😄 thanks in advance.

fujitatomoya:
Note for myself,
- enabling ContentFilteredTopic, we need https://github.com/rticommunity/rmw_connextdds to be merged before this.
- ideally, we need to follow-up with examples for https://github.com/ros2/demos/tree/master/demo_nodes_cpp.

ros-discourse:

https://discourse.ros.org/t/rmw-proposal-content-filtered-topic-suport/16113/7

articles/content_filtering.md
```diff
@@ -0,0 +1,315 @@
...
+
+#### DDS 
+
+the following create/delete API's are defined,
+
+- create_contentfilteredtopic()
+- delete_contentfilteredtopic()
+
+**According to the specification, filter_expression can be only initialized at constructor and it is read_only.
+Probably it is not possible to change the filter_expression at runtime based on the specification but [RTI v5.3.0](https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/api/connext_dds/api_cpp2/classdds_1_1topic_1_1ContentFilteredTopic.html#a76310bf0b7123dd89afbacf43dbabf4a) can support this requirement.
```
mitza-oci: filter_expression can be changed at runtime
If the spec is unclear you could consider filing an issue https://issues.omg.org/issues/create-new-issue?specification=DDS%231.4

articles/content_filtering.md
```diff
@@ -0,0 +1,315 @@
...
+  Basically it is likely that ContentFilteredTopicDataWriter does filtering on writer side, so that it can provide the improvement and efficiency.
+  But it is dependent on imnplementation if always ContentFilteredTopicDataWriter does filtering.
+  This is because if there is a lot of ContentFilteredTopicDataReader with complicated filtering expression and parameter, it would be probably more efficient just to send the messages to all ContentFilteredTopicDataReader.
+  And ContentFilteredTopicDataReader can filter out the message if not necessary based on filtering expression and parameter.
+  This is also said that ContentFilteredTopicDataReader does need to do filtering always just in case ContentFilteredTopicDataWriter ends up sending w/o filitering.
+  <img src="../img/content_filter/contents_filter_overview.png" width="600">
+- The filtering is done by means of evaluating a logical expression that involves the values some of the data-fields in the sample.
+  The logical expression is derived from the filter_expression and expression_parameters arguments.
+- Multiple filter_expression and expression_parameters can be supported on single ContentFilteredTopic.
```
mitza-oci: Each ContentFilteredTopic has a single filter_expression and a sequence of expression_parameters

articles/content_filtering.md
```diff
@@ -0,0 +1,315 @@
...
+- ContentFilteredTopicDataWriter does filtering for each ContentFilteredTopicDataReader endpoint.
+  Basically it is likely that ContentFilteredTopicDataWriter does filtering on writer side, so that it can provide the improvement and efficiency.
+  But it is dependent on imnplementation if always ContentFilteredTopicDataWriter does filtering.
+  This is because if there is a lot of ContentFilteredTopicDataReader with complicated filtering expression and parameter, it would be probably more efficient just to send the messages to all ContentFilteredTopicDataReader.
+  And ContentFilteredTopicDataReader can filter out the message if not necessary based on filtering expression and parameter.
+  This is also said that ContentFilteredTopicDataReader does need to do filtering always just in case ContentFilteredTopicDataWriter ends up sending w/o filitering.
+  <img src="../img/content_filter/contents_filter_overview.png" width="600">
+- The filtering is done by means of evaluating a logical expression that involves the values some of the data-fields in the sample.
+  The logical expression is derived from the filter_expression and expression_parameters arguments.
```
mitza-oci, line 83: When using parameters with multiple DDS implementations note that some differences have been encountered.  See https://issues.omg.org/issues/spec/DDS/1.4#issue-47484

articles/content_filtering.md
fujitatomoya: @mitza-oci thanks for the info. this was kinda confirmation since i was not clear on this.

articles/content_filtering.md
fujitatomoya: thanks, i borrowed your sentence here.

articles/content_filtering.md
fujitatomoya: appreciate for the heads-up. i will keep that in the note section.

asorbini:
I'd like to suggest renaming the two new RMW functions used to access and set the content-filter expression and parameters to be slightly more appropriate.

Quoting myself from a [comment](https://github.com/ros2/rmw_connextdds/pull/68#discussion_r735995595) in ros2/rmw_connextdds#68:

>  I feel like `[get|set]_cft_expression_parameters()` is a bit of a misnomer, since "parameters" are kind of an "advanced use case" and in many cases users might be interested in getting and setting only the expression.
>
>I think something like `[get|set]_content_filter()` (without "expression" or "parameter") would be more appropriate. So, we could rename the functions to:
>- `rmw_subscription_get_content_filter`
>- `rmw_subscription_set_content_filter`

Does that sound reasonable?

fujitatomoya:
@asorbini it does for me.

fujitatomoya:
@iuhilnehc-ynos do you happen to have a repo file for CFT development?

fujitatomoya:
> @iuhilnehc-ynos do you happen to have a repo file for CFT development?

@iuhilnehc-ynos friendly ping!

iuhilnehc-ynos:
@fujitatomoya 

Sorry, I missed the https://github.com/ros2/design/pull/282#issuecomment-989205246.
I need to re-check it. Some repositories might be rebased.
(After that, I'll provide it in gist.github.com)

iuhilnehc-ynos:
@fujitatomoya 

Please refer to https://raw.githubusercontent.com/iuhilnehc-ynos/ros2/topic-debug-connextdds-cft/ros2.repos
And I think we need to update the https://github.com/ros2/design/pull/282#issue-610568380

Remove <del>ros2/rmw_connext</del>
Update `rticommunity/rmw_connextdds` from https://github.com/rticommunity/rmw_connextdds/pull/12 to https://github.com/ros2/rmw_connextdds/pull/68

fujitatomoya:
@iuhilnehc-ynos thanks 👍

fujitatomoya:
@iuhilnehc-ynos 

which one is official fix for CFT? https://github.com/ros2/rmw_connextdds/tree/asorbini/cft or https://github.com/ros2/rmw_connextdds/pull/12? could you confirm and check?

iuhilnehc-ynos:
@fujitatomoya 

Please use https://github.com/ros2/rmw_connextdds/tree/asorbini/cft created in the https://github.com/ros2/rmw_connextdds/pull/68 by @asorbini , I'll close https://github.com/ros2/rmw_connextdds/pull/12.

fujitatomoya:
@ros2/team could you have some time to review this and shar comments?

ros-discourse:

https://discourse.ros.org/t/add-heartbeat-message-type/24162/15

articles/content_filtering.md
```diff
@@ -0,0 +1,308 @@
...
+`rclcpp::ParameterEventsFilter` is used to filter the parameter with specific name "use_sim_time".
+As user interface, `AsyncParametersClient::on_parameter_event` is provided to handle parameter events.
+
+<img src="../img/content_filter/parameter_event_overview.png" width="500">
+
+The problem can be broken into the following two parts,
+
+- Network resource consumption on `/parameter_events` topic
+  Since publisher does not know who needs to receive which messages, all of the parameter activity will be published and subscribed.
+  That said everyone publishes and subscribes all of the events within this giant topic `/parameter_events`.
```
ivanpauno:
```suggestion
  That said everyone publishes and subscribes all of the events in this high traffic `/parameter_events` topic.
```

articles/content_filtering.md
```diff
@@ -0,0 +1,308 @@
...
+  In expectation, it is likely that parameter event callback is interested on specific parameters only, so that there will be a lot of unnecessary message receives and filtering process needs to be done.
+
+As described above, we could imagine that if the number of node are 10, 20 and 100 and so are the parameters for each nodes.
+This will be a huge burden and pain for entire system, passing unnecessary messages via limited bandwidth network with edge devices and filtering messages to drop is not something user application wants to do.
+
+Besides, this problem can be applied to user defined topics.
+
+### Action Topics
+
+Each action server will provide two topics named `feedback` and `status` as followings,
```
ivanpauno:
```suggestion
Each action server provides two topics named `feedback` and `status` as followings,

```

articles/content_filtering.md
```diff
@@ -0,0 +1,308 @@
...
+
+<img src="../img/content_filter/action_architecture_overview.png" width="600">
+
+`feedback` and `status` are topics mapped with action name and published by action server, and action clients subscribe those topics to get feedback and status.
```
ivanpauno:
```suggestion
`feedback` and `status` are topics mapped with action name and published by action server, and action clients subscribe to those topics to get feedback and status.

```

articles/content_filtering.md
```diff
@@ -0,0 +1,308 @@
...
+- Each ContentFilteredTopic has a single filter_expression and a sequence of expression_parameters.
+  ```
+  [parameter use cases]
+  "node = %0 OR node = %1" spinal_node attention_node : subscribe all parameter events from spinal and attention nodes
+  "node = %0 OR name = %1" audio_node attention_level : subscribe all parameter events from audio_node and parameter named attention_level of all nodes.
+  "(node = %0 AND name = %1) OR (node = %2 AND name = %3)" eye_node brightness ear_node loudness : subscribe brightness parameter of eye_node and loudness parameter of ear_node only.
+  [action use case]
+  "uuid = %0 OR uuid = %1" DEADBEEF CAFEFEED : subscribe only interest goal ids which client handlers possess.
+  ```
+- A single filter_expression and a sequence of expression_parameters can be modified dynamically at runtime.
```
ivanpauno, line 93: I think it would be good to have a full specification of what's a valid filter_expression, even if it's a link to the specification used by DDS.

ivanpauno (commented):
The rationale looks good!
I didn't look a lot into the design bits, I have to take another look to that part.

SteveMacenski:
My primary issue with this is that none of the free Tier-1 supported RMWs have implementations of it. I agree this is not a niche feature and would make certain applications, especially related to multiple robots and saving bandwidth. However, given this is sufficiently "big", I would want this supported by at **minimum** the default, but with the expectation that shortly after it would be merged that all would have support for it.

With respect to the discussion on DDS-specificity, I think that's a valid concern, but not something that would be a blocker for me. If we had any Tier-1 supported RMW that was non-DDS or any intention in the short to medium term to add one, that would be a little different. Looking at the current list, there is no Tier-Any non-DDS implementation. 

While I haven't looked how specific DDS vendors implement this idea, is it unreasonable to assume that we could implement something analogous down the line should we want to address a non-DDS solution? Would it not be possible to rip out the concepts of the DDS vendors' implementation and have an optional version within an interface? I think it would make sense to support that interface, but use DDS' provided version as a default. 

In general, I support using DDS-specific versions of things if it (1) provides capabilities generally provided by the DDS standard or implementations we care about (2) would tangibly increase performance compared to the implementation within the ROS 2 stack and (3) it is possible to create a non-DDS specific way of doing it down the line should it be required. 

I know that would add some technical debt should we choose to move to another non-DDS solution down the line, but I suspect in the distant future, we'll already be having a heck of a time trying to move over to something with totally different concepts, I don't see adding some dependency in exchange for substantively improved performance or capabilities in the foreseeable future.

wjwwood:
> My primary issue with this is that none of the free Tier-1 supported RMWs have implementations of it. I agree this is not a niche feature and would make certain applications, especially related to multiple robots and saving bandwidth. However, given this is sufficiently "big", I would want this supported by at minimum the default, but with the expectation that shortly after it would be merged that all would have support for it.

This feature "works" even without support from the rmw implementation. It just happens on the receiving end, and so it is no more efficient than if the user did it themselves. However, it does remove a lot of boilerplate for the user if they want to filter a topic based on the content. So, even if none of the rmw implementations supported it, I think the proposal has value.

I would also like to see "publisher side filtering" in the default and other tier 1 implementations, but personally I don't feel that is a blocker, since it can be added in the future and shouldn't require changes to the user's code or the libraries above `rmw`.

If you still have concerns about this, we can discuss it in the MWWG, where we've discussed it in the past.

> If we had any Tier-1 supported RMW that was non-DDS or any intention in the short to medium term to add one, that would be a little different. Looking at the current list, there is no Tier-Any non-DDS implementation.

I think this is kind of a dangerous rationale in general, because I think it leads us to designing ourselves into a box, and it has been an uphill battle for years to avoid feature creep on the back this line of thinking.

Also, there are non-tiered implementations already like the eCal rmw implementation, and there will very likely be a Zenoh based implementation in the not so distant future. So, I think the chances of a non-DDS rmw implementation getting raised to tier-something is only a matter of time, but of course for now it is hypothetical.

But in the case of this feature, I think that it is something that can be reasonably implemented regardless of the middleware technology, because it can always be done in `rcl`, on the receiving end if necessary.

> While I haven't looked how specific DDS vendors implement this idea, is it unreasonable to assume that we could implement something analogous down the line should we want to address a non-DDS solution?

Yes, I think it's very possible to do without middleware support (this is already the case if the DDS implementation being used doesn't support it, unless I'm very mistaken).

> Would it not be possible to rip out the concepts of the DDS vendors' implementation and have an optional version within an interface?

I don't follow this, can you be more specific? We're not exposing the exact DDS interface in any part of the ROS 2 API in this proposal, and we're using the DDS CFT API under the hood, when it is available.

> In general, I support using DDS-specific versions of things if it (1) provides capabilities generally provided by the DDS standard or implementations we care about (2) would tangibly increase performance compared to the implementation within the ROS 2 stack and (3) it is possible to create a non-DDS specific way of doing it down the line should it be required.

I think all of those points are true for this particular proposal, @fujitatomoya do you agree?

fujitatomoya:
@SteveMacenski 

> My primary issue with this is that none of the free Tier-1 supported RMWs have implementations of it.

sorry, i should have mentioned this.

- `rmw_fastrtps` will be supported. currently it does not support, but see https://github.com/ros2/rmw/pull/302#issuecomment-1055561373. we have been working on that to support CFT via eProsima Fast-DDS.(we actually need Fast-DDS CFT as 1st step.)
- `rmw_connextdds` will be supported. (https://github.com/ros2/rmw_connextdds/pull/68)

> I would want this supported by at minimum the default

totally agree. that is what we are gonna do!

> If we had any Tier-1 supported RMW that was non-DDS or any intention in the short to medium term to add one, that would be a little different.

there has been discussion, https://github.com/ros2/rmw/pull/302#issuecomment-1055628593.
we can do this as follow-up, so that even if RMW implementation does not support this feature, we can provide this feature as ROS 2 (probably rcl) w/o any performance improvement.

fujitatomoya:
@SteveMacenski @wjwwood 


> This feature "works" even without support from the rmw implementation. It just happens on the receiving end, and so it is no more efficient than if the user did it themselves. However, it does remove a lot of boilerplate for the user if they want to filter a topic based on the content. So, even if none of the rmw implementations supported it, I think the proposal has value.

we have been thinking about that, and been waiting for the feedback. currently, we expect the rmw implementation supports the CFT.
but as discussed [here](https://github.com/ros2/rmw/pull/302#issuecomment-1055628593), we can do that in `rcl`. (expecting no interface change.)

> I would also like to see "publisher side filtering" in the default and other tier 1 implementations

this is exactly what we want to have for our use cases.
so that we have been working with eProsima to enable `DDS Writer Side Filtering` now. (Reader side is available, so that we can support ROS 2 CFT, and later we can enable Writer-side w/o interface change.)

> But in the case of this feature, I think that it is something that can be reasonably implemented regardless of the middleware technology, because it can always be done in rcl, on the receiving end if necessary.

precisely.

fujitatomoya:
originally we have a plan and history,

- Humble ROS 2 CFT comes with `rmw_fastrtps`(reader-filter) and `rmw_connextdds`. (these have been verified already.) this means 2/3 Tier-1 implementation can support RMW CFT interfaces.
- During Humble development, we can enable `rmw_fastrtps`(writer-filter) once it is ready. expecting no interface changes, but performance improvement is gonna be huge.
- (more than a year ago...this design is 2 years old...) we have talked with ADLINK that they are positive to support CFT in cyclonedds atm. if that happens, all tier-1 implementation can support ROS 2 CFT. but we are not sure about cyclonedds since that was more than a year ago.
- even if cyclonedds does not support CFT, this happens with other interfaces such as `LoanedMessages`.

so question is,

- Do we really need to implement `rcl` CFT? of course, we can do this in `rcl`, but no performance improvement, more maintenance cost.
- How about implementing `rcl` CFT when non-DDS tier-1 implementation is supported by ROS 2? it makes more sense?

we are totally open to this discussion, please share your thoughts.

SteveMacenski:
> How about implementing rcl CFT when non-DDS tier-1 implementation is supported by ROS 2? it makes more sense?

That would be my vote. And hopefully we do that when there are stakeholders that care enough about a non-DDS solution to put in some of their own time to helping at that point too. I'd make sure though that all of the interfaces and piping is done ahead of time so that someone could do it conceptually.

wjwwood:
I think we either need the CFT implemented in rcl as a fallback, or we need to require it in the rmw interface, meaning things like cyclone would need to implement it in the rmw layer on the receiving side, so basically what rcl would do.

Having it as a feature that may or may not be available from the perspective of rclcpp users is not good imo. That means that this feature can never really be relied on and used in core technologies like (just examples not necessarily real ones) tf or actions or something like that.

fujitatomoya:
> meaning things like cyclone would need to implement it in the rmw layer on the receiving side, so basically what rcl would do.

yeah, right.

> Having it as a feature that may or may not be available from the perspective of rclcpp users is not good imo.

okay, that also makes sense. i will put it down in the design and we will start the development.
since this does not block these PRs in interface aspect, we welcome more comments.

fujitatomoya:
> i will put it down in the design and we will start the development.

we will start the estimation and feasibility as 1st step since this needs to be implemented in C, it will be some work and complicated.

articles/content_filtering.md
fujitatomoya: that is actually good idea.

fujitatomoya:
### Proposals / Discussion

- `RCL` filtering is hard requirement for `RMW` ContentFilteredTopic?
- `RCL` filtering will be only `Reader-Filtering`. (no performance improvement.)

so far, we came up with 3 possible proposals.

1. integrate everything into `rcl` only. but this does not seems to be reasonable since it cannot access the field for `rclcpp`(typesupport_cpp). And, `sqlite3 like` parser and tokenizer needs to be implemented with C, this will be huge code maintenance. Any suggested approach? (expecting no external API changes.)
2. implement in client code such as `rclcpp` and `rclpy`. type support problem can be solved, but parsing and tokenizer will be maintained in each implementation. (expecting no external API changes.)
3. implement concrete functions into `rmw_dds_common` or other c++ package, and export C interface to `rcl`. So that typesupport_cpp problem can be resolved. And to address SQL parser and tokenizer, we can depend on PEGTL. (external API changes may be required.)

image

fujitatomoya:
Just posting the current status from today's MW WG.

> RCL filtering is hard requirement for RMW ContentFilteredTopic?

basically yes, but from MW WG discussion today, this needs to be re-discussed in OpenRobotics.
@wjwwood @ivanpauno please let us know that if we can have a go for RMW CFT w/o fallback function in RCL at this moment.

some notes,

- we do not really think that is feasible to implement whole fallback function in `rcl` in next few weeks.
- but we are good to follow up fallback function after RMW CFT Humble Release.
- we also will continue the feasibility that fallback function can be applied to Humble after release.
- (FYI) history, https://github.com/ros2/design/pull/282#issuecomment-1055938077

if we do release RMW CFT for Humble ,

- we should/will provide good example and document to show which implementation supports and how to manage in application code.
- continue to consider possible fallback implementation as described above. (https://github.com/ros2/design/pull/282#issuecomment-1062110693)

let me know if anything is missing here. thanks.

CC: @asorbini @MiguelCompany @alsora @gbiggs

fujitatomoya:
CI (CFT baseline interfaces):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16313/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10933/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16683/)

repo is https://raw.githubusercontent.com/iuhilnehc-ynos/ros2/topic-debug-connextdds-cft/ros2.repos. (**this does not include system enhancement with `/parameter_events` & action `feedback` topics with CFT.)

fujitatomoya:
@iuhilnehc-ynos could you check CI failure? probably we need to rebase for https://ci.ros2.org/job/ci_linux/16313/console?

iuhilnehc-ynos:
@fujitatomoya 

Can you run CI with build parameter ( '--packages-up-to rmw test_rmw_implementation rcl rclcpp' ) and test parameter (--packages-select rmw test_rmw_implementation rcl rclcpp)?

There are some other rmw (rmw_client_request_publisher_get_actual_qos, rmw_event_set_callback, etc) interfaces updated, and I can't rebase [`rmw_connextdds`](https://github.com/ros2/rmw_connextdds/pull/68).

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16315/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10935/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16685/)

fujitatomoya:
@iuhilnehc-ynos just started.

fujitatomoya:
@iuhilnehc-ynos CI is unstable, could you check the result? 

- https://ci.ros2.org/job/ci_linux/16315/
- https://ci.ros2.org/job/ci_linux-aarch64/10935/

iuhilnehc-ynos:
@fujitatomoya 

https://ci.ros2.org/job/ci_linux-aarch64/10935/ are not related https://github.com/ros2/rclcpp/pull/1561, the warning for uncrustify and cpplint had been fixed in the master. (I'll rebase the branch in the end and re-run CI in the future.)

For other unstable cases in https://ci.ros2.org/job/ci_linux/16315/ that happened in `rmw_connextdds`, I think it's better to ask @asorbini for help to check them. Because these tests have been passed in `rmw_fastrtps_cpp`.

ivanpauno:
> integrate everything into rcl only. but this does not seems to be reasonable since it cannot access the field for rclcpp(typesupport_cpp)

What message field do you need access to?
It's not clear to me why that's needed (I still need to take a deeper look to the fastrtps PR)

> And, sqlite3 like parser and tokenizer needs to be implemented with C, this will be huge code maintenance. Any suggested approach? (expecting no external API changes.)

It would be fine IMO to implement it using CPP and provide a C interface.
I don't know what other thinks though.

fujitatomoya:
> What message field do you need access to?
> It's not clear to me why that's needed (I still need to take a deeper look to the fastrtps PR)

based on the filtering expression and parameters, we need to check the content of each message, and then we can do filtering if this particular message should be dropped or not. these message types are user defined messages.
(off topic, Fast-DDS current Reader-Side CFT is much better performance than `rcl` filtering since it does not need to de-serialize the message before filtering. This is achieved by `TypeObject`, and i think RTI Connext DDS also has this advantage with `TypeCode`.)

> It would be fine IMO to implement it using CPP and provide a C interface.

👍 

> Compile option to opt-in CFT in client would be the idea

off the top of my head, how about using environment variable such as `ROS_DISABLE_LOANED_MESSAGES`?
by default, CFT is disabled atm, it will return `NOT SUPPORT`, so that user cannot create subscription with CFT option nor set/get filtering expression and parameters.
user can explicitly define environmental variable to enable CFT?
of course, documentation / tutorial will be provided, how to enable / which implementation supports.

wjwwood:
> based on the filtering expression and parameters, we need to check the content of each message, and then we can do filtering if this particular message should be dropped or not. these message types are user defined messages.
(off topic, Fast-DDS current Reader-Side CFT is much better performance than rcl filtering since it does not need to de-serialize the message before filtering. This is achieved by TypeObject, and i think RTI Connext DDS also has this advantage with TypeCode.)

We might be able to use `rosidl_typesupport_introspection_c{pp}` for this. We will have to have a filter implementation for each rosidl type we support, right now being C and CPP. We could use just the C typesupport for the purposes of evaluating the filter, and then deliver the messages that pass in CPP... but that would require extra conversions and stuff.

The alternative (which isn't necessarily mutually exclusive) is to implement the fallback filters in each client library. This would be the most maintenance, but with shared code perhaps the least work. It will be the most performant as well because it would not necessarily rely on runtime type introspection (e.g. rosidl_typesupport_introspection* or DDS's TypeCode API) and could even be done at compile time in C++, in some cases.

Maybe we should have a dedicated meeting for this?

wjwwood:
> off the top of my head, how about using environment variable such as ROS_DISABLE_LOANED_MESSAGES?
by default, CFT is disabled atm, it will return NOT SUPPORT, so that user cannot create subscription with CFT option nor set/get filtering expression and parameters.
user can explicitly define environmental variable to enable CFT?
of course, documentation / tutorial will be provided, how to enable / which implementation supports.

The idea behind this is actually more like a compiler definition, so if you wanted to use this feature you'd have to add a compiler definition to your package to enable the feature for rclcpp. That was people wondering why they need to do this are prompted to read about it, and so that pull requests using the feature are more readily identified so we don't have it slip in prematurely to other core packages. But that was just an idea someone in the ROS 2 meeting proposed. We can discuss it further.

fujitatomoya:
> Maybe we should have a dedicated meeting for this?

will set up the meeting.

fujitatomoya:
breaking-out discussion for CFT feature.

date: 03/17/2022 09:00 - 10:00 PDT

@ivanpauno @wjwwood @MiguelCompany @asorbini if i am mistaken, please let me know!


### RMW CFT Interface for Humble

- ***conclusion, we will take [current interfaces](https://github.com/ros2/rmw/pull/302).***
- discussion starts from https://github.com/ros2/rmw/pull/302#discussion_r823791348
- separating normal subscription from CFT subscription would be easier for maintenance.
- having CFT as Reader QoS would be fine, creating another entity is complicated.
- (not a justification but) besides, probably no time to adjust interfaces to create CFT subscription interfaces from now.
- any performance degression for discovery or normal subscription with CFT enhancement with empty filtering and expression? No, it would be no impact, just calling empty interfaces would be all.

### Compiler Option in rclcpp

- this is notification for user application, that `you are using CFT`.
- compiler definition to enable CFT in `rclcpp`, which is always ON for core libraries built.
- user must add definition to build the application with CFT feature. (document also needs to be provided.)
- having template or definition in header files to adjust the interfaces w/ or w/o CFT.
- details need to be discussed.

### Fallback function for Humble
  
- ***Sony to consider best proposal, and discuss in MW WG meeting or breakout meeting.***
- we seem to have 2 options where to implement, either client `rclcpp` or `rcl`. (https://github.com/ros2/design/pull/282#issuecomment-1068559583, https://github.com/ros2/design/pull/282#issuecomment-1062110693 option 2 and 3 since 1 is not really feasible.)
- internal interfaces can be changed, but not allowed to change data structure and type for user interfaces for ABI/API contract.
- if ABI/API changes are required to support fallback implementation, these changes must be ready for Humble. (Mon. April 4, 2022 - API Freeze)
- `rclcpp` efficiency w/o runtime introspection type support, but maintenance for each client library. (for example, when we support `rclpy`.) this would not be a right place to integrate the feature as design aspect, but we already have an exception such as `intra-process communication`.
- `rcl` maintenance cost effective, support all client libraries. runtime introspection type support c/cpp required. performance could be worse, especially for taking serialized message with filter. (de-serialize the message for the filter and user application will receive the serialized message and de-serialize again...)
- there could be difference about filtering expression and parameter between rmw implementation and fallback filter? we need to have clear strict specification in ROS 2 to keep the consistency about filtering expression and parameters between rmw CFT and fallback filtering.

#### Future Discussion / Note

- with DDS CFT reader side, the message will not go in HistoryCache. filtering process happens before HistoryCache w/o de-serializing the message.
- in case of fallback function, the message will be in HistoryCache w/o filtering if implementation itself does not support CFT.
- this behavior is really different between fallback and DDS CFT.
- to keep the consistency about this behavior, we could have a new interface provides user filter callback to determine the messages should be in HistoryCache?
- this would be really complicated to rmw implementation, because DDS does not provide the interfaces to register the function for filtering.
- even with fallback filtering, providing the same behavior not to wake up executer when message is filtered, this will be good performance for user application. (image that we need to take one message out of 1000 messages.)

ivanpauno:
> internal interfaces can be changed, but not allowed to change data structure and type for user interfaces for ABI/API contract.

In rcl we're using the pimpl pattern almost everywhere, so if we're going to implement the fallback in rcl and the API is in place before the freeze it should be possible to backport the feature.

MiguelCompany:
> * performance could be worse, especially for taking serialized message with filter. (de-serialize the message for the filter and user application will receive the serialized message and de-serialize again...)

This is not necessarily true. `rcl_take` will call `rmw_take`, which will desrialize the message on the user supplied destination. It will then proceed with the filtering, and then return if the filter passes or call `rmw_take` again and repeat if the filter does not pass.

fujitatomoya:
> This is not necessarily true. rcl_take will call rmw_take, which will desrialize the message on the user supplied destination. It will then proceed with the filtering, and then return if the filter passes or call rmw_take again and repeat if the filter does not pass.

if user wants to have serialized message with CFT via `rmw_take_serialized_message`, unnecessary de-serialization would be required?

MiguelCompany:
> 
> if user wants to have serialized message with CFT via `rmw_take_serialized_message`, unnecessary de-serialization would be required?

In that case, yes.

fujitatomoya:
@iuhilnehc-ynos @Barry-Xu-2018 i think we have misunderstanding for

- https://github.com/ros2/rcl/pull/894
- https://github.com/ros2/rclcpp/pull/1561

we will enable CFT interfaces for `rcl` and `rclcpp` for Humble Release, Mon. April 4, 2022 - Freeze with `Compile Option` and `Necessary Interface extension for fallback filtering`.

i might miss something, let me know if i got anything wrong.

fujitatomoya:
@MiguelCompany @asorbini 

Do you have any implementation specific filtering expression and parameters for ContentFitleredTopic? Or as implementation spec, [DDS v1.4 Annex B - Syntax for Queries and Filters](https://www.omg.org/spec/DDS/1.4/PDF) just applies?

MiguelCompany:
@fujitatomoya Fast DDS supports the following extensions from Annex B:

* Single dimension array / sequence indexing with `[<index>]`, where `<index>` is a decimal positive value.
* Boolean constants `FALSE`, `false`, `TRUE`, `true`.
* Operator `!=`, which is an alias of `<>`.
* Operator `like` supports both `_` and `?` as the single character wildcard.
* Operator `like` supports both `%` and `*` as the zero or more characters wildcard.
* Operator `match` which supports `std::regex` expression matching.

asorbini:
> Do you have any implementation specific filtering expression and parameters for ContentFitleredTopic? Or as implementation spec, [DDS v1.4 Annex B - Syntax for Queries and Filters](https://www.omg.org/spec/DDS/1.4/PDF) just applies?

The supported syntax is [described in the docs](https://community.rti.com/static/documentation/connext-dds/6.0.1/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/index.htm#UsersManual/SQL_Filter_Expression_Notation.htm#topics_3596570972_201895), but to provide some "highlights" of the extensions/limitations:

- Access to individual members of arrays and sequences is supported via `[<index>]` operator.
- A `hex()` function is provided to compare multiple bytes.
- A `MATCH` operator is supported for comparing strings, which accepts "regular expressions" using a syntax similar to `fnmatch()`.
- Standard `LIKE` operator is not supported.
- The "negated character set" operator (i.e. `[!charlist]` or `[^charlist]`) is not supported.

fujitatomoya:
@MiguelCompany @asorbini thanks for the information.

fujitatomoya:
@iuhilnehc-ynos i updated the header with current status and action items, https://github.com/ros2/design/pull/282#issue-610568380.

> add sample to [demo_nodes_cpp](https://github.com/ros2/demos/tree/master/demo_nodes_cpp)

can you make demonstration code for CFT? i will be working on docs, and once demo code is ready, i would like to link it to docs.

iuhilnehc-ynos:
@fujitatomoya 

I have created the demo for CFT. To make it simple, I just provided a simple way to set the content filter setting by using the SubscriptionOption, please refer to https://github.com/ros2/demos/pull/557

fujitatomoya:
@wjwwood @ivanpauno 
CC: @iuhilnehc-ynos 

Just FYI, all tasks are listed on https://github.com/ros2/design/pull/282#issue-610568380.

fujitatomoya:
@wjwwood @ivanpauno 
CC: @iuhilnehc-ynos 

I just made https://github.com/ros2/rcl/issues/982 for `Fallback filtering subscription`

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2023-01-19/29423/1

PR #289, opened 2020-06-26
Clarify ros2 arguments precedence

ivanpauno:
Adds a section to the `command line arguments` article, clarifying how precedence currently works.
Adds a discussion section exposing inconsistencies and weird interaction with launch files.

The goal of this PR is to state how precedence currently works to kick off a discussion on how it should work.
After that discussion takes place, I will update the document with the conclusion.

Follow up of https://github.com/ros2/rclcpp/issues/953.

### DIFF
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -344,7 +344,7 @@ LaunchDescription([
 ```
 
 Where `my_params.yaml` can be find in *Precedence order of command line arguments* section.
-The parameter file is assigning `my_int` to `2`.
+The parameter file is assigning `2` to `my_int`.
 Launch cannot know the fully qualified name of the node at launch time, as the node namespace wasn't specified.
 Thus, it needs to use a "wildcard" rule for the assignment made in the dictionary, and it won't override the parameter file assignment.
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -251,7 +251,7 @@ Assuming that `some_exec` only has a node named `my_node`, this will result in `
 
 In the case of parameter files, the following precedence apply:
 
-* Assignments targeting a specific node prevail to wildcard assignments.
+* Parameter assignments targeting a specific node prevail over those using wildcards.
 * Within a file, in case of the contradictory assignments, the last assignment applies (i.e.: the one nearer to the end of the file).
 * Parameter files have the same precedence that command line parameter assignments. They are parsed from left to right, and the rightmost argument prevails.
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -226,7 +226,7 @@ As is, this enclave assignment applies to each and every Domain Participant that
 The following precedence order applies to parameter assignments:
 
 * Arguments targeting a specific node prevail arguments targeting all the nodes in the executable.
-* Arguments are parsed from the leftmost to the rightmost. In case of contradictory assignments, the rightmost argument prevails.
+* Arguments are parsed left to right. In case of contradictory assignments, the rightmost argument prevails.
 
 As an example:
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -238,7 +238,7 @@ Supposing the executable only has a node named `my_node`, this will result in `m
 
 The following precedence order applies to remapping rules:
 
-* Arguments are parsed from the leftmost to the rightmost. In case of contradictory assignments, the leftmost argument prevails.
+* Arguments are parsed from left to right. In case of overlapping rules, the leftmost argument prevails.
 * Arguments targeting a specific node DO NOT prevail arguments targeting all the nodes in the executable.
 
 As an example:
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -252,7 +252,7 @@ Assuming that `some_exec` only has a node named `my_node`, this will result in `
 In the case of parameter files, the following precedence apply:
 
 * Parameter assignments targeting a specific node prevail over those using wildcards.
-* Within a file, in case of the contradictory assignments, the last assignment applies (i.e.: the one nearer to the end of the file).
+* Within a file, in case of the overlapping assignments, the last assignment applies (i.e.: the one closest to EOF).
 * Parameter files have the same precedence that command line parameter assignments. They are parsed from left to right, and the rightmost argument prevails.
 
 As an example:
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
 
 As is, this enclave assignment applies to each and every Domain Participant that `some_ros_executable` spawns unless explicitly ignored in code or overridden via security environment variables.  
 
+#### Precedence order of command line arguments
+
+The following precedence order applies to parameter assignments:
+
+* Arguments targeting a specific node prevail arguments targeting all the nodes in the executable.
+* Arguments are parsed from the leftmost to the rightmost. In case of contradictory assignments, the rightmost argument prevails.
+
+As an example:
+
+```sh
+ros2 run some_package some_exec --ros-args -r my_topic:=asd -r my_topic:=bsd -p my_node:my_param:=1 -p my_param:=2 -p another_param:=1 -p another_param:=2
+```
+
+Supposing the executable only has a node named `my_node`, this will result in `my_param` being set to `1` and `another_param` being set to `2`.
+
+The following precedence order applies to remapping rules:
+
+* Arguments are parsed from the leftmost to the rightmost. In case of contradictory assignments, the leftmost argument prevails.
+* Arguments targeting a specific node DO NOT prevail arguments targeting all the nodes in the executable.
+
+As an example:
+
+```sh
+ros2 run some_package some_exec --ros-args -r my_topic:=first -r my_topic:=second -r my_node:another_topic:=first -r another_topic:=second
+```
+
+Supposing the executable only has a node named `my_node`, this will result in `my_topic` being remapped to `first` and `another_topic` being remapped to `first`.
+
+In the case of parameter files, the following precedence apply:
+
+* Assignments targeting a specific node prevail to wildcard assignments.
+* Within a file, in case of the contradictory assignments, the last assignment applies (i.e.: the one nearer to the end of the file).
+* Parameter files have the same precedence that command line parameter assignments. They are parsed from left to right, and the rightmost argument prevails.
+
+As an example:
+
+```yaml
+# my_params.yaml
+my_node:
+    ros__parameters:
+        my_int: 1
+        my_int: 2
+        another_int: 1
+        my_float: 1.0
+        another_float: 1.0
+my_node:
+    ros__parameters:
+        another_int: 2
+/**:
+    ros__parameters:
+        my_int: 3
+        global_int: 1
+```
+
+```sh
+ros2 run some_package some_exec --ros-args --params-file my_params.yaml
+```
+
+will result in: my_int=2, another_int=2, global_int=1, my_float=1.0, another_float=1.0
+
+```sh
+ros2 run some_package some_exec --ros-args -p my_node:global_int:=2 -p my_float:=2.0 --params-file my_params.yaml -p my_float:=3.0 -p my_node:another_float:=2.0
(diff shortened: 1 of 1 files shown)


ivanpauno:
@hidmic can you make a first pass?

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
 
 As is, this enclave assignment applies to each and every Domain Participant that `some_ros_executable` spawns unless explicitly ignored in code or overridden via security environment variables.  
 
+#### Precedence order of command line arguments
+
+The following precedence order applies to parameter assignments:
+
+* Arguments targeting a specific node prevail arguments targeting all the nodes in the executable.
+* Arguments are parsed from the leftmost to the rightmost. In case of contradictory assignments, the rightmost argument prevails.
```
hidmic: @ivanpauno 
```suggestion
* Arguments are parsed from the leftmost to the rightmost. In case of overlapping assignments, the rightmost argument prevails.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
...
+
+As an example:
+
+```sh
+ros2 run some_package some_exec --ros-args -r my_topic:=asd -r my_topic:=bsd -p my_node:my_param:=1 -p my_param:=2 -p another_param:=1 -p another_param:=2
+```
+
+Supposing the executable only has a node named `my_node`, this will result in `my_param` being set to `1` and `another_param` being set to `2`.
```
hidmic: @ivanpauno 
```suggestion
Assuming that `some_exec` only has a node named `my_node`, this will result in `my_param` being set to `1` and `another_param` being set to `2`.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
 
 As is, this enclave assignment applies to each and every Domain Participant that `some_ros_executable` spawns unless explicitly ignored in code or overridden via security environment variables.  
 
+#### Precedence order of command line arguments
+
+The following precedence order applies to parameter assignments:
```
hidmic: @ivanpauno nit: consider using H5 headers to separate remapping rules precedence from parameter assignment precedence.

articles/150_ros_command_line_arguments.md
hidmic: @ivanpauno also
```suggestion
* Arguments are parsed left to right. In case of contradictory assignments, the rightmost argument prevails.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
...
+
+```sh
+ros2 run some_package some_exec --ros-args -r my_topic:=asd -r my_topic:=bsd -p my_node:my_param:=1 -p my_param:=2 -p another_param:=1 -p another_param:=2
+```
+
+Supposing the executable only has a node named `my_node`, this will result in `my_param` being set to `1` and `another_param` being set to `2`.
+
+The following precedence order applies to remapping rules:
+
+* Arguments are parsed from the leftmost to the rightmost. In case of contradictory assignments, the leftmost argument prevails.
```
hidmic: @ivanpauno 
```suggestion
* Arguments are parsed from left to right. In case of overlapping rules, the leftmost argument prevails.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
...
+* Arguments targeting a specific node DO NOT prevail arguments targeting all the nodes in the executable.
+
+As an example:
+
+```sh
+ros2 run some_package some_exec --ros-args -r my_topic:=first -r my_topic:=second -r my_node:another_topic:=first -r another_topic:=second
+```
+
+Supposing the executable only has a node named `my_node`, this will result in `my_topic` being remapped to `first` and `another_topic` being remapped to `first`.
```
hidmic: @ivanpauno 
```suggestion
Assuming that `some_exec` only has a node named `my_node`, this will result in `my_topic` being remapped to `first` and `another_topic` being remapped to `first`.
```

articles/150_ros_command_line_arguments.md
hidmic: I think it'd be also be nice to include a rationale for each precedence.

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
...
+
+In the case of parameter files, the following precedence apply:
+
+* Assignments targeting a specific node prevail to wildcard assignments.
```
hidmic: @ivanpauno 
```suggestion
* Parameter targeting a specific node prevail over those using wildcards.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -221,6 +221,72 @@ ros2 run some_package some_ros_executable --ros-args -e "/foo/bar"
...
+* Within a file, in case of the contradictory assignments, the last assignment applies (i.e.: the one nearer to the end of the file).
```
hidmic: @ivanpauno 
```suggestion
* Within a file, in case of the overlapping assignments, the last assignment applies (i.e.: the one closest to EOF).
```

articles/150_ros_command_line_arguments.md
```diff
@@ -255,3 +321,50 @@ This signficantly increases command line verbosity, but still avoids the need fo
 Remove the need for double dash tokens (`--`), conventionally used to signify the end of CLI options for a command, by adding the `--ros-` prefix to all ROS specific command line flags e.g. `--ros-remap`, `--ros-param`, etc.
 In exchange, it makes argument extraction slightly more difficult as all options must be known ahead of time, whereas `--ros-args`-based namespacing can achieve the same with a couple rules.
 It also increases command line verbosity.
+
+## Precedence order discussion
+
+There are some issues of the current precedence rules:
```
hidmic: @ivanpauno 
```suggestion
There are some issues with the current precedence rules:
```
Also, when you say *current* we need to be specific about when.

articles/150_ros_command_line_arguments.md
```diff
@@ -255,3 +321,50 @@ This signficantly increases command line verbosity, but still avoids the need fo
...
+  Node(
+    package='my_pkg',
+    exec='my_exec',
+    name='my_node',
+    parameters=['/path/to/my_params.yaml', {'my_int': '3'}]
+  )
+])
+```
+
+Where `my_params.yaml` can be find in *Precedence order of command line arguments* section.
```
hidmic, line 346: @ivanpauno I'd reproduce the example parameter file here.

articles/150_ros_command_line_arguments.md
```diff
@@ -255,3 +321,50 @@ This signficantly increases command line verbosity, but still avoids the need fo
...
+The parameter file is assigning `my_int` to `2`.
```
hidmic: @ivanpauno 
```suggestion
The parameter file is assigning `2` to `my_int`.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -255,3 +321,50 @@ This signficantly increases command line verbosity, but still avoids the need fo
...
+    parameters=['/path/to/my_params.yaml', {'my_int': '3'}]
+  )
+])
+```
+
+As the fully qualified node name is known at launch time in this last case, `my_int` will be set to `3`.
+
+Possible solutions:
+* Warn when parameter files and parameter dictionaries are combined in `Node`, and the fully qualified node name is not known.
+* Modify precedence between wildcard and not wildcard assignments.
```
hidmic, line 370: @ivanpauno shouldn't this design doc weigh pros and cons and make an executive decision?

articles/150_ros_command_line_arguments.md
ivanpauno: This is stating how it currently works in order to kick off a discussion on the topic.
This is not trying to rationalize why it works in the way it works.

articles/150_ros_command_line_arguments.md
ivanpauno: I will apply the following diff, which I find clearer:
```suggestion
* Parameter assignments targeting a specific node prevail over those using wildcards.
```

articles/150_ros_command_line_arguments.md
ivanpauno: > Also, when you say current we need to be specific about when.

With current, I mean the ones described above.
After a discussion on how the precedence rules should be, I will update this again with the conclusion.

articles/150_ros_command_line_arguments.md
ivanpauno: I will address this after we reach a conclusion on how precedence should work.
As my current goal is to start a discussion, having a clear reference to the file isn't important.

articles/150_ros_command_line_arguments.md
ivanpauno: as commented previously, I'm not trying to weight advantages and disadvantages at the moment.

clalancette:
As the starting point of what rules are currently in place, this is great.  Thanks @ivanpauno .

Where we want to go is harder to say.  I would say that there are 2 main problems pointed out here:
1.  The rules for parameter overrides and remapping overrides are different.
1.  The rules for launch are inconsistent with the command-line.

I'm going to concentrate on the second one for now (the first one is unfortunate, but at least documented now).  The problem as I understand it is that launch does not (and cannot) know the fully-qualified node name.  That's because the node may have hard-coded node names and even namespaces internally to itself.  Is that a good summary of the problem?

If that's the case, I'm wondering what would happen if we end up making launch a bit dumber.  That is, launch would just pass along the arguments exactly as it had them.  Then the precedence rules would be figured out by the rcl/rclcpp layer inside of the node, and the parameters applied.  This takes launch completely out of the equation when doing overrides, and concentrates all of the logic of overrides in one place (it needs to be in rcl/rclcpp for the command-line case anyway).

I'm sure there are complications to the above, so I'm looking for feedback on why that would or wouldn't work.

ivanpauno:
> If that's the case, I'm wondering what would happen if we end up making launch a bit dumber. That is, launch would just pass along the arguments exactly as it had them. Then the precedence rules would be figured out by the rcl/rclcpp layer inside of the node, and the parameters applied. This takes launch completely out of the equation when doing overrides, and concentrates all of the logic of overrides in one place (it needs to be in rcl/rclcpp for the command-line case anyway).

The current implementation tries to be as dumb as possible.
The problem is that `Node` action targets a single process with a single node, and `rcl` doesn't have any assumption about the number of nodes in a process. We could definitely have in launch a `MultiNodeExecutable` action, which would allow to set up a process with manually composed nodes (trying to use `Node` action for this doesn't work at all).

When launch generates a command from:

```python
Node(
  package='my_pkg',
  executable='my_exec',
  name = 'my_node',
  parameters=['path/to/file', {'my_param': 'my_value}]
)
```

the result can be:

(op A)
```sh
/path/to/executable/in/package --ros-args -r __node:=my_node --params-file path/to/file -p my_param:=my_value
```

or:

(op B)
```sh
/path/to/executable/in/package --ros-args -r __node:=my_node --params-file path/to/file --params-file path/to/tmp/param/file
```

where:

```yaml
# path/to/tmp/param/file contents
/**:
  ros__parameters:
    my_param: my_value
```

Currently, launch is doing **op B**, but `op A` would result in the same problem.

I'm not sure of what modifications in `rcl` and `launch_ros` could solve the problem without the user explicitly passing the namespace in the `Node` action.
i.e. I'm not sure if we can make launch more dumb of what currently is (I have tried to think of a solution, and I haven't come up with anything).

---

Support to wildcards in parameter files was added to solve the problem of not knowing the fully qualified node name at launch time (see https://github.com/ros2/launch_ros/pull/35, https://github.com/ros2/launch_ros/issues/4).

Considering that that was the main goal of wildcards, maybe we could make them have the same precedence than parameter assignments targeting a specific node. The problem is that that is a bit counter intuitive (I find it super counter intuitive, but it would solve the launch file use case).

jacobperron:
Reviewing proposed changes for additional wildcard support (https://github.com/ros2/rclcpp/issues/1265) brought me here.
I realized that wildcard support in parameter YAML files was not documented anywhere and so it would be nice to update this design doc.

I've opened https://github.com/ros2/design/pull/303 to document how wildcards can be used in parameter files. I have not addressed precedence (like this PR aims to do).

PR #290, opened 2020-06-28
Update for supporting multiple logging level settings

iuhilnehc-ynos:
Signed-off-by: Chen Lihui <Lihui.Chen@sony.com>

### DIFF
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -167,7 +167,7 @@ another_node:
      string_param: bar
 ```
 
-#### Multiple Logging level assignments
+#### Multiple logging level assignments
 
 Logging level assignments have two means of settings which are a global logging level and other logging levels for specific loggers. Minimum logging level can be externally set using the `--log-level` option.
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -184,6 +184,7 @@ ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG 
 ```
 
 The minimum logging level of a specific logger will override the globally specified minimum logger level.
+If a logging level is specified more than once in the passed command line arguments, the last one prevails.
 
 See `rcutils` and `rcl` logging documentation for reference on existing logging levels.
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -180,7 +180,7 @@ ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 Loggers can be set using the `--log-level` option as well:
 
 ```sh
-ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN rclcpp:=DEBUG
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN --log-level rclcpp:=DEBUG
 ```
 
 The minimum logging level of a specific logger will override the globally specified minimum logger level.
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -167,15 +167,23 @@ another_node:
      string_param: bar
 ```
 
-#### Logging level
+#### Multiple Logging level assignments
 
-Minimum logging level can be externally set using the `--log-level` option.
+Logging level assignments have two means of settings which are a global logging level and other logging levels for specific loggers. Minimum logging level can be externally set using the `--log-level` option.
 
-As an example, to set logging level to `DEBUG` for `some_ros_executable`, one may execute:
+As an example, to set a global logging level to `DEBUG` for `some_ros_executable`, one may execute:
 
 ```sh
 ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 ```
+
+Some specific loggers can be set using the `--log-level` option as well, the format as follows:
+
+```sh
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
+```
+
+Note that the logging level of a specific logger can override the global logging level for this logger.
 
 See `rcutils` and `rcl` logging documentation for reference on existing logging levels.
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -169,7 +169,7 @@ another_node:
 
 #### Multiple logging level assignments
 
-Logging level assignments have two means of settings which are a global logging level and other logging levels for specific loggers. Minimum logging level can be externally set using the `--log-level` option.
+Minimum logging level can be externally set either globally or per logger using the `--log-level` option.
 
 As an example, to set a global logging level to `DEBUG` for `some_ros_executable`, one may execute:
 
@@ -177,13 +177,13 @@ As an example, to set a global logging level to `DEBUG` for `some_ros_executable
 ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 ```
 
-Some specific loggers can be set using the `--log-level` option as well, the format as follows:
+Some specific loggers can be set using the `--log-level` option as well:
 
 ```sh
 ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
 ```
 
-Note that the logging level of a specific logger can override the global logging level for this logger.
+The minimum logging level of a specific logger will override the globally specified minimum logger level.
 
 See `rcutils` and `rcl` logging documentation for reference on existing logging levels.
 
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -177,10 +177,10 @@ As an example, to set a global logging level to `DEBUG` for `some_ros_executable
 ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 ```
 
-Some specific loggers can be set using the `--log-level` option as well:
+Loggers can be set using the `--log-level` option as well:
 
 ```sh
-ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN rclcpp:=DEBUG
 ```
 
 The minimum logging level of a specific logger will override the globally specified minimum logger level.
--- a/articles/150_ros_command_line_arguments.md
+++ b/articles/150_ros_command_line_arguments.md
@@ -167,7 +167,7 @@ another_node:
      string_param: bar
 ```
 
-#### Multiple logging level assignments
+#### Logging level assignments
 
 Minimum logging level can be externally set either globally or per logger using the `--log-level` option.
 


iuhilnehc-ynos:
To support multiple logging level settings, there might exist some other ways which could be convenient for users, such as

```
    a. use multiple items "--log-level", such as  
        
        --log-level debug --log-level talker1=error --log-level talker2=warn

    b. use one item, such as
        
        --log-level debug,talker1=error,talker2=warn
```

But to keep consistent with other arguments that using multiple items and use `:=` as a separator, such as

```
--ros-args --param key1:=value1 --param key2:=value2
```

I wonder if users would like to use the multiple logger level setting as follows,
```
--log-level debug,talker1:=error,talker2:=warn
```
or just use
```
--log-level debug --log-level talker1:=error --log-level talker2:=warn
```

I'd like to hear your opinions.

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,24 @@ another_node:
      string_param: bar
 ```
 
-#### Logging level
+#### Multiple Logging level assignments
```
fujitatomoya:
```suggestion
#### Multiple logging level assignments
```

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,24 @@ another_node:
...
 
-Minimum logging level can be externally set using the `--log-level` option.
+Logging level assignments have two means of settings which are a global logging level and other logging levels for specific loggers. Minimum logging level can be externally set using the `--log-level` option.
```
fujitatomoya: `a global logging level` means `a global minimal logging level`, right?. I think either global or not, `Minimum` is intended to described by author.

fujitatomoya:
> But to keep consistent with other arguments that using multiple items and use := as a separator, such as

for user experience, i'd like to go with `:=` separator to keep the consistency.

articles/150_ros_command_line_arguments.md
iuhilnehc-ynos: Yes, the logging level in ROS2 uses a minimum value to control the output message.
`minimum` is used for both the global logging level and the specific logger level.

iuhilnehc-ynos:
> > But to keep consistent with other arguments that using multiple items and use := as a separator, such as
> 
> for user experience, i'd like to go with `:=` separator to keep the consistency.

@fujitatomoya 
Thanks. I agree with you.

There are some ways to use `:=`, 
```
A. --log-level debug,talker1:=error,talker2:=warn
B. --log-level debug --log-level talker1:=error --log-level talker2:=warn
C. both A and B
```
I think I need to hear others' suggestions.

gbiggs:
Option A makes sense if you consider it to be the combination of all specifications given in option B. I think that option B is clearer and easier to parse, though.

gbiggs:
What is the order of precedence if someone gives two different logging levels for a single node?

iuhilnehc-ynos:
> What is the order of precedence if someone gives two different logging levels for a single node?

the last one wins

gbiggs:
Then the design document must say that.

iuhilnehc-ynos:
> Then the design document must say that.

Thanks. But I think it's the policy of all arguments, not just for --log-level. So maybe it should not be recorded here.

gbiggs:
If you choose option A then it matters because the argument parser will only see a single argument, and whatever parses that string will need to know what to do.

iuhilnehc-ynos:
> If you choose option A then it matters because the argument parser will only see a single argument, and whatever parses that string will need to know what to do.

After reconsideration, I think I will use option B. (Actually I use option B in the design file).
No need to break the existing rule ( --param doesn't support --param key1:=value1,key2:=value2 )

iuhilnehc-ynos:
@gbiggs 
Anyway, thanks for your suggestion. (option B is an easy way)

fujitatomoya:
I am really good to go with either one.

fujitatomoya:
@ivanpauno @hidmic @wjwwood 

friendly ping.

ivanpauno:
> What is the order of precedence if someone gives two different logging levels for a single node?

We currently have a lot of inconsistencies on how argument precedence works, I recently opened a [PR](https://github.com/ros2/design/pull/289) about that.

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,24 @@ another_node:
      string_param: bar
 ```
 
-#### Logging level
+#### Multiple logging level assignments
```
ivanpauno: I would use "Logging level assignments" as the title

articles/150_ros_command_line_arguments.md
ivanpauno: nit: I suggest rephrasing, as it sounds a bit confusing to me.
```suggestion
Minimum logging level can be externally set either globally or per logger using the `--log-level` option.
```

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,24 @@ another_node:
...
+Logging level assignments have two means of settings which are a global logging level and other logging levels for specific loggers. Minimum logging level can be externally set using the `--log-level` option.
 
-As an example, to set logging level to `DEBUG` for `some_ros_executable`, one may execute:
+As an example, to set a global logging level to `DEBUG` for `some_ros_executable`, one may execute:
 
 ```sh
 ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 ```
 
+Some specific loggers can be set using the `--log-level` option as well, the format as follows:
```
ivanpauno:
```suggestion
Some specific loggers can be set using the `--log-level` option as well:
```

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,24 @@ another_node:
...
+
+```sh
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
+```
+
+Note that the logging level of a specific logger can override the global logging level for this logger.
```
ivanpauno:
```suggestion
The minimum logging level of a specific logger will override the globally specified minimum logger level.
```

ivanpauno (approved):
As @gbiggs suggested, I think it's good to clarify precedence order.
Using a "last specified argument prevails" rule sounds like a good idea.

This proposal is pretty consistent with how we parse other kinds of arguments, :+1: to that!

ivanpauno:
> ```
> A. --log-level debug,talker1:=error,talker2:=warn
> B. --log-level debug --log-level talker1:=error --log-level talker2:=warn
> C. both A and B
> ```

I would also like to mention that option C sounds reasonable to me.
But in that case, we should support similar syntax for arguments and parameters:
```sh
ros2 run package_name exec_name --ros-args -p param1:=value1,param2:=value2
```

That would be a much bigger effort that only implementing the new log level arguments.
I would try to keep the proposal as minimal as possible.

ivanpauno:
@clalancette FYI

articles/150_ros_command_line_arguments.md
clalancette: Actually, I'd use something even different here:

```
Per-node log-level assignments
```

Which I think describes the feature better.

clalancette:
> A. --log-level debug,talker1:=error,talker2:=warn
> B. --log-level debug --log-level talker1:=error --log-level talker2:=warn
> C. both A and B

While I have my own personal preferences, I think consistency is more important here.  With that in mind, I'd definitely stick with the `:=` as the separator, as that is consistent with parameters.

I would also stick with option B here, as it more closely conforms to what parameters are doing.  If we want to expand in the future to support syntax A, we would do it across all of the parameters we support.

articles/150_ros_command_line_arguments.md
wjwwood: Is it per node or per logger?

wjwwood:
I agree with B, until we want to support something like A everywhere, aka C.

articles/150_ros_command_line_arguments.md
iuhilnehc-ynos: > Is it per node or per logger?

It's per logger.

iuhilnehc-ynos: I'll update it with 'Logging level assignments'.

fujitatomoya (approved):
LGTM

ivanpauno (approved):
I think it's worth clarifying the precedence in the design article.

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,24 @@ another_node:
...
 ```
 
+Some specific loggers can be set using the `--log-level` option as well:
+
+```sh
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
+```
+
+The minimum logging level of a specific logger will override the globally specified minimum logger level.
+
```
ivanpauno:
```suggestion
The minimum logging level of a specific logger will override the globally specified minimum logger level.
If a logging level is specified more than once in the passed command line arguments, the last one prevails.

```

ivanpauno (approved):

clalancette (approved):

clalancette:
Both Ivan and I are happy with this.  I'll let this sit another day before merging to see if there are any additional comments on this proposal, then I'll merge.  Pinging @wjwwood @gbiggs for any other thoughts.

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,25 @@ another_node:
...
+Minimum logging level can be externally set either globally or per logger using the `--log-level` option.
 
-As an example, to set logging level to `DEBUG` for `some_ros_executable`, one may execute:
+As an example, to set a global logging level to `DEBUG` for `some_ros_executable`, one may execute:
 
 ```sh
 ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 ```
 
+Some specific loggers can be set using the `--log-level` option as well:
```
wjwwood: So does `Some` here mean there are loggers you cannot set this way? If not, I'd drop it. If so, then I think we need some clarification on that.

articles/150_ros_command_line_arguments.md
```diff
@@ -167,16 +167,25 @@ another_node:
...
+
+```sh
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
```
wjwwood: It might be nice to have an example that's not just a node, like maybe `rclcpp` or `talker1.rclcpp` or something more interesting like that.

wjwwood (approved):
lgtm

I only had two nitpick comments that could be addressed or ignored.

gbiggs (approved):

gbiggs:
LGTM

fujitatomoya (approved):

articles/150_ros_command_line_arguments.md
```diff
@@ -177,10 +177,10 @@ As an example, to set a global logging level to `DEBUG` for `some_ros_executable
 ros2 run some_package some_ros_executable --ros-args --log-level DEBUG
 ```
 
-Some specific loggers can be set using the `--log-level` option as well:
+Loggers can be set using the `--log-level` option as well:
 
 ```sh
-ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN
+ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN rclcpp:=DEBUG
```
ivanpauno:
```suggestion
ros2 run some_package some_ros_executable --ros-args --log-level talker1:=DEBUG --log-level talker2:=WARN --log-level rclcpp:=DEBUG
```
?

iuhilnehc-ynos: Oh, bad mistake

clalancette:
I'm going to go ahead and merge this with the latest fixes now.  Thanks everyone for reviewing this and moving it forward.

PR #297, opened 2020-08-20
Added design document for remote and multi-machine launching

mlanting:
The document has been changed so much and it's been so long since the last time I submitted a PR for multi-machine launch changes I figured it'd be more appropriate just to create a new one.

### DIFF
--- a/articles/153_roslaunch_remote_launch.md
+++ b/articles/153_roslaunch_remote_launch.md
@@ -4,26 +4,15 @@ This article describes the design of the components of the ros2 launch system re
 ## Goals
 ROS 1's launch system provided the means to launch nodes remotely via SSH by defining a "machine" and referencing that machine in the declaration of each node meant to run on it.
 The machine declaration would contain the information necessary to execute processes remotely, such as address, username, password, etc.
-The goal of this design is to enable the same ability to launch nodes on remote machines, as well as to provide infrastructure for more advanced capabilities.
+These nodes would be executed on the specified machines and run until the launch process that started them was killed.
 
-- Monitor the status and manage lifecycles of nodes across systems
-- Shutdown nodes remotely
-- Ability to detach the machine used to launch a system and have the system continue running
-    - Be able to reconnect to the system to monitor and manage nodes running remotely
-
-
-To achieve our goals, this update will consist of two primary components: an ExecuteRemote action, and a LaunchServiceNode.
-
-The ExecuteRemote action will provide the basic capaiblity to start processes on remote machines.
-It will build off of the refactor that is described [here](https://github.com/ros2/design/pull/272) and include a machine object as an additional parameter providing implementation-specific connection information.
-Since we don't want to limit remote execution to a single protocol, ExecuteRemote will be a base class and we will create an implementation using SSH which we will call ExecuteRemoteSSH.
-The machine class will also be a base class for protocol-specific implementations.
+The goal of this design is to create an interface for similar remote execution capabilities to be added to the launch system for ROS2.
+This design will include an SSH implementation of the interface, but the design of the interface itself will allow for implementations using other protocols to be added as well.
+The primary component of this design will be the ExecuteRemote action.
+It will be similar to the ExecuteLocal action described in the design for refactoring ExecuteProcess into Execute and Executable (https://github.com/ros2/design/pull/272), with the addition of a machine parameter providing implementation-specific connection information.
+Since we don't want to limit remote execution to a single protocol, the ExecuteRemote and Machine classes will be base classes, and we will include implementations using SSH which we will call ExecuteRemoteSSH and SSHMachine, respectively.
 Together these components will provide the means to execute processes remotely and should cover most basic use cases.
 
-The [LaunchServiceNode(s)](154_roslaunch_remote_launch_service_node.md) will expose ROS2 services and topics for handling startup and shutdown of processes on remote machines.
-This component won't be necessary for basic execution of remote nodes, but would allow remote system to persist beyond the life of the initial LaunchService.
-The topics and services will allow remote tools to view and manage the processes running remotely.
-The LaunchServiceNode allows users to monitor and manage the status and lifecycles of nodes across systems, and enables the system to continue running if the machine used to start a system becomes disconnected.
 
 
 
@@ -40,7 +29,7 @@ Contains information needed to run a process on another machine via ssh
 |---|---|---|
 | hostname   | String | Hostname of the machine. |
 | port       | int | Port on the host to connect to (default 22) |
-| ssh\_keys  | String | Path to ssh key file |
+| ssh\_key   | String | Path to ssh key file |
 | passphrase | String | Optional passphrase if the key uses one. |
 
 #### launch.actions.ExecuteRemote
@@ -53,13 +42,13 @@ Base class for actions that use connection information contained in `machine` to
 | sigterm\_timeout     | List[Substitution] | Time until shutdown should escalate to `SIGTERM`, as a string or a list of strings and `Substitution`s to be resolved at runtime, defaults to the `LaunchConfiguration` called `sigterm_timeout` |
 | sigkill\_timeout     | List[Substitution] | Time until escalating to `SIGKILL` after `SIGTERM`, as a string or a list of strings and `Substitution`s to be resolved at runtime, defaults to the `LaunchConfiguration` called `sigkill_timeout` |
 | emulate\_tty         | boolean | Emulate a tty (terminal), defaults to `False`, but can be overridden with the `LaunchConfiguration` called `emulate_tty`, the value of which is evaluated as true or false according to `evaluate_condition_expression`. |
-| prefix               | List[Substitution] | A set of commands/arguments to preceed the `cmd`, used for things like `gdb`/`valgrind` and defaults to the `LaunchConfiguration` called `launch-prefix` |
+| prefix               | List[Substitution] | A set of commands/arguments to precede the `cmd`, used for things like `gdb`/`valgrind` and defaults to the `LaunchConfiguration` called `launch-prefix` |
 | output               | ?? | Configuration for process output logging. Defaults to `log` i.e. log both `stdout` and `stderr` to launch main log file and stderr to the screen. |
 | output\_format       | ?? | For logging each output line, supporting `str.format()` substitutions with the following keys in scope: `line` to reference the raw output line and `this` to reference this action instance. |
-| log\_cmd             | boolean | If `True`, prints the final cmd before executing the process, which is useful for debugging when substitutions are involved. |
+| log\_cmd             | boolean | If `True`, prints the final cmd before executing the process, which is useful for debugging when substitutions are involved, though implementations should take care to ensure sensitive information such as keys and passphrases are hidden. |
 | on\_exit             | List[LaunchDescriptionEntity] | List of actions to execute upon process exit.|
 | persistent\_connection | boolean | Whether the LaunchService should maintain a persistent connection to the ssh instance |
-| machine              | `launch.descriptions.SSHMachine` | The machine on which to execute the process |
+| machine              | `launch.descriptions.Machine` | The machine on which to execute the process |
 ##### Methods
 | Name               | Type | Description                                                  |
 | ------------------ | --- | ------------------------------------------------------------ |
@@ -79,13 +68,13 @@ Inherited
 | exec\_remote\_command | void | Executes a command on the remote machine
 
 #### launch.substitutions.RemoteSubstitution
-Uses a connection to a machine to run a command remotely and retuns the result as a string
+Uses a connection to a machine to run a command remotely and returns the result as a string
 ##### Members
 ##### Constructor
 | Argument | Type | Description |
 |---|---|---|
 | command | Text | The command to run on the remote machine |
-| machine | `launch.descriptions.Machine` | An object decribing the machine that the command should be run on. |
+| machine | `launch.descriptions.Machine` | An object describing the machine that the command should be run on. |
 ##### Methods
 | Name | Type | Description |
 |---|---|---|
--- a/articles/153_roslaunch_remote_launch.md
+++ b/articles/153_roslaunch_remote_launch.md
@@ -1,0 +1,185 @@
+# Multi-Machine and Remote Launching
+## Goals
+- Connect to a remote host and run nodes on it
+    - Ensure this capability does not compromise security
+- Support arbitrary remote execution methods
+- Monitor the status of and manage lifecycles of nodes across systems
+- Shutdown nodes remotely
+- Ability to detach the machine used to launch a system and have the system continue running
+    - Be able to reconnect to the system to monitor and manage nodes running remotely
+
+
+To achieve our goals, this update will consist of two primary components: an ExecuteRemote action, and a LaunchServiceNode.
+
+The ExecuteRemote action will build off of the refactor he is working on described here: https://github.com/ros2/design/pull/272
+Since we don't want to limit remote execution to a single protocol, ExecuteRemote will be a base class and we will create an implementation using SSH which we will call ExecuteRemoteSSH.
+Remote machines will be described by a Machine class which will also be a base class for protocol-specific implementations.
+Together these components will provide the means to execute processes remotely and should cover most basic use cases.
+
+The LaunchServiceNode(s) will expose ROS2 services and topics for handling startup and shutdown of processes on remote machines.
+This component won't be necessary for basic execution of remote nodes, but would allow remote system to persist beyond the life of the initial LaunchService.
+The topics and services will allow remote tools to view and manage the processes running remotely.
+To prevent execution of completely arbitrary processes, the node will be given a list of commands it is allowed to run at startup, each with a unique ID assigned by the user.
+Requests to start or stop processes would then use the unique ID to indicate to the node which process to start or stop.
+
+
+## New Classes
+
+#### launch.descriptions.Machine
+##### Members
+Members would be defined by child classes, since different protocols will require different types of information
+
+#### launch.descriptions.SSHMachine
+Contains information needed to run a process on another machine
+##### Members
+| Name | Type | Description |
+|---|---|---|
+| hostname   | String | Hostname of the machine. |
+| port       | int | Port on to connect to (default 22) |
+| ssh\_keys  | String | Path to ssh key file |
+| passphrase | String | Optional passphrase if the key uses one. |
+
+#### launch.actions.ExecuteRemote
+Base class for actions that use connection informationn contained in `machine` to start a process on a remote machine.
+##### Members
+| Name | Type | Description |
+|---|---|---|
+| process\_description  |`launch.descriptions.Executable` | An object describing the process to be executed.|
+| shell                 | boolean | If True, a shell is used to execute the process.|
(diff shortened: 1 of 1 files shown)


articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
+# Multi-Machine and Remote Launching
+## Goals
+- Connect to a remote host and run nodes on it
+    - Ensure this capability does not compromise security
```
kyrofa: We're currently working on launch integration for security. In order to make sure it scales to remote launching, we might want to consider ways to send security files over to the remote machine.

ros-discourse:

https://discourse.ros.org/t/ros2-tooling-wg-next-meeting/12545/36

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+|---|---|---|
+| describe | Text | Override method. Returns a description of this substitution as a string. |
+| perform  | Text | Override method. Perform the substitution, given the launch context, and return it as a string. |
+
+
+## ROS
+
+### Nodes
+#### launch\_ros.LaunchServiceNode
+This node runs on each machine in a system is is repsonsible for launching and managing remote processes.
```
fujitatomoya: nitpick:

```suggestion
This node runs on each machine in a system is is responsible for launching and managing remote processes.
```

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+
+To achieve our goals, this update will consist of two primary components: an ExecuteRemote action, and a LaunchServiceNode.
+
+The ExecuteRemote action will build off of the refactor he is working on described here: https://github.com/ros2/design/pull/272
+Since we don't want to limit remote execution to a single protocol, ExecuteRemote will be a base class and we will create an implementation using SSH which we will call ExecuteRemoteSSH.
+Remote machines will be described by a Machine class which will also be a base class for protocol-specific implementations.
+Together these components will provide the means to execute processes remotely and should cover most basic use cases.
+
+The LaunchServiceNode(s) will expose ROS2 services and topics for handling startup and shutdown of processes on remote machines.
+This component won't be necessary for basic execution of remote nodes, but would allow remote system to persist beyond the life of the initial LaunchService.
```
fujitatomoya: I am not clear on this use case, could you elaborate a little bit? especially on the difference from `LifecycleNode` except process statistics?

roger-strain: The design is intended to allow you to run a node pretty much like they currently would run, that is, bound to the lifespan of the launch process itself. When the launch process dies, the SSH connection dies, and anything still running on that SSH terminal dies with it.

The `LaunchServiceNode` is intended to work a little bit differently. It would be launched from the SSH terminal, but then detach and run without requiring the SSH connection to remain up. However, by doing that you lose the ability to directly kill a process if you actually want it to stop. Therefore, the `LaunchServiceNode` would provide a way to go back and ask for particular nodes on that server to be started/stopped ad hoc.

This is certainly somewhat similar to some of the functionality envisioned by `LifecycleNode`s, but would allow a way for us to remotely launch and manage nodes which were not natively designed to support the lifecycle feature. Lifecycle would provide nearly all the functionality we would need for this type of behavior, but should the node crash, Lifecycle won't let us bring it back for a full restart, and we also would be lacking other out-of-band control to deal with things such as unresponsive nodes.

(Obviously, these same arguments also apply to the `LaunchServiceNode` itself, so that particular entity should be as clean as possible to reduce the possibility of it getting into an unmanageable state itself.)

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+##### Members
+| Name | Type | Description |
+|---|---|---|
+| hostname   | String | Hostname of the machine. |
+| port       | int | Port on to connect to (default 22) |
+| ssh\_keys  | String | Path to ssh key file |
+| passphrase | String | Optional passphrase if the key uses one. |
+
+#### launch.actions.ExecuteRemote
+Base class for actions that use connection informationn contained in `machine` to start a process on a remote machine.
```
jaisontj: typo: information~n~

articles/153_roslaunch_remote_launch.md
jaisontj: nit: This node runs on each machine in a system ~is~ it is

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+To achieve our goals, this update will consist of two primary components: an ExecuteRemote action, and a LaunchServiceNode.
+
+The ExecuteRemote action will build off of the refactor he is working on described here: https://github.com/ros2/design/pull/272
+Since we don't want to limit remote execution to a single protocol, ExecuteRemote will be a base class and we will create an implementation using SSH which we will call ExecuteRemoteSSH.
+Remote machines will be described by a Machine class which will also be a base class for protocol-specific implementations.
+Together these components will provide the means to execute processes remotely and should cover most basic use cases.
+
+The LaunchServiceNode(s) will expose ROS2 services and topics for handling startup and shutdown of processes on remote machines.
```
emersonknapp: This LaunchServiceNode could be designed as an independent component - that is possible to use purely locally to detach a launch command from an ongoing launch context - it could then be used by the remote launching functionality.

articles/153_roslaunch_remote_launch.md
emersonknapp: Question to explicitly state in the doc: Who starts the LaunchServiceNode? How is its lifecycle expected to be managed?

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+
+## ROS
+
+### Nodes
+#### launch\_ros.LaunchServiceNode
+This node runs on each machine in a system is is repsonsible for launching and managing remote processes.
+#### Params
+| Name | Type | Description |
+|---|---|---|
+| allowable\_processes | List[Executable] | List of executables the node is allowed to launch |
```
emersonknapp: Explicit flag to allow remote launching _any_ process/node? Early in development this could be super useful for testing out various nodes via remote context from developer machine to robot.

emersonknapp: Or - is it not worth having this allow-list in here at all. Maybe SROS2 takes care of permissions well enough that we don't need to reimplement it. "If you use this you are exposing a vulnerability if you don't use SROS"

emersonknapp: Perhaps, only allow the LaunchServiceNode to only start known ros nodes on the remote system, rather than arbitrary commands. Restricting to ROS-specific in the remote context, since the local launch can execute arbitrary SSH commands anyways

dirk-thomas:
Some high level comments: while some of the aspects of the document are specific to remote execution:

* `ExecuteRemote`
* `RemoteSubstitution`

several other parts are applicable to a local launch invocation too:

* detach the main process and keep the launched processes running as well as continue handling event
* `Heartbeat
* `ProcessStatus`
* all the services like `QueryStatus`, `Start/StopProcess`, Shutdown.

It might be good to separate these to make each invidual part easier to design, implement and review.

emersonknapp:
Noting that we should call out explicitly the expectation that the LaunchServiceNode must be accessible within the current discoverable DDS network. The machine may be WAN-accessible via SSH but that case won't work because we use topics/services for all controls.

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+- heartbeat
+###### Received
+##### Services
+- query\_status
+- start\_process
+- stop\_process
+- shutdown
+
+### Messages
+#### Heartbeat
```
emersonknapp: Do all LaunchServiceNodes use a unique topic, or do all publish on the same topic (in which case would need to put machine/service ID in the messages)

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,185 @@
...
+|---|---|---|
+| timestamp       | `builtin_interface.Timestamp` | Time at which message was sent |
+| process\_name   | String | Name of the process |
+| process\_id     | int | A unique numerical identifier for the process. This value is generated by the launch system and should not be expected to be the same as the proc id on the machine. |
+| process\_status | int | An enumerated status value. STOPPED=0; RUNNING=1; SUSPENDED=2; |
+| mem\_usage      | double | Percent of memory used by the process |
+| processor\_load | double | Percent of processor load coming from this process |
+| err\_msg        | String | Error string |
+
+### Services
```
emersonknapp: Noting that each LaunchServiceNode will need to namespace itself to make service calls uniquely-fulfillable

articles/153_roslaunch_remote_launch.md
emersonknapp: Maybe this should be configurable, "heartbeat_topic", "process_status_topic"

emersonknapp:
When talking about multi-machine message passing, how does time sync come into play?

"Out of Scope" section in design is useful.

emersonknapp:
How do we support Mac + Windows? Do we think the SSH process will work out of the box for those platforms?

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-agenda-2020-09-17/16464/1

AlexKaravaev:
Hi everyone! For me as a ros2 developer, I really do miss that feature. Any timelines on this? I see that it's only a design doc and it's hard to say, but maybe it was included in some roadmap or so

emersonknapp:
So far there is not a roadmap that I know of. We discussed the design in one of the Tooling Working Group meetings, which is where my above comments came from, but I think the next step is just an updated split design - we identified that this is really 2 separate features that can be independently designed and implemented. 

@mlanting should be able to clarify if any proof-of-concept code exists already, I can't remember.

mlanting:
@AlexKaravaev, @pjreed made a simpler, temporary implementation of multi-machine launching that you can find here: https://github.com/pjreed/launch

I will try to get the next version of the design document out in the next week or so. @roger-strain is currently working on a refactor of the launch system. Once that is complete and this design is approved we can being implementation. I expect it will be a few months before we have anything implemented, but pj's implementation should cover basic use cases.

pjreed:
Just FYI, my changes in the `launch` repository (on the `multi-machine-launch` branch) basically just add in some changes necessary to abstract out the launching mechanism; I have another repository at https://github.com/pjreed/ssh_machine which adds an SSH-based launcher that can be used to launch nodes on remote machines via SSH.  The functionality is similar to how ROS1's remote launching mechanism works, but there are a few limitations to it; there's some more documentation and an example in that `ssh_machine` repo.

AlexKaravaev:
@mlanting @pjreed wow, thanks ! 
Will definitely check it out, because I was thinking of implementing similar ssh-based launcher. I understand the limitations and sure there is a need of multi-machine launch in ros2 native support, but still this repo can solve some current problems.

AlexKaravaev:
@mlanting @pjreed wow, thanks ! 
Will definitely check it out, because I was thinking of implementing similar ssh-based launcher. I understand the limitations and sure there is a need of multi-machine launch in ros2 native support, but still this repo can solve some current problems.

gavanderhoorn:
@mlanting et al.: this could have been discussed in the previous PR, so if it has, please just RTFM me, but has there been any thought about using existing orchestration frameworks for deployment, configuration and orchestration of multi-machine ROS applications?

I'm not immediately thinking of something like k8s, but perhaps something similar might exist, which is a little less 'heavy' but supports like operations ("infrastructure as code" and all that).

Ian Sherman in his keynote at ROSCon18 posited that "backend distributed systems and IoT" (as he phrased it) might have solutions to challenges we still see as problems. Deploying, starting and coordinating a multi-machine distributed application seems like it could be something they might have solutions or best practices for. It does seem like they might have similar requirements (ie: liveness tracking, monitoring, pushing updates, distributed configuration, failover, etc).

pjreed:
@AlexKaravaev: I haven't polished up my SSH-based launcher mostly because it was intended to be a stopgap solution that just implemented what we needed until a more robust solution is available, and I feel like some of the limitations (most significantly, the executable paths to nodes being resolved on the local computer before the command is sent to the remote computer) make it inappropriate for merging into the official ROS2 ecosystem at this time.  Still, hopefully it works for you; let me know if you have any issues, and if it doesn't take too much work to make it more robust, it might make sense to get it into ROS2 proper until we've got a better system.

@gavanderhoorn: When I made my `ssh_machine` launcher, actually, one of the considerations I had in mind is that it would be good if the remote launching mechanism wasn't tied to any particular implementation; in theory you could also write, say, a K8sMachine class that extends the `launch.machine.Machine` class to use k8s for launching instead of SSH.

gavanderhoorn:
@pjreed wrote:
> When I made my `ssh_machine` launcher, actually, one of the considerations I had in mind is that it would be good if the remote launching mechanism wasn't tied to any particular implementation; in theory you could also write, say, a K8sMachine class that extends the `launch.machine.Machine` class to use k8s for launching instead of SSH.

the abstraction is nice, but wouldn't that still mean that `launch` is doing the orchestration? It would be delegating some parts of it to k8s, but that would be it.

My assumption is that it's going to take "the ROS community" quite some time to get to the level of functionality which these existing solutions have already reached. Besides that, it would also seem to be duplication of effort, which is never very nice.

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+The machine declaration would contain the information necessary to execute processes remotely, such as address, username, password, etc.
+The goal of this design is to enable the same ability to launch nodes on remote machines, as well as to provide infrastructure for more advanced capabilities.
+
+- Monitor the status and manage lifecycles of nodes across systems
+- Shutdown nodes remotely
+- Ability to detach the machine used to launch a system and have the system continue running
+    - Be able to reconnect to the system to monitor and manage nodes running remotely
+
+
+To achieve our goals, this update will consist of two primary components: an ExecuteRemote action, and a LaunchServiceNode.
```
emersonknapp: Based on the re-scope of this design - it seems like this "Goals" section shouldn't be talking about the `LaunchServiceNode` since it's not specified within this design. If I remember correctly, we established that the `LaunchServiceNode` was independently interesting and designable. I do see that we want to talk about how these two things will interact, but I'm not sure if that belongs here in this doc anymore.

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+
+The ExecuteRemote action will provide the basic capaiblity to start processes on remote machines.
```
emersonknapp:
```suggestion
The ExecuteRemote action will provide the basic capability to start processes on remote machines.
```

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+It will build off of the refactor that is described [here](https://github.com/ros2/design/pull/272) and include a machine object as an additional parameter providing implementation-specific connection information.
```
emersonknapp: "Here" is never a very informative link title :)

```suggestion
It will build off of the refactor that is described in [Refactoring ExecuteProcess into Execute and Executable](https://github.com/ros2/design/pull/272) and include a machine object as an additional parameter providing implementation-specific connection information.
```

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+Since we don't want to limit remote execution to a single protocol, ExecuteRemote will be a base class and we will create an implementation using SSH which we will call ExecuteRemoteSSH.
+The machine class will also be a base class for protocol-specific implementations.
+Together these components will provide the means to execute processes remotely and should cover most basic use cases.
+
+The [LaunchServiceNode(s)](154_roslaunch_remote_launch_service_node.md) will expose ROS2 services and topics for handling startup and shutdown of processes on remote machines.
```
emersonknapp: It may be easier to evaluate this design tie-in if the linked document was available?

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+Members are defined by child classes, since different protocols require different types of information
+
+#### launch.descriptions.SSHMachine
+Contains information needed to run a process on another machine via ssh
+##### Members
+| Name | Type | Description |
+|---|---|---|
+| hostname   | String | Hostname of the machine. |
+| port       | int | Port on the host to connect to (default 22) |
+| ssh\_keys  | String | Path to ssh key file |
```
emersonknapp: should this be singular?
```suggestion
| ssh\_key  | String | Path to ssh key file |
```

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+| emulate\_tty         | boolean | Emulate a tty (terminal), defaults to `False`, but can be overridden with the `LaunchConfiguration` called `emulate_tty`, the value of which is evaluated as true or false according to `evaluate_condition_expression`. |
+| prefix               | List[Substitution] | A set of commands/arguments to preceed the `cmd`, used for things like `gdb`/`valgrind` and defaults to the `LaunchConfiguration` called `launch-prefix` |
+| output               | ?? | Configuration for process output logging. Defaults to `log` i.e. log both `stdout` and `stderr` to launch main log file and stderr to the screen. |
+| output\_format       | ?? | For logging each output line, supporting `str.format()` substitutions with the following keys in scope: `line` to reference the raw output line and `this` to reference this action instance. |
+| log\_cmd             | boolean | If `True`, prints the final cmd before executing the process, which is useful for debugging when substitutions are involved. |
```
emersonknapp: maybe should have a note that this will avoid printing any key passphrases or other sensitive information

articles/153_roslaunch_remote_launch.md
```diff
@@ -0,0 +1,94 @@
...
+| on\_exit             | List[LaunchDescriptionEntity] | List of actions to execute upon process exit.|
+| persistent\_connection | boolean | Whether the LaunchService should maintain a persistent connection to the ssh instance |
+| machine              | `launch.descriptions.SSHMachine` | The machine on which to execute the process |
```
emersonknapp: Want this to take the abstract interface, right?

```suggestion
| machine              | `launch.descriptions.Machine` | The machine on which to execute the process |
```

emersonknapp (changes requested):

emersonknapp:
I think overall this looks straightforward, but it still has remnants of the process management node, which shouldn't be necessary for the simple version of this design, right?

articles/153_roslaunch_remote_launch.md
mlanting: Oh, good point, I will include a note about that.

articles/153_roslaunch_remote_launch.md
mlanting: good catch

articles/153_roslaunch_remote_launch.md
mlanting: You are correct, thanks

mlanting:
Yeah, I think you're right. I've removed those references and fixed the errors you caught.

emersonknapp (approved):
This looks good to me now - though I'm not one with merge access here :)

ros-discourse:

https://discourse.ros.org/t/ros2-distributed-multi-pc-ros-launch-node-management-tool/39889/1

PR #299, opened 2020-09-22
Update remapping order of __ns and __node to not affect each other

iuhilnehc-ynos:
I encountered an issue about remapping two nodes that originally have a different name to the same new name but with a different namespace.

ex.
there are two nodes in `examples_rclcpp_minimal_composition`, and we want to remap node names
`publisher_node` -> `/publisher_different_ns/test_same_node`
`subscriber_node` -> `/subscriber_different_ns/test_same_node`

we want to do remapping names like the following way,
```
$ ros2 run examples_rclcpp_minimal_composition composition_composed \
    --ros-args \
    -r publisher_node:__ns:=/publisher_different_ns \
    -r publisher_node:__node:=test_same_node \
    -r subscriber_node:__ns:=/subscriber_different_ns \
    -r subscriber_node:__node:=test_same_node
```
rather than using
```
$ ros2 run examples_rclcpp_minimal_composition composition_composed \
    --ros-args \
    -r test_same_node:__ns:=/publisher_different_ns \ 
    -r publisher_node:__node:=test_same_node \
    -r test_same_node:__ns:=/subscriber_different_ns \
    -r subscriber_node:__node:=test_same_node
```
, since the existing remapping order that parsing `__node` before `__ns` causes the latter command to use the same `node_name prefix of __ns` parameter for two nodes.

Related to https://github.com/ros2/rcl/issues/806

Signed-off-by: Chen Lihui <Lihui.Chen@sony.com>

### DIFF
--- a/articles/141_static_remapping.md
+++ b/articles/141_static_remapping.md
@@ -255,7 +255,6 @@ Example rules are:
 - `foo:=bar`
 - `/foo/bar:=fiz/buzz`
 - `nodename:~/foo:=foo`
-- `/ns1/nodename:foo:=bar`
 - `**/foo:=\1/bar`
 
 #### Match Part of a Rule
@@ -266,7 +265,7 @@ The match part of a rule uses these operators:
 - `**` matches zero or more tokens delimited by slashes
 - `rosservice://` prefixed to the match makes the rule apply to only services
 - `rostopic://` prefixed to the match makes the rule apply to only topics
-- `nodename:` prefixed to the match makes it apply only to a node with that name, which could be a FQN
+- `nodename:` prefixed to the match makes it apply only to a node with that name
 
 The operators `*` and `**` are similar to the globbing behavior in bash.
 `**` behaves similar to its use in bash>=4.0 with the globstar option set.
@@ -325,8 +324,7 @@ The replacement must be a single token which will become the node's new name.
 
 Remapping rules are applied in the following order:
 
-1. Namespace remapping
-1. Node name remapping
+1. Namespace or Node name remapping
 1. All other rules
 
 Within each category, the rules are applied in the order in which the user gave them.
@@ -340,8 +338,8 @@ Within each category, the rules are applied in the order in which the user gave 
 **Example of node/namespace remapping order:**
 
 - A node has name `talker`
-- A user gives the rules `talker:__ns:=/my_namespace` then `/talker:__node:=foo`
-- The final node name is the default `talker` because the namespace remap is applied before the node name remap
+- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`
+- The final node name is `foo` because the namespace and node name remap are applied without affecting each other
 
 **Example of a default and node specific namespace remap:**
 
--- a/articles/141_static_remapping.md
+++ b/articles/141_static_remapping.md
@@ -339,7 +339,7 @@ Within each category, the rules are applied in the order in which the user gave 
 
 - A node has name `talker`
 - A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`, or vice versa
-- The final fully qualified name is `/my_namespace/foo` because the namespace and node name remap are applied without affecting each other
+- The final fully qualified name is `/my_namespace/foo` because the namespace and node name remap rules both match against the orignal node name before remappings are applied
 
 **Example of a default and node specific namespace remap:**
 
--- a/articles/141_static_remapping.md
+++ b/articles/141_static_remapping.md
@@ -339,13 +339,19 @@ Within each category, the rules are applied in the order in which the user gave 
 
 - A node has name `talker`
 - A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`, or vice versa
-- The final fully qualified name is `/my_namespace/foo` because the namespace and node name remap rules both match against the orignal node name before remappings are applied
+- The final fully qualified name is `/my_namespace/foo` because the namespace and node name remap rules both match against the original node name before remappings are applied
 
 **Example of a default and node specific namespace remap:**
 
 - A node has name `talker`
 - A user gives the rules `talker:__ns:=/foo` then `__ns:=/bar`
 - talker's namespace is `/foo` because that rule was given first
+
+**Example of node and topic/service remapping order:**
+
+- A node uses a name `/talker` and a topic/service name `/bar`
+- A user gives the rules `talker:__node:=foo` then `foo:bar:=/asdf`
+- The final fully qualified name is `/foo`, and the final topic/service name is `/asdf` because the node name remap rule is matched before other remappings are applied
 
 ### Applications of the syntax
 
--- a/articles/141_static_remapping.md
+++ b/articles/141_static_remapping.md
@@ -255,6 +255,7 @@ Example rules are:
 - `foo:=bar`
 - `/foo/bar:=fiz/buzz`
 - `nodename:~/foo:=foo`
+- `/ns1/nodename:foo:=bar`
 - `**/foo:=\1/bar`
 
 #### Match Part of a Rule
@@ -265,7 +266,7 @@ The match part of a rule uses these operators:
 - `**` matches zero or more tokens delimited by slashes
 - `rosservice://` prefixed to the match makes the rule apply to only services
 - `rostopic://` prefixed to the match makes the rule apply to only topics
-- `nodename:` prefixed to the match makes it apply only to a node with that name
+- `nodename:` prefixed to the match makes it apply only to a node with that name, which could be a FQN
 
 The operators `*` and `**` are similar to the globbing behavior in bash.
 `**` behaves similar to its use in bash>=4.0 with the globstar option set.
@@ -324,8 +325,8 @@ The replacement must be a single token which will become the node's new name.
 
 Remapping rules are applied in the following order:
 
-1. Node name remapping
 1. Namespace remapping
+1. Node name remapping
 1. All other rules
 
 Within each category, the rules are applied in the order in which the user gave them.
@@ -339,8 +340,8 @@ Within each category, the rules are applied in the order in which the user gave 
 **Example of node/namespace remapping order:**
 
 - A node has name `talker`
-- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`
-- The final namespace is the default `/` because the node name remap is applied before the namespace remap
+- A user gives the rules `talker:__ns:=/my_namespace` then `/talker:__node:=foo`
+- The final node name is the default `talker` because the namespace remap is applied before the node name remap
 
 **Example of a default and node specific namespace remap:**
 
--- a/articles/141_static_remapping.md
+++ b/articles/141_static_remapping.md
@@ -338,8 +338,8 @@ Within each category, the rules are applied in the order in which the user gave 
 **Example of node/namespace remapping order:**
 
 - A node has name `talker`
-- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`
-- The final node name is `foo` because the namespace and node name remap are applied without affecting each other
+- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`, or vice versa
+- The final fully qualified name is `/my_namespace/foo` because the namespace and node name remap are applied without affecting each other
 
 **Example of a default and node specific namespace remap:**
 


iuhilnehc-ynos:
I have no idea if we should apply for using the FQN prefix and remapping order. To push this commit is mainly for discussion.

articles/141_static_remapping.md
```diff
@@ -339,8 +340,8 @@ Within each category, the rules are applied in the order in which the user gave
 **Example of node/namespace remapping order:**
 
 - A node has name `talker`
-- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`
-- The final namespace is the default `/` because the node name remap is applied before the namespace remap
+- A user gives the rules `talker:__ns:=/my_namespace` then `/talker:__node:=foo`
+- The final node name is the default `talker` because the namespace remap is applied before the node name remap
```
fujitatomoya: what if user wants to remap `talker` into namespace `A` and `B`? I believe this is really common use case. I am not so convinced to change order. I think `Node name -> Namespace` would be understandable and usable for user.

iuhilnehc-ynos: Thanks for your response. @fujitatomoya 

> what if user wants to remap talker into namespace A and B?

Sorry, I don't catch it. How can we remap one node `talker` into two different namespaces `A` and `B`?

> Node name -> Namespace would be understandable

It has a limitation.
Imageing there are two nodes `publisher_node` and `subscriber_node` in an application.
/publisher_node
/subscriber_node

I want to remap these two nodes as following (https://github.com/ros2/rcl/issues/806),
`/publisher_node` -----> '/ns1/aaa'
`/subscriber_node` -----> '/ns2/aaa'

iuhilnehc-ynos: Maybe this case is suited for the real world.
/left_camera        -----> /left/camera
/right_camera      -----> /right/camera

It seems we can not remap them with current remapping order.

fujitatomoya: > How can we remap one node talker into two different namespaces A and B?

bad example, ignore it... sorry😢

> It seems we can not remap them with current remapping order.
> /left_camera -----> /left/camera
> /right_camera -----> /right/camera

I don't see this is because of remapping order, but implementation. I think we should have some kinda justification to change order to support two nodes with the same name but in different namespaces.

iuhilnehc-ynos: > I don't see this is because of remapping order, but implementation.

That's right, users can set two nodes with Node("camera", "/left") and Node("camera", "/right") in implementation.
but we can't ask users to do this, if somebody just wants to use remapping, there is a limitation.

>  some kinda justification 

Actually, the sample about the 'camera' above is the reason why I open this issue.

I found no discussion about why ros2 use this kind of remapping order(node name -> namespace), after checking https://github.com/ros2/rcl/pull/217 and https://github.com/ros2/design/pull/181, it seems that [ros2](https://github.com/ros2/rcl/blob/6f39edf904a6a041099c5deaf313ff975da4c3c8/rcl/src/rcl/node.c#L229-L254) follow the same implementation order to [ros1](https://github.com/ros/ros_comm/blob/84d413fa5689891d871680c18f217a6a16d12731/clients/roscpp/src/libros/this_node.cpp#L122-L134). (If I remember correctly, ros1 doesn't have the remapping order for __name and __ns because it just use one node in an application. )

clalancette:
@sloretz Since you wrote the original design document here, could you maybe weigh in on this design change?

articles/141_static_remapping.md
```diff
@@ -324,8 +325,8 @@ The replacement must be a single token which will become the node's new name.
 
 Remapping rules are applied in the following order:
 
-1. Node name remapping
 1. Namespace remapping
+1. Node name remapping
```
ivanpauno: Changing the order doesn't seem like a good idea.

What needs to happen to resolve https://github.com/ros2/rcl/issues/296 is to allow the prefix to be a fully qualified name, instead of only a name.

In that way, if you have both `/asd/talker` and `/bsd/talker` in a node, you can do:

```
$ ros2 run package exec_name \
    --ros-args \
    -r /asd/talker:__node:=talker1 \
    -r /bsd/talker:__node:=talker2 \
    -r /asd/talker1:__ns:=/ns1 \   # or "-r talker1:__ns:=/ns1"
    -r /asd/talker2:__ns:=/ns2     # or "-r talker2:__ns:=/ns2"
```

+ in this way backwards compatibility isn't broken.

iuhilnehc-ynos: Thank you for your response.

I know what you said, and also know that backward compatibility is important.
but there exists a limitation(actually, the example you provided is not what I want), it makes me feel uncomfortable.
Do you think we can remap the two node name in 'ros2 run examples_rclcpp_minimal_composition composition_composed' from ('publisher_node', 'subscriber_node') to ('/publisher/test_node', '/subscriber/test_node')?

fujitatomoya: > What needs to happen to resolve ros2/rcl#296 is to allow the prefix to be a fully qualified name, instead of only a name.

this is true, and requirement is covered.

> but there exists a limitation(actually, the example you provided is not what I want)

sorry, could you describe the limitation stays after?

iuhilnehc-ynos: The limitation is `remap the two node name in 'ros2 run examples_rclcpp_minimal_composition composition_composed' from ('publisher_node', 'subscriber_node') to ('/publisher/test_node', '/subscriber/test_node')`

fujitatomoya: okay if the node names are the same, confliction comes in 1st place even with https://github.com/ros2/rcl/issues/296, thanks.

ivanpauno:
With this proposal, how can the following example be solved:

remap `/ns1/node` `/ns2/node` to `/new_node1` `/new_node2` respectively.

?

This isn't currently possible, but it wasn't uncommon in ROS 1 I think.

articles/141_static_remapping.md
ivanpauno: > but there exists a limitation

I see, yeah my proposed solution doesn't solve the limitation pointed out here.
I think this proposal still has a limitation (see [here](https://github.com/ros2/design/pull/299#issuecomment-709330296)), and that is a problem I think users will more frequently run into.

iuhilnehc-ynos:
> remap /ns1/node /ns2/node to /new_node1 /new_node2 respectively.

That's interesting. I think this case should be also supported.

```
$ ros2 run package exec_name \
    --ros-args \
    -r /ns1/node:__ns:=/ \              # the FQN must use original value
    -r /ns1/node:__node:=new_node1 \    # the FQN must use original value
    -r /ns2/node:__ns:=/ \
    -r /ns2/node:__node:=new_node2
```

and `remap publisher_node subscriber_node to /publisher/test_node /subscriber/test_node`

```
$ ros2 run package exec_name \
    --ros-args \
    -r /publisher_node:__ns:=/publisher \
    -r /publisher_node:__node:=test_node \
    -r /subscriber_node:__ns:=/subscriber \
    -r /subscriber_node:__node:=test_node
```

----

after using the original value for prefix, it also breaks the backward compatibility.
ex. /ns1/node -> /new_node1
- before

```
$ ros2 run package exec_name \
    --ros-args \
    -r new_node1:__ns:=/ \
    -r node:__node:=new_node1 \
```

- after

```
$ ros2 run package exec_name \
    --ros-args \
    -r node:__ns:=/ \
    -r node:__node:=new_node1 \
```

Edit: If you agree with the above information, I'll update the design, otherwise I think it's time to close this issue.

iuhilnehc-ynos:
It means there is no order for __ns and __name.
Update source codes by moving https://github.com/ros2/rcl/blob/1bba348292f2c4cf937e5034ae561c8da5dd639a/rcl/src/rcl/node.c#L225-L227 after https://github.com/ros2/rcl/blob/1bba348292f2c4cf937e5034ae561c8da5dd639a/rcl/src/rcl/node.c#L241

ivanpauno:
> Edit: If you agree with the above information, I'll update the design, otherwise I think it's time to close this issue.

I think that a proposal that handle both cases above correctly would be acceptable.

iuhilnehc-ynos:
In this PR, I would not add the FQN prefix in it because it's different from the remapping order. If it's necessary, I'll create a new PR for it.

@ivanpauno 
I have updated the design, please review it.

ivanpauno:
@iuhilnehc-ynos I think the proposal is reasonable.

It would be great if you can edit the [post description](https://github.com/ros2/design/pull/299#issue-490654094) and include a motivation of the changes and examples of what the change enables (so future reviewers can understand the rationale of the change).

> In this PR, I would not add the FQN prefix in it because it's different from the remapping order. If it's necessary, I'll create a new PR for it.

Combined with allowing FQN in prefixes, this allow even more use cases.
I don't think that both changes need to be done at the same time, but it would be good to explain how it works in the description.

articles/141_static_remapping.md
```diff
@@ -340,7 +339,7 @@ Within each category, the rules are applied in the order in which the user gave
 
 - A node has name `talker`
 - A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`
-- The final namespace is the default `/` because the node name remap is applied before the namespace remap
+- The final node name is `foo` because the namespace and node name remap are applied without affecting each other
```
sloretz: What's the final fully qualified name? Current design says it would be `/foo` because the node name remapping happened first. Is this PR proposing the final output would be `/my_namespace/foo`?

iuhilnehc-ynos: > What's the final fully qualified name?

It's `/my_namespace/foo` which you had already answered. I should make the document clear, I'll update it.

iuhilnehc-ynos:
@ivanpauno 

> I don't think that both changes need to be done at the same time, but it would be good to explain how it works in the description.

Thanks for your understanding. I have updated the description.

articles/141_static_remapping.md
```diff
@@ -324,8 +324,7 @@ The replacement must be a single token which will become the node's new name.
 
 Remapping rules are applied in the following order:
 
-1. Node name remapping
-1. Namespace remapping
+1. Namespace or Node name remapping
```
ivanpauno, line 327: so now name are namespace rules are applied in the order they appear, right?

It would be great to make that clear here

ivanpauno: > order they appear

This is not the case, they order doesn't matter.
Their prefix is always the node name before remapping, maybe that can be made more clear.

sloretz: > This is not the case, they order doesn't matter.

They're still ordered within a rule, right? If I give `talker:__node:=foo` and `talker:__node:=bar`, then is the final name is `foo`?

articles/141_static_remapping.md
```diff
@@ -339,8 +338,8 @@ Within each category, the rules are applied in the order in which the user gave
 **Example of node/namespace remapping order:**
 
 - A node has name `talker`
-- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`
-- The final namespace is the default `/` because the node name remap is applied before the namespace remap
+- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`, or vice versa
+- The final fully qualified name is `/my_namespace/foo` because the namespace and node name remap are applied without affecting each other
```
sloretz: >  the namespace and node name remap are applied without affecting each other

I would change this wording to: `the namespace and node name remap rules both match against the orignal node name before remappings are applied`.

sloretz:
Implementation considerations - it seems possible to implement this by passing the original `name` arg to `rcl_remap_node_namespace()` - which could be done by calling `rcl_remap_node_namespace()` before `rcl_remap_node_name()` so the `name` variable is unchanged.

https://github.com/ros2/rcl/blob/e9c87e279bf7509c940b87fd4d6c5dff301bc735/rcl/src/rcl/node.c#L220-L240

sloretz (changes requested):
The idea itself seems good to me. I couldn't think of any use cases it breaks (doesn't seem to break any in the design doc), and it seems to enable a new use case of remapping nodes to have the same name but different FQN.

It looks a bit weird when swapping node names,

```
 -r alice:__node:=bob
 -r bob:__node:=alice
 -r alice:__ns:=/land
 -r bob:__ns:=/wonder
```

Before this change results in `/wonder/bob` and `/land/alice` while with this proposal it results in `/wonder/alice` and `/land/bob`.

But then it doesn't look so weird if I order them this way

```
 -r alice:__node:=bob
 -r alice:__ns:=/land
 -r bob:__node:=alice
 -r bob:__ns:=/wonder
```

I think it should be made clearer in the `Match Part of a Rule` section that whether `nodename:` refers to the pre-remaping name or the post-remapping name depends on the rule. For `__ns` and `__node` it's the pre-remapping node name that gets matched against, while for topic and service rules it's the post-remapping node name that matters.

articles/141_static_remapping.md
iuhilnehc-ynos: @ivanpauno 

>> name and namespace rules are applied in the order they appear, right?

> they order doesn't matter

the order of name and namespace rules doesn't matter.
I think users like to understand some rules by referring to an example description.
- A user gives the rules `talker:__ns:=/my_namespace` then `talker:__node:=foo`, or vice versa

@sloretz 
> They're still ordered within a rule, right?

I supposed you mean the order of multiple items about __node or __ns, the answer is yes.

> If I give talker:__node:=foo and talker:__node:=bar, then is the final name is foo?

Yes.
Actually, the example description already showed messages,
- talker's namespace is `/foo` because that rule was given first
though the above example uses `talker:__ns:=/foo` and `__ns:=/bar`

About this kind of information, I'll not update it in this PR.

articles/141_static_remapping.md
iuhilnehc-ynos: Thank you. I'll update it later.

iuhilnehc-ynos:
> But then it doesn't look so weird if I order them this way

Thanks for your example.

> I think it should be made clearer in the Match Part of a Rule section that whether nodename: refers to the pre-remaping name or the post-remapping name depends on the rule.

But there already exist information about this pre-remapping and post-remapping in `Order of Applying Remapping Rules`
`Within each category, the rules are applied in the order in which the user gave them.`
If adding this kind of information in the `Match Part of a Rule section`, it seems a little duplicated.

sloretz:
> But there already exist information about this pre-remapping and post-remapping in Order of Applying Remapping Rules
Within each category, the rules are applied in the order in which the user gave them.
If adding this kind of information in the Match Part of a Rule section, it seems a little duplicated.

If not the match rule section, how about another example showing node name remapping and topic name remapping together? I think it'd be helpful because this case looks a little confusing:

```
 -r alice:__node:=bob
 -r alice:foo:=baz
 -r bob:foo:=bar
```

ivanpauno:
Hi @iuhilnehc-ynos, here some feedback from the ros2 team meeting:

- Was this proposal the result of a real use case? i.e. were you actually trying to remap `/node1`, `/node2` to `/ns1/node` and `/ns2/node` in a real application?
- To avoid a breaking change (that in many cases will be a silent breakage), could this option been enabled with a prefix? e.g.:
```
ros2 run package_name exec_name --ros-args --inorder-name-ns \
    -r publisher_node:__ns:=/publisher_different_ns \
    -r publisher_node:__node:=test_same_node \
    -r subscriber_node:__ns:=/subscriber_different_ns \
    -r subscriber_node:__node:=test_same_node
```
In that way, we avoid breaking previous code (this use case seems to be not too common, so making it more verbose wouldn't be a big issue).

Bikeshedding: `--inorder-name-ns` could have a better name.

iuhilnehc-ynos:
> Was this proposal the result of a real use case? i.e. were you actually trying to remap /node1, /node2 to /ns1/node and /ns2/node in a real application?

No. These cases can be avoided in users' applications.

> To avoid a breaking change (that in many cases will be a silent breakage), could this option been enabled with a prefix?

Yes, adding this option can avoid the breaking change, but it seems not kindly for users because there are two kinds of rules.

If you don't mind, I want to close this issue.

fujitatomoya:
@iuhilnehc-ynos 
CC: @ivanpauno 

- `--inorder-name-ns` sounds okay to me.
- maybe there would be some corner cases, but it would be really rare.
- can avoid limitation with adding `--inorder-name-ns` (at least user can what they want to do.)

I do not have any strong justification to push this design either.

ivanpauno:
> Yes, adding this option can avoid the breaking change, but it seems not kindly for users because there are two kinds of rules.

Agreed, but "silent breakage" (the one that doesn't make your build fail and makes your program behave different than before) is a big issue for existing users.

> If you don't mind, I want to close this issue.

Sounds good to me.
If the current order of remapping rules is a problem for some use cases, adding the `--inorder-name-ns` option would be the best path forward.
It's not super nice to have two different orders for remapping rules, but this is a corner case that most users won't run into.

If you have any other alternatives to avoid breaking pre-existing code, that also would be acceptable.

Thanks for the proposal @iuhilnehc-ynos !

ivanpauno:
Based on the previous comments, I'm going to close this.
If there is an intention to push a non-breaking alternative of the proposal forward feel free to reopen.

PR #296, opened 2020-07-29
QoS profile files proposal

ivanpauno:
Fixes https://github.com/ros2/design/issues/280.

I have written down the ideas I have in a document, and I have included the points that were raised in previous discussions.

PR #300, opened 2020-09-23
Add design document discussing QoS reconfigurability

ivanpauno:
Alternative to https://github.com/ros2/design/pull/296.
Based on feedback, the proposal leverages ROS parameters instead of proposing an ad-hoc file format.

### DIFF
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -131,7 +131,7 @@ This perfectly matches the goal of the proposal of not making QoS settings recon
 Up to ROS 2 Foxy, read-only parameters generate a parameter event with its initial value when declared.
 If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will significantly increase.
 
-To avoid this issue, a way to opt-out parameter events for a given parameter could be added (should that be the default for read-only parameters??).
+To avoid this issue, a way to opt-out of parameter events for a given parameter could be added (should that be the default for read-only parameters??).
 
 ### Parameter callbacks
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -324,7 +324,7 @@ node->create_publisher(
         [](const QoSProfile qos) -> bool {
             return qos.is_keep_all() || qos.history_depth() > 10u;  // constrains involving more than one QoS in callbacks
         },
-        "my_id",  // id to disambiguate entities in same topic
+        "my_id"  // id to disambiguate entities in same topic
     });
 
 // nothing is reconfigurable
@@ -334,21 +334,32 @@ node->create_publisher(
 
 // allow reconfiguring durability
 node->create_publisher(
-    "another_topic",
+    "other_topic",
     user_qos_provided_in_code_3,
     QosOverridingOptions{QosPolicyKind::Durability});
+
+// "default" policies are reconfigurable
+node->create_publisher(
+    "another_topic",
+    user_qos_provided_in_code_4,
+    QosOverridingOptions{true});
 ```
 
-The above API mentions having some default set of QoS settings that are reconfigurable. 
-The intent is to provide a less verbose option, where the user can opt-in to reconfigurability of these default QoS settings with a single flag.
+The intent of being able to opt-in a set of "default" policies is to make the API easier to use and less verbose.
 
-Proposed defaults:
+All policies could be included in the "set" of default reconfigurable policies, with some exceptions:
+- Liviliness kind: It requires special care by the node's author.
+- Lifespan: It only applies to publishers, thus it shouldn't be declared for subscribers.
+  It also only applies for "transient local" durability, so it shouldn't be reconfigurable if durability isn't.
+
+We could also have a restricted list of default policies to be declared in parameters:
 - Reliability
 - History kind
 - History depth
-- Durability
 
-TBD: Do we want defaults?
+which are the ones that usually require reconfigurability.
+
+TBD: The defaults could be different to the ones proposed above.
 
 ### Reusing profiles
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -245,13 +245,13 @@ node->create_publisher(
       publisher_id1:  # {entity_type}_{id}
         'my/fully/qualified/topic/name/here': 
           reliability: reliable
-            history_depth: 100
-             history: keep_last
+          history_depth: 100
+          history: keep_last
       publisher_id2:  # {entity_type}_{id}
         'my/fully/qualified/topic/name/here': 
           reliability: best_effort
-            history_depth: 100
-            history: keep_last
+          history_depth: 100
+          history: keep_last
 ```
 
 ### Lowering the amount of declared parameters
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -326,7 +326,7 @@ node->create_publisher(
 ```
 
 The above API mentions having some "defaults". 
-The intend of that is to provide a less verbose option, in which case the user can opt-in qos reconfigurability by only opting-in a flag.
+The intent is to provide a less verbose option, where the user can opt-in to reconfigurability of these default QoS settings with a single flag.
 
 Proposed defaults:
 - Reliability
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -181,7 +181,7 @@ Parameters URIs currently look like:
 rosparam://my/full/namespace/node_name/param_group.nested_param_group.param_name
 ```
 
-For the following qos policy:
+For the following QoS policy:
 - Full node name: `/asd/bsd/my_node`
 - Entity: `publisher`
 - Topic name: `/foo/bar/topic_name`
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -118,6 +118,9 @@ node->create_publisher(
 
 The user provided QoS callback will be internally used as a [parameters callback](https://github.com/ros2/rclcpp/blob/3a4ac0ca2093d12035070443692798b0c9f9da3a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L183), but we provide a more user-friendly interface.
 
+Note: The examples through this document are using `QosOverridingOptions` as a new argument in `create_publisher`/`create_subscription`.
+Instead of that, `QosOverridingOptions` will likely be integrated in the existing `PublisherOptions`/`SubscriptionOptions` structures.
+
 ## Analysis of parameter API features and limitations
 
 ### Read-only parameters
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -129,7 +129,7 @@ This perfectly matches the goal of the proposal of not making QoS settings recon
 ### Parameter events
 
 Up to ROS 2 Foxy, read-only parameters generate a parameter event with its initial value when declared.
-If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will be highly increased.
+If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will significantly increase.
 
 To avoid this issue, a way to opt-out parameter events for a given parameter could be added (should that be the default for read-only parameters??).
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -57,7 +57,7 @@ This proposal was written with the following requirements in mind:
   - Rationale: Though you can currently pass a custom QoS profiles to services or actions, configuring them is quite fragile and some of the policies cannot be modified (e.g. best effort services don't work).
   Until a better QoS profile for services and actions is designed, this feature won't be exposed.
 - Shall only allow changing QoS at startup time, and not during runtime by default.
-  - Allowing runtime reconfiguration of QoS is a complex topic, and out of the goal of this proposal.
+  - Rationale: Allowing runtime reconfiguration of QoS is a complex topic, and purposefully left out of scope in this proposal.
 - Shall have a way of reusing the same QoS profile.
   - Rationale: Not having a way of setting the same QoS profile to different entities will make configuring entities which profiles must match (or be compatible) error prone.
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -66,7 +66,7 @@ This proposal was written with the following requirements in mind:
 A previous proposal (see PR [#296](https://github.com/ros2/design/pull/296)) suggested using a new file format for reconfiguring QoS policies.
 Some concerns were raised by the community:
 1. It adds a different mechanism to the ones that already exist.
-    Newcomers has yet to learn another thing.
+    Newcomers have to learn yet another thing.
 1. QoS settings are not conceptually different to other parameters, e.g.: publishing rate.
 1. It adds a new mechanism to configure QoS profiles.
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -262,7 +262,7 @@ The two mechanism are compatible with what it was said before and are not mutual
 #### Hidden parameters
 
 Currently, ROS 2 has the concept of hidden topics and services.
-Those aren't showed by the cli tools, except explicitly requested:
+Those aren't shown by the cli tools, except when explicitly requested:
 
 ```bash
 ros2 topic list  # Will not list hidden topics.
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -365,7 +365,7 @@ We could, for example, leverage yaml anchors to allow this:
           history_depth: 100  # override the history depth
 ```
 
-### Extending the ROS qos profile with a rmw vendor specific payload
+### Extending the ROS QoS profile with a rmw vendor specific payload
 
 TBD
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -25,7 +25,7 @@ Provide an standard way to configure QoS settings of different entities in the R
 
 ## Motivation
 
-Up to ROS 2 Foxy, there's no standard mechanism to reconfigure QoS settings at startup.
+Up to ROS 2 Foxy, there's no standard mechanism to override QoS settings when creating a node.
 It's quite common in the ROS 2 community to reuse nodes that other people have written and in a lot of use cases allowing QoS profiles to be changed make sense.
 
 Up to now, different workarounds have been used in order to provide this reconfigurability:
@@ -123,14 +123,17 @@ The user provided QoS callback will be internally used as a [parameters callback
 ### Read-only parameters
 
 ROS 2 currently provides read only parameters.
-They can be modified at startup time by providing an override (e.g. `--ros-args -p <param_name> <param_value>`), but they cannot be dynamically changed later.
+They can be modified when constructing a node by providing overrides (e.g. `--ros-args -p <param_name> <param_value>`), but they cannot be dynamically changed after the node was constructed.
 
 This perfectly matches the goal of the proposal of not making QoS settings reconfigurable during runtime.
+
+Note: parameter overrides can be constructed dynamically and then used to create a node, but it's not possible to modify the parameter after the node construction.
+This allows, for example, to pass parameter overrides when dynamically adding a node to a component container.
 
 ### Parameter events
 
 Up to ROS 2 Foxy, read-only parameters generate a parameter event with its initial value when declared.
-If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will significantly increase.
+If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will significantly increase, as the declaration of each parameter generates an event.
 
 To avoid this issue, a way to opt-out of parameter events for a given parameter could be added (should that be the default for read-only parameters??).
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -91,7 +91,7 @@ node->create_publisher(
   QosOverridingOptions{true});  // allow_reconfigurable all qos
 ```
 
-That will automatically declare the parameters for reconfiguring the qos policies that can be overridden in the following way:
+That will automatically declare the parameters for reconfiguring the QoS policies that can be overridden in the following way:
 
 ```yaml
 /my/ns/node_name:
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -254,6 +254,9 @@ node->create_publisher(
 
 ### Lowering the amount of declared parameters
 
+This two subsections describe some optional mechanism to lower the amount of declared parameters.
+The two mechanism are compatible with what it was said before and are not mutually exclusive.
+
 #### Hidden parameters
 
 Currently, ROS 2 has the concept of hidden topics and services.
@@ -265,7 +268,7 @@ ros2 topic --include-hidden-topics list  # Will list hidden topics.
 ```
 
 Similarly, hidden parameters could be added.
-All the parameters used for overriding QoS could be declared as hidden, thus avoiding noise for the user in commands like `ros2 param list`.
+All the parameters used for overriding QoS could be declared as hidden, thus avoiding "noise" in commands like `ros2 param list`.
 
 #### Only declaring some parameters
 
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -1,0 +1,379 @@
+---
+layout: default
+title: External configurability of QoS policies.
+permalink: articles/qos_configurability.html
+abstract:
+  This article describes a mechanism to allow reconfiguration of QoS settings at startup time.
+author: '[Ivan Santiago Paunovic](https://github.com/ivanpauno)'
+published: true
+---
+
+- This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## Summary
+
+Provide an standard way to configure QoS settings of different entities in the ROS graph at startup time, overriding the settings specified in code.
+
+## Motivation
+
+Up to ROS 2 Foxy, there's no standard mechanism to reconfigure QoS settings at startup.
+It's quite common in the ROS 2 community to reuse nodes that other people have written and in a lot of use cases allowing QoS profiles to be changed make sense.
+
+Up to now, different workarounds have been used in order to provide this reconfigurability:
+
+- Command line arguments.
+- ROS parameters to set particular policies.
+- Combining parameters, the system default QoS profile, and vendor specific QoS profiles.
+
+Some real example applying this mechanisms:
+
+- rosbag2 has an ad-hoc mechanism to override QoS profiles when recording or doing playback ([docs](https://index.ros.org/doc/ros2/Tutorials/Ros2bag/Overriding-QoS-Policies-For-Recording-And-Playback/)).
+- Image pipeline nodes use some parameters to allow changing some of the policies ([PR](https://github.com/ros-perception/image_pipeline/pull/521)).
+- Ouster drivers also use some parameters to allow changing some policies ([PR](https://github.com/ros-drivers/ros2_ouster_drivers/pull/26)).
+- Gazebo ROS packages allows configuring QoS profiles of the plugins in the SDF file ([issue](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1079)).
+
+Not providing an standard mechanism for reconfiguring QoS result in a lot of slightly different interfaces to do the same thing, thus making painful to end users to figure out how to reconfigure QoS in different nodes in their system.
+
+## Design goals
+
+This proposal was written with the following requirements in mind:
+
+- Shall be opt-in on a per entity basis.
+  - Rationale: Allowing to change all QoS policies could break a contract the node's implementer assumed.
+  For example, the node's implementer may have assumed that the data was durable and not volatile.
+  Allowing the node's author to only accept some reconfigurations will avoid user errors.
+- Shall allow the developer to accept/reject changes to a QoS profile with a validation callback.
+  - See rationale above.
+- Shall only apply to publisher/subscriber QoS.
+  - Rationale: Though you can currently pass a custom QoS profiles to services or actions, configuring them is quite fragile and some of the policies cannot be modified (e.g. best effort services don't work).
+  Until a better QoS profile for services and actions is designed, this feature won't be exposed.
+- Shall only allow changing QoS at startup time, and not during runtime by default.
+  - Allowing runtime reconfiguration of QoS is a complex topic, and out of the goal of this proposal.
+- Shall have a way of reusing the same QoS profile.
+  - Rationale: Not having a way of setting the same QoS profile to different entities will make configuring entities which profiles must match (or be compatible) error prone.
+
+## Adding a new mechanism vs using parameters
+
+A previous proposal (see PR [#296](https://github.com/ros2/design/pull/296)) suggested using a new file format for reconfiguring QoS policies.
+Some concerns were raised by the community:
+1. It adds a different mechanism to the ones that already exist.
+    Newcomers has yet to learn another thing.
+1. QoS settings are not conceptually different to other parameters, e.g.: publishing rate.
+1. It adds a new mechanism to configure QoS profiles.
+
+Based on that feedback, this PR will explore a design based on parameters.
+
+## Introduction
+
+Current code will not allow QoS reconfigurability if not changed:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast{10});
+```
+
+To make reconfigurability easy, only a flag is needed to automatically create the parameters:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast{10},
+  QosOverridingOptions{true});  // allow_reconfigurable all qos
+```
+
+That will automatically declare the parameters for reconfiguring the qos policies that can be overridden in the following way:
+
+```yaml
+/my/ns/node_name:
+  ros__parameters:
+    qos_profiles:  # group for all the qos settings
+      publisher:
+        'my/fully/qualified/topic/name/here': 
+          reliability: reliable
+          history_depth: 100
+          history: keep_last
+```
+
+A callback to accept/reject the configured profile can also be specified:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast(10),
+  [] (const QoSProfile & qos) -> bool {
+      return qos.is_reliable();  // only accept reliable communication, we should add getters to rclcpp::QoSProfile class for this.
+  });
+```
+
+The user provided QoS callback will be internally used as a [parameters callback](https://github.com/ros2/rclcpp/blob/3a4ac0ca2093d12035070443692798b0c9f9da3a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L183), but it has a more friendly interface for the user.
+
+## Analysis of parameter API features and limitations
+
+### Read-only parameters
+
+ROS 2 currently provides read only parameters.
+They can be modified at startup time by providing an override (e.g. `--ros-args -p <param_name> <param_value>`), but they cannot be dynamically changed later.
+
+This perfectly matches the goal of the proposal of not making qos settings reconfigurable.
+
+### Parameter events
+
+Up to ROS 2 Foxy, read-only parameters generate a parameter event with its initial value when declared.
+If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will be highly increased.
+
+To avoid this issue, a way to opt-out parameter events for a given parameter could be added (should that be the default for read-only parameters??).
+
+### Parameter callbacks
+
+If we want the following example to work:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast(10),
+  [] (const QoSProfile & qos) -> bool{
+      return qos.is_reliable();
+  });
+```
+
+We could completely ignore parameters [on set callbacks](https://github.com/ros2/rclcpp/blob/3defa8fc9d7410bd833ecd95b305ac94bb9b627a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L180-L192), the publisher creation code would do something like:
+
+```cpp
+rcl_interfaces::msg::ParameterDescription policy_x_description{};
+policy_x_description.read_only = true;
+policy_x_description.description = "<string with a description of the policy, including node and topic name>";
+auto policy_x_value = node->declare_parameter(
+    "<name of the parameter>", qos_provided_in_code.<getter_for_policy_x>(), policy_x_description);
+// Repeat for all policies
+rclcpp::QoS final_qos();
+// construct `final_qos` from the parameter values read above
+if (!user_provided_callback(final_qos) {
+    // qos profile rejected, throw error here
+}
+```
+
+In that way, we avoid installing a parameter callback that will be triggered when any parameter is modified.
+
+Note: this seems to be what you generally want in the case of read-only parameters.
+Note 2: We could also have a `declare_parameters_atomically` method and use:
+  ```cpp
+  auto callback_handle = node->add_on_set_parameters_callback(wrapper_for_qos_callback(user_provided_callback));
+  node->declare_parameters_atomically(...);  // names and defaults for all parameters given here
+  // callback_handle goes out of scope, the callback is auto-removed.
+  ```
+  `declare_parameters_atomically` would be needed, if not you cannot have conditions like `history == keep_all or history_depth > 10`.
+
+### Parameters URIs
+
+Parameters URIs currently look like:
+
+```
+rosparam://my/full/namespace/node_name/param_group.nested_param_group.param_name
+```
+
+For the following qos policy:
+- Full node name: `/asd/bsd/my_node`
+- Entity: `publisher`
+- Topic name: `/foo/bar/topic_name`
+- Policy: `reliability`
+
+The URI of the automatically generated parameter would be:
+
+```
+rosparam://asd/bsd/my_node/qos_profiles.publisher./foo/bar/topic_name.reliability
+```
+
+The `/` in one of the subgroups looks odd, but it's currently accepted.
+
+If in the future the addressing changes described in [#241](https://github.com/ros2/design/pull/241) are implemented, the URI would be:
+
+```
+rosparam://asd.bsd.my_node/qos_profiles/publisher/foo/bar/topic_name/reliability
+```
+
+but in both cases the previous parameters file example wouldn't change.
+
+## Caveats
+
+### Creating more than one publisher/subscription in the same topic with different QoS
+
+If only the topic name is used for the parameter name, it would be impossible to create two publishers with different QoS in the same topic.
+
+That could be solved by adding an optional extra identifier:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast(10),
+  QosOverridingOptions {
+    [] (const QoSProfile & qos) -> bool {
+        return qos.is_reliable();  // only accept reliable communication
+    },
+    "id1"
+  });
+node->create_publisher(
+  "chatter",
+  KeepLast(10),
+  QosOverridingOptions { 
+    [] (const QoSProfile & qos) -> bool {
+      return !qos.is_reliable();  // only accept best effort communication
+    },
+    "id2"
+  });
+```
+
+```yaml
+/my/ns/node_name:
+  ros__parameters:
+    qos_profiles:
+      publisher_id1:  # {entity_type}_{id}
+        'my/fully/qualified/topic/name/here': 
+          reliability: reliable
+            history_depth: 100
+             history: keep_last
+      publisher_id2:  # {entity_type}_{id}
+        'my/fully/qualified/topic/name/here': 
+          reliability: best_effort
+            history_depth: 100
+            history: keep_last
+```
+
+### Lowering the amount of declared parameters
+
+#### Hidden parameters
+
+Currently, ROS 2 has the concept of hidden topics and services.
+Those aren't listed in the cli tool commands, except when the user explicitly asks them to be included.
+
+Similarly, hidden parameters could be added.
+That will avoid showing a big list of all policies.
+
+#### Only declaring some parameters
+
+Instead of hidden parameters that correspond to the QoS policies, we could only declare parameters for the policies that the user want to be reconfigurable.
+That will lower the amount of declared parameters, but make publisher/subscription creation a bit more verbose.
+
+Example API:
+
+```cpp
+enum class QosPolicyKind {
+  Reliability,
+  ...,
+};
+
+/// Options that are passed in subscription/publisher constructor
+class QosOverridingOptions
+{
+    /// Constructor, allowing to declare all the "default" parameters. If `false` is passed, reconfiguring QoS is not allowed.
+    explicit QosOverridingOptions(bool declare_default_parameters, std::string id = "");
+    /// Implicit constructor, a parameter for each policy provided in the initializer list will be declared.
+    QosOverridingOptions(std::initializer_list<QosPolicyKind>, std::string id = "");
+    /// Same with a callback for accepting/rejecting the qos profile.
+    QosOverridingOptions(std::initializer_list<QosPolicyKind>, QosCallback callback, std::string id = "");
+    /// Declares the "default" parameters and pass a callback.
+    QosOverridingOptions(QosCallback callback, std::string id = "");
+};
+```
+
+```cpp
+// New method in Node class
+template<
+typename MessageT,
+typename AllocatorT = std::allocator<void>,
+typename PublisherT = rclcpp::Publisher<MessageT, AllocatorT>>
+std::shared_ptr<PublisherT>
+create_publisher(
+  const std::string & topic_name,
+  const rclcpp::QoS & qos,
+  const rclcpp::QosOverridingOptions & qos_options = QosOverridingOptions{false},
+  const PublisherOptionsWithAllocator<AllocatorT> & options = PublisherOptionsWithAllocator<AllocatorT>()
+);
+```
+
+```cpp
+// example usage
+node->create_publisher(
+    "my_topic",
+    user_qos_provided_in_code_1,
+    rclcpp::KeepLast(10),
+    {  // implicit, to lower verbosity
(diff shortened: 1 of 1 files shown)


ivanpauno:
@jacobperron this is ready for a first pass.

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+Up to ROS 2 Foxy, there's no standard mechanism to reconfigure QoS settings at startup.
+It's quite common in the ROS 2 community to reuse nodes that other people have written and in a lot of use cases allowing QoS profiles to be changed make sense.
+
+Up to now, different workarounds have been used in order to provide this reconfigurability:
+
+- Command line arguments.
+- ROS parameters to set particular policies.
+- Combining parameters, the system default QoS profile, and vendor specific QoS profiles.
+
+Some real example applying this mechanisms:
```
jacobperron:
```suggestion
Some real example applying this mechanisms:
```
```suggestion
Here are some examples applying these mechanisms:
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+- rosbag2 has an ad-hoc mechanism to override QoS profiles when recording or doing playback ([docs](https://index.ros.org/doc/ros2/Tutorials/Ros2bag/Overriding-QoS-Policies-For-Recording-And-Playback/)).
+- Image pipeline nodes use some parameters to allow changing some of the policies ([PR](https://github.com/ros-perception/image_pipeline/pull/521)).
+- Ouster drivers also use some parameters to allow changing some policies ([PR](https://github.com/ros-drivers/ros2_ouster_drivers/pull/26)).
+- Gazebo ROS packages allows configuring QoS profiles of the plugins in the SDF file ([issue](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1079)).
+
+Not providing an standard mechanism for reconfiguring QoS result in a lot of slightly different interfaces to do the same thing, thus making painful to end users to figure out how to reconfigure QoS in different nodes in their system.
```
jacobperron:
```suggestion
Not providing an standard mechanism for reconfiguring QoS results in a lot of slightly different interfaces to do the same thing, creating a pain point for end users when figuring out how to reconfigure QoS for nodes in their system.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+  - Rationale: Allowing to change all QoS policies could break a contract the node's implementer assumed.
+  For example, the node's implementer may have assumed that the data was durable and not volatile.
+  Allowing the node's author to only accept some reconfigurations will avoid user errors.
+- Shall allow the developer to accept/reject changes to a QoS profile with a validation callback.
+  - See rationale above.
+- Shall only apply to publisher/subscriber QoS.
+  - Rationale: Though you can currently pass a custom QoS profiles to services or actions, configuring them is quite fragile and some of the policies cannot be modified (e.g. best effort services don't work).
+  Until a better QoS profile for services and actions is designed, this feature won't be exposed.
+- Shall only allow changing QoS at startup time, and not during runtime by default.
+  - Allowing runtime reconfiguration of QoS is a complex topic, and out of the goal of this proposal.
```
jacobperron:
```suggestion
  - Rationale: Allowing runtime reconfiguration of QoS is a complex topic, and purposefully left out of scope in this proposal.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+- Shall have a way of reusing the same QoS profile.
+  - Rationale: Not having a way of setting the same QoS profile to different entities will make configuring entities which profiles must match (or be compatible) error prone.
+
+## Adding a new mechanism vs using parameters
+
+A previous proposal (see PR [#296](https://github.com/ros2/design/pull/296)) suggested using a new file format for reconfiguring QoS policies.
+Some concerns were raised by the community:
+1. It adds a different mechanism to the ones that already exist.
+    Newcomers has yet to learn another thing.
```
jacobperron:
```suggestion
    Newcomers have to learn yet another thing.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+To make reconfigurability easy, only a flag is needed to automatically create the parameters:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast{10},
+  QosOverridingOptions{true});  // allow_reconfigurable all qos
+```
+
+That will automatically declare the parameters for reconfiguring the qos policies that can be overridden in the following way:
```
jacobperron:
```suggestion
That will automatically declare the parameters for reconfiguring the QoS policies that can be overridden in the following way:
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast(10),
+  [] (const QoSProfile & qos) -> bool {
+      return qos.is_reliable();  // only accept reliable communication, we should add getters to rclcpp::QoSProfile class for this.
+  });
+```
+
+The user provided QoS callback will be internally used as a [parameters callback](https://github.com/ros2/rclcpp/blob/3a4ac0ca2093d12035070443692798b0c9f9da3a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L183), but it has a more friendly interface for the user.
```
jacobperron:
```suggestion
The user provided QoS callback will be internally used as a [parameters callback](https://github.com/ros2/rclcpp/blob/3a4ac0ca2093d12035070443692798b0c9f9da3a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L183), but we provide a more user-friendly interface.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Analysis of parameter API features and limitations
+
+### Read-only parameters
+
+ROS 2 currently provides read only parameters.
+They can be modified at startup time by providing an override (e.g. `--ros-args -p <param_name> <param_value>`), but they cannot be dynamically changed later.
+
+This perfectly matches the goal of the proposal of not making qos settings reconfigurable.
```
jacobperron:
```suggestion
This perfectly matches the goal of the proposal of not making QoS settings reconfigurable during runtime.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+### Parameter events
+
+Up to ROS 2 Foxy, read-only parameters generate a parameter event with its initial value when declared.
+If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will be highly increased.
```
jacobperron:
```suggestion
If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will significantly increase.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+To avoid this issue, a way to opt-out parameter events for a given parameter could be added (should that be the default for read-only parameters??).
```
jacobperron:
```suggestion
To avoid this issue, a way to opt-out of parameter events for a given parameter could be added (should that be the default for read-only parameters??).
```

jacobperron: Although opting out of parameter events for QoS parameters probably makes sense, I'm not sure that we want to default to no parameter events for read-only parameters; it feels like an arbitrary decision as it seems difficult to predict how read-only parameters may be used.

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+### Parameters URIs
+
+Parameters URIs currently look like:
+
+```
+rosparam://my/full/namespace/node_name/param_group.nested_param_group.param_name
+```
+
+For the following qos policy:
```
jacobperron:
```suggestion
For the following QoS policy:
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+  const rclcpp::QosOverridingOptions & qos_options = QosOverridingOptions{false},
+  const PublisherOptionsWithAllocator<AllocatorT> & options = PublisherOptionsWithAllocator<AllocatorT>()
+);
+```
+
+```cpp
+// example usage
+node->create_publisher(
+    "my_topic",
+    user_qos_provided_in_code_1,
```
jacobperron, line 324: I think this example, and the other examples below, have too many arguments.
```suggestion
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+// allow reconfiguring durability
+node->create_publisher(
+    "another_topic",
+    user_qos_provided_in_code_3,
+    rclcpp::KeepLast(10),
+    QosOverridingOptions{QosPolicyKind::Durability});
+```
+
+The above API mentions having some "defaults". 
+The intend of that is to provide a less verbose option, in which case the user can opt-in qos reconfigurability by only opting-in a flag.
```
jacobperron:
```suggestion
The intent is to provide a less verbose option, where the user can opt-in to reconfigurability of these default QoS settings with a single flag.
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+// allow reconfiguring durability
+node->create_publisher(
+    "another_topic",
+    user_qos_provided_in_code_3,
+    rclcpp::KeepLast(10),
+    QosOverridingOptions{QosPolicyKind::Durability});
+```
+
+The above API mentions having some "defaults".
```
jacobperron:
```suggestion
The above API mentions having some default set of QoS settings that are reconfigurable. 
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+The above API mentions having some "defaults". 
+The intend of that is to provide a less verbose option, in which case the user can opt-in qos reconfigurability by only opting-in a flag.
+
+Proposed defaults:
+- Reliability
+- History kind
+- History depth
+- Durability
+
+TBD: Do we want defaults?
```
jacobperron: IMO, letting all the settings be reconfigurable is fine (if a user opts in), unless there's a strong reason to exclude some settings.

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+## Reconfigurability of topics the ROS 2 core creates
+
+TBD: discuss reconfigurability of topics packages in the ROS 2 core creates here, e.g.: clock, rosout, tf, tf_static, etc.
+
+## Further enhancements
+
+### Reusing profiles
+
+The above proposal doesn't include a mechanism to reuse the same qos profile in the parameter files.
```
jacobperron: Seems like this *is* part of the proposal (it's written here 😄). We could probably move this section under "Analysis of parameter API features and limitations". WDYT?

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+/my/ns/node_name:
+  ros__parameters:
+    qos_profiles:
+      subscription:
+        'my/fully/qualified/topic/name/here': 
+          <<: *profileA,
+          history_depth: 100  # override the history depth
+```
+
+### Extending the ROS qos profile with a rmw vendor specific payload
```
jacobperron:
```suggestion
### Extending the ROS QoS profile with a rmw vendor specific payload
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,379 @@
...
+
+TBD
+
+## QoS resolution order
+
+- The QoS provided by the user in the publisher/subscription constructor is used as a default.
+- Those defaults are overridden with the values provided in parameters, if the user allowed this kind of reconfigurability.
+- If the policy was set to `RMW_QOS_*_POLICY_DEFAULT`, the qos resolution mechanism provided by the rmw vendor will be applied (e.g. DDS XML profile files). If not, the previous policy value is the final one.
```
jacobperron: We should probably clarify this point. Are you referring to the policy set *after* resolving any parameter overrides or *before*? IMO, this should happen after.

jacobperron (commented):
Nice! I think the motivation section really helps focus the article.

I've left some wording suggestions and other opinions.

articles/qos_configurability.md
ivanpauno: > Although opting out of parameter events for QoS parameters probably makes sense, I'm not sure that we want to default to no parameter events for read-only parameters; it feels like an arbitrary decision as it seems difficult to predict how read-only parameters may be used.

Sounds reasonable.
I think that the easiest thing is to add an option in `declare`/`set`/`unset` parameter functions, in which you can choose to generate a parameter event or not.

In that case, you're opting-out generating an event for that particular API call.

articles/qos_configurability.md
ivanpauno: good catch!

ivanpauno: Fixed in https://github.com/ros2/design/pull/300/commits/2d5d786d4893727ff1f3214d34c43daff8cbb998.

articles/qos_configurability.md
ivanpauno: Liveliness kind should be excluded, as `MANUAL_BY_TOPIC` requires special care by the node's author.
I don't have a particular concern about other QoS settings.

articles/qos_configurability.md
ivanpauno: Yes, it should happen after. I will rephrase.

articles/qos_configurability.md
ivanpauno: Reordered https://github.com/ros2/design/pull/300/commits/7ff679854145cde92591fb001572a7eee6a46f80.

articles/qos_configurability.md
ivanpauno: Rephrased https://github.com/ros2/design/pull/300/commits/9f5f17f94b009251b2da560daedf0f383f8d32fc.

articles/qos_configurability.md
jacobperron: Sounds okay to me.

articles/qos_configurability.md
jacobperron: Excluding liveliness kind by default sounds fine to me.

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+Here are some examples applying these mechanisms:
+
+- rosbag2 has an ad-hoc mechanism to override QoS profiles when recording or doing playback ([docs](https://index.ros.org/doc/ros2/Tutorials/Ros2bag/Overriding-QoS-Policies-For-Recording-And-Playback/)).
+- Image pipeline nodes use some parameters to allow changing some of the policies ([PR](https://github.com/ros-perception/image_pipeline/pull/521)).
+- Ouster drivers also use some parameters to allow changing some policies ([PR](https://github.com/ros-drivers/ros2_ouster_drivers/pull/26)).
+- Gazebo ROS packages allows configuring QoS profiles of the plugins in the SDF file ([issue](https://github.com/ros-simulation/gazebo_ros_pkgs/issues/1079)).
+
+Not providing an standard mechanism for reconfiguring QoS results in a lot of slightly different interfaces to do the same thing, creating a pain point for end users when figuring out how to reconfigure QoS for nodes in their system.
+
+## Design goals
```
mjeronimo: It's a minor nit, but "design goal" != "requirement." 

(A design goal is typically a higher-level principle used as guidance and to prioritize non-functional requirements, and is not itself verifiable, while a requirement is specific and verifiable. For example, a design goal might be that the system should be responsive. While a requirement would, for example, lay out a specific latency value for a particular operation.)

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+That will automatically declare the parameters for reconfiguring the QoS policies that can be overridden in the following way:
+
+```yaml
+/my/ns/node_name:
+  ros__parameters:
+    qos_profiles:  # group for all the qos settings
+      publisher:
+        'my/fully/qualified/topic/name/here':
```
mjeronimo: I don't think the topic name is sufficient. It seems to me that _topic+message_type_ would properly discriminate. In other words, not all message types using the same topic name should be expected to use the same QoS settings.

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+The user provided QoS callback will be internally used as a [parameters callback](https://github.com/ros2/rclcpp/blob/3a4ac0ca2093d12035070443692798b0c9f9da3a/rclcpp/include/rclcpp/node_interfaces/node_parameters_interface.hpp#L183), but we provide a more user-friendly interface.
+
+## Analysis of parameter API features and limitations
+
+### Read-only parameters
+
+ROS 2 currently provides read only parameters.
+They can be modified at startup time by providing an override (e.g. `--ros-args -p <param_name> <param_value>`), but they cannot be dynamically changed later.
```
mjeronimo: Startup/launch time is indeed when most nodes are typically created and when they read their parameters. However, it is possible to create nodes post-launch. A node created later will also need to read its parameters and so, technically, the user can _specify_ the parameter overrides on the command-line at startup/launch time, but they are _read/modified/overridden_ post-launch.

Also, one might be able to "dynamically change" a node's QoS parameters by destroying the node, updating the YAML file, and re-creating the node, causing it to reinitialize it's pub/subs with the new QoS settings. Not likely that anyone would do this, but it seems possible.

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+  // callback_handle goes out of scope, the callback is auto-removed.
+  ```
+  `declare_parameters_atomically` would be needed, if not you cannot have conditions like `history == keep_all or history_depth > 10`.
+
+### Parameters URIs
+
+Parameters URIs currently look like:
+
+```
+rosparam://my/full/namespace/node_name/param_group.nested_param_group.param_name
```
mjeronimo, line 190: Is the message type also required to fully discriminate (see previous comment)?

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+rosparam://asd.bsd.my_node/qos_profiles/publisher/foo/bar/topic_name/reliability
+```
+
+but in both cases the previous parameters file example wouldn't change.
+
+## Caveats
+
+### Creating more than one publisher/subscription in the same topic with different QoS
+
+If only the topic name is used for the parameter name, it would be impossible to create two publishers with different QoS in the same topic.
```
mjeronimo, line 219: Are these two publishers publishing the same message type or different types? I think an argument could be made that a particular topic/message type combo _should_ use the same QoS settings even if published by different publishers. On the other hand, the same topic name with different message types, I would assume would likely be unrelated (except that they happen to share the same topic name) and therefore should be able to use different QoS settings.

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+
+That could be solved by adding an optional extra identifier:
```
mjeronimo, line 221: Can two separate publishers, using the same topic/message_type combination use different QoS settings? If this is a valid use case, I can understand also introducing an ID to further discriminate. Personally, I would prefer a scheme where topic/message_type was required and ID was a further option.

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+```
+
+### Lowering the amount of declared parameters
+
+#### Hidden parameters
+
+Currently, ROS 2 has the concept of hidden topics and services.
+Those aren't listed in the cli tool commands, except when the user explicitly asks them to be included.
+
+Similarly, hidden parameters could be added.
```
mjeronimo, line 275: I assume you mean that any parameters added because of the QoS override mechanism would be hidden parameters. I think this section could be a bit clearer.

articles/qos_configurability.md
```diff
@@ -0,0 +1,374 @@
...
+             history: keep_last
+      publisher_id2:  # {entity_type}_{id}
+        'my/fully/qualified/topic/name/here': 
+          reliability: best_effort
+            history_depth: 100
+            history: keep_last
+```
+
+### Lowering the amount of declared parameters
+
```
mjeronimo, line 261: You could mention that the following sections describe implementation alternatives.

articles/qos_configurability.md
ivanpauno: Sounds good, I will try to rephrase.

articles/qos_configurability.md
ivanpauno: In the most common use cases, each topic has only one message type.

Always including the topic type in the parameter sounds a bit verbose to me, and the "publisher/subscription id" trick mentioned later should cover this use case.
Anyways, including the topic type in the parameter name sounds like an acceptable proposal to me (though it will get problematic to make the parameters URI unique for a given fq node name, entity type, fq topic name and topic type).

articles/qos_configurability.md
ivanpauno: > However, it is possible to create nodes post-launch. A node created later will also need to read its parameters and so, technically, the user can specify the parameter overrides on the command-line at startup/launch time, but they are read/modified/overridden post-launch.

I'm not sure I understand.
"startup-time" and "launch time" are vague terms, so I will try to rephrase this.

The idea is that you cannot modify the parameter after it was declared (because it's a *const* param), but you can have parameter overrides that are created at runtime (the same applies to other consts parameters).
Parameter overrides can only be provided at node construction time, so yes you would be able to tear down the node, modify the overrides and launch it again.

As another example, when loading a node in a container the message to load the node can have [parameter overrides](https://github.com/ros2/rcl_interfaces/blob/9b7bf4b30e2f51c09fdac86ba711fd6a0f1a5910/composition_interfaces/srv/LoadNode.srv#L27), so it's possible to specify the [parameter overrides at runtime](https://github.com/ros2/rclcpp/blob/5851eebdda5cc2ed9186a034df0ef75ccfe032d3/rclcpp/include/rclcpp/node_options.hpp#L131).

mjeronimo: Yes, agreed. Just wanted to be clear about the term "startup-time." Perhaps referring to node construction time specifically instead would avoid any potential misunderstandings (that is, misinterpreting "start-up time" as being "launch time").

articles/qos_configurability.md
mjeronimo: While it's true that each topic _typically_ has only one associated message type, I don't think we should build in this assumption since, in general, ROS2 allows for multiple types per topic. Regardless of particular solution, there should be a way to differentiate between TopicA/MessageTypeA and TopicA/MessageTypeB when specifying QoS settings. 

I'm wondering about using the ID approach (when creating a publisher)...  I suppose this change would propagate to other functions, such as get_publishers_info_by_topic, where you'd need to augment the TopicEndpointInfo class to have a method to get this ID.

articles/qos_configurability.md
ivanpauno: I will mark this one as resolved, to consolidate the discussion in https://github.com/ros2/design/pull/300#discussion_r499717543.
I will later modify with the result of that discussion.

articles/qos_configurability.md
ivanpauno: I would consider both subsections an "extension" and not an alternative, because they are compatible with the previous sections. Actually, the `Hidden parameters` and `Only declaring some parameters` subsections are not mutually exclusive.

After writing the `Only declaring some parameters` subsection, I would actually pick that option.
Declaring a parameter that cannot be overridden neither modified doesn't make much sense.
It adds some verbosity to publisher/subscription creation though.

mjeronimo: Ok, I see. I assumed that they were alternative approaches.

articles/qos_configurability.md
ivanpauno: Do you mean that all publishers in the same topic should have the same QoS or only all publishers in the same topic within a node? (the same would apply for subscriptions)

I would imagine that, in general, all publishers in the same topic would have the same QoS, regardless if they are in the same node or not. But I didn't want to make that mandatory.

P.S.: In all cases above, "same topic" means same topic name and type.

articles/qos_configurability.md
ivanpauno: > I'm wondering about using the ID approach (when creating a publisher)... I suppose this change would propagate to other functions, such as get_publishers_info_by_topic

I don't think that propagating the concept to graph functions is a good idea.

Having an ad-hoc id only to be able to distinguish publishers/subscribers when re-configuring QoS might not be a great idea either.
If we think that distinguishing publishers in the same topic within a node is not necessary (see [this comment](https://github.com/ros2/design/pull/300#discussion_r499884787)), the concept of "id" can be deleted and the topic type can be added to the parameter name.

It's a bit problematic to add the topic type to the parameter URI though, e.g.:

```yaml
  ros__parameters:
    qos_profiles:  # group for all the qos settings
      publisher:
        'my/fully/qualified/topic/name/here':
          'std_msgs.msg.String':
```

current addressing scheme:
> ros_param://my/full/namespace/ny_node_name/qos_profiles.publisher.my/fully/qualified/topic/name/here.std_msgs.msg.String

alternative addressing scheme
> ros_param://my.full.namespace.ny_node_name/qos_profiles/publisher/my/fully/qualified/topic/name/here.std_msgs.msg.String

well, that actually seems to work.

---

That will make the parameter names much more verbose, but I think that's ok.
The type cannot be optional, because that would imply declaring two parameters (except if we want to not use parameters for this, like in https://github.com/ros2/design/pull/296).

articles/qos_configurability.md
ivanpauno: > Can two separate publishers, using the same topic/message_type combination use different QoS settings?

Yes, they can.
I'm not sure if that has an use case or not.

articles/qos_configurability.md
mjeronimo: Yes, I was thinking that all publishers for the same topic (and type) would have the same QoS, regardless of which node they are in. As you say though, this wouldn't be mandatory.

articles/qos_configurability.md
ivanpauno: @wjwwood @jacobperron any opinions on this?
Should it be possible to reconfigure the QoS of different publishers in the same topic (name and type) to different QoS within a node?
Should it be assumed that all publishers in a topic within a node use the same QoS?

articles/qos_configurability.md
mjeronimo: Yes, it is more verbose, but I think it's consistent with the general idea of allowing for multiple types per topic name. 

Regarding this:
> If we think that distinguishing publishers in the same topic within a node is not necessary...

I would recommend starting with an addressing scheme that includes the type. Then, see if someone comes up with the requirement to distinguish between multiple publishers with the same topic and type in a single node. I'd like to hear what others have to say about this, though.

mjeronimo: By the way, I see from reading up a bit on DDS that a "topic" is defined as a triad composed of a name, a type, and a set of QoS policies (I've been a little sloppy in limiting the "topic" term to the name; I'll try to be more precise). If we're trying to assign QoS policies for topics, it would make sense to me that we allow the user to specify the name and type to which they apply.

articles/qos_configurability.md
ivanpauno: Rephrased in https://github.com/ros2/design/pull/300/commits/3fda43afeed2b436598a8bea653f3d552390ee8c.
Let me know if that looks good.

articles/qos_configurability.md
ivanpauno: > Yes, it is more verbose, but I think it's consistent with the general idea of allowing for multiple types per topic name.

I'm not sure if we guarantee that multiple types in a topic will work, or it's only working because the underlying middleware is DDS.
I think we don't miss any use case by requesting to have only one type per topic name, you can easily remap the colliding names to something else.

> I'd like to hear what others have to say about this, though.

:+1:

---

> on DDS that a "topic" is defined as a triad composed of a name, a type, and a set of QoS policies

A bit off topic, but I wouldn't consider the topic definition to be a triad.
e.g.: DW=data writer, DR=data reader

- DW A: reliable + defaults
- DW B: best effort + defaults
- DR A: reliable + defaults
- DR B: best effort + defaults
- All DW and DR use the same topic name and type.

DW A will match DR A and DR B.
DW B will match DR B.

If QoS are considered part of the topic definition, the above example is a bit strange.

IMHO, the easiest way to see it is that a name and a type defines a topic, but that doesn't guarantee that the endpoints will match. The DW offered QoS must satisfy the DR requested QoS in order to communicate.

P.S.: DDS has "Topic QoS", but those are used as a default for DR/DW and don't affect communication (but, can also be shared during discovery).

articles/qos_configurability.md
ivanpauno: Clarified in https://github.com/ros2/design/pull/300/commits/f8f27f6b902021b75eebbbf3cf6582ae484c0170.

articles/qos_configurability.md
ivanpauno: Rephrased in https://github.com/ros2/design/pull/300/commits/48174bfc5ee69552cd366dc6d47175c86915f3c8.

articles/qos_configurability.md
ivanpauno: Clarified in https://github.com/ros2/design/pull/300/commits/39ddf0d9166161d34e14ba65de5c05d589bed4bf.

articles/qos_configurability.md
ivanpauno: Clarified in https://github.com/ros2/design/pull/300/commits/3ce447de1583962478ff6208048e05a85efd96f2.

articles/qos_configurability.md
mjeronimo: @ivanpauno Good, thanks for the background info.

articles/qos_configurability.md
ivanpauno: Clarified in https://github.com/ros2/design/pull/300/commits/563ed642675bd1dbdefa14ac8b7950bcc1486ef1.

ivanpauno:
Another alternative to solve the discussions [here](https://github.com/ros2/design/pull/300#discussion_r499717543), [here](https://github.com/ros2/design/pull/300#discussion_r499734802) and [here](https://github.com/ros2/design/pull/300#discussion_r499759276), I was thinking that having a "profile id" that can be independent of the topic name and type may be a good idea.

i.e.: the user provides a construction time an id
```cpp
node->create_publisher(
  topic_name, qos,
  QosOverridingOptions::Builder()
    .id("my_profile_id")
    .policies({QoSPolicy::Reliability, QoSPolicy::History})
    .callback(callback)
    .build());
```

and then the file looks like:
```yaml
/my/node/name:
  ros__parameters:
    qos_profiles:
      publisher:
        my_profile_id:
          reliability: reliable
          history: keep_last
```

In that way, the node's author can use the same profile name in several entities that must share the qos profile.
I would use as the default id the topic name (after expansion and remapping), but allowing the node author's to change it covers all possible use cases without the need of adding more levels of nesting (the type and an optional id).

The profile id could be remappable (using the same remapping rules that services and topic names) or not, I think it doesn't matter because all parameters are "private" to a node.
Caveat: topic names can be remapped, so  if publisherA uses `id1`, and `publisherB` uses its topic name as the id and it gets remapped to `id1`, then you have a problem. But that can easily be avoided by the user doing systems integration.

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+  KeepLast{10});
+```
+
+To make reconfigurability easy, only a flag is needed to automatically create the parameters:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast{10},
+  QosOverridingOptions{true});  // allow_reconfigurable all qos
```
gbiggs, line 92: Is the intention to make this a member of the publisher options struct?

articles/qos_configurability.md
gbiggs: > I'm not sure if we guarantee that multiple types in a topic will work, or it's only working because the underlying middleware is DDS.

I don't think I've seen anything explicitly saying that in ROS the same topic name can have multiple types. From consistency and reliability and usability points of view I'd expect the opposite: In a single system, one topic, one type.

I think this is something those with more knowledge of the history of ROS 2 can definitively comment on.
@wjwwood @dirk-thomas @clalancette

gbiggs: I don't think QoS should be part of the topic definition. In DDS, that "triad" applies to a publisher or subscriber. It is possible for a publisher and a subscriber on the same topic to have different QoS settings, as long as those different settings are compatible (for example, the subscriber requires data at a minimum of 10Hz, and the publisher guarantees publishing at a minimum of 50Hz). The same should be true in ROS.

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+
+Note: parameter overrides can be constructed dynamically and then used to create a node, but it's not possible to modify the parameter after the node construction.
+This allows, for example, to pass parameter overrides when dynamically adding a node to a component container.
+
+### Parameter events
+
+Up to ROS 2 Foxy, read-only parameters generate a parameter event with its initial value when declared.
+If we start declaring parameters for each QoS policies of many entities, the amount of parameter events at startup time will significantly increase, as the declaration of each parameter generates an event.
+
+To avoid this issue, a way to opt-out parameter events when declaring a parameter could be added.
```
gbiggs, line 141: In general we probably want to opt-out of infrastructure parameter events by default, not just the QoS ones.

articles/qos_configurability.md
gbiggs: Although I can't think of a use case for it, as I recall in DDS it is possible to have multiple publishers with different QoS profiles on the same topic. I'm not sure if they have to be compatible or not, though I suspect that if they are not you would not be able to subscribe to the topic without an error occurring somewhere.

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+
+### Lowering the amount of declared parameters
+
+This two subsections describe some optional mechanism to lower the amount of declared parameters.
+The two mechanism are compatible with what it was said before and are not mutually exclusive.
+
+#### Hidden parameters
+
+Currently, ROS 2 has the concept of hidden topics and services.
+Those aren't showed by the cli tools, except explicitly requested:
```
gbiggs:
```suggestion
Those aren't shown by the cli tools, except when explicitly requested:
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+ros2 topic list  # Will not list hidden topics.
+ros2 topic --include-hidden-topics list  # Will list hidden topics.
+```
+
+Similarly, hidden parameters could be added.
+All the parameters used for overriding QoS could be declared as hidden, thus avoiding "noise" in commands like `ros2 param list`.
+
+#### Only declaring some parameters
+
+Instead of hidden parameters that correspond to the QoS policies, we could only declare parameters for the policies that the user want to be reconfigurable.
```
gbiggs, line 280: I think the earlier approach combined with hidden parameters is more usable. As a bonus, hidden parameters will be useful in many other places in the ROS infrastructure.

articles/qos_configurability.md
ivanpauno: I think that both ways are ok.
Using an independent argument can make construction a bit less verbose.

e.g. as a member of `PublisherOptions`:
```cpp
rclcpp::PublisherOptions options;
options.qos_overriding = QosOverridingOptions{true};
node->create_publisher(
  "chatter",
  KeepLast{10},
  options);
```

articles/qos_configurability.md
ivanpauno: I think that's reasonable.

Do you mean?
- Turn off parameter events for read-only parameters.
- Turn off by default parameter events generated when declaring a parameter, but continue generating events when a parameter is set.
- Turn off by default all parameter events.

`rclcpp` currently depends on parameters events to make `use_sim_time` works, but that's fixable.

articles/qos_configurability.md
ivanpauno: thanks

articles/qos_configurability.md
gbiggs: I mean to turn off parameter events for any parameters used internally by the ROS infrastructure, in the way these QoS ones are. I don't know how many of those there are at this time, though.

My reasoning is that a node author is going to care about events for parameters they declare themselves, but will find any additional events that they have to handle in some way to be noise.

ivanpauno: Ah, that makes sense.
I miss-read your previous comment.

articles/qos_configurability.md
mjeronimo: @gbiggs  One example is the ROS2 node's get_topic_names_and_types method, which returns a map of strings to vectors of strings:

```
  /// Return a map of existing topic names to list of topic types.
  /**
   * \sa rclcpp::Node::get_topic_names_and_types
   */
  RCLCPP_LIFECYCLE_PUBLIC
  std::map<std::string, std::vector<std::string>>
  get_topic_names_and_types(bool no_demangle = false) const;
```
Implying that each topic name can be associated with multiple types. 

I don't want to get off-topic, but if this is indeed the case I think this also will break any tools that make the assumption of one topic name per one topic type.  For example, rqt_bag, which I've been working on, has a row on the message timeline for each topic name. To properly support 1 topic name/n topic types, rqt_bag would need one row per combination.

articles/qos_configurability.md
jacobperron: I assume the case is rare, but I think it should be possible. E.g. by default one publisher may be using *transient local* durability and another using *volatile*. Perhaps the node author would like the history depth of each publisher to be externally configured independently. Having the optional identifier makes this possible.

jacobperron:
Using a profile ID as an alternative seems viable and perhaps simpler. Instead of defaulting the ID to the topic name, I would rather make the profile ID a hard requirement if an author wants to make their publisher or subscription QoS configurable. For example:

```c++
node->create_publisher(
  topic_name, qos,
  QosOverridingOptions::Builder("my_profile_id")
    .policies({QoSPolicy::Reliability, QoSPolicy::History})
    .callback(callback)
    .build());
```

This avoids any issues with topic remapping.

We'll probably want tooling to be able to query what pub/subs are associated with what profile IDs (post-remapping).

mjeronimo:
I agree with @jacobperron that using a profile ID but not defaulting to the topic name would a better way to go. It seems pretty straightforward and less error prone (remapping a topic name resulting in an expected change of QoS settings).

gbiggs:
If the ID is specified in source code, then it would need to be scoped to the node. Otherwise we could have two nodes from different authors using the same ID and clashing.

ivanpauno:
> If the ID is specified in source code, then it would need to be scoped to the node. Otherwise we could have two nodes from different authors using the same ID and clashing.

Parameter names are all "private" to a node (and not global, like in ROS 1), so that's not an issue.

---

@jacobperron @mjeronimo My main motivation to use the "resolved" topic name as the default id is that it makes pretty easy to set the QoS when you want is to set the same profile for all nodes for all entities in the same topic:

```yaml
/**:
  ros__parameters:
    qos_profiles:
      publisher:
        '/my/topic/name':
          reliability: reliable
          history_depth: best_effort
```

If desired, the qos parameters can be overridden for an specific node.
In the weird case that a node wants different qos parameters for the same topic, the you can override the id with something else.

But using the "resolved" topic name as the profile id makes configuration easy for the typical use case (and not impossible for strange use cases).

e.g.: node A publishes to `/my_topic`, node B publisher to `/another_topic`.
        Both are remapped to `/new_topic`.
        If the profile id was also modified to `/new_topic`, then you can configure the same qos easily using node wildcards (like the example above).

jacobperron:
Defaulting to the topic name sounds okay too. The re-usability argument is convincing.
In any case, I think using the (optional) pub/sub ID as a QoS profile ID makes sense.

articles/qos_configurability.md
wjwwood: I agree with having it in the options instead. We need to keep the create publisher signature simple, as in the past it got really long and hard to read.

We can make the publisher options more convenient (like using the parameter idiom) if it becomes annoying.

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+  KeepLast{10},
+  QosOverridingOptions{true});  // allow_reconfigurable all qos
+```
+
+That will automatically declare the parameters for reconfiguring the QoS policies that can be overridden in the following way:
+
+```yaml
+/my/ns/node_name:
+  ros__parameters:
+    qos_profiles:  # group for all the qos settings
```
wjwwood: This is a bit of bike-shedding, but are the items under this "group" actually "qos profiles"? or is actually more like `qos_overrides`? I know there is discussion later of defining ad-hoc qos profiles that can be used here using yaml anchors, but what's here (as I understand it) is putting those to use in order to override the existing qos settings being used in the code.

articles/qos_configurability.md
wjwwood: My personal opinion after all of this is that if you use more than one type per topic, and you want to configure these, then you should use a named entity (so called "id").

---

I think it's somewhat unlikely that within a single node there are two entities with the same topic name but separate types. I would consider this a bug and discourage it. It is more likely, in my opinion, that two entities in different nodes have the same name but different types, even though I'd still discourage that. But in that situation you could still configure them separately as they are namespaced by the node name and namespace.

It is true that we allow for more than one type per topic, but that's not because we encourage it, but because of the nature of distributed discovery, since there might be remote entities with the same name but different types. That's the main reason for the discovery API allowing for that.

In fact, unless it changed, most DDS vendors will not allow you to create a topic with the same name but different types within a single participant. It only comes up when the conflicting entity is remote or at least in a different participant. So their behavior is a bit asymmetric on that point.

I think adding the type to the name of the parameter could work, but I'd prefer keeping the common case simple (each topic has a different type) and letting users use the named entity mechanism if they need to do the more complicated thing (entities share a topic name but not a type).

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+
+A callback to accept/reject the configured profile can also be specified:
+
+```cpp
+node->create_publisher(
+  "chatter",
+  KeepLast(10),
+  [] (const QoSProfile & qos) -> bool {
+      return qos.is_reliable();  // only accept reliable communication, we should add getters to rclcpp::QoSProfile class for this.
+  });
```
wjwwood, line 116: This is fine for brevity of the example, but again I think this should be in the `PublisherOptions`.

articles/qos_configurability.md
wjwwood: I actually think we shouldn't do this. I think we should treat all of the parameters the same. It would be confusing to me as a user to see that some parameters cause events and others do not.

Answering the question "which parameters generate events" is not very trivial then.

That being said, if we do this, then it should be possible for the user to have parameters that do not generate events as well, i.e. we should dog food this feature.

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+In that way, we avoid installing a parameter callback that will be triggered when any parameter is modified.
+
+Note: this seems to be what you generally want in the case of read-only parameters.
+Note 2: We could also have a `declare_parameters_atomically` method and use:
+  ```cpp
+  auto callback_handle = node->add_on_set_parameters_callback(wrapper_for_qos_callback(user_provided_callback));
+  node->declare_parameters_atomically(...);  // names and defaults for all parameters given here
+  // callback_handle goes out of scope, the callback is auto-removed.
+  ```
+  `declare_parameters_atomically` would be needed, if not you cannot have conditions like `history == keep_all or history_depth > 10`.
```
wjwwood, line 183: You should at least set the QoS parameters for a single topic atomically, but honestly, maybe all of them should be set atomically. What if you wanted `publisherA.history == subscriptionB.history`?

So I agree we need a `declare_parameters_atomically()`.

articles/qos_configurability.md
wjwwood: This is supported. They might have different message content, e.g. imagine two tf broadcasters, publishing on the same topic name and using the same type, but one might want a larger history cache due to bursty publishing behavior.

articles/qos_configurability.md
wjwwood: I think a reasonable use case for this would be to have two subscriptions in a node, each with the same settings except for the history depth. Imagine two libraries created subscriptions to the same topic (and same type) using the same node (which was passed into the libraries). The two libraries are not coordinating and so may want different QoS. They may also do this to just have two different history caches (so one library taking a sample doesn't interfere with the other library).

So having two entities with the same name and type but different QoS is a valid use case from my perspective.

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+      publisher_id1:  # {entity_type}_{id}
+        'my/fully/qualified/topic/name/here': 
+          reliability: reliable
+            history_depth: 100
+             history: keep_last
+      publisher_id2:  # {entity_type}_{id}
+        'my/fully/qualified/topic/name/here': 
+          reliability: best_effort
+            history_depth: 100
+            history: keep_last
```
wjwwood: What about this instead?

```suggestion
/my/ns/node_name:
  ros__parameters:
    qos_profiles:
      'my/fully/qualified/topic/name/here':
        publisher_id1:  # {entity_type}_{id}
          reliability: reliable
          history_depth: 100
          history: keep_last
        publisher_id2:  # {entity_type}_{id}
          reliability: best_effort
          history_depth: 100
          history: keep_last
```

articles/qos_configurability.md
```diff
@@ -0,0 +1,399 @@
...
+
+```yaml
+/my/ns/node_name:
+  ros__parameters:
+    qos_profiles:
+      publisher_id1:  # {entity_type}_{id}
+        'my/fully/qualified/topic/name/here': 
+          reliability: reliable
+            history_depth: 100
+             history: keep_last
```
wjwwood: The indentation looks really weird here, maybe all of these three are supposed to be aligned?

articles/qos_configurability.md
ivanpauno: `qos_overrides` sounds good to me

articles/qos_configurability.md
ivanpauno: > What if you wanted publisherA.history == subscriptionB.history?

That's a good question, but I think in that case the user can control in which order they create entities:

```cpp
publisher = node->create_publisher(
  "my_topic",
  my_pub_default_qos,
  [] (const QoSProfile & qos) -> bool {
      return qos_profile_is_valid(qos);
  });

subscription = node->create_subscription(
  "my_topic",
  my_sub_default_qos,
  subscription_callback
  [required_history=publisher->get_qos().get_history()] (const QoSProfile & qos) -> bool {
      return qos.get_history() == required_history && qos_profile_is_valid_2(qos);
  });
```

Declaring all qos parameters atomically sounds like a valid alternative, but I'm not sure when that should happen.

articles/qos_configurability.md
ivanpauno: :+1:

If we add support for services/actions in the future and somebody is using the same name for a topic and a service, it would look like:

```
/my/ns/node_name:
  ros__parameters:
    qos_profiles:
      'my/fully/qualified/name':
        publisher_id1:  # {entity_type}_{id}
          ...
        service_id1:  # {entity_type}_{id}
          ...
```

It's a bit strange to have a publisher and a service nested in the same group, but it's also not an issue (and most people doesn't use the same name for topics and services).

articles/qos_configurability.md
ivanpauno: Fixed in https://github.com/ros2/design/pull/300/commits/0cee04571d97a0f8b746bbe0388e7a3aae284c82.

articles/qos_configurability.md
ivanpauno: Clarified in https://github.com/ros2/design/pull/300/commits/c5d88491678c9cba5073e8cbd9910d1e56d0ddbc.

articles/qos_configurability.md
ivanpauno: > I agree with having it in the options instead. We need to keep the create publisher signature simple, as in the past it got really long and hard to read.

Sounds good to me, added a note in https://github.com/ros2/design/pull/300/commits/c5d88491678c9cba5073e8cbd9910d1e56d0ddbc (for the sake of brevity, I didn't update the examples in the design doc, I think it doesn't matter much).

> We can make the publisher options more convenient (like using the parameter idiom) if it becomes annoying.

That's a good point, see for example https://github.com/ros2/rclcpp/issues/1392.

articles/qos_configurability.md
ivanpauno: Done in https://github.com/ros2/design/pull/300/commits/2e3576f569966f85f74a167edbf7e4d7a8823d8c.

articles/qos_configurability.md
ivanpauno: I have applied this suggestion in https://github.com/ros2/design/pull/300/commits/d4558599c9a744e79715d4a1d0b4d185b41005bf.

articles/qos_configurability.md
ivanpauno: > My personal opinion after all of this is that if you use more than one type per topic, and you want to configure these, then you should use a named entity (so called "id").

My last question about this: Do we want the name of the entity ("id") to be complementary to the topic name (like described in the [current proposal]) or to be a replacement for it (i.e. the default "id" is the topic name and that can be overridden)?

More details in [this comment](https://github.com/ros2/design/pull/300#issuecomment-705123339).

I don't have a strong preference about either option, using the "id" as an optional complement to the topic name is more restrictive, but that can be seen as an advantage (less ways of people getting creative on how to provide qos profile overrides).

[current proposal]: https://github.com/ros2/design/blob/d4558599c9a744e79715d4a1d0b4d185b41005bf/articles/qos_configurability.md

wjwwood: I don't have any opinion on `id` vs `<entity>_id`.

articles/qos_configurability.md
```diff
@@ -0,0 +1,402 @@
...
+  It also only applies for "transient local" durability, so it shouldn't be reconfigurable if durability isn't.
+
+We could also have a restricted list of default policies to be declared in parameters:
+- Reliability
+- History kind
+- History depth
+
+which are the ones that usually require reconfigurability.
+
+TBD: The defaults could be different to the ones proposed above.
```
jacobperron: The defaults above seem reasonable to me :+1:

ivanpauno:
FYI (to everyone in the thread), I have opened some PRs implementing the proposal:

* rclcpp: https://github.com/ros2/rclcpp/pull/1408
* rclpy: https://github.com/ros2/rclpy/pull/635
* demo: https://github.com/ros2/demos/pull/474

jacobperron (approved):
The document LGTM. I think most (or all?) comments have been addressed. I think we can merge this (maybe with a second approval, perhaps@mjeronimo). We can always make future amendments and additions where necessary.

articles/qos_configurability.md
ivanpauno: Removed the comment in https://github.com/ros2/design/pull/300/commits/d271d21620bfb63d000f1b79090b147b6d661c14.

ivanpauno:
Merging, thanks @gbiggs @jacobperron @mjeronimo @wjwwood @dejanpan for the feedback in https://github.com/ros2/design/pull/296 and here.

If there's any extra comments I can address them in a follow-up.

PR #301, opened 2020-10-05
Per-Package Documentation

maryaB-osr:
Signed-off-by: maryaB-osr <marya@openrobotics.org>


articles/per_package_documentation.md
```diff
@@ -0,0 +1,99 @@
...
+
+ROS 2 is lacking a process for the conglomeration of package documentation.
+Discoverability for package documentation (and documentation in general) is one of the most frequent user complaints.
+
+With a standardized package documentation process that encourages package maintainers to robustly document their packages, and presents the results alongside ROS 2's general documentation, issues of discoverability and consistency will be addressed.
+
+In general, our vision for the system is:
+
+- Package maintainers document their packages (API docs, manually written content, etc) in their packages' repositories following some guidelines or templates we recommend
+- Package documentation from the repositories is built on our documentation site by our buildfarm in an automatic process with no extra input from the maintainers
```
sloretz: Does no extra input mean nothing beyond adding a `doc` entry in the ROS distro's `distribution.yaml` to opt in?

articles/per_package_documentation.md
```diff
@@ -0,0 +1,99 @@
...
+</div>
+
+Original Author: {{ page.author }}
+
+## Background
+
+ROS 2 is lacking a process for the conglomeration of package documentation.
+Discoverability for package documentation (and documentation in general) is one of the most frequent user complaints.
+
+With a standardized package documentation process that encourages package maintainers to robustly document their packages, and presents the results alongside ROS 2's general documentation, issues of discoverability and consistency will be addressed.
```
sloretz: I'm having trouble parsing this sentence.

clalancette: Maybe something like:

```
The goal of this design is to standardize a documentation process for ROS 2 packages to encourage maintainers to robustly document their packages.  The results will be presented in a consistent way alongside the general ROS 2 documentation, making the documentation as a whole discoverable and consistent.
```

articles/per_package_documentation.md
```diff
@@ -0,0 +1,112 @@
...
+
+In general, our vision for the system is:
```
clalancette: The possessive tone may not be the best way to go.  Particularly for future readers, it may not be entirely clear who it is.  I'd go more for "the vision for the system is", and so on throughout the document.

articles/per_package_documentation.md
clalancette: Yeah, I think that this could be clarified a bit with:

```
Package documentation from the repositories is built on the ROS 2 infrastructure and deployed to the ROS 2 documentation site in an automatic process with no input from the maintainer besides an opt-in `doc` block in the distribution.yaml
```

Also, we should probably put a link to [REP 141](https://www.ros.org/reps/rep-0141.html) for context.

maryaB-osr: ~I think we wanted to keep implementation details out of the document at this stage, but maybe this point is generic enough to describe here, @tfoote?~ 

Got it

articles/per_package_documentation.md
```diff
@@ -0,0 +1,112 @@
...
+
+- All documentation will be hosted on `docs.ros.org`
+- All documentation will be versioned by ROS 2 distribution names
+- The URL structure will be `docs.ros.org/<lang>/<distro>/<page>` for generic documentation and `docs.ros.org/<lang>/<distro>/p/<package_name>/<page>` for package documentation
+  - For more context on where documentation will fall under this URL structure, see [this diagram](https://docs.google.com/drawings/d/1KxzDrcSZzwGgudk-kEXGWnoRuAtc1ffl_KBeIu-V60Y/edit?usp=sharing).
+
+## Requirements
+
+The "Primary requirements" are those that must be in place for the system to be functional and achieve its purpose.
+The "Secondary requirements" are important features and functionality that are not necessary to roll out the first stage of implementation, but should be carried out following the system's roll out.
```
clalancette:
```suggestion
The "Secondary requirements" are important features and functionality that are not necessary to roll out the first stage of implementation, but should be carried out as soon as possible following the initial roll out.
```

articles/per_package_documentation.md
```diff
@@ -0,0 +1,112 @@
...
+
+**The system must support C++ and Python API docs generation**
+
+The system will automatically extract API docs for these languages from the source code in package repositories and build the output on the docs site.
+
+**The system must allow package documentation to be versioned per ROS 2 distribution**
+
+The docs site and buildfarm will allow documentation for the latest version of a package corresponding to each ROS 2 distribution the docs site supports versioning for.
+This means no versioning for several versions of a package per ROS 2 distribution.
+The package repository can still maintain it's own docs for previous versions, they just won't be hosted on our site.
```
clalancette: I think we said here we wanted the URL structure to support this.  So I think we should clarify this a bit and say that while multiple versions isn't currently supported, the URL structure will support it and it is on the secondary requirements board below.

articles/per_package_documentation.md
```diff
@@ -0,0 +1,112 @@
...
+
+**The buildfarm must automatically build package documentation**
+
+Changes to the documentation in a package repository shouldn't require the maintainer to trigger a build or anything like that.
+The changes should be automatically updated to the site.
```
clalancette:
```suggestion
Changes to the documentation in a package repository shouldn't require the maintainer to manually trigger a build.
The changes should be automatically built and uploaded to the site.
```

clalancette (commented):
I've got a few things inline that I think could be clarified.

ros-discourse:

https://discourse.ros.org/t/request-for-comment-per-package-documentation-requrements/16699/1

articles/per_package_documentation.md
```diff
@@ -0,0 +1,115 @@
...
+ROS 2 is lacking a process for the conglomeration of package documentation.
+Discoverability for package documentation (and documentation in general) is one of the most frequent user complaints.
+
+The goal of this design is to standardize a documentation process for ROS 2 packages to encourage maintainers to robustly document their packages.
+The results will be presented in a consistent way alongside the general ROS 2 documentation, making the documentation as a whole discoverable and consistent.
+
+In general, our vision for the system is:
+
+- Package maintainers document their packages in their repositories following some guidelines or templates we recommend
+- Package documentation from the repositories is built on the ROS 2 infrastructure and deployed to the ROS 2 documentation site in an automatic process with no input from the maintainer besides an opt-in `doc` block in the distribution.yaml
```
SteveMacenski: "with no input from the maintainer besides an opt-in `doc` block", I'm not sure what this fragment is trying to get across. Of course the maintainer has some input, they would have needed to merge the PRs with the docs into the branch to build on the build farm. 

I think maybe what you mean isn't "input" but "action" but either way, I think the "deployed to the ROS 2 documentation site in an automatic process" makes that clear. Unless there was another point this was trying to get across,  I think it should be removed for clarity

articles/per_package_documentation.md
```diff
@@ -0,0 +1,115 @@
...
+- All documentation will be versioned by ROS 2 distribution names
+- The URL structure will be `docs.ros.org/<lang>/<distro>/<page>` for generic documentation and `docs.ros.org/<lang>/<distro>/p/<package_name>/<page>` for package documentation
+  - For more context on where documentation will fall under this URL structure, see [this diagram](https://docs.google.com/drawings/d/1KxzDrcSZzwGgudk-kEXGWnoRuAtc1ffl_KBeIu-V60Y/edit?usp=sharing).
+
+## Requirements
+
+The "Primary requirements" are those that must be in place for the system to be functional and achieve its purpose.
+The "Secondary requirements" are important features and functionality that are not necessary to roll out the first stage of implementation, but should be carried out as soon as possible following the initial roll out.
+
+### Primary requirements
```
SteveMacenski, line 51: An additional primary requirement I'd like to propose is that the format is a common markdown language (rst, md, etc) that can be rendered by GitHub - that will help catch a bunch of issues before release.

33thou:
As a newcomer to ROS2 coming from the Python ecosystem, it's often surprising and frustrating to come across github repos for packages with no documentation in their Readme files. In general the relationship between ROS packages, locations in ros.org, and repos in Github is super confusing and it's often not clear to me where to find documentation, file an issue, or ask a question. This is especially hard because of the highly federated mapping to many github organizations and child repositories, so sometimes the hierarchical structures are unclear.

Compare with the [PyPi website](https://pypi.org/). Packages have relatively simple introductions and metadata there, but in general package maintainers are encouraged to create robust docs directly in Github, or using a system of choice like readthedocs or a static site generator.

If docs.ros.org is being set up as the canonical way to find documentation that seems useful, but I would think two things are important:

1. a systematic way to link to an external homepage and docs page if the package maintainer chooses to put their docs elsewhere
2. a systematic way to generate README.md files in github which link to relevant locations in *.ros.org

thanks for considering!

P.S.: as an aside, I've found `index.ros.org` to be really awkward to use. I was browsing around looking for packages supported by Foxy related to kalman filters, sensor fusion, IMUs, 3D cameras, and visual odometry. Searching is super confusing because it just gives names of packages with no descriptions or any other useful context like popularity or frequency of updates.

articles/per_package_documentation.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Context
+
+The per-package documentation plan is an extension of the general ROS 2 documentation project.
+That project, in the implementation phase now, includes some keys points that affect the context of the per-package documentation.
+
+- All documentation will be hosted on `docs.ros.org`
+- All documentation will be versioned by ROS 2 distribution names
+- The URL structure will be `docs.ros.org/<lang>/<distro>/<page>` for generic documentation and `docs.ros.org/<lang>/<distro>/p/<package_name>/<page>` for package documentation
+  - For more context on where documentation will fall under this URL structure, see [this diagram](https://docs.google.com/drawings/d/1KxzDrcSZzwGgudk-kEXGWnoRuAtc1ffl_KBeIu-V60Y/edit?usp=sharing).
```
gbiggs: It would be better to embed this diagram in the page as a PNG stored in the repository.

articles/per_package_documentation.md
```diff
@@ -0,0 +1,115 @@
...
+The package repository can still maintain it's own docs for previous versions.
+
+While multiple package versions per distribution will not initially be supported, the URL structure will support the addition of this feature in the future (mentioned in the Secondary requirements below).
+
+**The buildfarm must automatically build package documentation**
+
+Changes to the documentation in a package repository shouldn't require the maintainer to manually trigger a build.
+The changes should be automatically built and uploaded to the site.
+
+**The system must automatically generate content for a package so it's listed on docs.ros.org even if the package maintainer does not explicitly set up anything in the package repository**
```
gbiggs: This requirement conflicts with this line from the background:

> Package documentation from the repositories is built on the ROS 2 infrastructure and deployed to the ROS 2 documentation site in an automatic process with no input from the maintainer besides an opt-in `doc` block in the distribution.yaml

While there's no conflict about not setting up anything in the package repository itself, the requirement says the system is opt-out while the line from the background is explicit that it's opt-in.

If the intention is that *some* page will be generated for every package but documentation will not be generated unless the package opts in, then perhaps clarify this requirement and the background so that is clear.

SteveMacenski: I’d like something to autogenerate given many packages would otherwise have nothing. In ROS1 if you asked it to setup docs in Bloom it would generate API docs even if there was no doxygen which was awesome. Something minimal like that and the readme would at least be something.

articles/per_package_documentation.md
```diff
@@ -0,0 +1,115 @@
...
+**All package documentation must be available with the rest of the ROS 2 documentation under a single domain**
+
+Package documentation will be treated the same as the generic ROS 2 documentation.
+Its presence as part of the docs site should be made well-known from the site's overview as well as from any entry point into the docs site, and be intuitive from the organization and layout of the site.
+
+**Package documentation must have an easily navigable and intuitive URL structure**
+
+Every package's "home page" should be reachable with minimal effort by the scheme `docs.ros.org/p/<package_name>/`, which will redirect to default language and version, `docs.ros.org/<lang>/<distro>/p/<package_name>/`  
+
+**Package documentation must be maintainable in its repository without going through a third party**
```
Timple, line 62: This is a good step forward :+1:

articles/per_package_documentation.md
```diff
@@ -0,0 +1,115 @@
...
+
+When writing package documentation, maintainers should not have to concern themselves with auto-generated content, like the current `/changelogs` and `/symbols` directories in the current `docs.ros.org` API docs structure for ROS 1.
+
+### Secondary requirements
+
+**Package maintainers should have the ability to include free form documentation, like tutorials, design docs, etc., alongside generated API docs on docs.ros.org**
+
+**The system should provide a standardized interface for generating an index page/landing page for packages, consistent across ROS 2**
+
+**The system should provide a standardized method for generating package documentation locally, including build error alerts**
```
mikeferguson, line 108: I feel like this should be a primary requirement - otherwise the round-trip time required to create properly formatted documentation is going to be quite long (merge a change into the distro branch, wait for the build farm to catch the problem, then update and repeat) - and the net result will likely be many people give up and have either broken or very little documentation.

Also, if there is a method for building package documentation locally, it can also be added to the package CI (I imagine that tools like industrial_ci would quickly integrate this) - otherwise the effort for maintainers to check over documentation changes is going to be quite high.

clalancette: While I agree that being able to build the documentation locally is important, we made it a secondary requirement because it is something we can add later.  That is, it's the difference between having a system at all, and having a system that is nice to use.

articles/per_package_documentation.md
maryaB-osr: Thanks for pointing this out, it's one of those things that we internally assume by default so we didn't consider, which is exactly why we wanted outside eyes on this!

We're going to default to rst

articles/per_package_documentation.md
maryaB-osr: We wanted to emphasize the limited amount of investment on the maintainers' end, but you're right it's not "no input". 
Making sure to incorporate @sloretz's point about [how one would opt in](https://github.com/ros2/design/pull/301#discussion_r499733468), I think the following satisfies your issue?

> Package documentation from the repositories is built on the ROS 2 infrastructure and deployed to the ROS 2 documentation site in an automatic process maintainers can opt in to by adding a `doc` block to the `distribution.yaml`

articles/per_package_documentation.md
maryaB-osr: I can do that, I just figured it's not really central to this proposal. However, that might be a side effect of how we internally separated the work for the docs into two parts. So, in it goes :)

articles/per_package_documentation.md
maryaB-osr: We think so too, thanks for reviewing!

articles/per_package_documentation.md
maryaB-osr: I agree that this should be something that gets implemented first just because it will greatly increase ease of use 

But our definition for secondary requirements is that they're important but "not necessary to roll out the first stage of implementation", which this does fall under. We need to ensure that we can move quickly through the initial implementation and not get tied up in anything that isn't an _absolutely_ necessity. So, we will definitely add this feature and all the others under "secondary requirements", just not at the first roll-out.

articles/per_package_documentation.md
maryaB-osr: This requirement is slightly different from the one in the background section, but you're right, it's not clear. The background section is talking about building any actual documentation like API docs and such which would be opt in. 

This point is talking about generating some generic content just for an index of packages on the package documentation homepage, that'll include the package name and maybe some other generic info like maintainer, activity, etc. That would be opt out

tfoote (approved):

articles/per_package_documentation.md
gbiggs: I think it's OK to not have it straight away provided we get it really soon after. GitHub Actions provides a good example of why having to go via the online version makes using the tool _extremely_ painful.

gavanderhoorn: > GitHub Actions provides a good example of why having to go via the online version makes using the tool *extremely* painful.

Off-topic, but: yes (although: [nektos/act](https://github.com/nektos/act)).

mikeferguson: Is it correct to state that the secondary requirements are still absolute requirements on how the system is architected? My concern is that if this is a secondary requirement, the system may end up being reliant on some cloud-based component - which means we might never be able to fulfill this secondary requirement.

(On the cloud-side of things, I've seen several projects where cloud-independence wasn't specified and you end up with people building something that uses all sorts of proprietary offerings from say AWS - because it is fast to implement, but then you're forever stuck using that proprietary offering or doing a major re-design)

clalancette: > Is it correct to state that the secondary requirements are still absolute requirements on how the system is architected?

Yes, exactly.  That's a great way to put it.

articles/per_package_documentation.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Context
+
+The per-package documentation plan is an extension of the general ROS 2 documentation project, which introduced some requirements that affect the context of the per-package documentation:
+
+- All documentation will be hosted on `docs.ros.org`
+- All documentation will be versioned by ROS 2 distribution names
+- The URL structure will be `docs.ros.org/<lang>/<distro>/...` for generic documentation and `docs.ros.org/<lang>/<distro>/p/<package_name>/...` for package documentation
+  - See the diagram below for more context on where documentation will fall under this URL structure:
+  ![](per_package_documentation/Version_package_flowchart.png)
```
DLu, line 43: I'm late to the game, but two things:
 * As a usability requirement, all images should have alt text. 
 * Should independent packages go under `docs.ros.org/<lang>/independent`?

maryaB-osr: @DLu thank you for pointing out the missing alt text! 
And yes, technically everything will fall under `docs.ros.org/<lang>`, we will make that clearer in the next iteration of the document.

ros-discourse:

https://discourse.ros.org/t/new-ros-ecosystem-paper/17261/8

PR #303, opened 2020-10-12
Describe wildcard usage in parameter files

jacobperron:
Wildcards were introduced in https://github.com/ros2/rclcpp/pull/762
And further usage proposed in https://github.com/ros2/rclcpp/issues/1265.

### DIFF
--- a/articles/160_ros_command_line_arguments.md
+++ b/articles/160_ros_command_line_arguments.md
@@ -183,8 +183,7 @@ For example,
 will set the parameter `string_param` on all nodes,
 
 ```yaml
-/**:
-  some_node:
+/**/some_node:
     ros__parameters:
       string_param: foo
 ```
@@ -192,8 +191,7 @@ will set the parameter `string_param` on all nodes,
 will set the parameter `string_param` on nodes named `some_node` in any namespace,
 
 ```yaml
-/foo:
-  "*":
+/foo/*:
     ros__parameters:
       string_param: foo
 ```
--- a/articles/160_ros_command_line_arguments.md
+++ b/articles/160_ros_command_line_arguments.md
@@ -167,6 +167,39 @@ another_node:
      string_param: bar
 ```
 
+Wildcards can be used for node names and namespaces as described in [Remapping Names](static_remapping.html#match-part-of-a-rule).
+`*` matches a single token delimeted by slashes (`/`).
+`**` matches zero or more tokens delimeted by slashes.
+Partial matches are not allowed (e.g. `foo*`).
+
+For example,
+
+```yaml
+/**:
+  ros__parameters:
+    string_param: foo
+```
+
+will set the parameter `string_param` on all nodes,
+
+```yaml
+/**:
+  some_node:
+    ros__parameters:
+      string_param: foo
+```
+
+will set the parameter `string_param` on nodes named `some_node` in any namespace,
+
+```yaml
+/foo:
+  "*":
+    ros__parameters:
+      string_param: foo
+```
+
+will set the parameter `string_param` on any node in the namespace `/foo`.
+
 #### Logging level assignments
 
 Minimum logging level can be externally set either globally or per logger using the `--log-level` option.


fujitatomoya (approved):

articles/160_ros_command_line_arguments.md
```diff
@@ -167,6 +167,39 @@ another_node:
...
+  some_node:
+    ros__parameters:
+      string_param: foo
+```
+
+will set the parameter `string_param` on nodes named `some_node` in any namespace,
+
+```yaml
+/foo:
+  "*":
```
ivanpauno: I would prefer:

```yaml
/foo/*:
  ros__parameters:
    string_param: foo
```

articles/160_ros_command_line_arguments.md
```diff
@@ -167,6 +167,39 @@ another_node:
...
+    string_param: foo
+```
+
+will set the parameter `string_param` on all nodes,
+
+```yaml
+/**:
+  some_node:
+    ros__parameters:
+      string_param: foo
```
ivanpauno: same here, I prefer the following:

```yaml
/**/some_node:
  ros__parameters:
    string_param: foo
```

ivanpauno (commented):
Both additions seem like reasonable options (matching all nodes in a namespace, matching all nodes with a name).

I'm not sure if this really has many use cases or not.
If not, I wouldn't add it.

hidmic:
Overall LGTM, but I will say that along with https://github.com/ros2/rclcpp/pull/762, https://github.com/ros2/rclpy/pull/370 was introduced. Whatever changes are proposed have to be applied to both client libraries IMHO.

jacobperron:
>  Whatever changes are proposed have to be applied to both client libraries

I'll follow-up to make sure that this happens.

articles/160_ros_command_line_arguments.md
jacobperron: That's more neat, thanks! cc6c0ff

articles/160_ros_command_line_arguments.md
jacobperron: cc6c0ff

jacobperron:
IMO, it would be nicer if the implementation for wildcard matching was done in `rcl`, then the client libraries would get it for free.

ivanpauno:
> IMO, it would be nicer if the implementation for wildcard matching was done in `rcl`, then the client libraries would get it for free.

I think that's possible, it only requires a refactor of the current implementation.

hidmic:
> I think that's possible, it only requires a refactor of the current implementation.

Hmm, I think I fiddled with it some time ago. Or @Lobotuerk did. We require extra API, as the wildcard matching cannot happen until we have the fully qualified node name.

ivanpauno (approved):
LGTM!

ivanpauno:
Checking https://github.com/ros2/rcl/pull/809, I have some doubts if some things are valid or not:

1. `/*`
1. `/asd/*/bsd`
1. `/asd/**/bsd`
1. `/*asd`
1. `/asd*`
1. `/a*sd`
1. `/**asd`
1. `/asd**`
1. `/a**sd`

4 to 9 should be all invalid IMO, `*` and `**` should always be between slashes.
1 should work, because `/asd/*` works (and I don't think namespace `/` should be special).
2, 3 also sound like reasonable things to support, but I'm not sure if they are really useful or not.

jacobperron:
@ivanpauno I completely agree with all your examples. IIUC, the logic added to rcl in ros2/rcl#809 *does* consider partial matching  invalid (e.g. examples 4 to 9).

jacobperron:
It might be the case that example 1 isn't considered valid in ros2/rcl#809, I'd have to check.

hidmic:
> It might be the case that example 1 isn't considered valid in ros2/rcl#809, I'd have to check.

We should follow-up.

hidmic (approved):

iuhilnehc-ynos:
@ivanpauno @jacobperron 

I didn't notice your comment before. https://github.com/ros2/design/pull/303#issuecomment-718743819

https://github.com/ros2/rcl/pull/809 expects items from 4 to 9 to be invalid (`EXPECT_FALSE(res)` with the `rcl_parse_yaml_file`).
image

Am I missing something?

fujitatomoya:
@iuhilnehc-ynos i think everything is covered as expected on https://github.com/ros2/design/pull/303#issuecomment-718743819.

@jacobperron @ivanpauno i will go ahead to merge this. if anything is missing, let us know!

PR #305, opened 2020-10-21
Events Executor design

ghost:
This PR presents a design for a new type of executor, with the purpose of addressing some of the long-standing issues of the current implementation.
This follows the idea presented in the discourse post
https://discourse.ros.org/t/ros2-middleware-change-proposal/15863

The new executor uses an events queue and a timers manager as opposed to waitsets, to efficiently execute entities with work to do.

Developed by iRobot
Mauro Passerino
Lenny Story
Alberto Soragna

----

Connects to:

- rmw interface:
  - https://github.com/ros2/rmw/pull/286
  - https://github.com/ros2/rmw_implementation/pull/161
- client libraries:
  - https://github.com/ros2/rcl/pull/839
  - https://github.com/ros2/rclcpp/pull/1579
- rmw implementations:
  - https://github.com/ros2/rmw_fastrtps/pull/468
  - https://github.com/ros2/rmw_cyclonedds/pull/256

### DIFF
--- a/articles/qos_deadline_liveliness_lifespan.md
+++ b/articles/qos_deadline_liveliness_lifespan.md
@@ -68,7 +68,6 @@ For Publishers it establishes the level of reporting that they will provide to S
 Topics will support the following levels of liveliness:
 - LIVELINESS_SYSTEM_DEFAULT - Use the ROS specified default for liveliness (which is LIVELINESS_AUTOMATIC).
 - LIVELINESS_AUTOMATIC - The signal that establishes a Topic is alive comes from the ROS rmw layer.
-- LIVELINESS_MANUAL_BY_NODE - The signal that establishes a Topic is alive is at the node level. Publishing a message on any outgoing channel on the node or an explicit signal from the application to assert liveliness on the node will mark all outgoing channels on the node as being alive.
 - LIVELINESS_MANUAL_BY_TOPIC - The signal that establishes a Topic is alive is at the Topic level. Only publishing a message on the Topic or an explicit signal from the application to assert liveliness on the Topic will mark the Topic as being alive.
 
 In order for a Subscriber to listen to a Publisher's Topic the level of liveliness tracking they request must be equal or less verbose than the level of tracking provided by the Publisher and the time until considered not alive set by the Subscriber must be greater than the time set by the Publisher.


ros-discourse:

https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/20

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+
+Many articles have discussed the CPU overhead of the default ROS 2 executor the `SingleThreadedExecutor` and of its improved version the `StaticSingleThreadedExecutor`.
+ - [Reducing ROS 2 CPU overhead by simplifying the ROS 2 layers](https://discourse.ros.org/t/reducing-ros-2-cpu-overhead-by-simplifying-the-ros-2-layers/13808)
+ - [SingleThreadedExecutor creates a high CPU overhead in ROS 2](https://discourse.ros.org/t/singlethreadedexecutor-creates-a-high-cpu-overhead-in-ros-2/10077)
+ - [ROS 2 Middleware Change Proposal](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863)
+
+We can identify some major contributors to this overhead:
+ - The cost of modifying and updating waitsets.
+   Currently this operation happens multiple times for every iteration of the executor, even if the majority of robotics systems are mostly static, i.e. the number of ROS 2 entities is fixed and known at startup.
+   A detailed walk-through of the waitset management process can be found [here](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/6).
```
ivanpauno: I think that a one of the things that makes the performance of our currents wait set based executors bad is this.

In each iteration, we're collecting entities again and adding them to the rcl wait set, that adds a lot of overhead.
If instead of "nulling" the entities that aren't ready in the input buffer we would have a separate output buffer, I think that much of the overhead could be avoided. The steps described [here](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/6) as `Clear wait set`, `Add each entity to the rcl wait set`, `Add each entity to the rmw wait set` won't be needed anymore. The only needed step would be to "zero" the output buffer. The "output buffer" could be a fixed size buffer, to avoid dynamic allocations which would damage performance.

I think that that change would avoid most of the "dance" we currently do to manage waitsets.
Was this approach explored?
I think it's something interesting to explore, in order to understand what is the real limitation of the current approach.

PS: I think that an event based executor is still a valid approach, but I would like to understand the limitations of the current executor better.

---

By the way, if you have done any profiling of the current executors in your testing it would be really interesting if you can share the resulting flamegraphs (or any other output).

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+ - Inefficiency of the timers management.
```
ivanpauno: :+1:
How timers are currently handled adds a lot of overhead, and it also doesn't work for simulated time well (particularly, RTF>1 isn't working at all AFAIR)

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+The `EventsExecutor` is characterized by two main components, i.e. the `EventsQueue` and the `TimersManager`.
+
+### Brief Overview
+
+In order to execute generic ROS 2 entities (i.e. everything except timers), the `EventsExecutor` relies on an event queue.
+Whenever a ROS 2 entity that is associated to the executor has work to do, it will push an event into the executor's queue.
+Then the executor can process these events in a FIFO manner, without need for expensive entities look-ups.
+Processing an event results in different operations depending on the entity that generated it.
+
+Timers are executed in a separate task by a timers manager, where they are kept in a priority queue sorted according to their expiration time. The task then has to monitor only the first element of the queue and can execute its callback as soon as it expires.
```
ivanpauno: I think this proposal is valid also in the case of the current wait set based executors.

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+The CPU overhead of ROS 2 with respect to other pub/sub systems (as the same DDS middlewares that it uses under the hood), makes it difficult to integrate ROS 2 on embedded platforms with constrained resources, as a not negligible amount of CPU is needed just for the executor to identify entities with work to do.
+ 
+## Requirements
+ 
+The following requirements have been taken account for the design of a new ROS 2 executor:
+ - Good performance: the executor shall not be the CPU bottleneck of a ROS 2 system, thus allowing to obtain performance comparable to the ones of the underlying middleware that is being used.
+ - Scalability: The executor must perform well independently of the system size, e.g. the latency for processing a fixed amount of messages must be independent from the number of entities.
+   This can flow into the "You don't pay for what you don't use" paradigm.
+ - Extensibility: The executor should set a base from which it is straight-forward to add more functionalities without impacting performances or requiring heavy modifications of the ROS2 core libraries.
+ - Ordering of events: The executor should process events in the same order they have occurred.
```
ivanpauno: This is an interesting point, and in the wait set based approach keeping events ordered is maybe more difficult (I'm not sure if it's even possible or not).

There are still some corner cases in an event based executor, as described at the end of the article and in [this comment](https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/18?u=ivanpauno).

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+
+![Overview](../img/events_executor/overview.png)
+
+
+### EventsQueue
+
+The `EventsExecutor` requires that all the entities (except ROS 2 timers) as soon as they have some work to do, they push an event into the executor's `EventsQueue`.
+The events are produced in the RMW layer, i.e. where the underlying middleware notifies the ROS 2 entity that there is new work to do, on the other hand, events are executed in the rclcpp layer by the `EventsExecutor`.
+The event data structure that is pushed into the `EventsQueue` must contain all what is needed for the executor to be able to process that particular event.
+The data structure so includes the type of the entity that generated the event and a handle to its corresponding rclcpp object.
```
ivanpauno: The description seems to be focused in `rclcpp`.
Can the same event based approach be used in other client libraries?

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+Alternative implementations may involve the `EventsExecutor` to keep weak pointers to the entities and lock them before executing any event.
+
+### Open Issues
+
+The `EventsQueue` does not have a knowledge of QoS settings, this may cause some issues in some particular situations.
+
+##### Unbound growth of event queue
+
+As long as events are pushed into the queue, this will keep growing, regardless of the actual history size of the underlying middleware entity.
+This may cause several events to accumulate, for example while the `EventsExecutor` is not spinning.
```
ivanpauno: I think that the big advantage that the wait set based executor has is that it doesn't suffer for this two issues (that's already handled by the queue of each entity based on its QoS settings).

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+Whenever a ROS 2 entity that is associated to the executor has work to do, it will push an event into the executor's queue.
+Then the executor can process these events in a FIFO manner, without need for expensive entities look-ups.
+Processing an event results in different operations depending on the entity that generated it.
+
+Timers are executed in a separate task by a timers manager, where they are kept in a priority queue sorted according to their expiration time. The task then has to monitor only the first element of the queue and can execute its callback as soon as it expires.
+
+![Overview](../img/events_executor/overview.png)
+
+
+### EventsQueue
```
ivanpauno: Based on the description, it seems that the event queue executor can be only single threaded.
Is my understanding right? Or can events been taken from different worker threads?

ivanpauno (commented):
Thanks for working on this!
I think that executor performance is something ROS 2 should really improve.

I have left some comments about the event based approach, but my overall feedback is that we should figure out what's a good default executor and only maintain one in `rclcpp`.
This doesn't mean that the one in `rclcpp` should be wait set based, I only mean that I would rather have one single threaded executor in `rclcpp` (which could be event queue based) instead of three.

If we're going to switch to a event based executor, it would be great to have more information to understand what's the real limit of the current wait set based approach (see my other comments).

I also think that `rmw`/`rcl` API should be flexible enough so you can implement an event queue/work queue/wait set based executor out of tree, and I think that would be possible.
Based on your experience, do you think it's possible to implement an event based executor out of tree? What additions would be needed on rmw/rcl to make that possible?
Off the top of my head, I think that if `rmw` API supported attaching listeners to entities, that would make this possible.

Some extra notes about simplifying code in rclcpp:
- AFAIR the `StaticSingleThreadExecutor`, isn't actually static. It rebuilds the wait set when an entity is added, but it uses a synchronization mechanism instead of always rebuiling it (which is much more efficient).
  Based on that, it sounds like the `StaticSingleThreadExecutor` is always a better choice than the `SingleThreadExecutor` so I think that replacing the later should be completely removed.
- We also have a `rclcpp::WaitSet` that we don't use in any of the wait set based executors.

articles/events_executor.md
alsora: I don't think that an approach with only a wait set can provide correct ordering of events at all.
In a wait set, each entity can only signal if it has work to do or not, but not how many "events" have to be processed.

If you have 2 ready subscriptions, one with 1 message and the other with 2 messages, it will be quite complex to figure out the correct order for processing those messages.
You will definitely need another structure on top of the wait set to order the messages correctly.

articles/events_executor.md
alsora: We are not really familiar with the state of other client libraries. That's why the design (and the PR with the implementation) are focusing on C++ client.

From my understanding the executor is a concept that is not shared among client libraries.
However, all the infrastructure for pushing events can definitely be used by all client libraries.

It is also possible to move the event queue itself and parts of the timers manager (e.g. the storage and sorting layer) to rcl, such that all client libraries can use them and have only to re-implement the executor APIs.

articles/events_executor.md
alsora: Yes, the wait set approach does not suffer from the unbound growth issue.
I'm sure it is possible to address that also in an event based executor, but is definitely not simple.

On the other hand, the second open issue mentioned here (i.e. that in some situations the order may be invalidated) is simply completely ignored by the wait set approach, where there are no guarantees at all about the order in which entities are processed.

articles/events_executor.md
alsora: We have a small note in the `Events Execution` section that mention that it is possible to have multiple workers thread.
If this overall proposal get through, one of our next action items would be to take the time to design also that part, but it should be quite straight-forward.

articles/events_executor.md
alsora: Yes, the TimersManager is on purpose a separate entity from the executor to have multiple parts of the system (or even application themselves) to take advantage of it.

alsora:
@ivanpauno thank you for taking the time to evaluate our proposal.

Here some replies to your comment.

> I have left some comments about the event based approach, but my overall feedback is that we should figure out what's a good default executor and only maintain one in `rclcpp`.
> This doesn't mean that the one in `rclcpp` should be wait set based, I only mean that I would rather have one single threaded executor in `rclcpp` (which could be event queue based) instead of three.
> 
> If we're going to switch to a event based executor, it would be great to have more information to understand what's the real limit of the current wait set based approach (see my other comments).

Yes, the ROS client library should really provide only 1 executor that works well.
The points that  we mentioned in the `Requirements` section, should be the guideline for how a ROS 2 executor should be.,
Our idea is that the `EventsExecutor` is a step forward in that direction.
The PR that we opened that implements this new executor (i.e. https://github.com/ros2/rclcpp/pull/1416) on purpose does not modify any of the existing code, but just adds new APIs.
We think that this will allow to more easily merge the changes and at the same time start the process to carefully evaluating what we have with the goal to have a single, good, executor that work in all use cases for the Galactic release.

> I also think that `rmw`/`rcl` API should be flexible enough so you can implement an event queue/work queue/wait set based executor out of tree, and I think that would be possible.
> Based on your experience, do you think it's possible to implement an event based executor out of tree? What additions would be needed on rmw/rcl to make that possible?
> Off the top of my head, I think that if `rmw` API supported attaching listeners to entities, that would make this possible.
> 

In order to implement the EventsExecutor, we had to only add some boilerplate code in rmw/rcl layers.
You can have a look at the specific PRs
rcl https://github.com/ros2/rcl/pull/839
rmw https://github.com/ros2/rmw/pull/286
rmw_fastrtps https://github.com/ros2/rmw_fastrtps/pull/468

It's just the same API copy-pasted for all the entities.

> 
>  * AFAIR the `StaticSingleThreadExecutor`, isn't actually static. It rebuilds the wait set when an entity is added, but it uses a synchronization mechanism instead of always rebuiling it (which is much more efficient).
>   Based on that, it sounds like the `StaticSingleThreadExecutor` is always a better choice than the `SingleThreadExecutor` so I think that replacing the later should be completely removed.
> 
>  * We also have a `rclcpp::WaitSet` that we don't use in any of the wait set based executors.

Yes, the StaticSingleThreadedExecutor is not really static and from our experience it always performs better than the SingleThreadedExecutor.
It has some limitations though, as it does not fully support all the executor APIs and the concept of an `ExecutableList` which keeps ownership of the entities may cause some issues.

From a discussion I had with @wjwwood I remember that the `rclcpp::WaitSet` can be configured such that it will become completely static. That should in theory allow to match the performance of the `StaticSingleThreadedExecutor`, but it will not allow to add new entities at all.

However, from our analysis, both these solutions still cause a considerable CPU overhead with respect to running your pub/sub application without ROS and directly using your middleware of choice.
On the other hand, the EventsExecutor allows to get very close to the performance of the middleware.

articles/events_executor.md
alsora: @mauropasse did a proof of concept where he took the `StaticSingleThreadedExecutor` and directly accessed the dds wait set from the rclcpp layer, thus removing all the maintenance.
He can give you more details about that.

The performance definitely improved, but not as much as we expected.
To give you an idea, we were running a 10 nodes system on a single core platform (RaspberryPi1), the CPU with `StaticSingleThreadedExecutor` was approximately 30%.
We first integrated the TimersManager and this improved to 25%.
Then, on top of this, we added the direct access to DDS wait set and the performance improved to 20%.
On the other hand, the same test with `EventsExecutor` results in 11% CPU usage.

I want to add that the numbers mentioned above were from an experiment, where we were definitely missing some mutexes, additional checks for safety and so on, that would all have added an impact on performance, while the `EventsExecutor` numbers are from our current implementation, which already includes these.

mauropasse: Hi Ivan. We have explored the "output buffer" approach, using a fixed size buffer which allowed to not rebuild the waitsets at every iteration. The output buffer was a simple ints array with the indexes of the ready-to-work entities matching the `ExecutableList` ordering (the `rmw_wait` of `rmw_cyclonedds` already works with indexes).
So, if the `ExecutableLlist` was made of:
```
rclcpp::subscription 1 // This subscription has an event (msg received)
rclcpp::subscription 2
rclcpp::subscription 3 // This subscription has an event (msg received)
```
the ouput buffer array was `[1,3]` , so we had to process subscription 1 and 3.
The waitset still has to be walked-through on `rmw_wait` though, to identify entities ready to work.
This approach lead to a improvement in CPU, but the amount of changes in the code in multiple layers didn't justify the not-so-great CPU improvement, so we discarded the idea. That's the point in which we decided that if multiple-layer code modifications were to be done, not only the CPU improvement should be noticeable but also we should fix some of the issues of the current waitset based executors. The "Requirements" section summarizes what we expected from a new design. 

Regarding your comment about profiling of the current executors, you can find a PDF `executors benchmark.pdf` here:
https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/20

Soon we'll find a way to share the benchmark files, as github doesn't allow me to attach them here.

Thanks for your time reviewing this!

ivanpauno: > The waitset still has to be walked-through on rmw_wait though, to identify entities ready to work.

Can you clarify this? I'm not completely sure what do you mean.

`rmw_wait` is currently doing two "walks" through all rmw entities:
- It's walking through all entities that were added to rmw wait set to get their corresponding DDS conditions, and attach them to the DDS wait set.
- After waiting, it walks **again** through all rmw entities to generate the output, comparing the conditions of each rmw object with **all** conditions that the DDS wait set marked as ready.

If we're not zeroing the rmw wait set in each iteration and re-adding entities, the first "walk-through" won't be needed anymore (you just store a wait set with all the conditions already attached in the rmw object).

The second "walk-through" is still needed, but how it's currently implemented is extremely sub-optimal.
We're walking through all rmw entities again, when we only need to walk through the conditions that the DDS wait set marked as ready. The number of ready conditions will be typically quite less than all the rmw entities in the wait set, so iterating over the ready conditions is much better.
e.g.: see [Connext example](http://d2vkrkwbbxbylk.cloudfront.net/sites/default/files/rti-examples/rticonnextdds-examples/examples/connext_dds/waitset_status_cond/c++/waitset_statuscond_subscriber.cxx).
A map from DDS conditions to rmw entities will do the rest of the work (in some DDS implementations supporting more modern cpp, you can even [attach a handler to a condition](https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/api/connext_dds/api_cpp2/classdds_1_1core_1_1cond_1_1StatusCondition.html#a474734a73f1093b68186c229bddb150f)).

Disclaimer: Talking about how to improve performance is much easier than actually improving it, so my analysis can be quite wrong.

Current implementations:

- [rmw_connext code](https://github.com/ros2/rmw_connext/blob/25721877d133a6ea09c1f3a864ed892d288fa918/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/wait.hpp#L341-L481)
- [rmw_fastrtps code](https://github.com/ros2/rmw_fastrtps/blob/f7fd0f56385a40bc344ba41c63eaac846c8933ac/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L196-L248)
- [rmw_cyclonedds code](https://github.com/ros2/rmw_cyclonedds/blob/3fa62f4f8fa9b3cc624879e9c085ad19aa9c1977/rmw_cyclonedds_cpp/src/rmw_node.cpp#L3366-L3385)

ivanpauno: > Regarding your comment about profiling of the current executors, you can find a PDF executors benchmark.pdf here:
https://discourse.ros.org/t/ros2-middleware-change-proposal/15863/20
> Soon we'll find a way to share the benchmark files, as github doesn't allow me to attach them here.


Ah I didn't see the pdf, thanks! I will take a closer look to it.

articles/events_executor.md
ivanpauno: > However, all the infrastructure for pushing events can definitely be used by all client libraries.

Thanks, I wanted to understand if that was possible or not.

> It is also possible to move the event queue itself and parts of the timers manager (e.g. the storage and sorting layer) to rcl, such that all client libraries can use them and have only to re-implement the executor APIs.

Though possible, I think we have kept all executor implementations in each client library, though I don't think that moving the event queue and the timers manager to rcl is needed.

articles/events_executor.md
ivanpauno: Thanks for clarifying!

articles/events_executor.md
mauropasse: Can you clarify this? I'm not completely sure what do you mean.    
    rmw_wait is currently doing two "walks" through all rmw entities:
In the current FastRTPS rmw_wait, there could be potentially four (4) whole "walks" through all waitset entities:
1. To [attach the `rmw_wait` condition variable to all entities](https://github.com/ros2/rmw_fastrtps/blob/f7fd0f56385a40bc344ba41c63eaac846c8933ac/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L123-L161) (whole waitset). There is a single condition variable shared by all entities.
2. Go again though all of them and [see if any has work to do](https://github.com/ros2/rmw_fastrtps/blob/f7fd0f56385a40bc344ba41c63eaac846c8933ac/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L169-L170) (potential whole waitset, if no entity is ready). 
3. When a condition variable has been triggered check predicate, so repeat step 2 to [check all of them until an entitiy with work to do is found](https://github.com/ros2/rmw_fastrtps/blob/f7fd0f56385a40bc344ba41c63eaac846c8933ac/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L178) (potential whole waitset, if the last guard condition of the waitset is the only with work to do).
4. Go through the whole waitset to [detach the condition variable and set to `0` the pointer of the entitiy with no work to do](https://github.com/ros2/rmw_fastrtps/blob/f7fd0f56385a40bc344ba41c63eaac846c8933ac/rmw_fastrtps_shared_cpp/src/rmw_wait.cpp#L196-L248).

As you said, the 1st one won't be needed with an outuput buffer approach.
The 2nd would be needed to run and potentially avoid the wait on the condition variable, as it's currently done.
The 3rd one would repeat step two, as works as the predicate for the triggered condition variable.
The 4th should set the output buffer (the detach part won't be needed either).

2nd, 3rd and 4th could be combined in a single function which takes care of check for work and at the same time sets the output buffer. In the worst case scenario you need to use this function twice, which is going through all the waitset.

    A map from DDS conditions to rmw entities will do the rest of the work (in some DDS implementations
    supporting more modern cpp, you can even attach a handler to a condition).

Currently there is a single DDS condition variable shared on all rmw entities, so this map won't work unless you create different condition variables for all. Attaching a handler to a condition could allow you to set the output buffer directly where the listener is notified about new data. So you end up with a "queue" of entities with work to do, which is quite similiar to what we do on the `EventsExecutor` where the handler would be "push an event in the queue" callback function.

ivanpauno: > In the current FastRTPS rmw_wait, there could be potentially four (4) whole "walks" through all waitset entities:

Thanks for the detailed explanation.
I focused my previous comment more in the `rmw_connext` implementation, which is a bit different (though it hides some of the complexity in the DDS waitset implementation).

> Then, on top of this, we added the direct access to DDS wait set and the performance improved to 20%.
On the other hand, the same test with EventsExecutor results in 11% CPU usage.

I didn't see @alsora comment before.
I think that's a good comparison to do, bare DDS waitset based performance vs the event based executor.

If CPU usage ends up being that main motivation of the change, I wouldn't like to be comparing a well optimized implementation with a badly optimized one.

gbiggs:
> I have left some comments about the event based approach, but my overall feedback is that we should figure out what's a good default executor and only maintain one in rclcpp.
This doesn't mean that the one in rclcpp should be wait set based, I only mean that I would rather have one single threaded executor in rclcpp (which could be event queue based) instead of three.

I strongly disagree with this. There is no reason that a client library should be limited to one executor. We should figure out what our use cases are and ensure that each is covered completely by one executor. If the collection of use cases requires more than one executor to completely cover, then so be it. There is also the possibility that a particular language may provide multiple ways to handle event queues and multi-threading, so one executor for each might make sense in that language's client library.

There should be one *default* executor, of course. When you call `spin()` without directly creating an executor, what executor is used inside that is an important choice we must make if we have multiple available.

articles/events_executor.md
budrus: Let me try align with your discussion

There are some differences in the rmw implementations. Depending on whether a DDS wait set or just a condition variable is used, the number of "walk-throughs" vary or are at different places. I feel that with the waitset approach there are at least these steps 

1. check all the attached conditions (rmw entities) for work to do. If there is already work go to 4. 
2. there is no work to do, so wait for work
3. after waiting check all the attached conditions (rmw entities) for work to do 
4. fill the output buffer 

As you said all the clearing, attaching and detaching with every wait can be optimized. When the the attaching and detaching to an underlying DDS wait set would be done in the context of calls like `rcl_wait_set_add_subscription()`, I'm wondering if `rmw_wait()` would no more need the rmw entities as parameter but only the rmw_wait_set

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+
+The `EventsExecutor` requires that all the entities (except ROS 2 timers) as soon as they have some work to do, they push an event into the executor's `EventsQueue`.
+The events are produced in the RMW layer, i.e. where the underlying middleware notifies the ROS 2 entity that there is new work to do, on the other hand, events are executed in the rclcpp layer by the `EventsExecutor`.
+The event data structure that is pushed into the `EventsQueue` must contain all what is needed for the executor to be able to process that particular event.
+The data structure so includes the type of the entity that generated the event and a handle to its corresponding rclcpp object.
+
+Considering the entities currently available in ROS 2, the content of the event data structure can be any of the following:
+ - `ExecutorEventType::SUBSCRIPTION_EVENT` and an identifier for a `rclcpp::SubscriptionBase` object.
```
budrus: This identifier for a SubscriptionBase is the subscription_handle that is provided with the `rmw_subscription_set_events_executor_callback` call?

alsora: Yes, exactly.
In the implementation that we proposed (see the other PRs) that's a raw pointer to an rclcpp subscription itself.
However anything that will allow the executor to get what is needed to call `execute_subscription()` will do the work: for example the executor may have a map of weak pointers to subscriptions and the event may include a key for that map.

mauropasse: Yes, it is.

articles/events_executor.md
ivanpauno: > I'm wondering if rmw_wait() would no more need the rmw entities as parameter but only the rmw_wait_set

Yeah, that sounds like it would be the case.

> Depending on whether a DDS wait set or just a condition variable is used, the number of "walk-throughs" vary or are at different places

A waitset is likely implemented with a condition variable + mutex, e.g. [here](https://github.com/eclipse-cyclonedds/cyclonedds/blob/75ba2102b342fbd42cf914bb14da8117377406b8/src/core/ddsc/src/dds__types.h#L346-L357) and [here](https://github.com/eclipse-cyclonedds/cyclonedds/blob/a65312d34ec3022e95cb9960a133c1c5d589897a/src/core/ddsc/src/dds_waitset.c#L219-L238) code partially showing what cyclonedds does.

articles/events_executor.md
budrus: I.e. it could happen that for a specific entity, there are more events in the executor queue than actually in the queue of the entity. Would the executor then try to take a message from a subscription and not execute the callback as it realized that the take did not return a message?

alsora: @budrus it is possible to do that, but it requires additional APIs.
At the moment an executor in the rclcpp layer can only call `execute_subscription(Subscription)` which will take the first element from the underlying DDS queue of the subscription.
The events executor design would allow to attach IDs to the messages/events, but then the rclcpp layer should also provide an API `execute_subscription(Subscription, id)` that will execute only if it's the correct element.
You can find more details on the problem at the bottom of the design page (i.e. `Invalid ordering corner case`)

budrus: Thanks @alsora. So it could happen that there is a `execute_subscription(Subscription)` without any action. 
This not an error when we end up [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executor.cpp#L541-L554)

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+While the `EventsExecutor` is busy processing events, the following events accumulates into the `EventsQueue`:
+ - Event 1 from Subscription A
+ - Event 2 from Subscription B
+ - Event 3 from Subscription A
+ 
+When the `EventsExecutor` is ready to process these new events, it will start from the first pushed, i.e. the event 1 from Subscription A.
+However, when calling the `execute_subscription()` API, the message that generated event 1 will not be available anymore as it has been overwritten from the message that generated event 3.
+The `execute_subscription()` API will not fail, but rather it will process the message that generated event 3.
+This violates the ordering as event 2 should have been processed before that.
+
```
budrus: The wait set and the event queue approach both indicate that there is work to do and the actual work is grabbed later. The event queue has "normally" already a temporal order and less "walk through" and sorting than the wait set, but is also not deterministic as you describe here. 

Determinism would be reached with a work queue or if the collected work is sorted before executed in case of  event queue and wait set. So determinism would be possible if you pay with some runtime. 

I'm wondering if it would be possible to have an intermediate API that provides a container with work to execute. If this is then optimized for performance, latency, determinism or however sorted is the job of the flexible part of the executor. But I'm not familiar enough with the upper layers of rcl and rclcpp to judge if this makes sense. This maybe goes in the direction of the out of tree proposal @ivanpauno had.

mauropasse: Determinism would be reached with a work queue or if the collected work is sorted
    before executed in case of event queue and wait set. So determinism would be possible
    if you pay with some runtime.

Determinism could be reached implementing what @alsora proposed above:

    The events executor design would allow to attach IDs to the messages/events,
    but then the rclcpp layer should also provide an API execute_subscription(Subscription, id)
    that will execute only if it's the correct element.

Regarding

    I'm wondering if it would be possible to have an intermediate API that provides a container with work to execute.

The issue with a work queue is that a message taken to be processed later could be invalid by the time of processing. If the QoS depth of the queue is 1 (one) and you take the message, but another message arrives before processing it, the stored message in the work queue should not be processed.

budrus: > Determinism could be reached implementing what @alsora proposed above:
> 
> ```
> The events executor design would allow to attach IDs to the messages/events,
> but then the rclcpp layer should also provide an API execute_subscription(Subscription, id)
> that will execute only if it's the correct element.
> ```

That's a pragmatic and efficient solution for a guaranteed temporal order. The middleware needs to provide a sneak preview for the id without taking the message but that's a solvable problem too  

> The issue with a work queue is that a message taken to be processed later could be invalid by the time of processing. If the QoS depth of the queue is 1 (one) and you take the message, but another message arrives before processing it, the stored message in the work queue should not be processed.

I see. That are the consequences of having two queues that are processed at different times. One in the middleware and one in the executor. The work queue looses the QoS depth, the event queue can fulfil it with the consequence of outdated NOP events as discussed above.   

What I meant with 

> ```
> I'm wondering if it would be possible to have an intermediate API that provides a container with work to execute.
> ```

is the question if it would make sense to have an executor top layer API that provides the work to do without the need to go back to the middleware again (and the races this entails). This allows the most determinism as the executor then would provide a plan that is guaranteed to be executed. But I fear this is too far away from where we are today

mauropasse: > What I meant with ... is the question if it would make sense to have an executor top layer API that provides the work to do without the need to go back to the middleware again (and the races this entails).

We have considered that approach, as it is one step closer to get better performances too. But for the sake of simplicity for a first proposal, and taking into account that we are already getting great performance improvements, we decided to leave it as a follow up improvement for the future. Looking at the CPU profiling flamegraphs we did (using rclcpp IPC) it doesn't seem to be much overhead of "go back to the middleware again" as the message passing occurs only on RCLCPP layer. The approach could have a bigger impact when not using rclcpp IPC.

articles/events_executor.md
hidmic: Though it reaches beyond the scope of this PR, bear in mind that eventually the `TimersManager` class will have to deal with different timers associated with different clocks. Perhaps a queue per clock type? Synchronous API for `spin_once()` implementation will probably suffer though.

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+Entities may be eventually added back again to the same executor or to a different one and they need to be able to start again producing events.
+This requires that the initialization procedure that is performed when entities are added to an executor is completely reversible.
+
+The events-based approach requires particular care in handling the case where entities are destroyed while still associated with an executor.
+An entity may push an event into the `EventsQueue` and then get immediately destroyed.
+This may happen before the `EventsExecutor` start to process those events or while it is processing them.
+
+The current implementation addresses this problem by providing an additional function pointer to each associated entity.
+This function pointer will be called whenever the entity's destructor is invoked and it will result in clearing the `EventsQueue` from any event that was generated by this entity. Note that this may block until the `EventsQueue` is done processing events.
```
hidmic: If the `EventsExecutor` does not own entities, what would prevent entity destruction during event processing (and the resulting deadlock)?

articles/events_executor.md
hidmic: I'm surprised statically allocated (bounded) buffers were not a requirement for resource constrained platforms -- often running time-sensitive applications. Large enough circular buffers would probably be OK for most such cases (whatever large enough means for any given application).

articles/events_executor.md
```diff
@@ -0,0 +1,233 @@
...
+The `EventsExecutor` requires that all the entities (except ROS 2 timers) as soon as they have some work to do, they push an event into the executor's `EventsQueue`.
+The events are produced in the RMW layer, i.e. where the underlying middleware notifies the ROS 2 entity that there is new work to do, on the other hand, events are executed in the rclcpp layer by the `EventsExecutor`.
+The event data structure that is pushed into the `EventsQueue` must contain all what is needed for the executor to be able to process that particular event.
+The data structure so includes the type of the entity that generated the event and a handle to its corresponding rclcpp object.
+
+Considering the entities currently available in ROS 2, the content of the event data structure can be any of the following:
+ - `ExecutorEventType::SUBSCRIPTION_EVENT` and an identifier for a `rclcpp::SubscriptionBase` object.
+ - `ExecutorEventType::SERVICE_EVENT` and an identifier for a `rclcpp::ServiceBase` object.
+ - `ExecutorEventType::CLIENT_EVENT` and an identifier for a `rclcpp::ClientBase` object.
+ - `ExecutorEventType::WAITABLE_EVENT` and an identifier for a `rclcpp::Waitable` object.
```
hidmic: Thinking out loud, I wonder why combining polymorphism and type tagging, as opposed to just using polymorphism (i.e. making everything a `Waitable`).

hidmic (commented):
> I also think that rmw/rcl API should be flexible enough so you can implement an event queue/work queue/wait set based executor out of tree, and I think that would be possible.

I'm with @ivanpauno here. Not against `EventsExecutor` being part of the core, but in favor of putting the machinery in place for out-of-tree implementations to be the norm.

articles/events_executor.md
alsora: Could you clarify what's the use-case for having timers with different clock coexist in the same executor?

articles/events_executor.md
alsora: The EventsExecutor does not own entities mainly to align with the current ROS 2 design paradigm where there are no APIs for removing/destroying entities.

There should not be a deadlock: the entity destructor and the event queue processing thread will all try to lock the same single mutex.

Assume the following: entity A and entity B are being destroyed and they both have events on the queue.
If A acquires the lock first, while holding this lock it can remove its own events from the queue.
Upon release the events executor may acquire the lock and it will execute B events.
Then B may acquire the lock and it will not have anything to remove.

There are definitely other solutions to handle this situation.

articles/events_executor.md
alsora: I agree with that, but it would be a quite big change for ROS as currently the waitable and the subscriptionBase, serviceBase, etc are very distinguished from each other.

We really tried to provide a design that allows to implement the new executor while not disrupting the current ROS system.

articles/events_executor.md
mauropasse: In the current implementation, mutex prevents entity destruction during event processing:
https://github.com/irobot-ros/rclcpp/blob/irobot/add-events-executor/rclcpp/include/rclcpp/executors/events_executor.hpp#L216
https://github.com/irobot-ros/rclcpp/blob/irobot/add-events-executor/rclcpp/include/rclcpp/executors/events_executor.hpp#L225

articles/events_executor.md
hidmic: > Could you clarify what's the use-case for having timers with different clock coexist in the same executor?

The simplest one I can think of is a node that unconditionally requires a timer to be associated with the system clock (e.g. to implement some I/O timeout) but otherwise associates all other timers to the ROS clock for proper operation in simulation whenever RTF != 1.

articles/events_executor.md
hidmic: > In the current implementation, mutex prevents entity destruction during event processing

Hmm, that's true when destruction and execution happen in separate threads. But unless I'm missing something from your `rclcpp` based prototype (which I might), if any entity associated with the event executor is destroyed within one of the callbacks that consume events [while `execution_mutex_` is being held](https://github.com/irobot-ros/rclcpp/blob/6c55f0bd79ef806825f0edb1e4abd7402ece75ac/rclcpp/src/rclcpp/executors/events_executor.cpp#L68), on entity removal [it will attempt to lock the `execution_mutex_` again](https://github.com/irobot-ros/rclcpp/blob/6c55f0bd79ef806825f0edb1e4abd7402ece75ac/rclcpp/include/rclcpp/executors/events_executor.hpp#L225). And `execution_mutex_` isn't recursive.

articles/events_executor.md
alsora: How do the existing executors handle this situation?
In general I think that mixing timers with different clocks in the same execution context can easily lead to errors.

articles/events_executor.md
alsora: Yes, that situation can cause a deadlock and making that mutex recursive should be enough to handle it.

However, again I'm not sure what the use-case for that could be.
In the current design the event types are used only to  distinguish between entities.
This means that you would have a regular subscription that, depending on the content of the message it receives, could decide to destroy itself or maybe to destroy another entity?
I think that ROS does not really support this workflow as entities are usually handled with shared ownership.

A different solution that we evaluated consisted in having additional event types to denote that we want to remove an entity from the executor.
However this does not really implements easily as ROS lacks several APIs for removing/destroying entities.

articles/events_executor.md
hidmic: They don't :). Thus the disclaimer above:

> Though it reaches beyond the scope of this PR, bear in mind that eventually the TimersManager class will have to deal with different timers...

It's a known issue that hopefully we can overcome at some point in the future.

articles/events_executor.md
hidmic: > I think that ROS does not really support this workflow as entities are usually handled with shared ownership.

As far as I can see, weak ownership of nodes and callback groups in executors would not prevent entity or even node destruction (e.g. node A tears down another node B while servicing a request). Whether that's useful or desirable is of no consequence: the implementation must behave in a sane way.

alsora: The `StaticSingleThreadedExecutor` has shared ownership of entities. This model does not allow a callback to destroy other entities, as the executor will always keep 1 reference to them. It actually does not really allow to destroy entities at all without first removing the node from the executor.
The `SingleThreadedExecutor` has weak ownership of entities. This model allows a callback to destroy other stuffs.

With the `EventsExecutor` we did something different from both.
The weak ownership model is not performant as the "fast path" require you to lock all the entities every time. I haven't measured this overhead, so maybe it's not really comparable to the other performance issues in ROS, but it's definitely something that eventually needs to be addressed.

I'm fully conscious that the ownership of entities is a weak point of the current implementation, but IMO is something not related to the `EventsExecutor` design in particular, but rather to its implementation or to the executors in general.
I would prefer a shared ownership model where additional APIs allow you to easily handle entities (e.g. add/remove from node and add/remove from executor).

Considering your scenario (a callback that destroys an entity) it could first remove that entity from the executor and then destroy it.
However this would require many more changes to ROS.

wjwwood: tl;dr I agree with @hidmic here:

> Whether that's useful or desirable is of no consequence: the implementation must behave in a sane way.

To merge this, we should keep the existing ownership behavior, and therefore preserve the ability to "delete" (or start asynchronously deleting) entities that might be being used by the executor from within a callback.

Specifically, the executor should have weak ownership on the entities associated with it, and strong ownership when using them.

----

> I'm fully conscious that the ownership of entities is a weak point of the current implementation, but IMO is something not related to the EventsExecutor design in particular, but rather to its implementation or to the executors in general.

If you want it to be as much of a "drop-in replacement" as possible (and I think that's a good thing for comparing them and making it easier to integrate) then the `EventsExecutor` should follow the existing behavior of shared ownership (i.e. user owns, executor has weak ownership). If it really is as decoupled as you say, then changing the executor implementation AND the sharing behavior is changing two things at once, and is generally not desired.

> The weak ownership model is not performant as the "fast path" require you to lock all the entities every time. I haven't measured this overhead, so maybe it's not really comparable to the other performance issues in ROS, but it's definitely something that eventually needs to be addressed.

I think that should be evaluated and addressed separately from this proposed change in the executor design. It's not clear to me that "but it's definitely something that eventually needs to be addressed." is true.

> I would prefer a shared ownership model where additional APIs allow you to easily handle entities (e.g. add/remove from node and add/remove from executor).

I don't believe this is substantially different from the current approach, because it still doesn't let you remove items from within a callback. You'd really only be able to have `remove` (which doesn't work in a callback) and `async_remove`/`remove_later` which would work in a callback, but that's not any different than releasing ownership of the shared pointer within a callback, imo.

hidmic: > The `StaticSingleThreadedExecutor` has shared ownership of entities. This model does not allow a callback to destroy other entities, as the executor will always keep 1 reference to them. It actually does not really allow to destroy entities at all without first removing the node from the executor.

That's a good point. I see shared ownership of entities in [`rclcpp::experimental::ExecutableList`](https://github.com/ros2/rclcpp/blob/9c62c1c9463cd2accee57fe157125950df50f957/rclcpp/include/rclcpp/experimental/executable_list.hpp#L67-L86).

> Considering your scenario (a callback that destroys an entity) it could first remove that entity from the executor and then destroy it.
However this would require many more changes to ROS.

I agree with you that we're digging a bit too much into implementation details here, but how so? Again, I could be wrong, but considering that callback groups have weak ownership and that removing an entity from the `EventsExecutor` effectively drops the push event callback and prunes related events, you just have to ensure you synchronize properly (as mentioned above). It won't trigger an event, sure, but that's a non-issue. Explicitly removing a node is safe too since you use guard conditions. Your design is pretty solid in that sense.

As @wjwwood says, we can discuss ownership later on. I'd much rather get your work reviewed and merged instead of growing the scope right now. With an emphasis on the `rmw` changes that enable this implementation and (potentially) future ones.

alsora: Ok, thank you all for the discussion.
We are updating the rclcpp implementation to have weak ownership of entities.
This will allow us also to measure the difference in performance between the two approaches.

mauropasse: image

These are the results including the implementation of the EventsExecutor having weak ownership of entities, and strong ownership when using them. We'll soon push this changes to the rclcpp PR.

mauropasse: **Update:** We've improved the weak pointer implementation for the new ownership model. The performances are now much better, using just 5% more CPU than the no-ownership approach (which used raw pointers thus no `lock()` needed to get ownership) for the worst case scenario (our biggest topology at 100hz in a single core platform). The graph above has been updated reflecting the latest performances, as well as the `rclcpp` PR https://github.com/ros2/rclcpp/pull/1416 including the improvements.

wjwwood:
Ok, I've been making progress on the rmw changes related to this. One big thing I did was remove the term `listener` from these pull requests. My reasoning being that we don't have any concept like "listener" in ros yet, and we might want it in the future, but for now, how the callbacks are called isn't really important. So I removed those in favor of names like `rmw_subscription_set_on_new_message_callback` rather than `rmw_subscription_set_listener_callback`. I'm open to feedback on this, but I think it was a good change.

I also improved the documentation and implementation for `rcl`.

I did a major refactor to `rclcpp`, making the callbacks type safe and exception safe. I did some other things too, like dropping the setter function for waitables, but I'll add more notes about that on the pr itself.

I still think we need some tests for `rcl` and/or `rclcpp`, and we need a implementation (even a dummy one) for `rmw_connextdds`. I'm not sure how much more I can get done tonight, but I'll at least run CI on it.

ros-discourse:

https://discourse.ros.org/t/ros2-speed/20162/9

adamdbrw:
@wjwwood is this work still ongoing?

wjwwood:
@adamdbrw Yes, but slowly. There are pull requests implementing needed features for some of what is included here, but they didn't make it into galactic, and have been mostly ideal since, which is my fault. I expect we'll start making progress again soon.

alsora:
@adamdbrw FYI we have branches where we implemented the whole design and that we are using for our testing/development.

The PRs mentioned in the first post implement the first part of the design (i.e. RMW listener APIs).
The executor itself can be found in this branch https://github.com/irobot-ros/rclcpp/tree/galactic-events-executor
This branch already includes the rclcpp PR mentioned in the first post and adds the events executor on top of it

articles/events_executor.md
ZhenshengLee: @alsora Hi, I am tring to reuse this great work about event based executor.
But could you show me some code to demostrate the mean of "direct access to DDS wait set"?

Because from all rmw codes what I have read, the user code actually get direct access to dds wait set, it does a `static_cast` type conversion,  for example , see rmw_iceoryx `rmw_wait()`

`https://github.com/ros2/rmw_iceoryx/blob/foxy/rmw_iceoryx_cpp/src/rmw_wait.cpp`

```cpp
rmw_wait(
  rmw_subscriptions_t * subscriptions,
  rmw_guard_conditions_t * guard_conditions,
  rmw_services_t * services,
  rmw_clients_t * clients,
  rmw_events_t * events,
  rmw_wait_set_t * wait_set,
  const rmw_time_t * wait_timeout)
{
  RCUTILS_CHECK_ARGUMENT_FOR_NULL(subscriptions, RMW_RET_ERROR);
  RCUTILS_CHECK_ARGUMENT_FOR_NULL(guard_conditions, RMW_RET_ERROR);
  RCUTILS_CHECK_ARGUMENT_FOR_NULL(services, RMW_RET_ERROR);
  RCUTILS_CHECK_ARGUMENT_FOR_NULL(clients, RMW_RET_ERROR);
  RCUTILS_CHECK_ARGUMENT_FOR_NULL(events, RMW_RET_ERROR);
  RCUTILS_CHECK_ARGUMENT_FOR_NULL(wait_set, RMW_RET_ERROR);

  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
    rmw_wait
    : waitset, wait_set->implementation_identifier,
    rmw_get_implementation_identifier(), return RMW_RET_ERROR);

  iox::popo::WaitSet<iox::MAX_NUMBER_OF_ATTACHMENTS_PER_WAITSET> * waitset =
    static_cast<iox::popo::WaitSet<iox::MAX_NUMBER_OF_ATTACHMENTS_PER_WAITSET> *>(wait_set->data);
//...
}

```

So I am confused about the difference between direct-access and non-direct-access to dds wait_set? .

Thanks for your time.

alsora:
Hi, we updated the design with some of the latest changes originated from the middelware WG discussions.

Joeyyzhao:
> @adamdbrw FYI we have branches where we implemented the whole design and that we are using for our testing/development.
> 
> The PRs mentioned in the first post implement the first part of the design (i.e. RMW listener APIs). The executor itself can be found in this branch https://github.com/irobot-ros/rclcpp/tree/galactic-events-executor This branch already includes the rclcpp PR mentioned in the first post and adds the events executor on top of it

Hi alsora, when I use this branch(https://github.com/irobot-ros/rclcpp/tree/galactic-events-executor) to build ros2, I meet the compiling error "rclcpp/rclcpp/src/rclcpp/timer.cpp:150:9: error: use of undeclared identifier 'rcl_timer_get_next_call_time', and there is no explicit definition of this function in whole project, how can I fix this error?

timonegk:
Hi, we recently ran into the performance issues in ROS2 and tried the EventsExecutor from the iRobot forks and it worked phenomenally.
@wjwwood @alsora What is currently blocking the progress of merging the EventsExecutor, especially to rclcpp? As far as I can see, part of the implementation in https://github.com/ros2/rclcpp/pull/1416 was merged in https://github.com/ros2/rclcpp/pull/1579, but the executor itself was not merged and there is not even a pull request for it.
As the current state of the ROS2 executors is not really usable for high frequency systems, this would be a great improvement and I / we would be glad to help move this forward wherever we can.

timonegk:
FYI, before someone else also does the same thing: We really needed the executor, so I rebased the work by @irobot-ros on the current master. This was mostly `rclcpp`, because most of the other stuff was already merged. I also needed a small number of additional commits for `rmw_fastrtps` and `rcl`.
The repositories can be found at 
- https://github.com/timonegk/rmw_fastrtps 
- https://github.com/timonegk/rclcpp
- https://github.com/timonegk/rcl

In addition, I created packages for ubuntu 22.04. They are available at the following release repositories:
- https://github.com/timonegk/ros-rolling-rcl-release
- https://github.com/timonegk/ros-rolling-rclcpp-release
- https://github.com/timonegk/ros-rolling-rmw-fastrtps-cpp-release
- https://github.com/timonegk/ros-rolling-rmw-fastrtps-dynamic-cpp-release
- https://github.com/timonegk/ros-rolling-rmw-fastrtps-shared-cpp-release

A strange bug I experienced was that I had to revert https://github.com/ros2/rmw_fastrtps/pull/583 because an invalid qos type was falsely triggered. I don't know why that happened but will investigate.

I hope this helps someone who also needs the executor, or it helps you when it will be merged.

alsora:
We plan to create soon an "events executor" repository that only contains the executor and it's compatible with standard rclcpp library.

This will allow ROS 2 Humble users to use the events executor in an easy way.

In the meantime, we will also open new PRs to rclcpp to merge this executor in the mainline.

Flova:
@alsora any updates on the executor repo? :)

mauropasse:
@timonegk this PR https://github.com/ros2/rmw_fastrtps/pull/625 fixes the issue you got about an invalid qos type falsely triggered. So you don't have to revert the FastRTPS PR https://github.com/ros2/rmw_fastrtps/pull/583

timonegk:
@mauropasse Thanks, works great!

Aposhian:
What are the next steps for this? Is this something that needs to be brought up in the middleware WG?

alsora:
Hi, thank you everyone for your interest in the events-executor.
You can find here a Github repository containing an implementation of the executor for ROS 2 Humble: https://github.com/irobot-ros/events-executor

We plan to release this as a debian package as soon as the known limitations and bugs mentioned in the readme are addressed.
Contributions and feedbacks are highly appreciated!

ijnek:
@alsora Thanks for sharing you and your teams amazing work! Is the plan to get this into rclcpp, or to keep it separate in this repo? 

Once the debians are out, I can imagine people getting confused about having to add a dependency on ``irobot_events_executor`` even though the way the events_executor should be included:

```cpp
#include "rclcpp/executors/events_executor/events_executor.hpp"
```

makes it look like its from ``rclcpp``.

alsora:
Yes the plan is definitely to merge this into rclcpp, the separate repo and the debian package are just intermediate solutions to let people play with it in the meantime.

It's now almost 2 years that we have a working implementation of the events executor (we first opened a rclcpp PR with it in 2020), but it may take a while before all the pieces and requirements are merged.

The reason for having the same paths as `rclcpp` was to simplify the transition, i.e. we could copy-paste the files into a PR for rclcpp and people that were using the executor from the separate repo wouldn't have to make any change to their source code.
However I see the point for the potential confusion.

We can probably just change the top-level directory from `rclcpp` to `irobot_rclcpp`.

P.S. if you have additional feedbacks and suggestions, feel free to open a ticket on the events executor repository or code PRs

alsora:
@wjwwood @clalancette @ivanpauno this PR is open since almost 2 years.
Anything to review/modify before merging it?

sandeepsuresh:
As i understand this Event Executor is a Single threaded executor.  Is there an update to this to support multithreaded EventExecutor with Callback Groups? Please revert back.

alsora:
The events executor by design can either be single thread or multi-thread.

The implementation currently available at https://github.com/irobot-ros/events-executor support two modes:
 - single thread
 - two threads: one thread is used to process timers callbacks and another thread is used to process everything else

Contributions are welcome to support an arbitrary number of threads.

ros-discourse:

https://discourse.ros.org/t/questions-about-the-intrinsic-acquisition-of-osrc/28763/15

russkel:
So... LGTM?

wjwwood:
We've recently had a discussion about what to do with this pr, and it's possible we can merge it and immediately mark it as historical. There are a few issues at play:

- this document is now out-of-date (the events executor has changed)
- we're trying to move away from design docs on this site and towards REPs and design docs in the code repositories (maybe a version of this should go into rclcpp as we bring the events executor into rclcpp)
- there are lots of tasks competing for our limited resources and improvements to the events executor and moving it into rclcpp are higher priority than resolving this long-open pr

alsora:
Hi, something strange happened to this PR.
It was merged last month, but apparently it doesn't include any changes

PR #307, opened 2020-11-14
Create ros2_ibmcloud_kubernetes.md

mm-nasr:
This article describes how to get ROS2 running on IBM Cloud Kubernetes using Docker files. It first gives a brief overview of docker images and how they work locally and then explores IBM Cloud and how the user can push their containers on it. 
Afterwards, a short description of how the user can use their own custom packages for ROS2 from github on IBM Cloud is provided.
A walkthrough of how to create a cluster and utilize Kubernetes on IBM Cloud is provided and finally the Docker image is deployed on the cluster to run ROS2.

### DIFF
--- a/articles/ros2_ibmcloud_kubernetes.md
+++ b/articles/ros2_ibmcloud_kubernetes.md
@@ -1,0 +1,505 @@
+---
+layout: default
+title: ROS2 on IBM Cloud Kubernetes
+abstract:
+  This article describes how to get ROS2 running on IBM Cloud Kubernetes using Docker files. It first gives a brief overview of docker images and how they work locally and then explores IBM Cloud and how the user can push their containers on it. 
+  Afterwards, a short description of how the user can use their own custom packages for ROS2 from github on IBM Cloud is provided.
+  A walkthrough of how to create a cluster and utilize Kubernetes on IBM Cloud is provided and finally the Docker image is deployed on the cluster to run ROS2.
+author: '[Mahmoud Nasr](https://github.com/mm-nasr)'
+published: false
+---
+
+- This will become a table of contents (this text will be scraped).
+{:toc}
+
+# {{ page.title }}
+
+<div class="abstract" markdown="1">
+{{ page.abstract }}
+</div>
+
+Original Author: {{ page.author }}
+
+## ROS2 on IBM Cloud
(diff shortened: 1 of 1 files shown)


articles/ros2_ibmcloud_kubernetes.md
```diff
@@ -0,0 +1,509 @@
...
+CMD ["ros2", "launch", "demo_nodes_cpp", "talker_listener.launch.py"]
+```
+
+* __FROM__: creates a layer from the ros:foxy Docker image
+* __RUN__: builds your container by installing vim into it and creating a directory called /ros2_home
+* __WORKDIR__: informs the container where the working directory should be for it
+
+Of course, you are free to change the ROS distribution (_foxy_ is used here) or change the directory name.
+The above docker file sets up ROS-foxy and installs the demo nodes for C++ and Python. Then it launches a file which runs a talker and a listener node. 
+We will see it in action in just a few, but they act very similar to the publisher-subscriber example found in the [ROS wiki](https://wiki.ros.org/ROS/Tutorials/WritingPublisherSubscriber%28c%2B%2B%29)
```
gavanderhoorn, line 109: You're linking to the ROS 1 wiki here.

gavanderhoorn:
Is this in the right place? `design.ros.org` is for documentation on design and design decisions of ROS 2. Should this have been opened against [ros2/ros2_documentation](https://github.com/ros2/ros2_documentation) instead?

clalancette:
> Is this in the right place? `design.ros.org` is for documentation on design and design decisions of ROS 2. Should this have been opened against [ros2/ros2_documentation](https://github.com/ros2/ros2_documentation) instead?

Agreed, this isn't the right place for this.  @mm-nasr could you open a pull request against ros2_documentation instead?

articles/ros2_ibmcloud_kubernetes.md
mm-nasr: Yes, just because it's the classic publisher-subscriber example which is found on that link. I'll check to see if a similar link exists on ROS2 Docs.

mm-nasr:
@gavanderhoorn @clalancette Thank you for the feedback. I'll open a pull request against documentation as suggested.

PR #312, opened 2021-02-25
Fix launch_xml examples

jacobperron:
Remove 'sep' attribute and use square brackets for lists.

As far as I know, this is the correct sytax for a list value and the 'sep' attribute was never implemented.

@hidmic @ivanpauno Does this look correct?

### DIFF
--- a/articles/151_roslaunch_xml.md
+++ b/articles/151_roslaunch_xml.md
@@ -154,9 +154,9 @@ either nested or brought from a YAML file to make a map.
 ```xml
 <node pkg="my_pkg" exec="my_node">
   <param name="some_numeric_param" value="100.2"/>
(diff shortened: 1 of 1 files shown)


ivanpauno:
> As far as I know, this is the correct sytax for a list value and the 'sep' attribute was never implemented.

IIRC, you have to use `value="Some phrase,'100.0','true'" value-sep=","`

jacobperron:
> IIRC, you have to use value="Some phrase,'100.0','true'" value-sep=","

Interesting. I found some docs about it [here](https://github.com/ros2/launch/tree/91536a96d699aeb2eb20b9f59e8ef7308255a46e/launch_xml).

I've updated this PR with the correct syntax, and also added an example that doesn't use the *-sep attribute.

ivanpauno (approved):

Issue #314, opened 2021-04-08
Logging Design Document


clalancette:
In this context, I'm going to define "logging" narrowly as "getting error, warning, info, and debug messages to the user".  These messages may either be printed to a tty, or written to a file, or both.

ROS 2 currently has a number of different logging pieces:
* `rcutils` has the RCUTILS_LOG family of macros
* `rclcpp` has the `RCLCPP_LOG` family of macros
* `rcl_logging` contains backend loggers for controlling log levels and writing data out to files
* `rcl` has infrastructure for initializing the loggers, changing the levels at startup time, and controlling what the log messages look like
* `launch_ros` has various infrastructure for capturing messages from the processes it is supervising, and printing them to the tty or to files (or both)

Those pieces are all useful, but we lack a coherent story around logging and how it should work for end users.  Ideally the outcome of this design document is that we identify gaps in the current infrastructure, and then work on filling those gaps.  We should also revamp the logging documentation (https://docs.ros.org/en/rolling/Concepts/About-Logging.html and https://docs.ros.org/en/rolling/Tutorials/Logging-and-logger-configuration.html), and likely expand it.

PR #315, opened 2021-05-06
Added initial version of logging.md file.

mlanting:
This PR is intended to address https://github.com/ros2/design/issues/314 and facilitate discussion around what we want the user experience to look like.

This is a work in progress as I'm still trying to understand some areas of the launch system so I can more completely and accurately describe the status quo, but I want to get something pushed out for people to start commenting on and providing suggestions or corrections.

### DIFF
--- a/articles/logging.md
+++ b/articles/logging.md
@@ -13,14 +13,28 @@ Based on our experiences with ROS 1, other logging systems, and community feedba
 ### Interaction Descriptions
 This section describes what each of the interactions in the list entails, and the steps users will have to take to complete them.
 #### Creating Loggers
-This interaction contitutes creating the objects that users will use to output information from their code.
+This interaction constitutes creating the objects that users will use to output information from their code.
+Users are able to create multiple named loggers which they can use to exercise fine-grained control over the logging output at runtime.
+Loggers are namespaced and can be controlled hierarchically by having their severity levels set individually.
+If a logger's level is not set explicitly, it will use the level of the nearest ancestor with an explicit severity, or the default if no ancestors have had their levels set.
+All named loggers are descendants of the default/unnamed logger.
 #### Setting Log Levels
 This is how users set the severity levels for the loggers in their code.
 There are a number of ways users might want to be able to change these levels, so this interaction is further broken down to describe each of those.
-##### Programatically
+##### Programmatically
+Severity levels for each logger can be set programmatically by calling a function that takes the desired severity level, and optionally a reference to or the name of a specific logger.
+If no name/reference is provided, function will apply the severity level as the default level.
 ##### Command Line
+When launching nodes from the command line, an argument can be provided setting the default log severity for that node.
+Programmatic logging severity changes within the node's code will override the severity set by the command line argument.
+##### Via Launch File
+Severity levels can be set for loggers by name via launch files.
+The severity levels will be applied to the appropriate loggers at startup, but may be overridden by programmatic changes.
 ##### Via ROS 2 Service Call
-##### Environment Variable?
+Users will be able to control severity levels of loggers during runtime via ROS 2 service calls.
+This level of control if often useful when debugging complex dynamic systems with many moving parts, especially when there is a high startup/restart cost.
+Severity levels set this way will override levels set at launch via command line arguments or configuration files.
+Programmatic changes to the severity may still occur after a level has been changed via the service call and will override the level set by the service call.
 #### Outputting Logs
 This is primary interaction where users output information from their code.
 #### Defining Output Destinations (Sinks)
@@ -35,6 +49,8 @@ Users may have requirements that none of the included backends support, so ROS 2
 Adding a sophisticated logging system to a complex code base can have performance impacts and are often mostly useful during development and debugging.
 ROS 2 provides a few ways to configure the logging system to eliminate this performance impact when the logs are not needed.
 This section describes how to make those configuration changes.
+#### Interacting with RMW Logging
+RMW implementations have their own built-in logging, and a function called `rmw_set_log_severity` which may need to be handled with special considerations.
 
 
 
@@ -68,7 +84,7 @@ Log level is *supposed* to be able to be set via the command line while launchin
 ### Outputting Logs
 Users that want to log information should use the `rclcpp` logging macros which cover a variety of use cases at 5 different severity levels.
 ### Specifying A Backend
-`rcl_logging` contians a couple different backend implementations, as well as an interface that can be used to add additional implementations. The environment variable `RCL_LOGGING_IMPLEMENTATION` can be set at compile time to select which implementation to compile in.
+`rcl_logging` contains a couple different backend implementations, as well as an interface that can be used to add additional implementations. The environment variable `RCL_LOGGING_IMPLEMENTATION` can be set at compile time to select which implementation to compile in.
 ### Avoiding Performance Impact
 `rcutils` and `rclcpp` include a flag called `RCLCPP_LOG_MIN_SEVERITY` which can be set to compile out all macros below a certain level. Constants of the form `RCLCPP_LOG_MIN_SEVERITY_[DEBUG|INFO|WARN|ERROR|FATAL|NONE]` are provided to facilitate setting the min severity that should be compiled out. (NOTE: in `rcutils` the `RCLCPP` is replaced with `RCUTILS`).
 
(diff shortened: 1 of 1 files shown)


clalancette (commented):
Thanks for getting this started!

What I would lead with in this article is an explanation of what we want the logging to look like from a users perspective.  That is, what are the pieces of functionality that we think the logging system needs to have to be useful to a wide range of users?  This section would naturally draw on what was in ROS 1, plus other logging systems and whatever we have learned in the past 14 years since ROS 1 was released.

After that, I would lay out what pieces exist in ROS 2 today, and how they fill in the "what" that we defined above.  That would give a great starting point of gaps that need to be filled in.

Finally, we should probably have a "How" document that explains how to use the parts that we have.  This would probably not live here (probably in https://github.com/ros2/ros2_documentation instead), but it is a natural extension of the first two sections here.

mlanting:
I added a new section at the beginning trying to break the issue down into a set of discrete interactions we want the logging system to support so we can then describe how each of those interactions should work. There's still plenty of detail to add, but I want to push out the "list of supported interactions" idea and my initial stab at breaking things down so others can weigh in.

articles/logging.md
```diff
@@ -0,0 +1,75 @@
+## Desired User Experience
+This section describes the desired user interface of the ROS2 logging system.
```
clalancette: A nitpick throughout the document: we prefer "ROS 2" over "ROS2".

articles/logging.md
```diff
@@ -0,0 +1,75 @@
...
+- defining output destinations
+- specifying a backend
+- implementing a custom backend
+- tailoring performance impact
+### Interaction Descriptions
+This section describes what each of the interactions in the list entails, and the steps users will have to take to complete them.
+#### Creating Loggers
+This interaction contitutes creating the objects that users will use to output information from their code.
+#### Setting Log Levels
+This is how users set the severity levels for the loggers in their code.
```
clalancette, line 22: This section isn't only about code, so I would reword this to something like:

```suggestion
This is how users set the severity levels for the logging messages at runtime.
```

articles/logging.md
```diff
@@ -0,0 +1,75 @@
...
+There are a number of ways users might want to be able to change these levels, so this interaction is further broken down to describe each of those.
+##### Programatically
+##### Command Line
+##### Via ROS2 Service Call
+##### Environment Variable?
```
clalancette: This seems more like a solution/implementation detail, rather than a thing we want users to be able to do.  As such, I'd leave it out of this section.

articles/logging.md
```diff
@@ -0,0 +1,75 @@
...
+### Interaction Descriptions
+This section describes what each of the interactions in the list entails, and the steps users will have to take to complete them.
+#### Creating Loggers
+This interaction contitutes creating the objects that users will use to output information from their code.
+#### Setting Log Levels
+This is how users set the severity levels for the loggers in their code.
+There are a number of ways users might want to be able to change these levels, so this interaction is further broken down to describe each of those.
+##### Programatically
+##### Command Line
+##### Via ROS2 Service Call
```
clalancette: How about in launch files?  That might be covered by the other sections, but I think it bears explicitly stating.

clalancette (commented):
Thanks for iterating here.  I left some comments inline (though I only reviewed the "Desired User Experience" section again.

Overall, I think that list covers most of the pieces I think about when thinking about logging.  But I'd like to get more feedback from others, so pinging @hidmic, @sloretz , @IanTheEngineer, @ecousineau (and anyone else who wants to comment).

Beyond that, it is probably worthwhile to start fleshing out some of these sections by giving them context and motivation.  Overall this is another step in the right direction!

articles/logging.md
mlanting: lol, oops. Good catch; having that capability was one of my main motivations when I started this, I can't believe I left it out.

articles/logging.md
```diff
@@ -0,0 +1,75 @@
...
+## Existing infrastructure
+ROS2 
+### `rcutils`
+Contains macros for logging at several different levels and under various constraints
+### `rclcpp`
+Contains similar macros to those in `rcutils`, which call down to the macros in `rcutils`
+### `rcl_logging`
+Contains a logging interface and several implementations including log4cxx, spdlog, and a no-op implementation
+### `rcl`
+Contains infrastructure to tie the logger implementations defined in `rcl_logging` to the macros in `rcutils`
```
christophebedard: I think "any rmw implementation's logging system" should be mentioned here (and included wherever else applicable). The "existing infrastructure" would be `rmw`'s `rmw_set_log_severity`. See:

* https://github.com/ros2/rcl/issues/915
* https://github.com/ros2/rcl/pull/918

As mentioned in the issues above, we still have to figure out if/how we want to use an rmw implementations's logging system. If we want `--log-level` to be ROS 2-only, then maybe we should add a separate option, e.g. `--rmw-log-level`?

clalancette: You know, thinking about it further, I'm wondering if we need configuration for that to be in ROS 2 at all.  That is, if you are enabling debugging on an RMW vendor, then you are obviously trying to debug something in that particular vendor.  The vendors all have vendor-specific ways to enable debugging, so wouldn't it make sense to use one of those?

(this line of reasoning also questions the `rmw_set_log_severity` API at all)

christophebedard: yeah, that's a valid question.

> That is, if you are enabling debugging on an RMW vendor, then you are obviously trying to debug something in that particular vendor.

This is true, but

> The vendors all have vendor-specific ways to enable debugging, so wouldn't it make sense to use one of those?

this also feels like an argument in favour of an interface like `rmw_set_log_severity`.

clalancette: > this also feels like an argument in favour of an interface like `rmw_set_log_severity`.

I think it depends on whether you consider the RMW vendors to be part of ROS 2 or not.  For instance, there are probably library-specific ways to enable debugging on `libyaml` (one of our dependencies), but I don't think anyone is arguing we should do that.  So it depends on where you want to draw the line.

christophebedard: good point. Some vendor-specific configuration is expected (like FastDDS's xml config files), so maybe logging can just be bundled with that?

The middleware WG is a good place to have a discussion about where the line should be drawn!

christophebedard: FYI @clalancette @mlanting I added this discussion/question to today's middleware WG meeting agenda. Sorry for the short notice.

christophebedard: We had a short discussion about this in the last middleware WG meeting (2021-05-19). I'd recommend checking out the meeting notes.

In short: `rmw_set_log_severity` has a purpose even if there are other ways to set an rmw/DDS implementation's logging level. We could use a "well-known"/special logger name to set the logging level through the CLI (using this syntax: https://docs.ros.org/en/rolling/Tutorials/Logging-and-logger-configuration.html#logger-level-configuration-command-line). However, we have to make sure that the rmw implementation does not produce log messages if they get filtered out at the ROS 2 level (in case it is set as a log consumer).

esteve:
@clalancette @christophebedard @mlanting what would be needed to continue with this document? Is there anything I can help with?

clalancette:
> @clalancette @christophebedard @mlanting what would be needed to continue with this document? Is there anything I can help with?

So we kind of went in a different direction here (which makes this document outdated, though not necessarily obsolete).

In particular, what we did is:
1.  Add end-user documentation to the documentation via https://github.com/ros2/ros2_documentation/pull/3025
2. Do some performance work on the logging via https://github.com/ros2/rcutils/pull/367, https://github.com/ros2/rcutils/pull/369, https://github.com/ros2/rcutils/pull/370, https://github.com/ros2/rcutils/pull/372, https://github.com/ros2/rcutils/pull/374, and https://github.com/ros2/rcutils/pull/381
3. Did some additional thinking about where we want the design to go in the future in https://github.com/ros2/rcl_logging/issues/92

With all of that said, I think where we want this document to go is to be documentation about the current design of logging, along with a future section where we lay out in more detail how we would accomplish https://github.com/ros2/rcl_logging/issues/92 .  That way someone could pick it up and understand the current design, and then if they were motivated could implement the new design.

fujitatomoya (commented):
I think contents in this PR are really helpful for user application, for me this looks more like user manual perspective than design document. Why don't we add some of these sections to https://github.com/ros2/ros2_documentation?

PR #318, opened 2021-06-30
Update launch XML schema to rename <node> attribute ns to namespace and add ros_args

### LINKED ISSUE #317: Update Launch XML Schema
Similar to #285 I believe the schema doesn't match the current implementation.

In the schema, the `<node>` element currently defines a `ns` attribute but this appears to have been renamed to `namespace`. At least if [here](https://github.com/ros2/launch_ros/blob/58334b4447daf1d86aa7e17bb5e1b9ebb74301b4/launch_ros/launch_ros/actions/node.py#L319) is where it is being parsed.

@ivanpauno Mentioning you as it's similar to the other issue you dealt with.

christophebedard:
Closes #317

This renames the `<node>` `ns` attribute to `namespace` (#317). This also adds a new `ros_args` attribute, see https://github.com/ros2/launch_ros/pull/253.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>

### DIFF
--- a/articles/151_roslaunch_xml.md
+++ b/articles/151_roslaunch_xml.md
@@ -81,7 +81,7 @@ The `<group>` tag allows for launch actions' grouping as well as optional launch
 
 ```xml
 <group scoped="true">
-  <node pkg="a_ros_package" name="dummy0" ns="my_ns" exec="dummy_node"/>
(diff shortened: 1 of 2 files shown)


ivanpauno (approved):
LGTM!

Thanks for the contribution @christophebedard !!

PR #319, opened 2021-07-28
ROS2 DDS Security PKCS#11 URI support

IkerLuengo:
The DDS-Security specification defines the use of Hardware Security Modules (HSM) and PKCS#11 URIs as an alternative to private keys and certificates stored in the file system. Current implementation only supports these tokens to be directly stored in the file system as `.pem` files. This is a design proposal to support PKCS#11 URIs.

The changes affect to the RMW implementations, as these are filling the DDS security attributes for the participant. However, it also affects the contents of the enclave directories in the keystore. Although the proposed changes are totally backwards compatible (meaning that current RMW implementations will continue working if no PKCS#11 URIS are used), description of the new enclave contents and the expected RMW behavior seems appropriate.


clalancette:
@ros2/security_working_group Thoughts about this proposal?

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-08-19/22008/1

IkerLuengo:
To give some context, this proposal was discussed an approved on the @ros2/security_working_group on June ([minutes here](https://github.com/ros-security/community/pull/21/files/32c36a3edd325f13161fb63d56aa43e0bc52df87?short_path=9cda9df#diff-9cda9df9da4cc61bc08a97de404c6ac72d258c1eeca3eb5ae8459d16c23002ce), although with Sid gone, they are still unmerged).

articles/ros2_dds_security_pkcs11_support.md
```diff
@@ -0,0 +1,69 @@
...
+    │           ├── cert.pem
+    │           ├── key.pem
+    │           ├── ...
+    └── public
+        ├── ...
+
+
+Note that it also requires the names of the files to be the ones expected. Specifically, all certificate and key files must have the `.pem` extension. In order to configure the security properties of a DDS participant, the path of the appropriate file in the enclave directory is added as the `file` URI of the corresponding property. For example, for the private key in the authentication plugin:
+
+    dds.sed.auth.private_key = file:<root>/enclaves/foo/bar/key.pem
```
MiguelCompany:
```suggestion
    dds.sec.auth.private_key = file:<root>/enclaves/foo/bar/key.pem
```

MiguelCompany (changes requested):
I found a typo on the DDS property name.

fujitatomoya (approved):

fujitatomoya (commented):
Sorry i was looking at only specific commit, i will take a look at the whole fix.

articles/ros2_dds_security_pkcs11_support.md
```diff
@@ -0,0 +1,69 @@
+---
+layout: default
+title: ROS2 DDS security PKCS#11 URI support
+abstract:
+  The [DDS-Security specification][dds_security] defines the use of Hardware Security Modules (HSM) and PKCS#11 URIs as an alternative to private keys and certificates stored in the file system. Current implementation only supports these tokens to be directly stored in the file system as `.pem` files. This is a design proposal to support PKCS#11 URIs.
```
fujitatomoya, line 5: one sentence per line, and so is everywhere else. (if i am not mistaken, this is required format for doc.)

```suggestion
  The [DDS-Security specification][dds_security] defines the use of Hardware Security Modules (HSM) and PKCS#11 URIs as an alternative to private keys and certificates stored in the file system.
  Current implementation only supports these tokens to be directly stored in the file system as `.pem` files.
  This is a design proposal to support PKCS#11 URIs.
```

articles/ros2_dds_security_pkcs11_support.md
```diff
@@ -0,0 +1,69 @@
...
+
+The problem then is how to let the RMW implementation when we want to use a `file` URI and when a `pkcs11` URI; and how to provide the values of these URIs (i.e., the file path in the case of `file` and the token name in the case of `pkcs11`).
+
+This can be solved if we allow for the key and certificate files in the enclave to have `.pem` or `.p11` extensions. Files with `.p11` will contain the PKCS#11 URI instead of the actual document data. Then, the RMW can be aware of the file extension and set the security property accordingly. Taking the private key in the authentication plugin as an example:
+
+1. The RMW will look for a file with name `key.pem` or `key.p11` in the enclave.
+1. If there is a `key.pem` file, it keeps the current behavior, and sets the value of the property to the path of the file.
+1. Otherwise, if there is a `key.p11` file, it must load the content of the file, and set this content as the value of the property. Some check can be done at this point, e.g., assert that the file contains a PKCS#11 URI (i.e., starts with `pkcs11:`).
```
asorbini, line 63: Maybe I'm misreading the description, but I think that the algorithm should have the `key.p11` file have higher priority (i.e. be checked first by the RMW), and then have `key.pem` be the default, "fallback" selection.

Are there reasons for forcing the use of `key.pem` over `key.p11` if both files are present? Maybe this approach will offer better "backwards compatibility" and avoid introducing sudden changes if the enclave material now contains a `.p11` file (since the `.p11` file will only be used if the user explicitly deletes `key.pem`)?

asorbini (commented):
I had been asked to review this PR during a recent RMW WG meeting, and I'm doing so with a bit of delay. Apologies for that.

Anyway, I think the proposal is sound from a technical point of view, and I only had a minor comment on the loading order of the security material.

articles/ros2_dds_security_pkcs11_support.md
MiguelCompany: I think this would make sense. I was not on the WG meetings where this was discussed. Perhaps @quinz or @ruffsl could bring us some light

kallekoivisto: Yep, I believe the main idea was to keep the "backwards compatibility" for existing enclaves. But on the other hand there wasn't much discussion on the loading order for pem vs p11 files. I agree with @asorbini that it would be more intuitive for me to have p11 have higher priority as it can be considered also as more secure option in most of the configurations.

MiguelCompany:
@quinz @fujitatomoya @asorbini I've open #332 since I don't have write access on the repo of the original contributor.

This PR can be closed.

clalancette:
Closing based on the last comment.

PR #324, opened 2022-02-05
add explanation for `ROS_DISABLE_LOANED_MESSAGES`.

fujitatomoya:
document for https://github.com/ros2/rcl/pull/949

Signed-off-by: Tomoya Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/articles/zero_copy.md
+++ b/articles/zero_copy.md
@@ -293,6 +293,17 @@ void
 rclcpp::Subscription::handle_loaned_message(void * loaned_message, const rmw_message_info_t & message_info)
 ```
 
+### ROS_DISABLE_LOANED_MESSAGES
+
+By default, `Loaned Messages` will try to borrow memory from the underlying middleware.
+The `ROS_DISABLE_LOANED_MESSAGES` environment variable is provided so the user can disable `Loaned Messages` and fallback to normal publisher / subscription without any code change or middleware configuration.
+
+How to disable `Loaned Messages`:
+
+```
+export ROS_DISABLE_LOANED_MESSAGES=1
+```
+
 ## Additional Considerations
 
 ### Loaning non-POD messages


fujitatomoya:
@wjwwood @clalancette requesting review.

articles/zero_copy.md
```diff
@@ -288,11 +288,22 @@ rclcpp::Publisher::can_loan_messages()
 bool
 rclcpp::Subscription::can_loan_messages()
 
-/// Subscription path for middlewares supporting loaned messages.
+/// Subscription path for middleware supporting loaned messages.
```
fujitatomoya: spelling check automatically detects this change, but no need to fix?

articles/zero_copy.md
```diff
@@ -288,11 +288,22 @@ rclcpp::Publisher::can_loan_messages()
...
 void
 rclcpp::Subscription::handle_loaned_message(void * loaned_message, const rmw_message_info_t & message_info)
 ```
 
+### ROS_DISABLE_LOANED_MESSAGES
+
+In default, `Loaned Messages` will try to borrow the memory from underlying middleware.
+`ROS_DISABLE_LOANED_MESSAGES` environment variable is provided to user that allows to disable `Loaned Messages` and fallback to normal publisher / subscription w/o any code change or middleware configuration.
```
clalancette:
```suggestion
By default, `Loaned Messages` will try to borrow memory from the underlying middleware.
The `ROS_DISABLE_LOANED_MESSAGES` environment variable is provided so the user can disable `Loaned Messages` and fallback to normal publisher / subscription without any code change or middleware configuration.
```

articles/zero_copy.md
clalancette: I think the existing word is correct.  That is, I think this should be "middlewares", since we are talking about more than one.

clalancette (changes requested):

articles/zero_copy.md
fujitatomoya: thanks for checking! appreciate it :+1:

wjwwood (approved):

PR #335, opened 2023-08-17
Lifecycle design

alsora:


clalancette:
@SteveMacenski Since Nav2 is a heavy user of lifecycle, would you mind giving this a look?  We'd be very interested in your feedback.

SteveMacenski:
It'll be next week at this point for me, but yes I will! The only thing I see immediately that I would ask a question about is the throwing of exceptions. Who, if anyone, is able to catch that error in the state transition? It would be unacceptable if that was uncatchable by the user / rclcpp such that the server crashed.

alsora:
@clalancette @SteveMacenski This wasn't supposed to be a public-facing PR, but rather to first go through a last round of internal reviews (I should have targeted our fork rather than the official repo)

 I won't take it down but I converted it to draft.

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+For clarity and consistency, these entities are referred to as "Managed Entities".
 
-- `Unconfigured`
-- `Inactive`
-- `Active`
-- `Finalized`
+## Lifecycle State Machine
+This section outlines the Lifecycle state machine with respective transition behaviors, `CallbackReturn` semantic meaning, and updated states.
 
-To transition out of a primary state requires action from an external supervisory process, with the exception of an error being triggered in the `Active` state.
```
SteveMacenski, line 48: I think this remark has value to keep (unless you just moved it elsewhere)

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
 - `deactivate`
 - `shutdown`
+- `raise_error`
 - `destroy`
 
-The behavior of each state is as defined below.
 
-### Primary State: Unconfigured
+### Primary State Definitions and Expected Behaviors
+Primary States allow for assumptions and expected behavior of the Lifecycle Node with respect to the  node’s [Managed Entities](#managed-entities).
```
SteveMacenski, line 87: I have no idea what this sentences are trying to communicate. I also don't understand what the second sentence in this block is trying to say that shouldn't have already been made clear when introducing the idea of transition states.

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
 
-### Primary State: Unconfigured
+### Primary State Definitions and Expected Behaviors
+Primary States allow for assumptions and expected behavior of the Lifecycle Node with respect to the  node’s [Managed Entities](#managed-entities). 
+Except when in terminal state, all outbound transitions from a Primary State must be to Transition States.
 
-This is the life cycle state the node is in immediately after being instantiated.
-This is also the state in which a node may be retuned to after an error has happened.
-In this state there is expected to be no stored state.
+Note "processing" refers to Managed Entities receiving/sending messages, those messages being propagated through the `ROS 2` stack, and user callbacks processing these messages with respect to the executor.
```
SteveMacenski, line 90: I understand fully what this is communicating, but I think its overly technocratic. 

"Processing in managed entities means messages, services, actions, etc are being processed, whereas not processing indicates that messages are not being sent or received by the middleware."

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+*Outbound State Transitions*
+- `Transition:CONFIGURE` → `Configuring`
+- `Transition:SHUTDOWN` → `ShuttingDown`
+- `RAISE_ERROR` → `ErrorProcessing`
 
-This state represents a node that is not currently performing any processing.
 
-The main purpose of this state is to allow a node to be (re-)configured (changing configuration parameters, adding and removing topic publications/subscriptions, etc) without altering its behavior while it is running.
+#### **Primary State: `Inactive`**
+ This state represents a node that is performing minimal processing with allocated Managed Entities.
```
SteveMacenski, line 109: What is "minimal processing" the original documents' seems more clear (maybe just update this with the entity language)

> This state represents a node that is not currently performing any processing.

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+ Thus, as defined in the [Managed Entities](#managed-entities) section, the Managed Entities will not be added to the `ROS` graph with data retention to topics following that of an unallocated Persistent Entity.
```
SteveMacenski, line 110: If pre-defined and within the definition, it needs not be restated repeatedly. Especially when the sentence prior basically just says "not processing"

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
-Data retention will be subject to the configured QoS policy for the topic.
+*Outbound State Transitions*
+- `Transition:ACTIVATE` → `Activating`
+- `Transition:CLEANUP` → `CleaningUp`
+- `Transition:SHUTDOWN` →  `ShuttingDown`
+- `RAISE_ERROR` → `ErrorProcessing`
 
-Any managed service requests to a node in the inactive state will not be answered (to the caller, they will fail immediately).
+#### **Primary State: `Active`**
+Equivalent to the default behavior of a created `Node`.
```
SteveMacenski, line 124: Misses a clear statement that this is the main operations state of the lifecycle state machine

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
-While in this state, the node performs any processing, responds to service requests, reads and processes data, produces output, etc.
+#### **Primary State: `Finalized`**
+The `Finalized` state is the state in which the node ends immediately before being destroyed.
+This state exists to support debugging and introspection.
+A node which has failed will remain visible to system introspection and may be potentially introspectable by debugging tools instead of directly destructing.
+There will be no Lifecycle internal/backend policy for automatically restarting a node.
 
-If an error that cannot be handled by the node/system occurs in this state, the node will transition to `ErrorProcessing`.
+*State Assumptions / Behaviors*
+- This state can only be reached via a `ShuttingDown` → `SUCCESS` → `Finalized`
```
SteveMacenski, line 144: Why is `SUCCESS` called out? That's not a state. I think that can be removed since none of the other discussions of state assumptions / behaviors introduces error conditions yet above so there's no reason for a reader to look at this differently

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
-### Primary State: Finalized
 
-The `Finalized` state is the state in which the node ends immediately before being destroyed.
-This state is always terminal the only transition from here is to be destroyed.
 
-This state exists to support debugging and introspection.
-A node which has failed will remain visible to system introspection and may be potentially introspectable by debugging tools instead of directly destructing.
-If a node is being launched in a respawn loop or has known reasons for cycling it is expected that the supervisory process will have a policy to automatically destroy and recreate the node.
+#### **Primary State: `UncleanFinalized`**
+This state exists to denote a node that failed recovery in `ErrorProcessing` and therefor the `LifecycleNode` cannot be guaranteed to be cleanly destroyed.
```
SteveMacenski, line 155:
```suggestion
This state exists to denote a node that failed recovery in `ErrorProcessing` and therefore the `LifecycleNode` cannot be guaranteed to be cleanly destroyed.
```

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+An external supervisor would therefor be responsible for handling of this node (i.e., keeping it around in an unclean state or cleaning it up such as killing the process it is in).
```
SteveMacenski, line 156:
```suggestion
An external supervisor would therefore be responsible for handling of this node (i.e., keeping it around in an unclean state or cleaning it up such as killing the process it is in).
```

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+
+*State Assumptions / Behaviors*
+- This state can only be reached upon  `ErrorProcessing` → `FAILURE` → `UncleanFinalized` 
+- This state indicates failed recovery and thus is in an unclean / uncertain state w.r.t. Managed Entities / memory / cpu (i.e., node is in an uncertain finalized state where we cannot guarantee it can be cleanly destroyed)
+
+**Terminal state**: No Outbound Transitions
```
SteveMacenski, line 162: So there's no way out of this state? That's really not good. There needs to be _some_ way to recover or why not simply crash?

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+
+
+### Transition State Definitions and Expected Behaviors
+
+Transition States exist to transition a node's [Managed Entities](#managed-entities) while allowing for node set up, teardown, and error recovery. As such, these states cannot be terminal states.
```
SteveMacenski, line 167: This isn't as confusingly technocratic as the other time I made that remark, but I think this entire introduction section should be rewritten in more easily parsed language. I don't think so many words are needed to communicate what you want to communicate.

SteveMacenski: As well for "Definition of a Transition"

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
-This method is expected to clear all state and return the node to a functionally equivalent state as when first created.
-If the cleanup cannot be successfully achieved it will transition to `ErrorProcessing`.
+By default, all user transition functions return `CallbackReturn::SUCCESS` **except the `ErrorProcessing` callback which returns `CallbackReturn::FAILURE` by default**. 
+This avoids the unintentional situation of returning a `CallbackReturn::ERROR` that silently succeeds/transitions into `Unconfigured`.
+With this design, the node will transition to `UncleanFinalized` giving a more clear debugging pathway.
 
-#### Valid transitions out if CleaningUp
+#### Synchronous and Asynchronous Transitions
+By default, a transition is synchronous (i.e., thread-blocking requiring a return value from the user transition function). 
+Asynchronous (i.e., response deferrable) transition functions will also be supported.
```
SteveMacenski, line 222: Oh? That seems like a terrible idea

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+This follows the same principles of [concurrent `PUT` requests for REST APIs](https://medium.com/swlh/handling-concurrent-requests-in-a-restful-api-5a25a4b81a1). A cancel request is considered valid if the ongoing transition matches this `id` as well as there is not already an ongoing cancel request (similar to the [Definition of a Transition](#definition-of-a-transition)). 
+
+Once a cancel is valid and active, it is up to the user code to check on this cancel request (similar to a `ROS 2 Action`).
+This is designed as only the user can demarcate appropriate times and code paths to clean up an ongoing transition (i.e., only the user knows how to safely clean up a request at any given part of the ongoing transition).
+The user can opt to ignore this request (i.e., the transition continues as if no cancel request existed) or attempt to handle the request.
+The user can denote a successful handling of the request or a failed handling.
+
+If the user accepts and successfully handles the cancellation request, the state will follow the `CallbackReturn::FAILURE` path of the state machine.
```
SteveMacenski, line 247: I would think that if we cancel a transition, it wouldn't go all the way back to unconfigured but remain at the last known state. But I suppose that puts the burden on user-space code to "undo" any "partially done" stuff before the cancel was handled so I can understand the impulse.

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+
+Once a cancel is valid and active, it is up to the user code to check on this cancel request (similar to a `ROS 2 Action`).
+This is designed as only the user can demarcate appropriate times and code paths to clean up an ongoing transition (i.e., only the user knows how to safely clean up a request at any given part of the ongoing transition).
+The user can opt to ignore this request (i.e., the transition continues as if no cancel request existed) or attempt to handle the request.
+The user can denote a successful handling of the request or a failed handling.
+
+If the user accepts and successfully handles the cancellation request, the state will follow the `CallbackReturn::FAILURE` path of the state machine. 
+This follows the "fallback" behavior of a transition.
+If the user accepts and unsuccesfully handles the cancellation request, the state will follow the `CallbackReturn::ERROR` path of the state machine.
```
SteveMacenski, line 249:
```suggestion
If the user accepts and unsuccessfully handles the cancellation request, the state will follow the `CallbackReturn::ERROR` path of the state machine. 
```

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+
+Once a cancel is valid and active, it is up to the user code to check on this cancel request (similar to a `ROS 2 Action`).
+This is designed as only the user can demarcate appropriate times and code paths to clean up an ongoing transition (i.e., only the user knows how to safely clean up a request at any given part of the ongoing transition).
+The user can opt to ignore this request (i.e., the transition continues as if no cancel request existed) or attempt to handle the request.
+The user can denote a successful handling of the request or a failed handling.
+
+If the user accepts and successfully handles the cancellation request, the state will follow the `CallbackReturn::FAILURE` path of the state machine. 
+This follows the "fallback" behavior of a transition.
+If the user accepts and unsuccesfully handles the cancellation request, the state will follow the `CallbackReturn::ERROR` path of the state machine. 
+An ignored cancellation request becomes inactive and failed the instant a transition is inactive.
```
SteveMacenski, line 250:
```suggestion
An ignored cancellation request becomes inactive and failed the instant a transition is inactive.
```

I would think that we'd basically transition back down to what the cancel would have done for us if it were processed properly. E.g. If transitioning to Active and cancel is ignored, we just force the state machine back down to Configured. 

Or, if cancel accepted + processed puts it always back in the unconfigured state, then an ignored cancel request would transition the system back down the unconfigured automatically so it seems as if it were handled regardless

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
 
-- If the `onShutdown` callback raises or results in any other result code the node will transition to `Finalized`.
+#### **Operation `destroy()`**
+This is a stateless operation to be taken from the `Finalized` state.
+This operation will simply cause the deallocation of the node.
+This operation should always succeed if taken from the `Finalized` state.
 
-### Destroy Transition
+#### **Operation `raise_error()`**
+This operation transitions the node from the primary states `{Unconfigured, Inactive, Active}` to `ErrorProcessing`.
```
SteveMacenski, line 327: What exceptions can I throw that will do this? I think that needs to be explicit (e.g. std::runtime_error)

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+This is used to attempt a recovery to `Unconfigured` from a Primary State in the case of an internal error.
+Thus, this operation is only available internally to the node.
 
-This transition will simply cause the deallocation of the node.
-In an object oriented environment it may just involve invoking the destructor.
-Otherwise it will invoke a standard deallocation method.
-This transition should always succeed.
 
-### Create Transition
+### Return Values
```
SteveMacenski, line 332: I get that for transition states, but is that true for primary states as well? So no exception catching to cause us to go into error processing if an uncaught exception is thrown in a primary state? I think that's untrue given how the primary states work, so I think this needs clarification that we're only referring to transition states.

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
 
-### Lifecycle events
-
-A topic should be provided to broadcast the new life cycle state when it changes.
-This topic must be latched.
-The topic must be named `lifecycle_state` it will carry both the end state and the transition, with result code.
-It will publish every time that a transition is triggered, whether successful or not.
+### Supported Interfaces
+
+This section outlines Persistent Entities created by the backend of Lifecycle nodes. These Persistent Entities make up the external facing management interface of a Lifecycle node. Each of these external interfaces must also include an internal equivalent (i.e., a function the user can call from within the node itself).
```
SteveMacenski, line 373: L373 - L383 I have no idea what you're trying to communicate or why its an important thing to distinctly discuss

SteveMacenski: Actually, this entire subsection "Supported Interfaces" I don't understand. Something is really unclear about what the intention is. There's some stuff about the ROS interfaces (e.g. topics/services) which I ack as standard practice, but no idea what anything before "The following interface behaviors will be supported" seems unrelated. 

Also, all the interfaces are CamelCase and if these are supposed to be topics, then theey should be snake_case. If these are interface message types, that's unclear.

articles/node_lifecycle.md
```diff
@@ -27,224 +27,447 @@ Last Modified: {% if page.last_modified %}{{ page.last_modified }}{% else %}{{ p
...
+While not dictated by this design doc, implementation should strongly consider including an `failure_msg` or equivalent for services that include a `bool success` (e.g., [rclcpp::#2154](https://github.com/ros2/rclcpp/issues/2154) and [rclcpp::#2231](https://github.com/ros2/rclcpp/issues/2231)).
+This message would be to indicate why the service failed (e.g., `ChangeState` failing due to an already ongoing transition).
+It should be considered that if a service does not contain a `bool success`, it should not fail from the Lifecycle client library perspective.
+
+An additional implementation consideration would be to group the `ChangeState`, `CancelTransition`, and `OnStateMachineTransitioned` to be grouped into a single `ROS 2 Action` as it covers all desired behaviors.
+This would reduce needed code duplication if implemented as separate services (e.g., [rclcpp::#2213](https://github.com/ros2/rclcpp/issues/2213)).
+
+## Managed Entities
+
+Managed entities transition with the state machine, being automatically allocated and spun.
```
SteveMacenski, line 412: Huh (on this whole subsection)

SteveMacenski:
More general thoughts

- I'm not sure why ROS 2 is in code-styling, its quite distracting.
- I don't know if Persistent Entities is already part of this technology's lexicon. If its not, I might suggest changing it. I'm not sure it describes well what you want to differentiate. May I suggest _Unmanaged Entities_. 
- The T:Shutdown in the error processing marked half of the diagram is out of place and should be on the right hand side
- Also, the diagram should be inverted, the left (first) half should be the main state machine. Its confusing to inspect it in reverse
- I think some more of the concepts needs to be introduced up top before going into describing all the states. I don't think we should wait until introducing all the transition states to even tell someone what that means. Same with primary states, although its less an issue since its the first up. I think there needs to be a concepts section after Background to talk about the types of states like you do types of entities. Then you can lump all the states + expected behaviors in one section.

There are lots of things in here that are overly technocratic that even someone that uses lifecycle nodes every day it takes me 2-3 sometimes 4 times reading through to actually understand what you mean. I'd recommend looking over the original document where the same points were addressed previously that were sufficiently technical (might benefit from more detail) but written in language that I can grokk.

PR #337, opened 2024-01-07
fixed a typo

harderthan:


articles/intraprocess_communication.md
```diff
@@ -133,7 +133,7 @@ The executor can then pop the message from the buffer and trigger the callback o
 The choice of having independent buffers for each `Subscription` leads to the following advantages:
 
  - It is easy to support different QoS for each `Subscription`, while, at the same time, simplifying the implementation.
- - Multiple `Subscription`s can extract messages from their own buffer in parallel without blocking each other, thus providing an higher throughput.
+ - Multiple `Subscription`s can extract messages from their own buffer in parallel without blocking each other, thus providing a higher throughput.
```
clalancette: If we are being pedantic about this:

```suggestion
 - Multiple `Subscription`s can extract messages from their own buffer in parallel without blocking each other, thus providing higher throughput.
```

clalancette (changes requested):

articles/intraprocess_communication.md
harderthan: Thanks.  
Done.

ahcorde (approved):

PR #338, opened 2024-04-03
Use title instead of name in config

christophebedard:
`title:` seems to be preferred over `name:` now.

My goal is to try to change the title that Google uses:

image

It's not really clear how Google chooses the title of a link, though: https://developers.google.com/search/docs/appearance/title-link


christophebedard:
Also, the repo's description should probably be changed to rename "ROS 2.0" to "ROS 2"

image

mjcarroll (approved):

PR #341, opened 2024-08-06
history_depth is actually just depth

mikeferguson:
There isn't much documentation on QoS overrides - and many resources that I came across still link back to this (somewhat outdated) design document. Hopefully updating the parameter name here will save some people time in the future.

### DIFF
--- a/articles/qos_configurability.md
+++ b/articles/qos_configurability.md
@@ -107,7 +107,7 @@ That will automatically declare the parameters for reconfiguring the QoS policie
       'my/fully/qualified/topic/name/here':
         publisher:
           reliability: reliable
-          history_depth: 100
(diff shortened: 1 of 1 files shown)


mikeferguson:
For easy reference - this is the commit that implemented the design: https://github.com/ros2/rclcpp/commit/71a58d40f19160b1d1ba8f1f408702acaea94fde and the qos_overriding_options.hpp has not been altered since

fujitatomoya (approved):
lgtm

MichaelOrlov:
@mjcarroll @sloretz Friendly ping to merge this PR.

sloretz (approved):
Thank you for the PR!

PR #344, opened 2025-10-20
Clarify publish behavior for shared_ptr and unique_ptr

### LINKED ISSUE #2872: rclcpp does not support zero-copy intra-process communication for shared_ptr or std::move raw messages
### Generated by Generative AI

_No response_

### Operating System:

```Ubuntu 22```

### ROS version or commit hash:

humble

### RMW implementation (if applicable):

_No response_

### RMW Configuration (if applicable):

_No response_

### Client library (if applicable):

_No response_

### 'ros2 doctor --report' output

_No response_

### Steps to reproduce issue

1. Enable intra-process communication in a node:

`rclcpp::NodeOptions().use_intra_process_comms(true)`

2. Create a publisher:

`auto pub = node->create_publisher<std_msgs::msg::Int32>("topic", 10);`

3. Create a subscriber that accepts shared_ptr or shared ownership.

Publish a message using a shared_ptr, possibly moved:

```
    auto msg = std::make_shared<std_msgs::msg::Int32>();
    pub->publish(std::move(msg));
```

4. Observe whether the subscriber receives the same memory address as the published message.

Additionally, inspect the test function in [rclcpp/test/rclcpp/test_intra_process_manager.cpp](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/test_intra_process_manager.cpp#L531-L582) (lines 531–582).
The comment says it publishes a shared_ptr message, but the actual code only publishes std::unique_ptr.

### Expected behavior

Publishing a shared_ptr (or std::move(shared_ptr)) should avoid deep copying if intra-process communication is enabled, and all subscribers use shared ownership.

### Actual behavior

- In rclcpp, publishing a shared_ptr message always results in a deep copy, even if std::move() is used.
- The test case test_intra_process_manager.cpp incorrectly claims that the share_ptr message memory address is the same in intra processing communication, but the code publishes std::unique_ptr, so the intended case is not tested at all.

### Additional information

Tutorial [Intra-Process Communication](https://docs.ros.org/en/rolling/Tutorials/Demos/Intra-Process-Communication.html) documents that only std::unique_ptr can enable zero-copy communication, which is correct.

But this contradicts the design document, which implies that shared_ptr with same ownership can also be passed without copying.

It is theoretically possible to support zero-copy for shared_ptr in safe cases (e.g., use_count() == 1), or provide an opt-in behavior to avoid unnecessary copies.

### PR DESCRIPTION
**Pull request by Yuchen966**
## Description

Updates the intra-process communication design document to reflect the actual current behavior with respect to function publish():
shared_ptr messages: Cannot be published (removed incorrect zero-copy claims)
unique_ptr messages: Zero-copy transfer achieved 
raw messages: Both msg and std::move(msg) result in copying
**Key changes:**
- Add notes about current behaviour
- Removed the entire "Publishing SharedPtr" table section since shared_ptr cannot be published

Fixes [# (issue)](https://github.com/ros2/rclcpp/issues/2872)

### Is this user-facing behavior change?
No

### Did you use Generative AI?
No

### DIFF
--- a/articles/intraprocess_communication.md
+++ b/articles/intraprocess_communication.md
@@ -329,7 +329,7 @@ If all the `Subscription`s want ownership of the message, then a total of `N-1` 
 The last one will receive ownership of the published message, thus saving a copy.
 
 If none of the `Subscription`s want ownership of the message, `0` copies are required.
-It is possible to convert the message into a `std::shared_ptr<MessageT> msg` and to add it to every buffer.
+It is possible to convert the message into a `std::shared_ptr<const MessageT> msg` and to add it to every buffer.
 
 If there is 1 `Subscription` that does not want ownership while the others want it, the situation is equivalent to the case of everyone requesting ownership:`N-1` copies of the message are required.
 As before the last `Subscription` will receive ownership.
@@ -344,11 +344,6 @@ A copy of the message will be given to all the `Subscription`s requesting owners
 The following tables show a recap of when the proposed implementation has to create a new copy of a message.
 The notation `@` indicates a memory address where the message is stored, different memory addresses correspond to different copies of the message.
 
-**Note on current behavior**: The actual implementation behavior differs from the proposed implementation as described in [rclcpp issue #2872](https://github.com/ros2/rclcpp/issues/2872):
-- **shared_ptr messages**: Cannot be published.
-- **unique_ptr messages**: When using `std::move()`, zero-copy transfer is achieved.
-- **raw messages**: Both `msg` and `std::move(msg)` result in copying the message.
-
 #### Publishing UniquePtr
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: articles/intraprocess_communication.md [hunk 1]
```diff
@@ -352,22 +357,10 @@ The notation `@` indicates a memory address where the message is stored, differe
 | unique_ptr\<MsgT\> @1   | unique_ptr\<MsgT\>      |                     @1  |
 | unique_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> unique_ptr\<MsgT\>     |     @1 <br> @2 |
 | unique_ptr\<MsgT\> @1   | shared_ptr\<MsgT\>      |                     @1  |
-| unique_ptr\<MsgT\> @1   | shared_ptr\<MsgT\> <br> shared_ptr\<MsgT\>     |     @1 <br> @1 |
+| unique_ptr\<MsgT\> @1   | shared_ptr\<MsgT\> <br> shared_ptr\<MsgT\>     |     @1 <br> @2 |
```
alsora: your change here is wrong
if there is 1 publisher with unique ptr and two subscribers with shared pointers, both receive the same publisher pointer (`@1` in this case)

articles/intraprocess_communication.md
```diff
@@ -352,22 +357,10 @@ The notation `@` indicates a memory address where the message is stored, differe
...
 | unique_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> shared_ptr\<MsgT\>     |     @1 <br> @2 |
-| unique_ptr\<MsgT\> @1   | unique_ptr\<MsgT\> <br> shared_ptr\<MsgT\>   <br> shared_ptr\<MsgT\>   |  @1 <br> @2 <br> @2|
```
alsora, line 357: i don't think these should be removed

alsora (changes requested):
> ~~shared_ptr messages: Cannot be published (removed incorrect zero-copy claims)~~

~~Shared pointers CAN be published, although they don't achieve zero copy.~~
~~This is exactly what's indicated in the `#### Publishing SharedPtr` section that you removed.~~

My bad, i forgot that publishing shared_ptr has been deprecated long ago.

However, your PR is also making incorrect changes.

I think we should just remove the section about `Publishing shared_ptr`.

articles/intraprocess_communication.md
```diff
@@ -344,6 +344,11 @@ A copy of the message will be given to all the `Subscription`s requesting owners
 The following tables show a recap of when the proposed implementation has to create a new copy of a message.
 The notation `@` indicates a memory address where the message is stored, different memory addresses correspond to different copies of the message.
 
+**Note on current behavior**: The actual implementation behavior differs from the proposed implementation as described in [rclcpp issue #2872](https://github.com/ros2/rclcpp/issues/2872):
```
alsora: I would remove this section.
The fact that you can't publish shared ptr is not necessarily related to this design.

Moreover, i don't really see many references to publishing shared ptrs in this page.
 - https://github.com/ros2/design/blob/gh-pages/articles/intraprocess_communication.md?plain=1#L231
 - https://github.com/ros2/design/blob/gh-pages/articles/intraprocess_communication.md?plain=1#L360-L370

these should be removed

articles/intraprocess_communication.md
Yuchen966: Thanks for your suggestions. I just took a deeper look into the code, and I noticed that only `shared_ptr<const MsgT>` would share the stored object pointer and avoid extra copy. Every sharedptr containing non-const msg would require intra process manager do another copy. The changes I made is based on my experiments on `shared_ptr<MsgT>`, which is non-const.

I think the current implementation makes sense, as non-const sharedptr could be changed and an extra copy is necessary for safety reasons. But the table doesn't show the constness of MsgT. It also seems weird that the first part of design documentation specifies `const MsgT`, but at some point this gets lost. I will update the document to emphasize the constness of shared_ptr.

